PR #5, opened 2018-07-20
Recognize cmake files of jsoncpp library

ksaito7:
Modified cmake files to recognize jsoncpp library for statistics.


mbutrovich:
jsoncpp has apt and Homebrew packages. Should we be linking from there rather than adding to third_party?

apavlo:
@mbutrovich We should link things when possible.

ksaito7:
Got it. I'll change to add its installation into packages.sh and find_package for it into dependencies.cmake.

ksaito7:
I did it and checked it on Ubuntu, but I can't check on mac (homebrew), could @mbutrovich check it?

mbutrovich:
There was still a reference to the local files subdirectory and include path, so I removed those. It also wasn't finding the package on macOS. It's working for me now, but you can check these changes on your end?

ksaito7:
Thanks so much! It could find the package on Linux. But it seems the include path differs between Linux and Mac. I modified it to be changed by macro.

PR #9, opened 2018-07-23
Include <cstddef> to use std::byte on Ubuntu.

lmwnshn:
I think it worked on Mac because we had an explicit stdlib=libc++.
This allows cmake to build on Ubuntu with gcc/g++ 7.


lmwnshn:
Curiously, Travis Mac isn't happy about the missing include too.

PR #14, opened 2018-07-24
Change Printable class to json value based

aaron-tian:


src/include/common/printable.h
```diff
@@ -1,25 +1,54 @@
...
-  virtual const std::string GetInfo() const = 0;
+  /** @brief Get the styled string from the Json value. */
+  std::string GetInfo ();
 
-  friend std::ostream &operator<<(std::ostream &os, const Printable &printable) {
-    return os << printable.GetInfo();
-  }
+ protected:
+  /** The Json value about the printable object */
+  Json::Value json_value_;
```
apavlo: I don't think we want to allocate and store a JSON object for every object. Why can't we just generate this thing any time `GetInfo` is invoked?

src/include/common/printable.h
```diff
@@ -1,25 +1,54 @@
...
+
+  /** @brief Get the Json value about the printable object. */
+  Json::Value GetPrintable ();
 
-  /** @brief Return information about the object in human-readable way. */
-  virtual const std::string GetInfo() const = 0;
+  /** @brief Get the styled string from the Json value. */
+  std::string GetInfo ();
 
-  friend std::ostream &operator<<(std::ostream &os, const Printable &printable) {
```
apavlo: Why did you get rid of this?

aaron-tian: We don't need it anymore since we can get the embedding feature and styled string for free once we have json values.

apavlo:
Instead of `Printable`, we should rename this to `JSONSerializable`.

I would also keep the `<<` override operator

src/include/common/printable.h
aaron-tian: JSON library also overload the `<<` operator even if we want to directly call it, which is not necessary though.

aaron-tian:
JSON library also overrides the << operator even if we want to directly call it, which is not necessary though.

src/include/common/printable.h
aaron-tian: Will do, but we have to keep a JSON object for statistics, otherwise it would be difficult or even infeasible to print all the stats in gdb.

apavlo:
I would get rid of the std::string `GetInfo` and rename it to `GetJSON`. Dynamically create the JSONValue on the fly inside of that function.

PR #21, opened 2018-07-24
Varlen pool

aaron-tian:


test/storage/varlen_pool_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+#include <pthread.h>
+
+#include "storage/varlen_pool.h"
+#include "gtest/gtest.h"
+
+namespace terrier {
+class VarlenPoolTests : public ::testing::Test {};
+
+// disable unused const variable warning for clang
+#ifdef __APPLE__
```
tli2: Why do you need this? What unused const variable is there that you cannot get rid of?

test/storage/varlen_pool_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+#pragma clang diagnostic ignored "-Wunused-const-variable"
+#endif
+    const size_t str_len = 1000; // test string length
+#ifdef __APPLE__
+#pragma clang diagnostic pop
+#endif
+
+// Allocate and free once
+TEST_F(VarlenPoolTests, AllocateOnceTest) {
+  std::unique_ptr<VarlenPool> pool(new VarlenPool());
```
tli2: Why do you need a unique ptr here?

test/storage/varlen_pool_test.cpp
```diff
@@ -0,0 +1,45 @@
...
+  void *p = nullptr;
+  size_t size;
```
tli2: const uint64_t size = 40;

test/storage/varlen_pool_test.cpp
```diff
@@ -0,0 +1,45 @@
+//===----------------------------------------------------------------------===//
```
tli2: Maybe you want to write more tests. Even though the implementation may change soon, these tests are not implementation specific and can be reused.

At least test for concurrency.

src/include/storage/varlen_pool.h
```diff
@@ -0,0 +1,78 @@
...
+/**
+ * @brief A varlen pool that can quickly allocate chunks of memory.
+ * */
+class VarlenPool {
+ public:
+  VarlenPool() = default;
+
+  ~VarlenPool();
+
+  void *Allocate(size_t size);
```
tli2: We are using uint64_t instead of size_t in the new codebase

src/include/storage/varlen_pool.h
```diff
@@ -0,0 +1,78 @@
...
+#include "common/spin_latch.h"
+
+namespace terrier {
+
+/**
+ * @brief A varlen pool that can quickly allocate chunks of memory.
+ * */
+class VarlenPool {
+ public:
+  VarlenPool() = default;
```
tli2: default empty constructors and destructors should be omitted

src/include/storage/varlen_pool.h
```diff
@@ -0,0 +1,78 @@
...
+
+  ~VarlenPool();
+
+  void *Allocate(size_t size);
+
+  void Free(void *ptr);
+
+ public:
+  /** Location list */
+  std::unordered_set<char *> locations_;
```
tli2: use (byte *) in the new code base.

src/include/storage/varlen_pool.h
```diff
@@ -0,0 +1,78 @@
...
+  SpinLatch pool_lock_;
+};
+
+////////////////////////////////////////////////////////////////////////////////
+///
+/// Implementation below
+///
+////////////////////////////////////////////////////////////////////////////////
+
+VarlenPool::~VarlenPool() {
```
tli2: Inline these in the declaration in the new code base.

src/include/storage/varlen_pool.h
```diff
@@ -0,0 +1,78 @@
...
+#include <unordered_set>
+
+#include "common/spin_latch.h"
+
+namespace terrier {
+
+/**
+ * @brief A varlen pool that can quickly allocate chunks of memory.
+ * */
+class VarlenPool {
```
tli2: Write doxygen docs for every public method

src/include/common/spin_latch.h
```diff
@@ -0,0 +1,57 @@
+//===----------------------------------------------------------------------===//
```
tli2: Remove this designation as we haven't decided on what to do with them in the new codebase yet.

src/include/common/spin_latch.h
```diff
@@ -0,0 +1,57 @@
...
+
+#include "common/macros.h"
+
+//===--------------------------------------------------------------------===//
+// Cheap & Easy Spin Latch
+//===--------------------------------------------------------------------===//
+
+namespace terrier {
+
+enum class LatchState : bool { Unlocked = 0, Locked };
```
tli2: enum members should be all caps snake case.

src/include/common/spin_latch.h
```diff
@@ -0,0 +1,57 @@
...
+    state_.store(LatchState::Unlocked, std::memory_order_release);
+  }
+
+ private:
+  /* the exchange method on this atomic is compiled to a lockfree xchgl
+   * instruction*/
+  std::atomic<LatchState> state_;
+};
+
+}  // namespace terrier
```
tli2: missing new line at the end of file

src/include/common/spin_latch.h
```diff
@@ -0,0 +1,57 @@
...
+//===--------------------------------------------------------------------===//
+// Cheap & Easy Spin Latch
+//===--------------------------------------------------------------------===//
+
+namespace terrier {
+
+enum class LatchState : bool { Unlocked = 0, Locked };
+
+class SpinLatch {
+ public:
```
tli2: doxygen comments for class and methods.

src/include/storage/varlen_pool.h
```diff
@@ -0,0 +1,78 @@
...
+  locations_.insert(location);
+  pool_lock_.Unlock();
+
+  return location;
+}
+
+void VarlenPool::Free(void *ptr) {
+  auto *cptr = (char *)ptr;
+  pool_lock_.Lock();
+  auto result = locations_.erase(cptr);
```
tli2: Per discussion with Matt we should check the return value of erase here,

aaron-tian: We do check if the result is not 0 in the following `if` statement, which means we only free `cptr` when we can find it in our `locations_`.

src/include/storage/varlen_pool.h
aaron-tian: Written

aaron-tian:
Thanks for the review. Done.

PR #25, opened 2018-07-25
Added bitmap and bitmap_test

mbutrovich:
We need a bitmap for our tuple buffers being passed around, but don't need it to be concurrent. This is  basically ConcurrentBitmap without the CAS.

Also removed UnsafeSet from ConcurrentBitmap and ran formatter.

### DIFF
--- a/src/include/common/container/bitmap.h
+++ b/src/include/common/container/bitmap.h
@@ -1,6 +1,7 @@
 #pragma once
 #include <memory>
 #include "common/common_defs.h"
+#include "common/concurrent_bitmap.h"
 
 #ifndef BYTE_SIZE
 #define BYTE_SIZE 8u
@@ -49,7 +50,8 @@ class RawBitmap {
    * @param size number of bits in the bitmap
    * @return ptr to new RawBitmap
    */
-  static RawBitmap *Allocate(uint32_t size) {
+  static RawBitmap *Allocate(uint32_t num_elements) {
+    auto size = BitmapSize(num_elements);
     auto *result = new uint8_t[size];
     PELOTON_MEMSET(result, 0, size);
     return reinterpret_cast<RawBitmap *>(result);
@@ -97,9 +99,6 @@ class RawBitmap {
       Set(pos, true);
     return *this;
   }
-
-  // TODO(Tianyu): We will eventually need optimization for bulk checks and
-  // bulk flips. This thing is embarrassingly easy to vectorize.
 
  private:
   uint8_t bits_[0];
--- a/src/include/common/container/bitmap.h
+++ b/src/include/common/container/bitmap.h
@@ -93,10 +93,7 @@ class RawBitmap {
   }
 
   RawBitmap &Flip(uint32_t pos) {
-    if (Test(pos))
-      Set(pos, false);
-    else
-      Set(pos, true);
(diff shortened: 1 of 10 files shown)


src/include/common/container/bitmap.h
```diff
@@ -0,0 +1,113 @@
+#pragma once
+#include <memory>
+#include "common/common_defs.h"
+#include "common/concurrent_bitmap.h"
```
tli2: Why is this a dependency?

src/include/common/container/bitmap.h
```diff
@@ -0,0 +1,113 @@
...
+
+#ifndef BYTE_SIZE
+#define BYTE_SIZE 8u
+#endif
+
+// Some platforms would have already defined the macro. But its presence is
+// not standard and thus not portable. Pretty sure this is always 8 bits.
+// If not, consider getting a new machine, and preferably not from another
+// dimension :)
+static_assert(BYTE_SIZE == 8u, "BYTE_SIZE should be set to 8!");
```
tli2: Seeing how this is duplicated, maybe the right way to do this is to have ConcurrentBitmap inherit from Bitmap, (but then it becomes unclear whether a method should be thread-safe or not.).

At the very least, it makes sense that concurrent_bitmap.h can include the bitmap header to get the same #defines.

src/include/common/container/bitmap.h
```diff
@@ -0,0 +1,113 @@
...
+  RawBitmap &Set(uint32_t pos, bool val) {
+    if (val)
+      bits_[pos / BYTE_SIZE] |= ONE_HOT_MASK(pos % BYTE_SIZE);
+    else
+      bits_[pos / BYTE_SIZE] &= ONE_COLD_MASK(pos % BYTE_SIZE);
+    return *this;
+  }
+
+  RawBitmap &Flip(uint32_t pos) {
+    if (Test(pos))
```
tli2: ^=

test/common/bitmap_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+
+namespace terrier {
+template<uint32_t num_elements>
+void CheckReferenceBitmap(const common::RawBitmap &tested,
+                          const std::bitset<num_elements> &reference) {
+  for (uint32_t i = 0; i < num_elements; ++i) {
+    EXPECT_EQ(reference[i], tested[i]);
+  }
+}
+
```
tli2: nit: description?

test/common/bitmap_test.cpp
```diff
@@ -0,0 +1,113 @@
...
+#include <thread>
+#include <unordered_set>
+
+#include "gtest/gtest.h"
+#include "common/test_util.h"
+#include "common/container/bitmap.h"
+
+namespace terrier {
+template<uint32_t num_elements>
+void CheckReferenceBitmap(const common::RawBitmap &tested,
```
tli2: Maybe again put this function into test util, since it is now used in two places.

src/include/common/container/bitmap.h
mbutrovich: BitmapSize(), but in the next comment you're suggesting I reverse it and have bitmap.h contain the common stuff and have ConcurrentBitmap get it from there.

Issue #31, opened 2018-07-26
Package cleanup


lmwnshn:
The minimum dependencies appear to be:

clang-format-6.0
clang-tidy-6.0
cmake
git
g++-7
libjemalloc-dev
libjsoncpp-dev
libtbb-dev
libz-dev
llvm-6.0

TODO:
1. Rewrite packages.sh
2. Rewrite as brew file
3. Rewrite Dockerfile (need to manually set /usr/bin/c++)

PR #34, opened 2018-07-26
Packages cleanup, new Travis check

lmwnshn:
Addresses #31.

We now only support Ubuntu 18.04 and MacOS.

1. Simplifies and removes the packages installed by packages.sh to a bare minimum.
2. We are not using Brewfiles. We will specify version in brew install whenever possible.
3. Added a new machine to Travis that will install requirements via packages.sh instead of Dockerfile.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,26 +11,46 @@ matrix:
       install:
         - ./script/installation/packages.sh
         - export PATH=/usr/local/opt/llvm/bin:$PATH
-      script:
-        - mkdir build
-        - cd build
-        - cmake ..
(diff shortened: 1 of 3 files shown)


lmwnshn:
The build passes, but there's an error in the logs: 
`/usr/include/wchar.h:35:10: error: 'stddef.h' file not found [clang-diagnostic-error]`

@mbutrovich Ideas? I'm not sure if I'm missing a package or if clang is being fussy.

mbutrovich:
Maybe this for the stddef.h issue?

https://ubuntuforums.org/showthread.php?t=674047

PR #35, opened 2018-07-27
Add JsonSerializable class for printing state information

aaron-tian:


src/include/common/json_serializable.h
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#ifdef __APPLE__
```
tli2: Put a comment around it so others know what's going on.

PR #51, opened 2018-08-01
Documentation and Doxygen check

lmwnshn:
Largely documentation, please check that the comments make sense.

Not documentation:
1. corrected the type of string_util::Indent (from int to size_t).
2. made RawBitmap::Clear() consistent in taking bits rather than bytes.
3. Enabled documentation check on Travis.

### DIFF
--- a/src/include/common/constants.h
+++ b/src/include/common/constants.h
@@ -14,6 +14,6 @@ struct Constants {
   /**
    * Block size.
    */
-  static const uint32_t BLOCK_SIZE = 1048576u; // Should only ever be a power of 2.
+  static const uint32_t BLOCK_SIZE = 1048576u;  // Should only ever be a power of 2.
 };
 }  // namespace terrier
--- a/src/include/common/constants.h
+++ b/src/include/common/constants.h
@@ -12,8 +12,8 @@ namespace terrier {
  */
 struct Constants {
   /**
-   * Should only ever be a power of 2.
+   * Block size.
    */
-  static const uint32_t BLOCK_SIZE = 1048576u;
+  static const uint32_t BLOCK_SIZE = 1048576u; // Should only ever be a power of 2.
 };
 }  // namespace terrier
--- a/src/include/common/constants.h
+++ b/src/include/common/constants.h
@@ -11,7 +11,9 @@ namespace terrier {
  * in code.
  */
 struct Constants {
-  // Should only ever be a power of 2
+  /**
+   * Should only ever be a power of 2.
+   */
   static const uint32_t BLOCK_SIZE = 1048576u;
 };
 }  // namespace terrier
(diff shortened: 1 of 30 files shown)


src/include/common/constants.h
```diff
@@ -11,7 +11,9 @@ namespace terrier {
  * in code.
  */
 struct Constants {
-  // Should only ever be a power of 2
+  /**
+   * Should only ever be a power of 2.
```
tli2: This shouldn't be a doxygen comment as that is an implementation detail. Back to inline?

src/include/common/container/bitmap.h
```diff
@@ -42,16 +42,19 @@ class RawBitmap {
   // Always reinterpret_cast from raw memory.
   RawBitmap() = delete;
   ~RawBitmap() = delete;
+  /**
```
tli2: Does doxygen complain about this?
If so, maybe it's better to add this generic comment into the macro.

src/include/common/typedefs.h
```diff
@@ -67,15 +67,42 @@ template <class Tag, typename T>
 class StrongTypeAlias {
  public:
   StrongTypeAlias() : val_() {}
+  /**
+   * Constructs a new StrongTypeAlias.
+   * @param val pointer to the underlying type.
```
tli2: const reference instead of pointer

src/include/storage/storage_defs.h
```diff
@@ -15,19 +15,42 @@ namespace terrier::storage {
...
+   * Number of columns.
+   */
   const uint16_t num_cols_;
+  /**
+   * Vector of attribute sizes.
+   */
   const std::vector<uint8_t> attr_sizes_;
-  // Cached so we don't have to iterate through attr_sizes every time
+  /**
+   * Cached tuple size so that we don't have to iterate through attr_sizes_ every time.
```
tli2: again, cached or not is an implementation detail better left for inline comments. (intended for developer in this file rather than users of the file.)

src/include/storage/storage_defs.h
```diff
@@ -242,16 +290,36 @@ class ProjectedRow {
...
+  /**
+   * Pointer to the next delta record.
+   */
   DeltaRecord *next_;
+  /**
+   * Timestamp at which the old projected row was visible.
+   */
   timestamp_t timestamp_;
+  /**
+   * The old projected row.
```
tli2: before-image of the modified attributes on the row

src/include/storage/storage_defs.h
```diff
@@ -242,16 +290,36 @@ class ProjectedRow {
...
+/**
+ * A DeltaRecord points to the old projected row and the timestamp at which it was visible.
+ */
 struct DeltaRecord {
+  /**
+   * Pointer to the next delta record.
+   */
   DeltaRecord *next_;
+  /**
+   * Timestamp at which the old projected row was visible.
```
tli2: up to which

PR #61, opened 2018-08-02
Add RawConcurrentBitmap::FirstUnsetPos, use it in TAS::Allocate

### LINKED ISSUE #24: storage::TupleAccessStrategy::Allocate is inefficient
Right now we cmpxchg every single bit (8 times on a byte) linearly until successful. Realistically we can test for bits 64 at a time and only cmpxchg on the ones that tested false.

lmwnshn:
Fix #24.

1. Adds FirstUnsetPos to find the first unset bit in a concurrent bitmap.
We search for the first unset bit, narrowing down 64, 32, 16, and finally 8 bits at a time.
We sidestep endianness issues by testing if the entire word is -1.

Because our result can be stale and we might end up calling it a lot, I added a start_pos parameter to start searching from a given bit position. Technically, we could reuse out_pos as an in-out, but that seems less explicit. Open to suggestions.

2. Minor documentation fix-up for bitmap's Allocate (rename size to num_bits).


src/include/common/concurrent_bitmap.h
```diff
@@ -88,6 +88,78 @@ class RawConcurrentBitmap {
...
+   * @param[out] out_pos the position of the first unset bit will be written here, if it exists.
+   * @return true if an unset bit was found, and false otherwise.
+   */
+  bool FirstUnsetPos(uint32_t num_bits, uint32_t start_pos, uint32_t *out_pos) {
+    uint32_t num_bytes = BitmapSize(num_bits);  // maximum number of bytes in the bitmap
+    uint32_t byte_pos = start_pos / BYTE_SIZE;  // current byte position
+    uint32_t search_width;                      // number of bits searched at a time
+
+    // optimization: don't search wider than the number of bits we have
+    if (num_bits <= 8) {
```
tli2: nit: Can you get rid of these magic constants? (Use sizeof(uint8_t), etc. I know this sounds redundant.)

src/include/common/concurrent_bitmap.h
```diff
@@ -88,6 +88,78 @@ class RawConcurrentBitmap {
...
+      search_width = 32;
+    } else {
+      search_width = 64;
+    }
+
+    while (byte_pos < num_bytes) {
+      // each case loads a word of search_width bits and casts it as a signed type.
+      // if all bits are set (equals -1), then we can refine our search.
+      switch (search_width) {
+        default:
```
tli2: default?

src/include/common/concurrent_bitmap.h
```diff
@@ -88,6 +88,78 @@ class RawConcurrentBitmap {
...
+        case 64:
+          if (static_cast<std::atomic<int64_t>>(bits_[byte_pos]).load() != -1) {
```
tli2: You can probably templatize this on T, considering that earlier you would do sizeof(T), and otherwise these branches look identical.

src/include/common/concurrent_bitmap.h
```diff
@@ -88,6 +88,99 @@ class RawConcurrentBitmap {
...
+    }
+    return true;
+  }
+
+  /**
+   * Returns the position of the first unset bit, if it exists.
+   * Note that this result is immediately stale.
+   * @param num_bits number of bits in the bitmap.
+   * @param start_pos start searching from this bit location.
+   * @param[out] out_pos the position of the first unset bit will be written here, if it exists.
```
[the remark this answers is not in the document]
lmwnshn: By this nit, you mean: don't use `@param[out]`, just use `param`?

lmwnshn:
Too much performance variance on my laptop to check if it actually sped up TAS or otherwise.

tli2:
A quick hacked together benchmark shows significant speed up. (fill up 20 pages with layout {8, 8})

speedybits: ~17s

master: ~10 min

@lmwnshn good job

src/include/common/concurrent_bitmap.h
```diff
@@ -88,6 +88,106 @@ class RawConcurrentBitmap {
...
+   * Looks for an unset bit in a T-sized word from the bitmap, starting at byte_pos.
+   * If an unset bit is found, returns true.
+   * Otherwise updates byte_pos to be the next place we should search.
+   * @tparam T signed fixed width integer type.
+   * @param[in,out] byte_pos invariant: next byte position we should search.
+   * @param[in,out] bits_left the number of valid bits remaining.
+   * @return true if an unset bit was found, false otherwise.
+   */
+  template <class T>
+  bool FindUnsetBit(uint32_t *byte_pos, uint32_t *bits_left) {
```
tli2: nit: move this to private.

src/include/common/concurrent_bitmap.h
```diff
@@ -88,6 +88,106 @@ class RawConcurrentBitmap {
...
+    // for a signed integer, -1 represents that all the bits are set
+    if (reinterpret_cast<std::atomic<T> *>(&bits_[*byte_pos])[0].load() == -1) {
```
tli2: nit: maybe do this load outside the if.
```c++
T bits = reinterpret_cast<std::atomic<T> *>(&bits_[*byte_pos])->load();
if (bits == static_cast<T>(-1)) {
...
```

src/include/common/concurrent_bitmap.h
```diff
@@ -88,6 +88,106 @@ class RawConcurrentBitmap {
...
+      *byte_pos += static_cast<uint32_t>(sizeof(T));
+      // saturate bits_left at 0
```
tli2: nit: "saturate" has unclear meaning here. maybe just say you want to prevent underflow

src/include/common/concurrent_bitmap.h
```diff
@@ -88,6 +88,106 @@ class RawConcurrentBitmap {
...
+    uint32_t num_bytes = BitmapSize(bitmap_num_bits);  // maximum number of bytes in the bitmap
+    uint32_t byte_pos = start_pos / BYTE_SIZE;         // current byte position
+    uint32_t search_width;                             // number of bits searched at a time
+    uint32_t bits_left = bitmap_num_bits;              // number of bits remaining
+    bool found_unset_bit = false;                      // whether we found an unset bit previously
+
+    while (byte_pos < num_bytes && bits_left > 0) {
+      // as soon as we find an unset bit, we go back to looking at single bytes
+      if (found_unset_bit) {
+        search_width = sizeof(uint8_t);
```
tli2: nit: inline this. i.e.

```c++
while (...) {
  if (found_unset_bit || search_width < sizeof(uint16_t)) {
    ...
  } else if (...) {
    found_unset_bit = FindUnsetBit<int64_t>(...);
  } ...
}
```

PR #62, opened 2018-08-02
Coveralls experiments

lmwnshn:
**Do not merge.**

Messing with coveralls.


lmwnshn:
Closed since it appears to have fixed itself.

PR #76, opened 2018-08-06
Clang tidy fixes

### LINKED ISSUE #50: check-clang-tidy, clang-tidy possibly broken
make check-clang-tidy is producing errors (e.g. /repo/src/include/common/concurrent_map.h:95:24: error: no member named 'emplace' in ...) without failing the build. It is producing an exit code of 0 (both inside and outside Docker images), which is why Travis doesn't fail it.

There is a related issue [here](https://github.com/01org/tbb/issues/22), but attempts at defining TBB_USE_GLIBCXX_VERSION and __TBB_CPP11_VARIADIC_TEMPLATES_PRESENT have not solved the issue. The gist is that it can't detect support for C++11 features.

lmwnshn:
Fix #50.

1. Linux + clang + TBB = TBB cannot detect C++11 features.
We can manually tell TBB where the C++11 features are by setting -DTBB_USE_GLIBCXX_VERSION=V, where for gcc compiler version M.m.p, we have V = 10000 M + 100 m + p.
As a workaround, we know that clang-tidy reads the contents of compile_commands.json, which is itself generated by CMake. So we add the above TBB_USE_GLIBCXX_VERSION as a CXX flag.

2. The clang-tidy we modified from Arrow was pretty broken. We settled for bugprone, clang-analyzer, google, modernize, performance, portability, and readability checks. We also check header files now.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -18,5 +18,18 @@
 # Modified from the Apache Arrow project for the Terrier project.
 #
 ---
-Checks:          'clang-diagnostic-*,clang-analyzer-*,-clang-analyzer-alpha*,google-.*,modernize-.*,performance-.*,readablity-.*'
+Checks:     '
+            bugprone-*,
+            clang-analyzer-*,
+            google-*,
+            modernize-*,
+            performance-*,
+            portability-*,
+            readability-*,
+            -google-readability-braces-around-statements,
+            -google-readability-namespace-comments,
+            -readability-braces-around-statements,
+            '
+WarningsAsErrors: '*'
+HeaderFilterRegex: '.*'
 AnalyzeTemporaryDtors: true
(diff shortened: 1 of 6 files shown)


.clang-tidy
```diff
@@ -18,5 +18,18 @@
...
+            bugprone-*,
+            clang-analyzer-*,
+            google-*,
+            modernize-*,
+            performance-*,
+            portability-*,
+            readability-*,
+            -google-readability-braces-around-statements,
+            -google-readability-namespace-comments,
+            -readability-braces-around-statements,
```
mbutrovich: I'd like to formally register my protest against omitting this check. :)

cmake_modules/FindTBB.cmake
```diff
@@ -94,6 +94,16 @@ include(FindPackageHandleStandardArgs)
 
 if(NOT TBB_FOUND)
 
+    #############################################
+    # Hack to make TBB and clang-tidy play nicely
+    # https://www.threadingbuildingblocks.org/docs/help/reference/environment/feature_macros.html
+    # https://www.threadingbuildingblocks.org/docs/help/reference/appendices/known_issues/linux_os.html
+    #############################################
+    if (NOT TBB_USE_GLIBCXX_VERSION AND UNIX AND NOT APPLE)
```
mbutrovich: LINUX instead of UNIT AND NOT APPLE?

PR #77, opened 2018-08-07
TupleAccessStrategy benchmarks

mbutrovich:
Once these go in, we can model the DataTable benchmarks on them and come up with more interesting scenarios.

With initial profiling of these benchmarks, as you'd expect with 8 threads doing tuple-at-a-time inserts, 50% of our time is spent in the bitmap finding a slot.

Summary of changes:
- Added storage_benchmark_util.
- Added ConcurrentInsert benchmark.
- Combined TAS_test_util with storage_test_util.


mbutrovich:
The reason for storage_benchmark_util is to remove the correctness checks and STL overhead incurred by the storage_test_util stuff. By making these changes, we went from this:
```
BM_ConcurrentInsert/repeats:10/real_time_mean           48.5164MB/s
BM_ConcurrentInsert/repeats:10/real_time_median         48.487MB/s
BM_ConcurrentInsert/repeats:10/real_time_stddev          1.57963MB/s
```
to this:
```
BM_ConcurrentInsert/repeats:10/real_time_mean           78.7985MB/s
BM_ConcurrentInsert/repeats:10/real_time_median         79.6141MB/s
BM_ConcurrentInsert/repeats:10/real_time_stddev          1.84267MB/s
```
I believe that this model where we write util functions for the Google Tests that stress correctness, and then duplicate the logic with the checks removed for benchmarking is a good way to use Google Benchmark going forward.

Issue #47, opened 2018-08-01
Performance counters and stats


tli2:
Migrated from earlier discussions. See #30.

Replicated below (from @pervazea ):

I don't think the design in this PR is suitable for building upon. In other words, I'm saying we should not merge this PR. So, rather than comment on the implementation, I'm going to step back to design, requirements and use cases. I don't have answers for everything. 

Questions
1. Do we want statistics only in debug mode or also in release mode? IMO even in release mode you need stats for monitoring and understanding what the system is doing. Depending on the amount of overhead in debug mode, the behavior can be quite a bit different in release mode. Further, it would possible and reasonable to feed stats into the self-learning portion, if we so chose.

1. How are stats collected, reported, used? IMO the current design is lacking in this area.

1. How are stats aggregated?

1. Do we need to clear stats? Normally yes, you'd want a method to clear stats. You can cheat and build in something to collect and diff stats. Not as nice though. 

1. Do we want only counters, or also also include support for time durations? One could usefully start with just the former.

Requirements
1. Cost of keeping stats should be minimized.
1. If stats are "disabled" or compiled out, the resulting overhead on the system should be zero.
1. Stats should be available in human friendly form as well as in API / machine consumable form.
1. A mechanism should be provided for determining available stats.
1. A mechanism should be provided for dumping selected stats.
1. It is preferred that stats aggregation is part of the design and built into the system rather than bolted on externally.

Examples of stats usage
1. Codegen cache. Stats kept for this might include:
    1. No. of inserts into cache
    1. No. of deletions from the cache
    1. Lookup in cache
    1. Found in cache (so one can determine cache hit rate)

There would (I assume) be a single instance of the codegen cache class, so no stats aggregation would be needed. 

2. Network module stats. First, my assumption would be that the system has multiple network threads, so aggregation of stats would be needed. Each network thread or network class instance might keep:
   1. No. of requests received
   1. No. of requests completed
   1. No. of JDBC requests
   1. No. of prepared statements handled
   1. Text mode / marshaling request
   1. Binary mode / marshaling request

A possible scenario might be that one has a test that uses prepared statements:
1.  Run the test
1.  Query for and dump out aggregate network stats. These display that we successfully execute the expected number of JDBC, prepared statement requests. All of them text mode.
1. Clear the stats. 
1. Modify the test and increase the number of prepared statement invocations. Run the test.
1. Dump network aggregate stats again. Expected result would be that now you see some number of text mode requests and some number of binary mode requests. 

[...]

PR #82, opened 2018-08-08
Add performance counters

ksaito7:
The performance counter is so simple version. It simply increments/decrements counters by inserting these functions into target places. As a preparation to use this, the class collecting counters needs to add a performance counter object as a member variable when the constructor is called. See`performance_counters_test.cpp` and `object_pool.h` as a use case.


ksaito7:
@pervazea Thanks your comments. This is just a primary pc implementation. Now I'm taking #47 into this. I will also change tbb map to atomic variable.

ksaito7:
Fixed according to the design in #47.
- Counters for a target class is defined as an object of stats in the class.  (`object_pool.h`)
- A collector can output all of collected counters as JSON. To collect counters, stats objects register itself within the collector. The collector collects the counters when a collection function of the collector is called or stats object is deleted. (`stats_collector.h` and `stats_test.cpp`)
- Stats class has performance counters with its update functions and a synchronization function with a collector. Each counter is defined as atom for multi-threading. So that the collector can collect some kind of stats objects, there is an abstract stats class and each stats class has to implement a synchronization function and a destructor. (`object_pool_stats.h` and `abstract_stats.h`)

tli2:
See #47 for closure

PR #88, opened 2018-08-09
Concurrency Control

tli2:
Implements all transaction semantics related to execution (begin, commit, abort, maintenance of versions, etc.)

Parts related to garbage collection (global table for running transactions) is not implemented.


tli2:
More tests will come in in a separate PR. We want to get this in since this modifies API, and hopefully before any other PRs come in relying on stale API.

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -33,22 +37,19 @@ static void BM_SimpleInsert(benchmark::State &state) {
   byte *redo_buffer = new byte[redo_size_];
   storage::ProjectedRow *redo = storage::ProjectedRow::InitializeProjectedRow(redo_buffer, all_col_ids_, layout);
   StorageTestUtil::PopulateRandomRow(redo, layout, 0, &generator);
-
-  // generate an undo DeltaRecord to populate on Insert
-  byte *undo_buffer = new byte[undo_size_];
-  storage::DeltaRecord *undo =
-      storage::DeltaRecord::InitializeDeltaRecord(undo_buffer, timestamp_t(0), layout, all_col_ids_);
+//  transaction::TransactionContext txn(timestamp_t(0), timestamp_t(0), &buffer_pool);
```
lmwnshn: Dead code? Maybe remove or leave a comment on why its there

tli2: Removed

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -85,15 +86,9 @@ static void BM_ConcurrentInsert(benchmark::State &state) {
   while (state.KeepRunning()) {
     storage::DataTable table(&block_store, layout);
     auto workload = [&](uint32_t id) {
-      // generate an undo DeltaRecord to populate on Insert
-      byte *undo_buffer = new byte[undo_size_];
-      storage::DeltaRecord *undo =
-          storage::DeltaRecord::InitializeDeltaRecord(undo_buffer, timestamp_t(0), layout, all_col_ids_);
-
+      transaction::TransactionContext txn(timestamp_t(0), timestamp_t(0), &buffer_pool);
```
lmwnshn: Maybe use txn_manager like with SimpleInsert, begin and commit

benchmark/storage/tuple_access_strategy_benchmark.cpp
```diff
@@ -43,6 +43,7 @@ static void BM_SimpleInsert(benchmark::State &state) {
   // We'll say it 2x RawBlock because we zero it, and then populate it. This is likely an underestimation
   size_t bytes_per_repeat = 2 * sizeof(storage::RawBlock);
   state.SetBytesProcessed(state.iterations() * bytes_per_repeat);
+  state.SetItemsProcessed(state.iterations() * layout.num_slots_);
```
lmwnshn: Maybe document why we're counting this

benchmark/storage/tuple_access_strategy_benchmark.cpp
```diff
@@ -84,6 +85,7 @@ static void BM_ConcurrentInsert(benchmark::State &state) {
...
```
lmwnshn: Copypaste documentation from above

benchmark/storage/data_table_benchmark.cpp
mbutrovich: Honestly we shouldn't even be using the txn_manager for SimpleInsert. That's overhead we shouldn't be measuring for a DataTable benchmark.

benchmark/storage/tuple_access_strategy_benchmark.cpp
tli2: Maybe in a separate PR?

src/include/common/object_pool.h
```diff
@@ -25,7 +26,7 @@ struct ByteAllocator {
    * Reuse a reused chunk of memory to be handed out again
    * @param reused memory location, possibly filled with junk bytes
    */
-  void Reuse(T *reused) { PELOTON_MEMSET(reused, 0, sizeof(T)); }
+  void Reuse(T *reused) {}
```
lmwnshn: More of a style thing, do we need this function?

tli2: Some one could perceivably want to do something with the reused pointer.

We don't have a use case for now.

src/include/storage/data_table.h
```diff
@@ -36,39 +37,36 @@ class DataTable {
...
    */
-  void Select(timestamp_t txn_start_time, TupleSlot slot, ProjectedRow *out_buffer) const;
+  void Select(transaction::TransactionContext *txn, TupleSlot slot, ProjectedRow *out_buffer) const;
 
   /**
    * Update the tuple according to the redo slot given, and update the version chain to link to the given
    * delta record. The delta record is populated with a before-image of the tuple in the process. Update will only
    * happen if there is no write-write conflict, otherwise, this is equivalent to a noop and false is returned,
    *
+   * @param txn the calling transaction
```
lmwnshn: Previously, we documented the undo record:
"the undo record to maintain and populate. It is expected that the projected row has the same structure as the redo, but the contents need not be filled beforehand, and will be populated with the before-image after this method returns."

We probably don't need the full details, but is it worth mentioning that the undo record is updated, or is it a no-brainer given the function name?

benchmark/storage/data_table_benchmark.cpp
lmwnshn: Maybe change that instead then.

src/include/storage/storage_defs.h
```diff
@@ -343,13 +363,24 @@ class DeltaRecord {
...
-  DeltaRecord *next_;
+  DeltaRecord *&Next() { return next_; }
+
+  /**
+   * @return Timestamp up to which the old projected row was visible.
+   */
+  std::atomic<timestamp_t> &Timestamp() { return timestamp_; }
+
+  /**
+   * @return the DatTable this DeltaRecord points to
```
lmwnshn: dat's a datatable

benchmark/storage/tuple_access_strategy_benchmark.cpp
lmwnshn: I get it now, should be fine. (that you're only telling gbench about another metric, not doing anything new)

src/include/storage/data_table.h
tli2: Since the entire update behavior is now invisible to the caller (they don't need to know what happens in side transaction context), I think it's okay we leave that out.

src/include/transaction/transaction_context.h
```diff
@@ -0,0 +1,251 @@
...
+  TransactionContext(timestamp_t start, timestamp_t txn_id, common::ObjectPool<UndoBufferSegment> *buffer_pool)
+      : start_time_(start), txn_id_(txn_id), undo_buffer_(buffer_pool) {}
+
+  /**
+   * @return start time of this transaction
+   */
+  timestamp_t StartTime() const { return start_time_; }
+
+  /**
+   * @return id of this tranaction
```
lmwnshn: transaction

PR #92, opened 2018-08-13
Debug logging (spdlog)

pervazea:
Initial PR, to see what Travis and CI report.


lmwnshn:
clang-tidy should be ignoring the third_party folder, but it isn't. This is causing Travis to fail.

It looks like HeaderFilterRegex in the .clang-tidy file is what we should be modifying, but I have not been able to appease it yet. I'll keep trying.

lmwnshn:
#95 should fix third_party CI issues.

mbutrovich:
Feedback I gave in person but don't want to forget: I think the logger instantiation methods are small enough that we can put their definitions in the headers. The .cpp files are pretty sparse. As far as actually instantiating these, I think we'll need to have them in a Google Test harness so we can log in Google Test. Instantiating loggers in the main() function doesn't help us for Google Tests. We'll have to verify that the proposed harness solution works.

pervazea:
To summarize most recent changes:

- Log to console (rather than a file)
- Add class for the tests to use, that initializes the logger(s)
- Add wiki page with short description of how to use and how to add modules

mbutrovich:
Wiki page is here for reference:

https://github.com/cmu-db/terrier/wiki/Using-debug-logging

mbutrovich:
How much did we end up changing the spdlog code (if any) from the master repo? Is it easy to summarize those changes in case we want to bring in a newer version in the future?

test/include/util/test_harness.h
```diff
@@ -0,0 +1,30 @@
...
+// Created by pakhtar on 8/15/18.
+//
+
+#pragma once
+#include "gtest/gtest.h"
+#include "loggers/main_logger.h"
+#include "loggers/storage_logger.h"
+
+namespace terrier {
+namespace test {
```
mbutrovich: I think we're just doing the terrier namespace for the test subfolder, rather than terrier::test.

src/loggers/main_logger.cpp
```diff
@@ -0,0 +1,18 @@
+//
```
mbutrovich: We can ditch these comments at the top.

pervazea:
The changes to spdlog were small in number and localized, so should be easy to bring in new versions.

PR #97, opened 2018-08-14
cmake cleanup

lmwnshn:
1. Remove infer as a make target as we don't foresee ourselves using it.
2. Rename lint target to check-lint for consistency.


lmwnshn:
Note: coveralls is not reporting status properly, hence we could not merge earlier.

PR #98, opened 2018-08-14
Large transaction tests and associated DataTable and CC bug fixes

### LINKED ISSUE #112: Transaction layer randomized testing
Large, randomized multithreaded testing for the transaction layer to ensure correctness

tli2:
fix #112


src/include/storage/storage_defs.h
```diff
@@ -220,7 +220,7 @@ using BlockStore = common::ObjectPool<RawBlock, BlockAllocator>;
  * --------------------------------------------------------
  * Would be the row: { 0 -> 15, 1 -> 721, 2 -> nul}
  */
-class ProjectedRow {
+class PACKED ProjectedRow {
```
mbutrovich: Add a comment why this is necessary.

src/include/transaction/transaction_manager.h
```diff
@@ -33,12 +33,13 @@ class TransactionManager {
    * Commits a transaction, making all of its changes visible to others.
    * @param txn the transaction to commit
    */
-  void Commit(TransactionContext *txn) {
+  timestamp_t Commit(TransactionContext *txn) {
     tbb::reader_writer_lock::scoped_lock guard(commit_latch_);
```
mbutrovich: Change to the new wrapped scoped rw_latch?

src/storage/data_table.cpp
```diff
@@ -21,16 +20,21 @@ void DataTable::Select(transaction::TransactionContext *txn,
...
+    for (uint16_t i = 0; i < out_buffer->NumColumns(); i++)
+      StorageUtil::CopyAttrIntoProjection(accessor_, slot, out_buffer, i);
+  } while (version_ptr != AtomicallyReadVersionPtr(slot, accessor_));
+
 
   // TODO(Tianyu): Potentially we need a memory fence here to make sure the check on version ptr
   // happens after the row is populated. For now, the compiler should not be smart (or rebellious)
   // enough to reorder this operation in or in front of the for loop.
-  DeltaRecord *version_ptr = AtomicallyReadVersionPtr(slot, accessor_);
+//  DeltaRecord *version_ptr = AtomicallyReadVersionPtr(slot, accessor_);
```
mbutrovich: Dead code.  I think we can remove the TODO above as well (re: memory fence).

src/storage/data_table.cpp
```diff
@@ -101,8 +105,7 @@ TupleSlot DataTable::Insert(transaction::TransactionContext *txn,
   DeltaRecord *undo = txn->UndoRecordForInsert(this, accessor_.GetBlockLayout(), result);
 
   // Populate undo record with the before image of presence column
-  // TODO(Tianyu): This is projection list id, not col id, fix eventually
-  undo->Delta()->SetNull(VERSION_VECTOR_COLUMN_ID);
+  undo->Delta()->SetNull(0);
```
mbutrovich: Are we not using the VERSION_VECTOR_COLUMN_ID def anymore?

src/storage/data_table.cpp
```diff
@@ -120,12 +123,12 @@ TupleSlot DataTable::Insert(transaction::TransactionContext *txn,
 void DataTable::Rollback(timestamp_t txn_id, terrier::storage::TupleSlot slot) {
   DeltaRecord *version_ptr = AtomicallyReadVersionPtr(slot, accessor_);
   // We do not hold the lock. Should just return
-  if (version_ptr->Timestamp().load() != txn_id) return;
+  if (version_ptr == nullptr || version_ptr->Timestamp().load() != txn_id) return;
```
mbutrovich: Nice catch.

src/storage/data_table.cpp
```diff
@@ -143,7 +146,7 @@ void DataTable::AtomicallyWriteVersionPtr(const TupleSlot slot,
                                           DeltaRecord *desired) {
   byte *ptr_location = accessor.AccessWithNullCheck(slot, VERSION_VECTOR_COLUMN_ID);
   PELOTON_ASSERT(ptr_location != nullptr, "Only write version vectors for tuples that are present.");
-  reinterpret_cast<std::atomic<DeltaRecord *> *>(ptr_location)->store(desired);
+  reinterpret_cast<std::atomic<DeltaRecord *> *>(ptr_location)->store(desired, std::memory_order::memory_order_release);
```
mbutrovich: Do we know if we still need this memory_order after the loop we added to Select? Maybe we remove it?

test/transaction/large_transaction_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+#include "storage/data_table.h"
+#include "transaction/transaction_context.h"
+#include "transaction/transaction_manager.h"
+#include "common/container/concurrent_vector.h"
+#include "util/storage_test_util.h"
+#include "gtest/gtest.h"
+
+namespace terrier {
+
+class RandomWorkloadTransaction {
```
mbutrovich: Is any of this suitable for the storage_test_util file?

test/transaction/large_transaction_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+class LargeTransactionTests : public ::testing::Test {
+ public:
+  storage::BlockStore block_store_{1000};
+  common::ObjectPool<transaction::UndoBufferSegment> buffer_pool_{1000};
+  std::default_random_engine generator_;
+};
+
+// This test case generates random update-selects in concurrent transactions on a pre-populated database.
+// Each transaction logs their operations locally. At the end of the run, we can reconstruct the snapshot of
+// a database using the updates at every timstamp, and compares the reads with the reconstructed snapshot versions
```
mbutrovich: typo: timestamp

test/transaction/large_transaction_test.cpp
```diff
@@ -0,0 +1,286 @@
...
+// to make sure they are the same.
+// NOLINTNEXTLINE
+TEST_F(LargeTransactionTests, MixedReadWrite) {
```
mbutrovich: We'll want to add more test cases (update/select ratio, tuples, etc.) but this is good for now. That can be done next week during the cleanup.

src/storage/data_table.cpp
tli2: As commented, this is the 0th index in the projection list id, which we assume to have col_id of 1.

This is confusing as hell. We will need to standardize these magic constants and document these assumptions later in the clean up.

src/storage/data_table.cpp
tli2: We do not. I forgot. Nice catch.

tli2:
Summary of the bugs fixed in this PR in case any of them ever reproduced:

1. If txn1 aborted because of txn2 held a write lock, and txn2 aborted also but before txn1 is rolled back, the rollback logic can potentially dereference a nullptr.
2. Select could read an in-place version that is being rolled back, where the read of tuple happens before the roll back but the read of the version chain happens after the roll back is complete, resulting in read-uncommitted.
3. Concurrent atomic read/write on unaligned addresses apparently could result in a partial read, albeit very rarely.

It's hard to write regression tests for these as they happen only rarely under certain workloads.

src/include/transaction/transaction_context.h
```diff
@@ -240,6 +242,7 @@ class TransactionContext {
     // Pretty sure we want 1, the primary key column?
     uint32_t size = storage::DeltaRecord::Size(layout, {1});
     storage::DeltaRecord *result = undo_buffer_.NewEntry(size);
+    //    PELOTON_ASSERT(reinterpret_cast<uintptr_t>(result) % 8 == 0, "not aligned");
```
mbutrovich: Did you mean to remove this or uncomment this?

src/include/transaction/transaction_manager.h
```diff
@@ -55,10 +57,9 @@ class TransactionManager {
   common::ObjectPool<UndoBufferSegment> *buffer_pool_;
   // TODO(Tianyu): Timestamp generation needs to be more efficient
   std::atomic<timestamp_t> time_{timestamp_t(0)};
-  std::atomic<timestamp_t> txn_id_{timestamp_t(static_cast<uint64_t>(INT64_MIN))};  // start from "negative" value
+  //  std::atomic<timestamp_t> txn_id_{timestamp_t(static_cast<uint64_t>(INT64_MIN))};  // start from "negative" value
```
mbutrovich: This needs to be uncommented, doesn't it?

src/storage/data_table.cpp
```diff
@@ -21,16 +20,21 @@ void DataTable::Select(transaction::TransactionContext *txn,
...
+    for (uint16_t i = 0; i < out_buffer->NumColumns(); i++)
+      StorageUtil::CopyAttrIntoProjection(accessor_, slot, out_buffer, i);
+  } while (version_ptr != AtomicallyReadVersionPtr(slot, accessor_));
+
 
   // TODO(Tianyu): Potentially we need a memory fence here to make sure the check on version ptr
   // happens after the row is populated. For now, the compiler should not be smart (or rebellious)
   // enough to reorder this operation in or in front of the for loop.
-  DeltaRecord *version_ptr = AtomicallyReadVersionPtr(slot, accessor_);
+  //  DeltaRecord *version_ptr = AtomicallyReadVersionPtr(slot, accessor_);
```
mbutrovich: Dead code.

src/storage/data_table.cpp
```diff
@@ -70,11 +74,9 @@ bool DataTable::Update(transaction::TransactionContext *txn,
   // Update the next pointer of the new head of the version chain
   undo->Next() = version_ptr;
 
-  // TODO(Tianyu): Is it conceivable that the caller would have already obtained the values and don't need this?
-  // Populate undo record with the before image of attribute
-  for (uint16_t i = 0; i < redo.NumColumns(); i++)
+  // Store before-image before making any changes or grabbing lock
+  for (uint16_t i = 0; i < undo->Delta()->NumColumns(); i++)
```
mbutrovich: Good change for readability, even if the condition is equivalent.

PR #100, opened 2018-08-14
Upgrade docker images on Travis.

lmwnshn:
I don't know if we particularly care about Docker's version.


lmwnshn:
Yeah, we don't care.

PR #103, opened 2018-08-15
Add tests for ProjectedRow, DeltaRecord and StorageUtil

yangjuns:


mbutrovich:
You'll have to rebase with @pervazea's changes today. It shouldn't be too bad from a conflict standpoint. Any test fixtures you modified or added will probably just need to inherit from the new test harness.

test/storage/delta_record_test.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include "common/object_pool.h"
+#include "storage/data_table.h"
+#include "storage/storage_util.h"
+#include "util/storage_test_util.h"
+#include "storage/storage_defs.h"
+#include "util/test_harness.h"
+
+namespace terrier {
+
+struct DeltaRecordTests : public ::terrier::TerrierTest {
```
tli2: nit: you are already under terrier namespace so `struct DeltaRecordTests : public TerrierTest` would suffice

test/storage/projected_row_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+#include "common/object_pool.h"
+#include "storage/data_table.h"
+#include "storage/storage_util.h"
+#include "util/storage_test_util.h"
+#include "storage/storage_defs.h"
+#include "util/test_harness.h"
+
+namespace terrier {
+
+struct ProjectedRowTests : public ::terrier::TerrierTest {
```
tli2: same nit

test/storage/projected_row_test.cpp
```diff
@@ -0,0 +1,110 @@
...
+    void *lower_bound = update->ColumnIds();
+    void *upper_bound = update + update->Size();
+    // check the first value is in bound
+    StorageTestUtil::CheckInBounds(update->AccessForceNotNull(0), lower_bound, upper_bound);
+    // check the rest values memory addresses don't overlapping previous addresses.
+    for (uint16_t col = 1; col < update->NumColumns(); col++) {
+      lower_bound = update->AccessForceNotNull(static_cast<uint16_t >(col - 1));
+      upper_bound = StorageTestUtil::IncrementByBytes(update->AccessForceNotNull(static_cast<uint16_t >(col - 1)),
+                                                      layout.attr_sizes_[col]);
+      // check if the value address is in bound
```
tli2: not in bound?

The use of CheckNotInBounds is probably correct judging by your comment outside the loop, but maybe it's most straightforward to instead advance the lower bound incrementally and use CheckInBounds against a constant upper bound.

test/storage/storage_util_test.cpp
```diff
@@ -0,0 +1,211 @@
...
+#include "common/object_pool.h"
+#include "storage/data_table.h"
+#include "storage/storage_util.h"
+#include "util/storage_test_util.h"
+#include "storage/storage_defs.h"
+#include "util/test_harness.h"
+
+namespace terrier {
+
+struct StorageUtilTests : public ::terrier::TerrierTest {
```
tli2: same nit

test/storage/projected_row_test.cpp
yangjuns: I saw that was what you did for TAS test. My concern was that if you fix the upper bound to be the very end, then the bound is too big and the address can still be incorrect even if it's in bound.  
On second thought, I think your way of checking bound is still correct because of some sort of "induction" reason. 

Correct me if I'm wrong.

yangjuns: What I'm doing here is to,  for each column value,  identify the memory space used by the previous value, which is [lower_bound, upper_bound) and make sure that the current value's memory address don't start within this range.

yangjuns: Never mind. I agree that advancing the lower bound and keeping upper bound constant serve the same purpose and are more straightforward.

yangjuns:
Quick Summary: This PR

- added 3 test files for testing ProjectedRow, DeltaRecord and StorageUtil.
- removed redundant namespace in `tuple_access_strategy_test.cpp`, `data_table_test.cpp`, `data_table_concurrent_test.cpp` as pointed out by @tli2

PR #107, opened 2018-08-16
Performance counters

### LINKED ISSUE #47: Performance counters and stats

lmwnshn:
Fix #47.

An alternative implementation of performance counters as discussed.

**Summary:**
- PerformanceCounters are really fast (single class member access) and have boilerplate generated by macros.
- StatisticsRegistry supports arbitrary module nesting and uses string names to identify different performance counters. It was not designed for speed, under the assumption that you're not calling it often (if you ask it for a pointer to a performance counter, keep it around somewhere, otherwise I mainly see it being used interactively).

**Per requirements,**
1. cost of keeping stats: once you get to the performance counter, every operation is a single class member access to an atomic integral type. **Unfortunately, this cannot be made constexpr because gcc is stupid about those until gcc 9.**
2. compiled out: if compiled out, all the methods are replaced with (void)0. Except for GetAttribute() which has to return an int, so it returns 0 instead.
3. human friendly: every performance counter can be trivially serialized to/from JSON.
4. determining available stats: the stats registry is essentially a JSON object that you can use on the fly, e.g. GetRegistryListing({"Network", "JDBC"}) would return a list of everything registered under Network/JDBC. Modules can be arbitrarily nested.
5. dumping stats: supported, dumps out to a JSON string
6. aggregation: as the stats registry speaks primarily JSON, we can add/subtract corresponding fields together before returning a dump if desired.

**Main changes:**
1. Use [nlohmann::json](https://github.com/nlohmann/json) instead of [libjsoncpp](https://github.com/open-source-parsers/jsoncpp). In [benchmarks](https://github.com/miloyip/nativejson-benchmark): [parsing] it is 10% more conformant, twice as fast, and uses 2.5 times less memory; [stringifies] at the same speed; [executable size] compiles to 6% less code. Most importantly, it is actively maintained, has a ton of testing, it is simple to distribute (single header) and has a very friendly API (almost Python dicts), which makes it easy to interact with.

2. Performance counters are generated by macros. Each counter is a simple class member, so the cost of incrementing/decrementing/fetching is just a class member access. 

Example definition:
```cpp
#define CACHE_MEMBERS(f) \
  f(uint64_t, num_insert) \
  f(uint32_t, num_hit) \
  f(uint16_t, num_failure) \
  f(uint8_t, num_user)

MAKE_PERFORMANCE_COUNTER(CacheCounter, CACHE_MEMBERS)
```
This generates a CacheCounter class with those four attributes and all the methods e.g. ZeroCounters(), ToJson(), FromJson(), etc. automagically.

3. The statistics registry is responsible for aggregation. It stores strings for module names and pointers (as integers) to PerformanceCounter objects. When stats are dumped, it follows the pointers to get the actual PerformanceCounters.

```cpp
CacheCounter cc;
CacheCounter cc2;
StatisticsRegistry reg;
reg.register({}, cc); // registers "CacheCounter" at root level
reg.register({"Cache"}, cc); // registers "CacheCounter" at "Cache" level
reg.register({"Cache", "Local"}, cc); // registers "CacheCounter" at "Cache/Local" level
reg.DumpStats({"Cache"}, 4); // dumps the stats rooted at "Cache" with 4 spaces for indents
```

Whether you want to store the performance counter in the object itself or elsewhere is left up to you. 

**Todo:**
1. Add an example usage for DataTable - this requires storing the stat_registry somewhere.
2. Mock up the scenario described in the issue - should essentially be adapting the tests.
3. Think about same-counter-level aggregation more. When do we decide what we want to aggregate together?
    - e.g. four separate NetworkCounters that got added and automatically assigned to NetworkCounter, NetworkCounter1, NetworkCounter2, NetworkCounter3)
    - when should we return the aggregate vs returning them separately? I don't think it makes sense to always sum, but that means we would need a per-member decision...

I just rebased, grabbing food for now.


test/common/performance_counter_test.cpp
```diff
@@ -0,0 +1,136 @@
...
+   */
+  void Equal() const {
+    EXPECT_EQ(cc->Get_num_insert(), num_insert);
+    EXPECT_EQ(cc->Get_num_hit(), num_hit);
+    EXPECT_EQ(cc->Get_num_failure(), num_failure);
+    EXPECT_EQ(cc->Get_num_user(), num_user);
+  }
+
+  std::vector<double> work_probs = std::vector<double>(8, 0.125);
+  std::vector<std::function<void()>> workloads = {
```
tli2: auto should be fine here?

test/common/performance_counter_test.cpp
```diff
@@ -0,0 +1,136 @@
...
+   * Runs random operations, automatically checking that all the state remains consistent.
+   * @param generator random number generator
+   * @param num_operations number of operations to run
+   */
+  void RandomOperation(std::default_random_engine generator, uint32_t num_operations) {
+    MultiThreadedTestUtil::InvokeWorkloadWithDistribution(workloads, work_probs, &generator, num_operations);
+  }
+};
+
+#ifdef NDEBUG
```
tli2: Gtest has a DISABLED flag, though I don't remember clearly how that works.

We might want to put a macro for this since it is perceivable we want this kind of behavior for other tests as well.

test/include/util/performance_counter_test_util.h
```diff
@@ -0,0 +1,23 @@
+#pragma once
+#include "common/performance_counter.h"
+
+namespace terrier {
```
tli2: https://github.com/cmu-db/terrier/wiki/Coding-Guideline

We basically set a rule that only a directory gets its own test util.

src/include/common/performance_counter.h
```diff
@@ -0,0 +1,188 @@
...
+ * You can use the resulting PerformanceCounter as a class in its own right. For example,
+ *      #define MAKE_PERFORMANCE_COUNTER(NetworkCounter, NETWORK_MEMBERS)
+ * will make the following code valid:
+ *      NetworkCounter nc;
+ *      nc.requests_received++;
+ *      nc.connections_opened.store(100);
+ *
+ * Note that every class member is wrapped in std::atomic.
+ */
+#define MAKE_PERFORMANCE_COUNTER(ClassName, MemberList)                               \
```
pervazea: Name could use adjustment.   MAKE should probably be DEFINE (as in declaration vs. definition)
COUNTER should be COUNTERS, since the argument is a list.

test/common/performance_counter_test.cpp
lmwnshn: error: 'auto' not allowed in non-static class member

test/common/performance_counter_test.cpp
lmwnshn: It appears the GTest way is to do something like
TEST(PerformanceCounterTests, DISABLED_SerializationTest) {

lmwnshn:
I have added an example usage for datatable in main/terrier.cpp. I think we might want to shake out the API a little before we actually merge it, but the overall implementation of this approach should be there.

src/storage/data_table.cpp
```diff
@@ -7,6 +9,8 @@
 
 namespace terrier::storage {
 DataTable::DataTable(BlockStore *store, const BlockLayout &layout) : block_store_(store), accessor_(layout) {
+  data_table_counter_ = new DataTableCounter;
```
tli2: Why not inline these counters? Do we support inlining these counters in the class themselves?

src/storage/data_table.cpp
```diff
@@ -80,6 +85,7 @@ bool DataTable::Update(transaction::TransactionContext *txn,
   // Update in place with the new value.
   for (uint16_t i = 0; i < redo.NumColumns(); i++) StorageUtil::CopyAttrFromProjection(accessor_, slot, redo, i);
 
+  data_table_counter_->Inc_num_update();
```
tli2: nit: CamelCase?

tli2:
Overall looks good

src/include/common/stat_registry.h
```diff
@@ -0,0 +1,227 @@
...
+
+  /**
+   * Deregister and remove the named performance counter from the module specified.
+   *
+   * @param module_path path to destination module
+   * @param name name of the performance counter to be deregistered
+   * @param free_pc true if the performance counter should be freed, false otherwise
+   * @return true if the performance counter was successfully removed, false otherwise
+   */
+  bool Deregister(const std::vector<std::string> &module_path, const std::string &name, bool free_pc) {
```
tli2: Do we always want these to go out of scope, or maybe retain them if needed?

lmwnshn: Retain the removed performance counters? They remain in whatever object they were in, they will just not be tracked by the registry any more.

lmwnshn:
I suspect yesterday's CMake fixes may have borked linking when externs are involved. PR on hold while I sort it out.

yangjuns:
I get core dump when running terrier

src/storage/data_table.cpp
```diff
@@ -43,6 +45,7 @@ void DataTable::Select(transaction::TransactionContext *const txn, const TupleSl
     StorageUtil::ApplyDelta(accessor_.GetBlockLayout(), *(version_ptr->Delta()), out_buffer);
     version_ptr = version_ptr->Next();
   }
+  data_table_counter_.IncNumSelect();
```
tli2: nit: don't use abbreviation.

Second thought: does it make sense to add something not 1 in a call to Increment? Maybe the right way is to take in an `increment_by` argument that defaults to 1 (don't tell Andy)

mbutrovich: Don't do the default argument :)

mbutrovich:
What’s the easiest way to compile it out?

lmwnshn:
I just remembered why we had increment and decrement. This is so it would be compiled out as do-nothing.

mbutrovich:
Is it easy to change the DEBUG flag as being the setting? We’ll want some of these in Release mode since they’ll be used as metrics.

lmwnshn:
The flag is set here https://github.com/cmu-db/terrier/blob/master/src/include/common/performance_counter.h#L43 for all performance counters.

Currently no support for only some performance counters, I believe it could be done though.

PR #115, opened 2018-08-17
Garbage Collection

### LINKED ISSUE #110: Garbage Collection
A simple garbage collection that runs on its own thread and consumes the out-of-scope transaction contexts form transaction manager should suffice at this point. Correctness first, optimization later.

mbutrovich:
Fix #110.

This is a simple garbage collector for the storage layer that will unlink obsolete versions from version chains and then free TransactionContexts once no one else can hold a reference to them.

There are also a few API changes to support GC, as well as minor bug fixes.

### DIFF
--- a/src/include/storage/data_table.h
+++ b/src/include/storage/data_table.h
@@ -74,6 +74,8 @@ class DataTable {
   void Rollback(timestamp_t txn_id, TupleSlot slot);
 
  private:
+  friend class GarbageCollector;
+
   BlockStore *block_store_;
   // TODO(Tianyu): this is here for when we support concurrent schema, for now we only have one per DataTable
   // common::ConcurrentMap<layout_version_t, TupleAccessStrategy> layouts_;
--- a/src/include/storage/garbage_collector.h
+++ b/src/include/storage/garbage_collector.h
@@ -12,12 +12,34 @@
 
 namespace terrier::storage {
 
+/**
+ * The garbage collector is responsible for processing a queue of completed transactions from the transaction manager.
+ * Based on the contents of this queue, it unlinks the DeltaRecords from their version chains when no running
+ * transactions can view those versions anymore. It then stores those transactions to attempt to deallocate on the next
+ * iteration if no running transactions can still hold references to them.
+ */
 class GarbageCollector {
  public:
   GarbageCollector() = delete;
-  explicit GarbageCollector(transaction::TransactionManager *txn_manager) : txn_manager_(txn_manager), last_run_{0} {}
+  /**
+   * Constructor for the Garbage Collector that requires a pointer to the TransactionManager. This is necessary for the
+   * GC to invoke the TM's function for handing off the completed transactions queue.
+   * @param txn_manager pointer to the TransactionManager
+   */
+  explicit GarbageCollector(transaction::TransactionManager *txn_manager) : txn_manager_(txn_manager), last_run_{0} {
+    PELOTON_ASSERT(txn_manager_->GCEnabled(),
+                   "The TransactionManager needs to be instantiated with gc_enabled true for GC to work!");
+  }
   ~GarbageCollector() = default;
 
+  /**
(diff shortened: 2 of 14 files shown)


mbutrovich:
I’ll fix the coveralls number by initializing the txn logger in main. I didn’t end up using it after adding it.

src/include/storage/data_table.h
```diff
@@ -74,6 +74,8 @@ class DataTable {
   void Rollback(timestamp_t txn_id, TupleSlot slot);
 
  private:
+  friend class GarbageCollector;
```
tli2: No need to fix it in this PR, but we should be more consistent with this and the rollback function from transactions (whether to write a function in DataTable or make friend class)

src/include/storage/garbage_collector.h
```diff
@@ -0,0 +1,180 @@
...
+
+/**
+ * The garbage collector is responsible for processing a queue of completed transactions from the transaction manager.
+ * Based on the contents of this queue, it unlinks the DeltaRecords from their version chains when no running
+ * transactions can view those versions anymore. It then stores those transactions to attempt to deallocate on the next
+ * iteration if no running transactions can still hold references to them.
+ */
+class GarbageCollector {
+ public:
+  GarbageCollector() = delete;
```
tli2: don't need this delete, should be implicitly deleted after you declared a non-default constructor.

src/include/storage/garbage_collector.h
```diff
@@ -0,0 +1,180 @@
...
+  /**
+   * Constructor for the Garbage Collector that requires a pointer to the TransactionManager. This is necessary for the
+   * GC to invoke the TM's function for handing off the completed transactions queue.
+   * @param txn_manager pointer to the TransactionManager
+   */
+  explicit GarbageCollector(transaction::TransactionManager *txn_manager) : txn_manager_(txn_manager), last_run_{0} {
+    PELOTON_ASSERT(txn_manager_->GCEnabled(),
+                   "The TransactionManager needs to be instantiated with gc_enabled true for GC to work!");
+  }
+  ~GarbageCollector() = default;
```
tli2: Also okay to omit.

src/include/storage/garbage_collector.h
```diff
@@ -0,0 +1,180 @@
...
+class GarbageCollector {
+ public:
+  GarbageCollector() = delete;
+  /**
+   * Constructor for the Garbage Collector that requires a pointer to the TransactionManager. This is necessary for the
+   * GC to invoke the TM's function for handing off the completed transactions queue.
+   * @param txn_manager pointer to the TransactionManager
+   */
+  explicit GarbageCollector(transaction::TransactionManager *txn_manager) : txn_manager_(txn_manager), last_run_{0} {
+    PELOTON_ASSERT(txn_manager_->GCEnabled(),
```
tli2: Yes

src/include/storage/garbage_collector.h
```diff
@@ -0,0 +1,180 @@
...
+    if (!txns_to_unlink_.empty()) {
+      // Append to our non-empty unlink queue
+      while (!from_txn_manager.empty()) {
+        txn = from_txn_manager.front();
+        from_txn_manager.pop();
+        txns_to_unlink_.push(txn);
+      }
+    } else {
+      // Overwrite our empty unlink queue
+      txns_to_unlink_ = from_txn_manager;
```
tli2: nit: This probably copies anyway, so this if branch can probably go away. (simply iterate through the `from_txn_manager` and add them to `txns_to_unlink`)

src/include/storage/data_table.h
mbutrovich: Friend class for GC definitely makes sense since it’s in the same namespace and abstraction layer and we don’t want to expose anything extra from the DT that GC might need to other layers.

It gets weird with CC. Friend class might make sense there too if we’re trying encapsulate everything in the backend (storage and CC) from the ref of the system — I.e. not expose something like Rollback to the execution layer.

PR #116, opened 2018-08-18
LargeTransactionTest framework refactor + large randomized GC tests

tli2:
The framework we used to test concurrency control is adpated for GC tests as well.

Caveat: It is too expensive to test GC for a long period in the current framework because of the extra book-keeping we do for correctness checks. We don't have to check correctness only at the end though, so I could technically put in a producer-consumer model and fix that.

Will probably fix that in a different PR. When that is done, this thing can in theory run forever.


src/include/transaction/transaction_manager.h
```diff
@@ -61,9 +61,7 @@ class TransactionManager {
     size_t result UNUSED_ATTRIBUTE = curr_running_txns_.erase(start_time);
     PELOTON_ASSERT(result == 1, "committed transaction did not exist in global transactions table");
     txn->TxnId() = commit_time;
-    if (gc_enabled_ && !undos.Empty()) {
-      completed_txns_.push(txn);
-    }
+    if (gc_enabled_) completed_txns_.push(txn);
```
mbutrovich: Nice catch! leak check to the rescue.

src/storage/garbage_collector.cpp
```diff
@@ -45,45 +45,44 @@ uint32_t GarbageCollector::Unlink() {
...
-
   // Append to our local unlink queue
   while (!from_txn_manager.empty()) {
     txn = from_txn_manager.front();
     from_txn_manager.pop();
     txns_to_unlink_.push(txn);
   }
 
-  uint32_t txns_cleared = 0;
+  uint32_t txns_unlinked = 0;
```
mbutrovich: Good rename

src/util/transaction_test_util.cpp
```diff
@@ -0,0 +1,265 @@
...
+
+RandomWorkloadTransaction::~RandomWorkloadTransaction() {
+  if (!test_object_->gc_on_) delete txn_;
+  if (test_object_->bookkeeping_) delete[] buffer_;
+  for (auto &entry : updates_)
+    delete[] reinterpret_cast<byte *>(entry.second);
+  for (auto &entry : selects_)
+    delete[] reinterpret_cast<byte *>(entry.second);
+}
+template<class Random>
```
mbutrovich: It's not clear to me why this works in a .cpp without explicit instantiation but sure!

Issue #105, opened 2018-08-16
Cleanup week!


lmwnshn:
It's been a month of rapid engineering by a very small team, and now we need a few days to reflect on what we've accomplished and get the codebase ready for a lot more people to work on it. To that end, we're going to complete the following next week:

- [x] Arrow's codebase assumed .cc files. A quick search shows that we should change cpplint and run_clang_format. Previously, it was only doing that for .h files. We can also try to make cpplint.py work for Python 3. We’ll do this at the end of the week to avoid merge conflicts on other PRs. (PR is open #117, ready to go whenever we are).
- [x] Revisit every TODO in the codebase. We sprinkled these liberally through the codebase when we didn't know the answer to questions, and some are now irrelevant.
- [x] Revisit code documentation to make sure it's still correct. APIs, behavior, and assumptions might have changed since first written. Some hidden assumptions will also need to be made explicit and clearly documented. Suggest turning on -Wdocumentation locally on Clang for extra help finding issues.
- [x] Add debug logging statements where appropriate.
- [x] Break some class definitions into their own files. Closely related classes were often written in the same headers since that was easier for engineering, but we should make the files more accessible now.
- [x] Make sure definitions are in .cpp files rather than .h files when appropriate. Again, this was easier for engineering but we want to make sure we're following our own guidelines.
- [ ] Revisit and expand the wiki documentation. Design decisions in the storage and CC layers have been in flux for a month, but we've more or less settled on our design and should document them (issue #46). Some infrastructure stuff has also changed (lint target, CI, etc.) and we want to make sure that's current.
- [ ] Get our Continuous Integration (both Travis and Jenkins) running the checks that we want. This mostly involves shifting work to Jenkins and getting the dependency in place for Jenkins to trigger after Travis.
- [x] Make unit test and benchmark parameters consistent, in particular number of threads, and put these in a common harness.
- [x] Magic numbers, names, and other readability improvements.
- [ ] Get a benchmark machine set up to collect nightly numbers so we can spot regressions.
[...]

PR #117, opened 2018-08-20
Fix clang-format for .cpp files

lmwnshn:
Part of #105. 

Comments:
1. There is a reference to .cc files in [asan_symbolize.py](https://github.com/cmu-db/terrier/blob/master/build-support/asan_symbolize.py#L27), but I'm not sure if it should be changed.

2. I tried making cpplint Python-3 compatible. It mostly involves removing references to unicode, iteritems, itervalues. But I do not think this is worth doing; most systems will have Python 2 as the default env python. At most a mild inconvenience for people like me who have a Python 3 default.

Whenever we're ready, we can rebase and run make format.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -219,7 +219,7 @@ endif ()
 message(STATUS "CMAKE_CXX_FLAGS: ${CMAKE_CXX_FLAGS}")
 
 ###########################################################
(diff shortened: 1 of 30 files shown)


mbutrovich:
Add the do-not-merge tag until the end of the cleanup week.

PR #119, opened 2018-08-20
Add thread pool to Google Tests to avoid constant creation and reaping of threads

### LINKED ISSUE #102: Change MultiThreadedTestUtil to use a thread pool
Some funky thing is going on where c++ threads are not garbage-collected after joined in long running tests, resulting in slow but noticeable "leak".

This is not an issue for unit tests. This is an issue if we want to scale up the same tests to do stress/soak/fuzz nightly. We should reuse threads with a pool inside the test util class. (or use a harness)

lmwnshn:
Fix #102.

To use RunUntilThreadsFinish, we now need to construct an instance of MultiThreadedTestUtil. The tests have been fixed up accordingly. All other functionality remains the same.

Not implemented: shutting down a subset of threads. Would we want this?

### DIFF
--- a/benchmark/storage/data_table_benchmark.cpp
+++ b/benchmark/storage/data_table_benchmark.cpp
@@ -71,6 +71,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, SimpleInsert)(benchmark::State &state) {
 // Insert the num_inserts_ of tuples into a DataTable concurrently
 // NOLINTNEXTLINE
 BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentInsert)(benchmark::State &state) {
+  MultiThreadedTestUtil mtt_util;
   // NOLINTNEXTLINE
   for (auto _ : state) {
     storage::DataTable table(&block_store_, layout_);
@@ -80,7 +81,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentInsert)(benchmark::State &state
       for (uint32_t i = 0; i < num_inserts_ / num_threads_; i++)
         table.Insert(&txn, *redo_);
     };
-    MultiThreadedTestUtil::RunThreadsUntilFinish(num_threads_, workload);
+    mtt_util.RunThreadsUntilFinish(num_threads_, workload);
   }
 
   state.SetItemsProcessed(state.iterations() * num_inserts_);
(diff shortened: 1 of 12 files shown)


mbutrovich:
Good stuff.

lmwnshn:
TBB does have a task_group mechanism. If we are okay with [up to N threads running] instead of [guarantee N threads running], and the scheduler being bound per thread, their task_group is pretty easy to use.

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -71,6 +71,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, SimpleInsert)(benchmark::State &state) {
 // Insert the num_inserts_ of tuples into a DataTable concurrently
 // NOLINTNEXTLINE
 BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentInsert)(benchmark::State &state) {
+  MultiThreadedTestUtil mtt_util;
```
tli2: Since it is no longer stateless, maybe rename it to `MultiThreadedTestHarness`

test/include/util/multi_threaded_test_util.h
```diff
@@ -1,18 +1,68 @@
...
 #include "common/object_pool.h"
 #include "gtest/gtest.h"
+#include "tbb/task_group.h"
+#include "tbb/task_scheduler_init.h"
 
 namespace terrier {
 /**
  * Static utility class for common code for multi-threaded tests.
  */
 struct MultiThreadedTestUtil {
```
tli2: Change to class?

test/include/util/multi_threaded_test_util.h
```diff
@@ -1,18 +1,68 @@
...
+
+  void AddThread() {
+    thread_pool_.emplace_back([this] {
+      // keep the thread alive
+      while (true) {
+        // grab the lock
+        std::unique_lock<std::mutex> lock(work_lock_);
+        // try to get work
+        work_cv_.wait(lock, [this] { return shutdown_ || !work_pool_.empty(); });
+        // woke up! time to work or time to die?
```
tli2: Cruel life

benchmark/storage/data_table_benchmark.cpp
mbutrovich: Test harness has a very specific meaning to me when referring to Google Tests (something the tests inherit from). I'm not sure this applies.

mbutrovich:
I split this into a TestThreadPool class and a RandomTestUtil class. MultithreadedTestUtil didn't properly encompass its functionality anymore.

PR #124, opened 2018-08-21
Allow external iteration control on benchmarks

pervazea:
For the micro-benchmarks, we'll want to control the iteration count via the command line. 
Remove the hard-coded iteration count.


tli2:
I think google benchmark will run the loop enough times within each repetition to get accurate and meaningful numbers.

Repetitions will be good to account for variability, but only for longer benchmarks. This is what google's documentation says: https://github.com/google/benchmark#reporting-the-mean-median-and-standard-deviation-by-repeated-benchmarks

PR #127, opened 2018-08-22
Make Object Pool Size and Reuse Limit Tunable

yangjuns:
Main Changes:
- assign a size limit to an object pool to control the maximum objects an object pool can manage
- throw an exception when an object pool cannot handout more objects
- make object pool size_limit and reuse_limit tunable
- add more object pool tests

I also changed AlignedByteAllocator implementation so that it doesn't allocate unnecessary memory when the size of the object is multiple of 8.


yangjuns:
I also changed delta_record_test to avoid object pool exception introduced by the new code

src/include/common/object_pool.h
```diff
@@ -38,6 +39,20 @@ class AlignedByteAllocator {
...
+ * An exception thrown by object pools when they reach their size limits and
+ * cannot give more memory space for objects.
+ */
+class NoMoreObjectException : public std::exception {
+ public:
+  /***
+   * Describe the exception.
+   * @return a string of exception description
+   */
+  const char *what() const noexcept override { return "Object Pool have no object to hand out\n"; }
```
tli2: Let's add the current limit in here as well as a field

src/include/common/object_pool.h
```diff
@@ -38,6 +39,20 @@ class AlignedByteAllocator {
   // We believe otherwise, hence we're telling it to shut up.
 };
 
+// TODO(Yangjun): this class should be moved somewhere else.
+/***
```
tli2: extra `*`. We only use 2.

src/include/common/object_pool.h
```diff
@@ -55,12 +70,25 @@ class AlignedByteAllocator {
 template <typename T, class Allocator = AlignedByteAllocator<T>>
 class ObjectPool {
  public:
-  /**
+  /***
```
tli2: same

src/include/common/object_pool.h
```diff
@@ -55,12 +70,25 @@ class AlignedByteAllocator {
...
    * Initializes a new object pool with the supplied limit to the number of
    * objects reused.
-   * @param reuse_limit
+   *
+   * @param size_limit the maximum number of objects the object pool controls
+   * @param reuse_limit the maximum number of reusable objects, which needs be to
+   * not greater than size_limit
    */
-  explicit ObjectPool(const uint64_t reuse_limit) : reuse_limit_(reuse_limit) {}
+  explicit ObjectPool(const uint64_t size_limit, const uint64_t reuse_limit)
```
tli2: no need for `explicit` keyword since the constructor takes in more than 1 parameter.

src/include/common/object_pool.h
```diff
@@ -55,12 +70,25 @@ class AlignedByteAllocator {
...
+      : size_limit_(size_limit), reuse_limit_(reuse_limit), current_size_(0) {}
+
+  /**
+   * Initializes a new object pool with the supplied limit to the number of
+   * objects reused. The reuse_limit is set to be the same as size limit.
+   *
+   * @param size_limit the number of objects the object pool controls
+   */
+  explicit ObjectPool(const uint64_t size_limit)
```
tli2: This is basically a default argument. I say we remove it.

src/include/common/object_pool.h
```diff
@@ -76,17 +104,72 @@ class ObjectPool {
   /**
    * Returns a piece of memory to hold an object of T.
    *
+   * It throws exception if the object pool fails to fetch memory.
```
tli2: @throw or @throws ? I don't remember, but there's a directive you should use for these things.

src/include/common/object_pool.h
```diff
@@ -76,17 +104,72 @@ class ObjectPool {
...
+   *
    * @return pointer to memory that can hold T
    */
   T *Get() {
     T *result = nullptr;
-    if (!reuse_queue_.Dequeue(&result))
-      result = alloc_.New();
-    else
+    if (!reuse_queue_.Dequeue(&result)) {
+      curr_add_lock_.Lock();
```
tli2: we're database people, let's call them latches

test/common/object_pool_test.cpp
```diff
@@ -7,14 +7,30 @@
...
+#include "util/test_harness.h"
 
 namespace terrier {
+
+struct ObjectPoolTests : public TerrierTest {
+  std::default_random_engine generator_;
+  std::uniform_int_distribution<uint64_t> size_dist_;
+
+ protected:
+  void SetUp() override {
```
tli2: You can leave these two things out if they don't do anything special

src/include/common/object_pool.h
```diff
@@ -76,17 +104,72 @@ class ObjectPool {
...
+      size_limit_lock_.Lock();
```
tli2: Use ScopedSpinLatch for these

src/include/common/object_pool.h
```diff
@@ -95,15 +178,38 @@ class ObjectPool {
...
       reuse_queue_.Enqueue(std::move(obj));
+    }
+  }
+
+  /***
+   * Check if current_size_ is always no greater than size_limit_
+   *
+   * @return true if current size <= size_limit; false otherwise.
+   */
+  bool CheckInvariance() {
```
tli2: This should not be a public method if you only expect to check them in tests.

It probably qualifies more as asserts in the functions.

src/include/common/object_pool.h
```diff
@@ -18,7 +19,7 @@ class AlignedByteAllocator {
    * @return a pointer to the byte array allocated.
    */
   T *New() {
-    auto *result = reinterpret_cast<T *>(new uint64_t[sizeof(T) / 8 + 1]);
+    auto *result = reinterpret_cast<T *>(new uint64_t[(sizeof(T) + 7) / 8]);
```
mbutrovich: Does it make sense to use posix_memalign here?

mbutrovich:
Seeing latches and concurrent data structures mixing together in the control flow concerns me. Let's white board this tomorrow and see if we can streamline the logic a bit.

src/include/common/object_pool.h
tli2: problem with `posix_memalign` is that it's more of a C-style function. I think if the two are equivalent and not too much extra code is required we should stick to a more modern C++ code style.

I agree that this could stand to get some readability improvements though, either by using the Padding function in `StorageUtil` or comments.

pervazea: I understand what this is doing, but the code and intent would be much more transparent if we used posix_memalign. 

While this is technically C++, we are going to some effort to instruct the compiler and tell it what to do, which feels wrong.  IMO, we should use posix_memalign.

src/include/common/object_pool.h
pervazea: First, in addition to tli2's observation, I don't see why explicit is deemed necessary here. What breaks if is invoked with a compatible type?

Second, if it is required that reuse_limit <= size_limit, don't just document, enforce it with an assertion or other check and exception. I think assertion appropriate in this case.

src/include/common/object_pool.h
yangjuns: @pervazea  check-clang-tidy requires that every single argument constructor needs to be marked as explicit.

tli2:
By the way, at @mbutrovich's relentless request we are using `posix_memalign` like on this branch https://github.com/cmu-db/terrier/tree/tianyu-cleanup

src/include/common/object_pool.h
yangjuns: I don't think ScopedSpinLatch fit in this case. The latches are members in an object pool. If we use ScopedSpinLatch, then latches will only be unlocked when the object pool is destroyed, which is not something we want.

src/include/common/object_pool.h
```diff
@@ -76,17 +104,72 @@ class ObjectPool {
...
+      curr_add_lock_.Lock();
+      size_limit_lock_.Lock();
+      if (current_size_ < size_limit_) {
+        result = alloc_.New();
+        current_size_++;
+      }
+      size_limit_lock_.Unlock();
+      curr_add_lock_.Unlock();
+
+      if (result == nullptr) {
```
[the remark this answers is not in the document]
yangjuns: You are right. I assumed the allocation was successful. 

But this brings up a question I had before. Do we have rules on how an allocator should behave when the allocation fails? Should it return a null pointer or raise an exception? If it raises an exception, it certainly shouldn't be the same exception thrown by an object pool when it fails to give objects.

src/include/common/object_pool.h
tli2: I was more suggesting that we always take in an explicit reuse_limit as second argument

src/include/common/object_pool.h
```diff
@@ -1,11 +1,33 @@
...
 
 namespace terrier::common {
+// TODO(Yangjun): this class should be moved somewhere else.
+/**
+ * An exception thrown by object pools when they reach their size limits and
+ * cannot give more memory space for objects.
+ */
+class NoMoreObjectException : public std::exception {
+ public:
+  explicit NoMoreObjectException(uint64_t limit) : limit_(limit) {}
```
tli2: doxygen is complaining about this not documented.

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -41,8 +41,8 @@ class DataTableBenchmark : public benchmark::Fixture {
 
   // Test infrastructure
   std::default_random_engine generator_;
-  storage::BlockStore block_store_{1000};
-  common::ObjectPool<storage::BufferSegment> buffer_pool_{num_inserts_};
+  storage::BlockStore block_store_{1000, 1000};
+  common::ObjectPool<storage::BufferSegment> buffer_pool_{num_inserts_, num_inserts_};
```
tli2: let's actually have descriptive variable names here instead of reusing `num_inserts_`

benchmark/storage/tuple_access_strategy_benchmark.cpp
```diff
@@ -41,7 +41,7 @@ class TupleAccessStrategyBenchmark : public benchmark::Fixture {
 
   // Test infrastructure
   std::default_random_engine generator_;
-  storage::BlockStore block_store_{num_blocks_};
+  storage::BlockStore block_store_{num_blocks_, num_blocks_};
```
tli2: same

src/include/common/object_pool.h
```diff
@@ -38,23 +68,74 @@ class ObjectPool {
...
    */
   T *Get() {
+    SpinLatch::ScopedSpinLatch guard(&latch_);
     T *result = nullptr;
-    if (!reuse_queue_.Dequeue(&result))
-      result = alloc_.New();
-    else
+    if (reuse_queue_.empty()) {
+      if (current_size_ < size_limit_) result = alloc_.New();
+      if (result == nullptr) {
```
tli2: This is convoluted logic. It should suffice to use an else branch here. Alternatively, flatten the if-statement and throw exception either at the end of function or before the if-block to keep cyclomatic complexity down.

src/include/common/object_pool.h
```diff
@@ -38,23 +68,74 @@ class ObjectPool {
...
+        // out of memory
+        throw NoMoreObjectException(size_limit_);
+      }
+      current_size_++;
+    } else {
+      result = reuse_queue_.front();
+      reuse_queue_.pop();
       alloc_.Reuse(result);
+    }
+    TERRIER_ASSERT(current_size_ <= size_limit_, "object pool size exceed it's size limit");
```
tli2: nit: its?

src/include/common/object_pool.h
```diff
@@ -38,23 +68,74 @@ class ObjectPool {
...
+   *
+   * @param new_size the new object pool size
+   * @return true if new_size is successfully set and false the operation fails
+   */
+  bool SetSizeLimit(uint64_t new_size) {
+    SpinLatch::ScopedSpinLatch guard(&latch_);
+    if (new_size >= current_size_) {
+      // current_size_ might increase and become > new_size if we don't use lock
+      size_limit_ = new_size;
+      TERRIER_ASSERT(current_size_ <= size_limit_, "object pool size exceed it's size limit");
```
tli2: same

src/include/common/object_pool.h
```diff
@@ -38,23 +68,74 @@ class ObjectPool {
...
+      return true;
+    }
+    return false;
+  }
+
+  /**
+   * Set the reuse limit to a new value.
```
tli2: also document what the function will do. (It doesn't have to be an exact description of behavior, but some level of contract is needed: does it always happen immediately? Or is it not guaranteed? Is it guaranteed to be done by some point?)

test/common/object_pool_test.cpp
```diff
@@ -9,12 +9,13 @@
...
 namespace terrier {
+
 // Rather minimalistic checks for whether we reuse memory
 // NOLINTNEXTLINE
 TEST(ObjectPoolTests, SimpleReuseTest) {
   const uint32_t repeat = 10;
-  const uint64_t reuse_limit = 1;
-  common::ObjectPool<uint32_t> tested(reuse_limit);
+  const uint64_t size_limit = 1;
+  common::ObjectPool<uint32_t> tested(size_limit, size_limit);
```
tli2: same

test/common/object_pool_test.cpp
```diff
@@ -54,24 +89,37 @@ TEST(ObjectPoolTests, ConcurrentCorrectnessTest) {
   TestThreadPool thread_pool;
   // This should have no bearing on the correctness of test
   const uint64_t reuse_limit = 100;
-  common::ObjectPool<ObjectPoolTestType> tested(reuse_limit);
+  common::ObjectPool<ObjectPoolTestType> tested(reuse_limit, reuse_limit);
```
tli2: same

test/common/object_pool_test.cpp
```diff
@@ -54,24 +89,37 @@ TEST(ObjectPoolTests, ConcurrentCorrectnessTest) {
...
     std::default_random_engine generator;
     // Store the pointers we use.
     std::vector<ObjectPoolTestType *> ptrs;
-    auto allocate = [&] { ptrs.push_back(tested.Get()->Use(tid)); };
+    auto allocate = [&] {
+      try {
+        ObjectPoolTestType *temp = tested.Get();
+        ptrs.push_back(temp->Use(tid));
+      } catch (common::NoMoreObjectException) {
+      }
```
tli2: Do you really want to swallow the exception here? If the answer is yes, add a comment explaining why.

tli2:
FYI you still have doxygen failures on Travis

src/include/common/object_pool.h
```diff
@@ -63,15 +144,23 @@ class ObjectPool {
...
+      reuse_queue_.push(obj);
+    }
   }
 
  private:
   Allocator alloc_;
-  ConcurrentQueue<T *> reuse_queue_;
-  const uint64_t reuse_limit_;
+  SpinLatch latch_;
+  std::queue<T *> reuse_queue_;
```
mbutrovich: Add a TODO if we should explicitly pass a second template parameter to define the backing container for the std::queue. We could extend the ObjectPool tests to behave more like a benchmark and then measure each backing container.

yangjuns: So you are saying that we don't need to use std::queue here and, instead, we can have a second template argument that defines the underlying container for holding reusable objects. 

If I understand correctly, I agree with you. We don't need to reuse objects in a FIFO pattern. All we need is a container that can store whatever we give to him and give something back when we ask for it. I'll add a TODO

src/include/common/object_pool.h
yangjuns: clang-tidy will complain if you have an else-clause after throw. I'm not sure how to flatten the if-statement

tli2: Yes, because they want you to do:

```c++
if (...) throw foo;
...
```

Because you will exit the function immediately after the throw(or return), the else is completely useless.

yangjuns: I'm not sure what you want me to do here. Do you want me to do

```c++
if (result != nullptr) { 
    current_size_++;
} else {
    // out of memory
    throw NoMoreObjectException(size_limit_);
}
...
```

Is this better?

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -38,11 +38,12 @@ class DataTableBenchmark : public benchmark::Fixture {
...
   const uint32_t num_inserts_ = 10000000;
   const uint32_t num_threads_ = TestThreadPool::HardwareConcurrency();
+  const uint64_t bp_reuse_limit_ = 10000000;
 
   // Test infrastructure
   std::default_random_engine generator_;
-  storage::BlockStore block_store_{1000};
-  common::ObjectPool<storage::BufferSegment> buffer_pool_{num_inserts_};
+  storage::BlockStore block_store_{1000, 1000};
+  common::ObjectPool<storage::BufferSegment> buffer_pool_{num_inserts_, bp_reuse_limit_};
```
tli2: nit: don't use abbreviations

benchmark/storage/tuple_access_strategy_benchmark.cpp
```diff
@@ -38,10 +38,11 @@ class TupleAccessStrategyBenchmark : public benchmark::Fixture {
   const uint32_t num_inserts_ = 10000000;
   const uint32_t num_threads_ = TestThreadPool::HardwareConcurrency();
   const uint32_t num_blocks_ = num_inserts_ / layout_.NumSlots();
+  const uint64_t bs_reuse_limit_ = num_blocks_;  // block store reuse limit
```
tli2: nit: don't use abbreviations

src/include/common/object_pool.h
```diff
@@ -85,11 +85,12 @@ class ObjectPool {
     T *result = nullptr;
     if (reuse_queue_.empty()) {
       if (current_size_ < size_limit_) result = alloc_.New();
-      if (result == nullptr) {
+      if (result != nullptr) {
+        current_size_++;
+      } else {
         // out of memory
         throw NoMoreObjectException(size_limit_);
```
tli2: I think we said we should put this at the end of the function?

PR #130, opened 2018-08-23
Define TransactionQueue to be backed by std::list, add TRACE logging to GC

mbutrovich:
By default the std::queues were backed by std::dequeue which I believe to be inefficient for the GC's use case. Also added basic TRACE logging to GC to be able to watch its state.

### DIFF
--- a/src/include/transaction/transaction_defs.h
+++ b/src/include/transaction/transaction_defs.h
@@ -4,5 +4,11 @@
 #include <queue>
 namespace terrier::transaction {
 class TransactionContext;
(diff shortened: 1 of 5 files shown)


src/include/transaction/transaction_defs.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+#include <list>
+#include <queue>
+namespace terrier::transaction {
+class TransactionContext;
+using TransactionQueue = std::queue<transaction::TransactionContext *, std::list<transaction::TransactionContext *>>;
```
tli2: nit: Maybe add a comment on why this is necessary?

PR #131, opened 2018-08-23
Add small-scale GC tests and optimization for read-only txns

mbutrovich:
Adapted the MVCC tests to basic GC tests. Also added an optimization to GC that allows it to deallocate a read-only txn in a single pass.

### DIFF
--- a/test/storage/garbage_collector_test.cpp
+++ b/test/storage/garbage_collector_test.cpp
@@ -93,6 +93,7 @@ struct GarbageCollectorTests : public ::terrier::TerrierTest {
   const uint16_t max_columns_ = 100;
 };
 
+// Run a single txn that performs an Insert. Confirm that it takes 2 GC cycles to process this tuple.
 // NOLINTNEXTLINE
 TEST_F(GarbageCollectorTests, SingleInsert) {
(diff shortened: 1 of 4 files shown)


test/transaction/mvcc_test.cpp
```diff
@@ -628,7 +628,7 @@ TEST_F(MVCCTests, InsertUpdate1) {
     select_tuple = tested.SelectIntoBuffer(txn0, slot, tested.all_col_ids_);
     EXPECT_FALSE(StorageTestUtil::ProjectionListEqual(tested.Layout(), select_tuple, insert_tuple));
 
-    txn_manager.Commit(txn0);
+    txn_manager.Abort(txn0);
```
tli2: Is this change intended? Does it make a difference?

test/storage/garbage_collector_test.cpp
```diff
@@ -94,7 +94,7 @@ struct GarbageCollectorTests : public ::terrier::TerrierTest {
 };
 
 // NOLINTNEXTLINE
-TEST_F(GarbageCollectorTests, BasicTest) {
+TEST_F(GarbageCollectorTests, SingleInsert) {
```
tli2: nit: Add comments to these test cases about what they do

PR #132, opened 2018-08-23
Fix stupid leak in the test harness

tli2:
I am retarded. I should not write code at 11 p.m. There is no fancy leaks in clang and lambdas. GC was not too slow. The bug was in the performance testing harness.


mbutrovich:
Confirmed running a test scaled to 10 million txns on my laptop no longer balloons memory usage to multiple GBs. It now hovers between 400-700MB (thanks GC!)

mbutrovich:
memory
It's beautiful.

Issue #141, opened 2018-08-28
New scenarios for large_transaction_test (w/ GC)


tli2:
We have large scale tests that generate randomized workloads to run against the storage layer. The general framework is in place, and it has helped us find many bugs, but we have discovered that certain bugs only appear under certain configurations (e.g. high abort rates, large number of read-only transactions).

Currently we only have one test scenario for each of the test cases, with numbers that largely don't mean anything. The goal would be to come up with a variety of meaningful configurations that will help us smoke out different classes of bugs. 

The tests themselves are under `test/storage/large_garbage_collector_test.cpp` and  `test/transaction/large_transaction_test.cpp`. The small framework we use for them can be found in `test/util/transaction_test_util.cpp` and `test/include/util/transaction_test_util.h`. The tunable parameters exposed are documented and the existing test cases should serve as good examples. 

Come up with a configuration, describe the scenario it models, and if you are curious, verify that it can find bugs by injecting the kind of bug you think this would lead to into the transaction system.

mbutrovich:
Example scenarios that I think would be good:
### High abort rate
```
initial_table_size = 1000;
txn_length = 20;
const std::vector<double> update_select_ratio = {0.7, 0.3};
```
The hope would be that the high txn length (number of statements), large ratio of updates, and small table size would yield a high number of write-write conflicts (thus, aborts). We should be able to test this with the return values from `SimulateOltp()` which is a pair of vectors: one for committed txns and the other for aborted txns. Just take a look at the size and make sure our abort rate is over 50% -- maybe even higher. We really want to hammer the `Rollback()` correctness behavior, as well as for the GC test ensure that the GC doesn't mess up `Rollback()`. Tweak the parameters as you see fit to get the abort rate up.

### Low abort rate - high throughput
```
initial_table_size = 1000;
txn_length = 1;
const std::vector<double> update_select_ratio = {0.5, 0.5};
```
The hope would be that the low txn length (number of statements), split ratio of updates, and large table size would yield a low number of write-write conflicts (thus, aborts). We should be able to generate some long version chains here. Verify the abort rate in the same manner as above, except we're looking for a small number here (<1% maybe).


### TPC-C-like
```
initial_table_size = 1000000;
txn_length = 5;
const std::vector<double> update_select_ratio = {0.4, 0.6};
```
[...]

PR #142, opened 2018-08-28
Remove hardcoded num_threads from test

tli2:


mbutrovich:
So I think that this is probably the right move, but I’ll say that this does mean we may have test scenarios that pass locally but fail on Jenkins and it may be non-obvious to users why. I’m not sure where to leave a comment though telling them to hardcode the thread count to something higher than their machine supports.

pervazea:
This is reasonable for now. A hardcoded value is going to be wrong in many places.
TestThreadPool::HardwareConcurrency();  is better but not correct either. 

For functional tests, the precise number of threads is less important that testing with a range of threads. The problem we've seen is that when you increase the concurrency level (on a machine that can support more threads), bugs show up. A good approach would be to iterate over tests using a range of thread counts.

For benchmarks, we want to get a consistency to the results. The approach that I'm planning to use, will partition resources on the benchmark machine as follows:

- <n> cores allocated to the OS and supporting applications such as the Jenkins agent
- <m> cores allocated to benchmark threads.

It is unlikely that TestThreadPool::HardwareConcurrency(); is going to provide the right setting. The change is fine for now, I expect to have to change how this works.

tli2:
This is more in preparation for #141. It should be relatively easy for us to change behavior now that we wrapped it in a function that we own. (this is also why I did not directly call `std::thread::hardware_concurrency`). Sorry should have made that clear.

PR #143, opened 2018-08-29
Strong typedef col_id to avoid confusion. Associated minor fixes

tli2:
col_id and projection_list_index are both uint16_t but mean different things. This is very error prone since the two classes of values often coexist and are simply named generally as `col` or `i` in the code,

With strong typedef, the compiler shall slap you if you try to use one for the other.

`StrongTypeAlias` is also modified to have most of the integer operators, because it's pretty clear now that we almost always wrap things transparently (they still make sense as integers), as opposed to originally where we thought maybe we want to stop people from not only using them as integer types, but also do integer operations on them.

I have verified that this has no performance impact on benchmarks, as expected, since strong typedefs are compile-time only.


test/storage/projected_row_test.cpp
```diff
@@ -131,7 +133,7 @@ TEST_F(ProjectedRowTests, Alignment) {
     storage::BlockLayout layout = StorageTestUtil::RandomLayout(common::Constants::MAX_COL, &generator_);
 
     // generate a random projectedRow
-    std::vector<uint16_t> all_col_ids = StorageTestUtil::ProjectionListAllColumns(layout);
+    std::vector<col_id_t> all_col_ids = StorageTestUtil::ProjectionListAllColumns(layout);
```
mbutrovich: Do we want to replace these with `auto` while we're in here? The function name seems descriptive enough to imply what's going on.

tli2: as discussed we can make things explicit when it's not too much trouble, and when tidy is not yelling.

PR #150, opened 2018-08-30
ScopedTimer

mbutrovich:
This is more of a debugging tool at the moment, but eventually we might want to use this performance profiling of various components at runtime in the complete system. I decided to throw this together while trying to time each stage of the garbage collection process. We should prefer to use Google Benchmarks when actually characterizing performance, and also avoid using this as some sort of Google Test pass/fail metric. I just think it's useful for debugging and comparing performance changes for aspects of the system that don't have full benchmarks. I'm happy to not merge it in if others don't think it'll prove useful though.
```
#include <iostream>
#include <thread>  // NOLINT
#include "common/scoped_timer.h"

int main() {
  uint64_t elapsed_ms = 0;
  {
    terrier::common::ScopedTimer timer(&elapsed_ms);
    // do some stuff
    std::this_thread::sleep_for(std::chrono::seconds(5));
  }

  std::cout << "that took " << elapsed_ms << " milliseconds" << std::endl;
}
```
The output is:
`that took 5000 milliseconds`

### DIFF
--- a/src/include/common/scoped_timer.h
+++ b/src/include/common/scoped_timer.h
@@ -13,11 +13,13 @@ namespace terrier::common {
 /**
  * The ScopedTimer provided an easy way to collect the elapsed time for a block of code. It stores the current time when
  * it is instantiated, and then when it is destructed it updates the elapsed milliseconds time in the value pointed to
- * by the constructor argument.
+ * by the constructor argument. This is meant for debugging the performance impact of components that don't have
+ * dedicate benchmarks, and possibly for use with performance counters in the future. Its output shouldn't be used as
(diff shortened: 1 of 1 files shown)


src/include/common/scoped_timer.h
```diff
@@ -0,0 +1,41 @@
...
+// suitable.
+
+/**
+ * The ScopedTimer provided an easy way to collect the elapsed time for a block of code. It stores the current time when
+ * it is instantiated, and then when it is destructed it updates the elapsed milliseconds time in the value pointed to
+ * by the constructor argument.
+ */
+class ScopedTimer {
+ public:
+  ScopedTimer() = delete;
```
tli2: this line is not necessary. Default constructor will be implicitly deleted once you declare a different one.

PR #158, opened 2018-09-02
Document and enforce the alignment assumption in RawConcurrentBitmap.

lmwnshn:
Added to class header and to the function that assumes this.

### DIFF
--- a/src/include/common/container/concurrent_bitmap.h
+++ b/src/include/common/container/concurrent_bitmap.h
@@ -24,7 +24,9 @@ namespace terrier::common {
  * discipline to ensure safe access.
  *
  * For easy initialization in tests and such, use the static Allocate and
- * Deallocate methods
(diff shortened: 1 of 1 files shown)


mbutrovich:
Why doesn't Allocate() enforce this requirement? I realize we currently only using it in testing, but it seems odd not to enforce alignment requirement in its own internal Allocate() function.

tli2:
Actually, better still, TERRIER_ASSERT this requirement (`this % 8 == 0`) where the correctness of code depends on this

lmwnshn:
@mbutrovich 

Will we ever use Allocate outside of testing? I added an assert there, but technically the Allocate function itself should ensure it returns some 64-bit aligned memory. Though I'm not sure how that works right now.

tli2:
@lmwnshn You can use `AllocationUtil`

PR #161, opened 2018-09-07
Refactor BlockLayout constructor to take single argument

mbutrovich:
The num_attrs argument was redundant.

### DIFF
--- a/src/include/storage/storage_defs.h
+++ b/src/include/storage/storage_defs.h
@@ -46,7 +46,7 @@ struct BlockLayout {
    * @param attr_sizes vector of attribute sizes.
    */
   explicit BlockLayout(std::vector<uint8_t> attr_sizes)
-      : num_cols_(attr_sizes.size()),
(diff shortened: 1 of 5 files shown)


src/include/storage/storage_defs.h
```diff
@@ -43,16 +43,15 @@ struct BlockLayout {
    * attr_sizes, as the constructor applies optimizations based on sizes. It is up to the caller to then
    * associate these "column ids" with the right upper level concepts.
    *
-   * @param num_attrs number of attributes.
    * @param attr_sizes vector of attribute sizes.
    */
-  BlockLayout(const uint16_t num_attrs, std::vector<uint8_t> attr_sizes)
-      : num_cols_(num_attrs),
+  explicit BlockLayout(std::vector<uint8_t> attr_sizes)
+      : num_cols_(static_cast<uint16_t>(attr_sizes.size())),
```
tli2: This field is probably also redundant. I believe I went through the code and removed any direct reference to it already, so it should be easy to remove this as well?

PR #162, opened 2018-09-09
Write Ahead Log

### LINKED ISSUE #111: Write Ahead Logging
While the logging mechanism will not be complete until execution engine is in, it is possible to build most of the code with DataTable and concurrency control already in.

We need:
- `RedoRecord`, mirroring the ones we use for undo in transaction manager
- `LogManager` (I hate the word `Manager`), in charge of flushing buffers and signaling buffers flushed
- I/O code to serialize `RedoRecord`s out to a file. This include inlining of varlen fields (how such fields should be denoted is an open question)

tli2:
The basic hooks into the rest of the storage engine is done. Serialization format will definitely need to change, as we add in varlen fields. (Because we will probably rewrite that, I am not adding checksums) A bunch of other code needs to be written for handling errors on reading records back in.

Tests will be added soon. fix #111


src/include/storage/record_buffer.h
```diff
@@ -32,22 +32,150 @@ class BufferSegment {
...
+   * Delete the given buffer segment and frees the memory
+   * @param ptr the buffer to delete
+   */
+  void Delete(BufferSegment *const ptr) { delete ptr; }
 };
 
+/**
+ * Type alias for an object pool handing out buffer segments
+ */
+using RecordBufferSegmentPool = common::ObjectPool<BufferSegment, RecordBufferSegmentAllocator>;
```
mbutrovich: Maybe just BufferSegmentPool?

src/include/storage/data_table.h
```diff
@@ -94,7 +100,7 @@ class DataTable {
   // new one when the current one is full. Needless to say, we will need to revisit this when extending GC to handle
   // deleted tuples and recycle slots
   std::vector<RawBlock *> blocks_;
-  common::SpinLatch blocks_latch_;
+  mutable common::SpinLatch blocks_latch_;
```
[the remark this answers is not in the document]
tli2: Eventually in sequential scan. I will remove this here.

src/include/common/typedefs.h
```diff
@@ -214,6 +214,7 @@ using byte = std::byte;
 STRONG_TYPEDEF(timestamp_t, uint64_t);
 STRONG_TYPEDEF(layout_version_t, uint32_t);
 STRONG_TYPEDEF(col_id_t, uint16_t);
+STRONG_TYPEDEF(oid_t, uint32_t);
```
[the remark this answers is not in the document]
tli2: Interestingly, postgres define oid_t to represent any "object" in the database. We can have that conversation at 4:30 today

src/include/storage/record_buffer.h
tli2: I'd rather be specific in case we have other buffers (network, e.g.).

src/include/common/allocator.h
```diff
@@ -54,7 +54,7 @@ class ByteAlignedAllocator {
    * Deletes the byte array.
    * @param ptr pointer to the byte array to be deleted.
    */
-  void Delete(T *const ptr) { delete[] ptr; }  // NOLINT
+  void Delete(T *const ptr) { delete[] reinterpret_cast<byte *>(ptr); }  // NOLINT
```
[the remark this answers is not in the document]
tli2: Nope, Clang-tidy still hates us.

test/util/transaction_test_util.cpp
```diff
@@ -44,6 +45,11 @@ void RandomWorkloadTransaction::RandomUpdate(Random *generator) {
     }
     updates_[updated] = update;
   }
+  // TODO(Tianyu): Hardly efficient, but will do for testing.
+  if (test_object_->wal_on_) {
+    auto *record = txn_->StageWrite(nullptr, updated, initializer);
+    TERRIER_MEMCPY(record->Delta(), update, update->Size());
```
yangjuns: Is record ever used? I don't think the function StageWrite() has any side effect.

test/util/transaction_test_util.cpp
```diff
@@ -192,9 +199,14 @@ void LargeTransactionTestObject::PopulateInitialTable(uint32_t num_tuples, Rando
                                                : reinterpret_cast<storage::ProjectedRow *>(redo_buffer);
     StorageTestUtil::PopulateRandomRow(redo, layout_, 0.0, generator);
     storage::TupleSlot inserted = table_.Insert(initial_txn_, *redo);
+    // TODO(Tianyu): Hardly efficient, but will do for testing.
+    if (wal_on_) {
+      auto *record = initial_txn_->StageWrite(nullptr, inserted, row_initializer_);
+      TERRIER_MEMCPY(record->Delta(), redo, redo->Size());
```
yangjuns: same question here.

src/include/storage/record_buffer.h
```diff
@@ -3,15 +3,15 @@
 #include "common/constants.h"
 #include "common/object_pool.h"
 #include "common/typedefs.h"
-#include "storage/delta_record.h"
+#include "storage/undo_record.h"
 
 namespace terrier::storage {
 
 /**
  * An UndoBufferSegment is a piece of (reusable) memory used to hold undo records. The segment internally keeps track
```
yangjuns: I think the name UndoBufferSegment should be changed to BufferSegment since both RedoBuffer and UndoBuffer use this.

src/include/storage/record_buffer.h
```diff
@@ -32,22 +32,150 @@ class BufferSegment {
...
+  Iterator end() { return {segment_, segment_->size_}; }
+
+ private:
+  BufferSegment *segment_;
+};
+
+/**
+ * Custom allocator used for the object pool of buffer segments
+ */
+class RecordBufferSegmentAllocator {
```
yangjuns: Maybe this class should go to common/allocator.h?

src/include/storage/record_buffer.h
```diff
@@ -169,19 +297,53 @@ class UndoBuffer {
...
+   * @param log_manager the log manager this redo buffer talks to, or nullptr if logging is disabled
+   * @param buffer_pool The buffer pool to draw buffer segments from. Must be the same buffer pool the log manager uses.
+   */
+  explicit RedoBuffer(LogManager *log_manager, RecordBufferSegmentPool *buffer_pool)
+      : log_manager_(log_manager), buffer_pool_(buffer_pool) {}
+
+  /**
+   * Reserve a redo record with the given size. The returned pointer is guaranteed to be valid until NewEntry is called
+   * again, or when the buffer is explicitly flushed by the call Finish().
+   * @param size the size of the redo record to allocate
```
yangjuns: The documentation is kind of confusing since we use RedoBuffer to allocate both RedoRecord and CommitRecord. Maybe we should rename RedoBuffer to be something like LogBuffer?? But since it's a very large PR, we can do refactoring in a separate PR

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -0,0 +1,105 @@
...
+  // TODO(Tianyu): This can be changed later to be include things that are not necessarily backed by a disk
+  // (e.g. logs can be streamed out to the network for remote replication)
+  BufferedLogWriter out_;
+  RecordBufferSegmentPool *buffer_pool_;
+
+  // These need to be thread-safe since various execution threads will modify these
+  common::ConcurrentQueue<BufferSegment *> flush_queue_;
+
+  // TODO(Tianyu): Might not be necessary, since commit on txn manager is already protected with a latch
+  common::SpinLatch callbacks_latch_;
```
yangjuns: Mixed usage of concurrent data structure and latches. Is it possible to use just one kind?

mbutrovich:
Any guess why this isn't running Jenkins?

crd477:
The configuration option to periodically scan the repository got destroyed after a recent security update for a plug-in. 
I’ve restored it now and Jenkins  appears to building the commit. 
On a related note, I did discover the mechanism do the webhook trigger from Travis but I still have to work out the details.

mbutrovich:
Thanks for looking into it @crd477!

crd477:
Is this the first commit to get rejected because of the benchmark step?

mbutrovich:
I believe so, yes. We'll have to investigate if that's a real failure, but I believe that it is.

pervazea:
The benchmark result is real, but not a regression. I have a PR to submit, that controls how the benchmark is run. With that change, I expect the performance to come in line with historical average.

Will submit shortly, if possible.

src/include/storage/record_buffer.h
```diff
@@ -3,15 +3,15 @@
...
 
 /**
  * An UndoBufferSegment is a piece of (reusable) memory used to hold undo records. The segment internally keeps track
  * of its memory usage.
  *
- * This class should not be used by itself. @see UndoBuffer.
+ * This class should not be used by itself. @see UndoBuffer, @see RedoBuffer
  *
  * Not thread-safe.
  */
```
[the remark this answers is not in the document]
tli2: Yes

src/storage/record_buffer.cpp
```diff
@@ -0,0 +1,36 @@
...
+}
+
+byte *RedoBuffer::NewEntry(const uint32_t size) {
+  if (buffer_seg_ == nullptr) {
+    // this is the first write
+    buffer_seg_ = buffer_pool_->Get();
+  } else if (!buffer_seg_->HasBytesLeft(size)) {
+    // old log buffer is full
+    if (log_manager_ != LOGGING_DISABLED) log_manager_->AddBufferToFlushQueue(buffer_seg_);
+    buffer_seg_ = buffer_pool_->Get();
```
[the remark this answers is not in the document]
tli2: the old buffer seg is reclaimed when the log manager is done flushing it

src/transaction/transaction_manager.cpp
```diff
@@ -10,42 +10,55 @@ TransactionContext *TransactionManager::BeginTransaction() {
...
   timestamp_t txn_id = txn->TxnId().load();  // will not change
-  for (auto &it : undos) Rollback(txn_id, it);
+  for (auto &it : txn->undo_buffer_) Rollback(txn_id, it);
   table_latch_.Lock();
   const timestamp_t start_time = txn->StartTime();
   size_t ret UNUSED_ATTRIBUTE = curr_running_txns_.erase(start_time);
   TERRIER_ASSERT(ret == 1, "Aborted transaction did not exist in global transactions table");
   if (gc_enabled_) completed_txns_.push_front(txn);
   table_latch_.Unlock();
+  if (log_manager_ != LOGGING_DISABLED) txn->redo_buffer_.Discard();
```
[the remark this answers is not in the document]
tli2: Yes. Good catch

test/util/transaction_test_util.cpp
tli2: its memory content would have changed in the memcpy

src/include/storage/record_buffer.h
tli2: I think a specialized one like this lives with the typedef for the object pool

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -0,0 +1,200 @@
...
+   * Wrapper around the posix write call, where a single function call will always write the entire buffer out.
+   * (unlike posix write, which can write arbitrarily many bytes less than the given amount)
+   * @param fd posix fildes arg
+   * @param buf posix buf arg
+   * @param nbyte posix nbyte arg
+   * @throws runtime_error if the underlying posix call failed
+   */
+  static void WriteFully(int fd, const void *buf, size_t nbyte);
+};
+// TODO(Tianyu): Apparently, c++ fstream is considered slow and inefficient. Additionally, we
```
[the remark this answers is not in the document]
tli2: I'll shut up without a solid benchmark number. The control itself should be sufficient reason for using posix

src/include/storage/write_ahead_log/log_manager.h
tli2: I will use only a latch for now and add a TODO for later

PR #164, opened 2018-09-11
Add DataTable read benchmark

yangjuns:
Add DataTable benchmarks for 
- sequential reads
- random reads
- concurrent random reads

### DIFF
--- a/benchmark/storage/data_table_benchmark.cpp
+++ b/benchmark/storage/data_table_benchmark.cpp
@@ -98,7 +98,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentInsert)(benchmark::State &state
 // NOLINTNEXTLINE
 BENCHMARK_DEFINE_F(DataTableBenchmark, SequentialRead)(benchmark::State &state) {
   storage::DataTable read_table(&block_store_, layout_, layout_version_t(0));
-  // populate read_table by inserting tuples
+  // Populate read_table by inserting tuples
   // We can use dummy timestamps here since we're not invoking concurrency control
   transaction::TransactionContext txn(timestamp_t(0), timestamp_t(0), &buffer_pool_);
   std::vector<storage::TupleSlot> read_order;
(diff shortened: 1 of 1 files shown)


benchmark/storage/data_table_benchmark.cpp
```diff
@@ -85,8 +94,93 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentInsert)(benchmark::State &state
...
+  // We can use dummy timestamps here since we're not invoking concurrency control
+  transaction::TransactionContext txn(timestamp_t(0), timestamp_t(0), &buffer_pool_);
+  std::vector<storage::TupleSlot> read_order;
+  for (uint32_t i = 0; i < num_reads_; ++i) {
+    read_order.emplace_back(read_table.Insert(&txn, *redo_));
+  }
+  // NOLINTNEXTLINE
+  for (auto _ : state) {
+    // We can use dummy timestamps here since we're not invoking concurrency control
+    transaction::TransactionContext txn(timestamp_t(0), timestamp_t(0), &buffer_pool_);
```
mbutrovich: Redundant?

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -85,8 +94,93 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentInsert)(benchmark::State &state
...
+  std::vector<storage::TupleSlot> read_order;
+  for (uint32_t i = 0; i < num_reads_; ++i) {
+    read_order.emplace_back(read_table.Insert(&txn, *redo_));
+  }
+  // Create random reads
+  std::shuffle(read_order.begin(), read_order.end(), generator_);
+  // NOLINTNEXTLINE
+  for (auto _ : state) {
+    // We can use dummy timestamps here since we're not invoking concurrency control
+    transaction::TransactionContext txn(timestamp_t(0), timestamp_t(0), &buffer_pool_);
```
mbutrovich: Same as previous comment.

yangjuns: I'm viewing this benchmark as each thread is running a transaction. Do we want multiple threads executing the same transaction? Well... we didn't do this way for ConcurrentInsert benchmark.

benchmark/storage/data_table_benchmark.cpp
yangjuns: If we want everything to be in the same transaction, yes, it's redundant. I'll fix it.

benchmark/storage/data_table_benchmark.cpp
yangjuns: Sry. Ignore my comment... I thought you commented on the concurrent benchmark.

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -85,8 +94,98 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentInsert)(benchmark::State &state
...
+  // Generate random read orders and read buffer for each thread
+  std::vector<std::vector<storage::TupleSlot>> read_orders;
+  std::vector<byte *> read_buffers;
+  std::vector<storage::ProjectedRow *> reads;
+
+  for (uint32_t i = 0; i < num_threads_; i++) {
+    // Create random reads
+    std::shuffle(read_order.begin(), read_order.end(), generator_);
+    read_orders.emplace_back(
+        std::vector<storage::TupleSlot>(read_order.begin(), read_order.begin() + (num_reads_ / num_threads_)));
```
yangjuns: This portion of code takes some time although it's not included in the benchmark timer. It basically shuffles the 10-million-long vector *num_threads_* times. I do this so that each thread can read a random subset of the table and these subsets can potentially overlap each other. 

If we don't want to enforce the overlapping condition, I can just shuffle once and let threads use different portions of the vector. I just don't know if that's something we want.

yangjuns: never mind. Just think of a better way to do it.

PR #165, opened 2018-09-14
Micro benchmark changes

pervazea:
- micro benchmark execution is now restricted to a single NUMA node. This has a significant effect on concurrent benchmarks. Performance is improved and much more consistent.

- convert time to seconds, for reporting via JUnit Jenkins plugin.


pervazea:
If this passes, then I suggest merging this first.
PR 162, the write ahead logging, after pulling in these change, "should" pass.

mbutrovich:
Just curious: do you have results from a few runs on the target machine to demonstrate the reduced variance? As we discussed yesterday, we expect that this should reduce variance, I just always like to see numbers prove out our theories when possible. :)

pervazea:
Yes, but not in a form that I can quickly insert here.  

1. I ran approx. 15 runs on Jenkins today, so that the results comparison will be with current, consistent results, rather than the older data points where any CPU was permitted. 
2. That data is in Jenkins, in pa_terrier / microbench. The results there are much more consistent.
3. I expect to see the run of this PR and the re-run of write-ahead, logging to confirm better consistency  by passing the checks.

pervazea:
From this PR's logs. You can see the desired effect in the p_diff (percentage diffference) being small.

```
RES. value     reference % tol. p_diff ref type nres suite                        test                 

----|---------|---------|------|------|--------|----|----------------------------|--------------------|

pass 3.017e+07 0         0       +0    config   0    DataTableBenchmark           ConcurrentRandomRead

pass 5.525e+06 5.296e+06 10      +4    history  10   DataTableBenchmark           ConcurrentInsert

pass 2.299e+07 0         0       +0    config   0    DataTableBenchmark           SequentialRead

pass 5.828e+06 0         0       +0    config   0    DataTableBenchmark           RandomRead

pass 8.222e+06 8.177e+06 10      +0    history  10   DataTableBenchmark           SimpleInsert

pass 2.393e+07 2.452e+07 10      -2    history  10   TupleAccessStrategyBenchmark SimpleInsert

pass 5.333e+06 5.158e+06 10      +3    history  10   TupleAccessStrategyBenchmark ConcurrentInsert
```

mbutrovich:
Those numbers look great! Nice job!

mbutrovich:
Do we only envision the microbenchmark suite for Jenkins running on an Ubuntu host? I think that's probably fine, but just want to point out that this won't work for macOS hosts (no lscpu).

pervazea:
The short answer is, yes, this specific script is only intended for use on a Linux system. My opinion is that having consistent benchmarks along with historical data is the priority, for our PRs. Having the infrastructure be portable to other environments is secondary. 

The underlying Google Benchmark based benchmarks, are of course runnable on a Mac, and manually running them with processor affinity, if that is possible, would be preferred.

PR #169, opened 2018-09-20
Added simple Column and Schema classes

### LINKED ISSUE #146: Schema
SqlTable will be responsible for translating the SQL notion of schema to the storage layer's BlockLayout. We'll likely need to bring in Peloton's representation of schema. We also want to add the notion of layout_version to the DataTable level.

mbutrovich:
Fix #146.

Need these since SqlTable will be instantiated with Schema rather BlockLayout. This seems like the minimum functionality we need now, but happy to discuss if I overlooked something.

### DIFF
--- a/src/include/catalog/schema.h
+++ b/src/include/catalog/schema.h
@@ -39,7 +39,7 @@ class Schema {
           nullable_(nullable),
           inlined_(true),
           oid_(oid) {
-      if (attr_size_ == type::VARLEN) {
+      if (attr_size_ == 0) {
         // this is a varlen attribute
         attr_size_ = 8;
         inlined_ = false;
@@ -71,11 +71,11 @@ class Schema {
     /**
      * @return internal unique identifier for this column
      */
-    col_oid_t &GetOid() { return oid_; }
+    col_oid_t GetOid() const { return oid_; }
     /**
      * @return id used at the storage layer for identifying this column in a layout
      */
-    col_id_t &GetId() { return id_; }
+    col_id_t GetId() const { return id_; }
 
    private:
     const std::string name_;
@@ -102,7 +102,7 @@ class Schema {
    * @return description of the schema for a specific column
    */
   Column GetColumn(const col_id_t col_id) const {
-    TERRIER_ASSERT(col_id < columns_.size(), "column id is out of bounds for this Schema");
+    TERRIER_ASSERT(!col_id < columns_.size(), "column id is out of bounds for this Schema");
     return columns_[static_cast<uint16_t>(col_id)];
   }
   /**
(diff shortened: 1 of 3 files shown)


src/include/catalog/schema.h
```diff
@@ -0,0 +1,107 @@
...
+     * @param oid internal unique identifier for this column
+     */
+    Column(std::string name, const type::TypeId type, const bool nullable, const col_oid_t oid)
+        : name_(std::move(name)),
+          type_(type),
+          attr_size_(type::TypeUtil::GetTypeSize(type_)),
+          nullable_(nullable),
+          inlined_(true),
+          oid_(oid) {
+      if (attr_size_ == 0) {
```
tli2: maybe `#define VARLEN 0`?

src/include/catalog/schema.h
```diff
@@ -0,0 +1,107 @@
...
+
+   private:
+    const std::string name_;
+    const type::TypeId type_;
+    uint8_t attr_size_;
+    const bool nullable_;
+    bool inlined_;
+    const col_oid_t oid_;
+    // TODO(Matt): default value would go here
+    // Value default_;
```
tli2: Should this Column also provide mapping to col_id_t in the storage layer?

src/include/common/typedefs.h
```diff
@@ -214,6 +214,7 @@ using byte = std::byte;
 STRONG_TYPEDEF(timestamp_t, uint64_t);
 STRONG_TYPEDEF(layout_version_t, uint32_t);
 STRONG_TYPEDEF(col_id_t, uint16_t);
+STRONG_TYPEDEF(col_oid_t, uint32_t);
```
tli2: Not in your PR, but this `col_id_t`, `col_oid_t` thing is making me sad.

I will probably later refactor all these strong typedefs into namespaces. (that way we will get `storage::col_id_t` and `catalog::col_oid_t`)

src/include/catalog/schema.h
mbutrovich: Possibly! We don't have a catalog yet. I almost think the mapping from col_oit_t to col_id should exist at the SqlTable though, where we're handle stuff like multiple layouts...

src/include/common/typedefs.h
mbutrovich: That sounds reasonable.

mbutrovich: Want me to add a todo in my PR?

src/include/type/type_util.h
```diff
@@ -2,6 +2,8 @@
 
 namespace terrier::type {
 
+#define VARLEN 0
```
tli2: nit: does this belong in type_util? I think it probably belongs in common_def?

PR #174, opened 2018-09-23
Delete() added to DataTable

### LINKED ISSUE #173: Add Delete() to DataTable
- [x] Constructor needs to guarantee that the first 2 columns are reserved for version pointer and then logical delete
- [x] Modify Select() to return a boolean: false if not present or deleted, true if visible and ProjectedRow has been populated with data
- [x] Modify Update() to not allow writes to the first 2 column_ids (hidden from user). Maybe can't assert this if I want to call Update() from Delete() to flip that column's null bit.
- [x] Add Delete(transaction::TransactionContext *txn, TupleSlot slot) method
- [x] Modify tests to support new semantics
- [x] Add new tests to verify functionality

mbutrovich:
Fix #173.

Need to beef up the testing but I'm ready to see what CI looks like.


mbutrovich:
This is ready for a first round of comments next week. The mvcc_test stuff covers the basic behavior, but we should figure out next week how to test this at large scale in another PR (perhaps another beginner task).

mbutrovich:
I had to change the BlockLayout on the benchmarks (add more columns) since the old 2-column layout doesn't really work when you reserve 2 columns for metadata in the real system. I made them 4 columns so I expect a regression in the numbers, but when I run master with 4 columns the numbers look about the same so my changes don't directly affect performance. Either way, I expect this will fail the benchmarks delta check on Jenkins.

We should discuss if we want to benchmark multiple layouts or if we think there is one layout that would be general enough to spot regressions. I think I'm okay with just 1 BlockLayout for benchmarks.

tli2:
Alternatively I think it would also be okay to just not create null for the first column

mbutrovich:
Benchmark numbers from this PR:
```
RES. value     reference % tol. p_diff ref type nres suite                        test                 
----|---------|---------|------|------|--------|----|----------------------------|--------------------|
pass 2.084e+07 0         0       +0    config   0    DataTableBenchmark           ConcurrentRandomRead
pass 5.557e+06 5.296e+06 10      +4    history  10   DataTableBenchmark           ConcurrentInsert
pass 1.551e+07 0         0       +0    config   0    DataTableBenchmark           SequentialRead
pass 4.025e+06 0         0       +0    config   0    DataTableBenchmark           RandomRead
FAIL 6.936e+06 8.177e+06 10     -15    history  10   DataTableBenchmark           SimpleInsert
FAIL 1.633e+07 2.452e+07 10     -33    history  10   TupleAccessStrategyBenchmark SimpleInsert
pass 5.266e+06 5.158e+06 10      +2    history  10   TupleAccessStrategyBenchmark ConcurrentInsert
```
Benchmark numbers from #177 that just changes the parameters to match this PR:
```
RES. value     reference % tol. p_diff ref type nres suite                        test                 
----|---------|---------|------|------|--------|----|----------------------------|--------------------|
pass 1.787e+07 0         0       +0    config   0    DataTableBenchmark           ConcurrentRandomRead
pass 5.51e+06  5.296e+06 10      +4    history  10   DataTableBenchmark           ConcurrentInsert
pass 1.228e+07 0         0       +0    config   0    DataTableBenchmark           SequentialRead
pass 3.332e+06 0         0       +0    config   0    DataTableBenchmark           RandomRead
FAIL 6.144e+06 8.177e+06 10     -24    history  10   DataTableBenchmark           SimpleInsert
FAIL 1.221e+07 2.452e+07 10     -50    history  10   TupleAccessStrategyBenchmark SimpleInsert
pass 5.441e+06 5.158e+06 10      +5    history  10   TupleAccessStrategyBenchmark ConcurrentInsert
```
This PR is actually _faster_ under the same benchmark parameters, so I don't think failing the benchmarks on Jenkins is a reason not to merge this after review.

src/storage/data_table.cpp
```diff
@@ -29,31 +46,50 @@ void DataTable::Select(transaction::TransactionContext *const txn, const TupleSl
...
   while (version_ptr != nullptr &&
          transaction::TransactionUtil::NewerThan(version_ptr->Timestamp().load(), txn->StartTime())) {
-    StorageUtil::ApplyDelta(accessor_.GetBlockLayout(), *(version_ptr->Delta()), out_buffer);
+    auto modifies_logical_delete_column = StorageUtil::DeltaModifiesLogicalDelete(*(version_ptr->Delta()));
+    if (modifies_logical_delete_column == LogicalDeleteModificationType::NONE) {
+      // Normal delta to be applied. Does not modify the logical delete column.
+      StorageUtil::ApplyDelta(accessor_.GetBlockLayout(), *(version_ptr->Delta()), out_buffer);
+    } else if (modifies_logical_delete_column == LogicalDeleteModificationType::INSERT) {
+      // Applying the undo of an INSERT makes the tuple invisible to this txn.
+      visible = false;
```
tli2: I am not convinced we need to special case inserts and deletes. We still end up with the correct buffer, just need to check the resulting buffer at the end for visibility. Also, we don't really do extra work because as you said most likely inserts and deletes are at the end and beginning of a version chain, respectively.

mbutrovich: The idea is that the output buffer doesn't have a column for the logical delete column, so we're not actually applying deltas in that case. The notion of the logical delete column shouldn't be exposed above this layer, so the ProjectedRow won't hold an attribute for it.

src/storage/data_table.cpp
```diff
@@ -29,31 +46,50 @@ void DataTable::Select(transaction::TransactionContext *const txn, const TupleSl
     // Here we will need to check that the version pointer did not change during our read. If it did, the content
     // we have read might have been rolled back and an abort has already unlinked the associated undo-record,
     // we will have to loop around to avoid a dirty read.
+    visible = Visible(slot, accessor_);
```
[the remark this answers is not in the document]
tli2: Yeah sure

mbutrovich:
Coverage drop is expected because I just added the IsVisibleI() method that @tli2 will use for columnar access.

PR #180, opened 2018-09-25
Logical expressions and Values

lmwnshn:
Initial port of (logical) expressions.

I brought over the bare minimum required to work with [postgresparser](https://github.com/cmu-db/peloton/blob/master/src/parser/postgresparser.cpp), and it should work with that right now; though more will need to be added in future e.g. FunctionExpression will eventually need to store function types.

Right now, everything is basically just a container holding stuff from the parser. Executable/visitor pattern version coming hopefully by next meeting. I wanted to do a second pass over this current PR to see if I was being inefficient (unnecessary copying, not std::move'ing where I should). But might not get to that until Wednesday night earliest.

Sorry for the delay, forgot to push on Friday.

Also, should we distinguish Logical/Executable in the class name?
LogicalAbstractExpression, LogicalCaseExpression, ...
ExecutableAbstractExpression, ExecutableCaseExpression, ...


pervazea:
I've been looking at this. This is not the review, just some comments.

- It needs some unit tests (see further down).
- It is very timely, as this PR and the "value" PR I'm working on, are related. I was trying to define the right caller API for the value object, and this PR helps do that. As a result, I changed what I had as the API.
- So, I'm integrating my PR with this one, and have started writing some expression unit tests, which also have the effect of exercising "values".

src/include/type/expression/abstract_expression.h
```diff
@@ -0,0 +1,191 @@
...
+   * @param expression_type what type of expression we have
+   * @param return_value_type the type of the expression's value
+   * @param children the list of children for this node
+   */
+  AbstractExpression(const ExpressionType expression_type, const TypeId return_value_type,
+                     const std::vector<AbstractExpression *> children)
+      : expression_type_(expression_type), return_value_type_(return_value_type) {
+    for (auto &child : children) {
+      children_.emplace_back(std::unique_ptr<AbstractExpression>(child));
+      if (child->is_nullable_) {
```
pervazea: The is_nullable_, has_parameter_ and has_subquery_ fields don't appear to have default initialization anywhere, so they are initialized by the default constructor. The rules around what is initialized for you and what is not, are quite complicated. Therefore, much clearer if we add code to initialize these explicitly.

src/include/type/expression/abstract_expression.h
```diff
@@ -0,0 +1,191 @@
...
+        is_nullable_ = true;
```
pervazea: What does is_nullable_ mean? I'm not clear, even after looking at Peloton and doing some reading.
1. Add comments to the declaration in abstract_expression.h to spell out what this is.
2. In Peloton, is_nullable is method that walks the expression tree. Here it is a flag that just looks at the first level child. I suspect that what is needed is the Peloton implementation.
3. Ditto for has_parameter_ and has_subquery_

src/include/type/expression/expression_defs.h
```diff
@@ -0,0 +1,76 @@
...
+#include <cstdint>
+
+namespace terrier::type::expression {
+
+/**
+ * All possible expression types.
+ * TODO(WAN): Do we really need some of this stuff? e.g. HASH_RANGE for elastic
+ */
+enum class ExpressionType : uint8_t {
+  INVALID,
```
pervazea: Not clear which ones we need to support and which not. What some of the types mean (semantics) is also not clear. Can we flag the questionable ones for discussion / review?

src/include/type/value.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include "common/typedefs.h"
+#include "type/type_id.h"
+
+namespace terrier::type {
+/**
+ * A Value is a tagged union of a SQL type ID and the contents.
+ */
+struct Value {
```
pervazea: This Value placeholder will likely change. I have a simple implementation of Value that I will submit into this PR. It can better evaluated in the context of the expression code.

src/include/common/typedefs.h
```diff
@@ -211,12 +211,14 @@ namespace terrier {
 using byte = std::byte;
 }
 
-STRONG_TYPEDEF(timestamp_t, uint64_t);
-STRONG_TYPEDEF(layout_version_t, uint32_t);
+STRONG_TYPEDEF(boolean_t, int8_t);
 // TODO(Matt): consider namespacing strong typedefs so we have storage::col_id_t and catalog::col_oid_t
 STRONG_TYPEDEF(col_id_t, uint16_t);
 STRONG_TYPEDEF(col_oid_t, uint32_t);
+STRONG_TYPEDEF(date_t, int32_t);
```
mbutrovich: Signed seems odd to me, but there may be a good reason for this.

src/include/type/expression/abstract_expression.h
```diff
@@ -0,0 +1,191 @@
...
+ protected:
+  /**
+   * Instantiates a new abstract expression. Because these are logical expressions, everything should be known
+   * at the time of instantiation, i.e. the resulting object is immutable.
+   * @param expression_type what type of expression we have
+   * @param return_value_type the type of the expression's value
+   * @param children the list of children for this node
+   */
+  AbstractExpression(const ExpressionType expression_type, const TypeId return_value_type,
+                     const std::vector<AbstractExpression *> children)
```
mbutrovich: const & for the vector?

src/include/type/expression/abstract_expression.h
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  /**
+   * (Convenience) Instantiates a new abstract expression for the 2-child case.
+   * @param expression_type what type of expression we have
+   * @param return_value_type the type of the expression's value
+   * @param left left child
+   * @param right right child
+   */
+  AbstractExpression(const ExpressionType expression_type, const TypeId return_value_type, AbstractExpression *left,
```
mbutrovich: Why do we need multiple constructors to handle the cases for number of children, rather than being able to just use the constructor with the vector and just put 0,1,2, etc. children in the vector?

src/include/type/expression/case_expression.h
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Represents a logical case expression.
+ */
+class CaseExpression : public AbstractExpression {
+ public:
+  /**
+   * A unique pointer to an abstract expression.
+   */
+  using AbsExprPtr = std::unique_ptr<AbstractExpression>;
```
mbutrovich: A lot of files in this folder use std::unique_ptr<AbstractExpression>. Seems odds to me to only have it in this one file. I'd remove this or add it to expression_defs.h and use it consistently in all of these other files.

src/include/type/expression/case_expression.h
```diff
@@ -0,0 +1,128 @@
...
+    AbsExprPtr then;
+  };
+
+  /**
+   * Instantiate a new case expression.
+   * @param return_value_type return value of the case expression
+   * @param when_clauses list of when clauses
+   * @param default_expr default expression for this case
+   */
+  explicit CaseExpression(const TypeId return_value_type, std::vector<WhenClause> &&when_clauses,
```
mbutrovich: r-value references here seem odd to me. I'd add a TODO to revisit this once we actually use it to see if that signature makes sense. My gut says no.

src/include/type/expression/case_expression.h
```diff
@@ -0,0 +1,128 @@
...
+
+  bool operator!=(const AbstractExpression &rhs) const override { return !(*this == rhs); }
+
+  AbstractExpression *Copy() const override {
+    std::vector<WhenClause> copied_clauses;
+    for (auto &clause : when_clauses_) {
+      copied_clauses.push_back(WhenClause{AbsExprPtr(clause.condition->Copy()), AbsExprPtr(clause.then->Copy())});
+    }
+    return nullptr;
+    // return new CaseExpression(GetReturnValueType(), copied_clauses, default_expr_ == nullptr ? nullptr :
```
mbutrovich: Dead code.

src/include/type/expression/comparison_expression.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <vector>
+#include "type/expression/abstract_expression.h"
+#include "type/expression/expression_defs.h"
+#include "type/type_id.h"
+
+namespace terrier {
+namespace type {
+namespace expression {
```
mbutrovich: @tli2 where did we land on namespace depth? Are we making a global change to allow more, or just an exception for the execution folder?

test/expression/expression_test.cpp
```diff
@@ -0,0 +1,136 @@
+#include "gtest/gtest.h"
```
mbutrovich: Do we have many more tests we can pull from Peloton?

test/type/value_type_test.cpp
```diff
@@ -0,0 +1,27 @@
+#include "gtest/gtest.h"
```
mbutrovich: Do we have many more tests we can pull from Peloton?

test/CMakeLists.txt
```diff
@@ -10,5 +10,7 @@ target_link_libraries(test_util ${TERRIER_LINK_LIBS})
 add_dependencies(test_util gtest)
 
 add_subdirectory(common)
+add_subdirectory(expression)
```
mbutrovich: In theory since expression is a subfolder of type, the tests should be under type as well.

src/include/common/typedefs.h
pervazea: Looking at the Peloton code, it has an inconsistent approach to date and time stamp values, with respect to signed / unsigned. I think we fix it now and have them both be unsigned.

src/include/type/value.h
```diff
@@ -0,0 +1,266 @@
...
+     */
+    timestamp_t timestamp;
+    /**
+     * Date values
+     */
+    date_t date;
+    /**
+     * variable length strings, without a null terminator. Length specified by var_len
+     */
+    char *varchar;
```
mbutrovich: We prefer byte * over char * in the storage layer, but maybe it makes sense at this layer since we're actually reasoning about a VARCHAR at this point?

test/CMakeLists.txt
```diff
@@ -10,5 +10,7 @@ target_link_libraries(test_util ${TERRIER_LINK_LIBS})
...
 add_subdirectory(storage)
 add_subdirectory(transaction)
+
```
mbutrovich: I think we need add_subdirectory(type) for the other tests as well.

lmwnshn:
Quick thoughts - strong_typedef for boolean seems strange, it might cause there to be 256 distinct boolean values.

I think logical expressions are ready. The executable versions have some discussion.

mbutrovich:
We could try just using a real bool and a static_assert that the size of it is 1 byte. It's technically not guaranteed by the spec, but I'd be surprised if we fail that.

lmwnshn:
@pervazea @mbutrovich Did we come to a consensus on the bool_t strongtypedef?

@wenxuanqiu The consensus for today's discussion on expressions, new system looks like

Parser -> ParserExpressions (as-is) -> Binder1 [col_oids] -> ParserExpression + vec<col_oid>, which is what is cached, then -> Binder2 [values] -> ParserExpression + vec<col_oid> + vec<values>.

src/include/parser/expression/abstract_expression.h
```diff
@@ -0,0 +1,122 @@
...
+   * @param rhs other
+   * @return true if the two expressions are logically equal
+   */
+  virtual bool operator==(const AbstractExpression &rhs) const {
+    if (expression_type_ != rhs.expression_type_ || children_.size() != rhs.children_.size()) {
+      return false;
+    }
+
+    for (size_t i = 0; i < children_.size(); i++) {
+      if (*children_[i].get() != *rhs.children_[i].get()) {
```
mbutrovich: * should be overloaded on unique_ptr such that the .get() to retrieve the raw pointer first shouldn't be necessary. This isn't necessarily a required change and is probably legacy behavior -- just want to point this out as we write more code.

src/include/parser/expression/abstract_expression.h
```diff
@@ -0,0 +1,122 @@
...
+   * @return number of children
+   */
+  size_t GetChildrenSize() const { return children_.size(); }
+
+  /**
+   * Returns a child of the abstract expression.
+   * @param index index of child, NOT BOUNDS CHECKED
+   * @return child
+   */
+  AbstractExpression *GetChild(size_t index) const { return children_[index].get(); }
```
mbutrovich: We should add an @warning here that we're handing out raw pointers to what is internally a managed object. You should never call this with the belief that you are taking ownership of this object.

src/include/parser/expression/case_expression.h
```diff
@@ -0,0 +1,127 @@
...
+                                          std::unique_ptr<AbstractExpression>(clause.then->Copy())});
+    }
+    return new CaseExpression(
+        GetReturnValueType(), &copied_clauses,
+        default_expr_ == nullptr ? nullptr : std::unique_ptr<AbstractExpression>(default_expr_->Copy()));
+  }
+
+  /**
+   * Get the number of when clauses.
+   * @return the number of when clauses
```
mbutrovich: For simple methods like this you can omit the description and just fill out the @return for a getter or the @param for a setter. That should be enough to make Doxygen happy without being too redundant/burdensome for the author.

src/include/parser/expression/comparison_expression.h
```diff
@@ -0,0 +1,38 @@
...
+#include <memory>
+#include <utility>
+#include <vector>
+#include "parser/expression/abstract_expression.h"
+#include "parser/expression/expression_defs.h"
+#include "type/type_id.h"
+
+namespace terrier {
+namespace parser {
+namespace expression {
```
mbutrovich: Do we want a third-level namespace? @tli2

src/type/value.cpp
```diff
@@ -0,0 +1,130 @@
+#include <type/value.h>
```
mbutrovich: Was this one of the targets we were gonna consolidate under a src/sql folder? @yangjuns 

That doesn't have to be in this PR, just want to make sure we either document it or have it as part of an issue.

src/include/parser/expression/aggregate_expression.h
```diff
@@ -0,0 +1,32 @@
...
+ * An AggregateExpression is only used for parsing, planning and optimizing.
+ */
+class AggregateExpression : public AbstractExpression {
+ public:
+  /**
+   * Instantiates a new aggregate expression.
+   * @param type type of aggregate expression
+   * @param children children to be added
+   */
+  explicit AggregateExpression(ExpressionType type, std::vector<std::unique_ptr<AbstractExpression>> &&children)
```
tli2: No need for `explicit` here

src/include/parser/expression/case_expression.h
```diff
@@ -0,0 +1,127 @@
...
+    std::unique_ptr<AbstractExpression> then;
+  };
+
+  /**
+   * Instantiate a new case expression.
+   * @param return_value_type return value of the case expression
+   * @param when_clauses list of when clauses
+   * @param default_expr default expression for this case
+   */
+  explicit CaseExpression(const type::TypeId return_value_type, std::vector<WhenClause> *when_clauses,
```
tli2: same

src/include/parser/expression/comparison_expression.h
tli2: I think because codegen have them, it's fine.

However, I do not like the use of third-level namespace here. Adding `expression::` to every expression doesn't tell people anything they don't know and is just code clutter. Consider `expression::Comparison` or just drop the namespace and use `ComparisonExpression`.

src/include/parser/expression/abstract_expression.h
```diff
@@ -0,0 +1,122 @@
...
+   * Logical inequality check.
+   * @param rhs other
+   * @return true if the two expressions are not logically equal
+   */
+  virtual bool operator!=(const AbstractExpression &rhs) const { return !(*this == rhs); }
+
+  /**
+   * Creates a copy of the current AbstractExpression.
+   */
+  virtual AbstractExpression *Copy() const { return new AbstractExpression(*this); }
```
tli2: I find this abstraction weird since we use smart pointers most other places but raw pointers here.

src/include/parser/expression/comparison_expression.h
```diff
@@ -0,0 +1,38 @@
...
+   */
+  explicit ComparisonExpression(const ExpressionType cmp_type,
+                                std::vector<std::unique_ptr<AbstractExpression>> *children)
+      : AbstractExpression(cmp_type, type::TypeId::BOOLEAN, std::move(*children)) {}
+
+  AbstractExpression *Copy() const override {
+    std::vector<std::unique_ptr<AbstractExpression>> children;
+    children.emplace_back(std::unique_ptr<AbstractExpression>(GetChild(0)));
+    children.emplace_back(std::unique_ptr<AbstractExpression>(GetChild(1)));
+    return new ComparisonExpression(GetExpressionType(), &children);
```
tli2: General comment for later as well: since we are using C++17, we can use `std::make_unique` for these.  Also, won't be too hard to turn this whole chunk of code into an initializer list (`return std::make_unique<AbstractExpression>({std::make_unique(GetChild(0), std::make_unique(GetChild(1))`)

src/include/parser/expression/comparison_expression.h
```diff
@@ -0,0 +1,38 @@
...
+   * @param cmp_type type of comparison
+   * @param children vector containing exactly two children, left then right
+   */
+  explicit ComparisonExpression(const ExpressionType cmp_type,
+                                std::vector<std::unique_ptr<AbstractExpression>> *children)
+      : AbstractExpression(cmp_type, type::TypeId::BOOLEAN, std::move(*children)) {}
+
+  AbstractExpression *Copy() const override {
+    std::vector<std::unique_ptr<AbstractExpression>> children;
+    children.emplace_back(std::unique_ptr<AbstractExpression>(GetChild(0)));
```
tli2: This is a bug I believe and will result in double-freeing, since you now have two unique pointers pointing to the same underlying data.

I believe you need to call `GetChild(0)->Copy()` instead of `GetChild(0)`, like in peloton and in the rest of this PR

src/include/parser/expression/function_expression.h
```diff
@@ -0,0 +1,66 @@
...
+  AbstractExpression *Copy() const override { return new FunctionExpression(*this); }
+
+  /**
+   * Returns the name of the function.
+   * @return function name
+   */
+  const std::string &GetFuncName() const { return func_name_; }
+
+ protected:
+  // TODO(WAN): clang-tidy wants me to =default, how does that work?
```
tli2: clang-tidy wants you to default because I believe you are doing the same thing a default copy-constructor does. Declaring something default might allow the compiler to reason some more about things.

Speaking of copy constructors. Why do we need one here and nowhere else?

tli2:
There's a lot of suboptimal C++ inherited in this PR. We should probably get this in anyways in order to make forward progress, but make it clear that this code is not exemplar, to stop bad practices from propagating in the codebase like they did in peloton. (Some sort of warning header?)

tli2:
It turns out some of these code was never compiled since they are not included anywhere and are headers. I made a cleanup path across the PR and added all these headers temporarily to the main file so at least they are picked up by the compiler and the static checks.

mbutrovich:
It looks like there's a Google Test that includes those files, so we should have picked those up already. Did you rerun CMake after checking out the branch?

pervazea:
If I hear no objections, I will merge this in.

- Increasing the test coverage requires more functionality and integration and will come as we get more infrastructure.
- The expression code is needed for the next step(s) with the parser. The interface to libpg_query in postgresparser.h, uses the expression definitions.

PR #181, opened 2018-09-25
Add BwTree 

yangjuns:
1. Add bwtree implementation to third_party
2. Add basic tests


test/CMakeLists.txt
```diff
@@ -12,3 +12,5 @@ add_dependencies(test_util gtest)
 add_subdirectory(common)
 add_subdirectory(storage)
 add_subdirectory(transaction)
+add_subdirectory(index)
```
mbutrovich: I think index will just live in storage, maybe as a subdirectory if we get enough files for that.

yangjuns: ok :)

PR #183, opened 2018-09-27
Travis/Jenkins trigger support

### LINKED ISSUE #1: Set up Jenkins
Set up Jenkins for Terrier.
As stated in Peloton PR #1458, re-purpose the MemSQL machines. These can be the starting point for systems used for Terrier. Machines currently used for Peloton can be migrated over to Terrier as activity increases.

Ubuntu 18.04 only.
- C++ 17 (= gcc7, clang 6)
- LLVM 6.0
- cmake 3.2

At the moment the repo has just a skeleton setup in it. Matt Butrovich has been setting up the Cmake components.

There are no code review requirements, so feel free to commit Jenkins file and update as needed.

### LINKED ISSUE #33: CI Improvements
The plan for our CI setup is:
**Travis:**

- [x] packages.sh
- [x] check-format
- [x] check-clang-tidy
- [x] lint
- [x] Doxygen
- [x] DEBUG: make

**Jenkins:**

- [x] DEBUG: make unittest (-DTERRIER_USE_ASAN=ON)
- [x] DEBUG: make unittest (-DTERRIER_GENERATE_COVERAGE=ON)
- [x] RELEASE: make unittest (-DBUILD_WARNING_LEVEL=Production)
- [x] RELEASE: make runbenchmark (-DBUILD_WARNING_LEVEL=Production)
- [x] performance and memory
- [ ] coveralls

TODO (me): quick script to fail build if coverage drops by 5%

crd477:
Fix #1.
Fix #33.

Setup a trigger from Travis to Jenkins when the preliminary checks in Travis succeed.  Also, the targets identified as, "Will be replaced by Jenkins", in `.travis.yml` have been moved into `Jenkinsfile`.  This should definitely close #1 and accurately reflect the current state of #33.  I still need to check on the unaddressed coveralls check in that issue.

Once this is merged, the "Periodically if not otherwise run" option should be unchecked in the terrier Jenkins pipeline configuration.


Jenkinsfile
```diff
@@ -1,19 +1,39 @@
...
+        GenericTrigger(causeString: 'Travis builds succeeded', token: '5331146c7f6f99b995e8')
+    }
     stages {
         stage('Build') {
             parallel {
 
+                stage('macOS 10.13/Apple clang-902.0.39.2/llvm-6.0.1 (Debug/ASAN)') {
+                    agent { label 'macos' }
+                    environment {
+                        PATH="/usr/local/opt/llvm/bin:$PATH"
```
mbutrovich: Does it work if we don't have this PATH line? Our FIndClangTools cmake stuff actually expect that not to be the case (it's written to seek out those folders to get clang-tidy and clang-format) and in fact adding that to PATH might mean we end up building with the clang version in there, rather than the stock Xcode AppleClang version that we expect.

Jenkinsfile
```diff
@@ -36,15 +56,29 @@ pipeline {
                     }
                 }
 
+                stage('macOS 10.13/Apple clang-902.0.39.2/llvm-6.0.1 (Release/unittest)') {
+                    agent { label 'macos' }
+                    environment {
+                        PATH="/usr/local/opt/llvm/bin:$PATH"
```
mbutrovich: Same comment as above.

mbutrovich:
Thanks for helping us out, Chad! This looks good.

mbutrovich:
The Coveralls on Jenkins stuff is looking increasingly daunting. I'd recommend we merge this with Coveralls still on Travis for now while we explore migrating Coveralls to Jenkins.

crd477:
Sounds good to me.

mbutrovich:
So right now Jenkins runs when the PR is generated, and then runs again when the web hook executes on Travis. Regarding your comment:

> Once this is merged, the "Periodically if not otherwise run" option should be unchecked in the terrier Jenkins pipeline configuration.

Is that what will stop the first build from running, leaving us with just one? Just trying to get a better understanding of the pipelines.

crd477:
Yes, that will stop the double builds. 

    -Chad

mbutrovich:
The CMake script we’re using to communicate to Coveralls is written for Travis. I tried changing the environment variables to look at Jenkins’ in a way that made sense, but it didn’t work. Coveralls doesn’t document the format of input for Jenkins, and I fell down a “coveralls jenkins” Google rabbit-hole of deprecated plugins.

We can look more, but I’m not seeing a quick fix anywhere just yet.

mbutrovich:
Coveralls knows how to talk to Travis so we don't need to embed a repo_token in our scripts. That's not the case with Jenkins. In theory we'd need to store it on a file in the Jenkins image, and then read it out to send with the Coveralls data.

mbutrovich:
I'm out of ideas after that last commit. I think the next step would be figuring out a way to pass the Coveralls token along from a file in the Jenkins box.

Not convinced switching to Codecov isn't a better use of the engineering time though.

PR #185, opened 2018-09-28
Add benchmarks for GC unlinking and deallocating

### LINKED ISSUE #153: Add GarbageCollector benchmark
Look at #149 to see how we measured the performance of the GarbageCollector, and then implement a Google Benchmark that reflects this.

mbutrovich:
Went after some low hanging fruit.

Fix #153.


mbutrovich:
Will need to chat with @pervazea on Monday about the ManualTime stuff.

mbutrovich:
Thanks for pushing that change, @pervazea. Now I'm noticing that the execution time for the benchmark script went from about 2.5 minutes to about 10 minutes. It seems the iteration count is extremely high for these 2 new tests. One ended up with almost 200 iterations, and the other almost 500. Do we think that the execution time of the timed section is too short? Should I try to make the workload larger for it in order to get the iteration count down?

PR #186, opened 2018-09-28
SqlTable

### LINKED ISSUE #140: SqlTable
- [x] `tuple_id` discussion
  - whether we need it to represent logical deletes. **Not really**
  - whether having a layer of indirection has any benefits (in indexes, logging, elsewhere) **Probably not**
  - secondary indexes, logical or physical? **Physical, but HyPer did logical, so we should ask them.**
- [x] Should `SqlTable` own indexes or `oid`s?  **Own**
  - might make sense to just have`SqlTable` own indexes until we have Catalog, and make decision then
- [x] Column access? (immutable flag / synopsis needed) **Punt**
- [x] Recycling can be written now that logical deletes are useful
    - this includes potential optimization to the `DataTable` to reuse blocks. It didn't make sense because we effectively never remove tuples before.

mbutrovich:
Fix #140.

It's mostly a wrapper of DataTable except that SqlTable is instantiated from a Schema, which then generates a BlockLayout to instantiate a DataTable from, generating a map from Column col_oids to BlockLayout col_ids for use in the binding phase.

I also added ProjectedRowInitializer and ProjectedColumnInitializers that can be called with just col_oids. We want to keep col_id as a purely storage abstraction. I also added constraints that columns must be distinct in a projection. If there are repeats in the original query, handle that early in the execution pipeline and only materialize once.

It's not clear how much testing I should have at this layer right now. We'd really just be exercising the DataTable. I added testing for the Schema -> BlockLayout and corresponding ColumnMap.

Once indexes and such are in, the logic will get more complicated at this layer.

### DIFF
--- a/src/include/storage/sql_table.h
+++ b/src/include/storage/sql_table.h
@@ -1,8 +1,11 @@
 #pragma once
+#include <list>
 #include <utility>
 #include <vector>
 #include "catalog/schema.h"
 #include "storage/data_table.h"
+#include "storage/projected_columns.h"
+#include "storage/projected_row.h"
 #include "storage/storage_defs.h"
 
 namespace terrier::storage {
@@ -10,9 +13,22 @@ namespace terrier::storage {
 /**
  * A SqlTable is a thin layer above DataTable that replaces storage layer concepts like BlockLayout with SQL layer
  * concepts like Schema. This layer will also handle index maintenance, and possibly constraint checking (confirm when
- * we bring in execution layer).
+ * we bring in execution layer). The goal is to hide concepts like col_id_t and BlockLayout above the SqlTable level.
+ * The SqlTable API should only refer to storage concepts via things like Schema and col_oid_t, and then perform the
+ * translation to BlockLayout and col_id_t to talk to the DataTable and other areas of the storage layer.
  */
 class SqlTable {
+  /**
+   * Contains all of the metadata the SqlTable needs to reference a DataTable. We shouldn't ever have to expose these
+   * concepts to anyone above the SqlTable level. If you find yourself wanting to return BlockLayout or col_id_t above
+   * this layer, consider alternatives.
+   */
+  struct DataTableVersion {
+    DataTable *const data_table;
+    const BlockLayout layout;
+    const ColumnMap column_map;
+  };
+
  public:
   /**
    * Constructs a new SqlTable with the given Schema, using the given BlockStore as the source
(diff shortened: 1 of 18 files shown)


mbutrovich:
#187 will merge first, then I'll update this.

src/storage/data_table.cpp
```diff
@@ -20,7 +20,7 @@ bool DataTable::Select(terrier::transaction::TransactionContext *txn, terrier::s
   return SelectIntoBuffer(txn, slot, out_buffer);
 }
 
-void DataTable::Scan(transaction::TransactionContext *const txn, SlotIterator *start_pos,
+void DataTable::Scan(transaction::TransactionContext *const txn, SlotIterator *const start_pos,
```
tli2: Fine...

test/include/util/storage_test_util.h
```diff
@@ -78,6 +80,27 @@ struct StorageTestUtil {
     return storage::BlockLayout(attr_sizes);
   }
 
+  // Returns a random Schema that is guaranteed to be valid for a single test instance. Multiple Schemas instantiated
+  // with this function may have non-unique oids, but within a Schema, Column oids are unique.
+  template <typename Random>
+  static catalog::Schema RandomSchema(const uint16_t max_cols, Random *const generator) {
```
tli2: Should this belong in `storage_test_util`, `Schema` being in the namespace `catalog`?

test/storage/storage_util_test.cpp
```diff
@@ -198,4 +197,45 @@ TEST_F(StorageUtilTests, ApplyDelta) {
   }
 }
 
+// Verifies that we properly generate both a valid BlockLayout and a valid mapping from col_oid to col_id for a given
+// random Schema
+// NOLINTNEXTLINE
+TEST_F(StorageUtilTests, BlockLayoutFromSchema) {
```
tli2: Same here. It looks like it's small, but I think we should make it its own test file under `catalog`

src/include/storage/sql_table.h
```diff
@@ -0,0 +1,207 @@
...
+   * of its storage blocks.
+   *
+   * @param store the Block store to use.
+   * @param schema the initial Schema of this SqlTable
+   * @param oid unique identifier for this SqlTable
+   */
+  SqlTable(BlockStore *const store, const catalog::Schema &schema, const table_oid_t oid)
+      : block_store_(store),
+        oid_(oid),
+        table_({new DataTable(block_store_, StorageUtil::BlockLayoutFromSchema(schema).first, layout_version_t(0)),
```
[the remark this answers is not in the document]
mbutrovich: I tried that but BlockLayout's default constructor is deleted, so that doesn't work to not initialize it in the initializer list. I can remove the const label on the fields in BlockLayout and then it works.

src/include/storage/sql_table.h
```diff
@@ -0,0 +1,207 @@
...
+
+  /**
+   * Materializes a single tuple from the given slot, as visible at the timestamp of the calling txn.
+   *
+   * @param txn the calling transaction
+   * @param slot the tuple slot to read
+   * @param out_buffer output buffer. The object should already contain projection list information. @see ProjectedRow.
+   * @return true if tuple is visible to this txn and ProjectedRow has been populated, false otherwise
+   */
+  bool Select(transaction::TransactionContext *const txn, const TupleSlot slot, ProjectedRow *const out_buffer) const {
```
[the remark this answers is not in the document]
mbutrovich: Not sure yet! We'll have a better idea as we being in executors. I'm just guessing on a lot of this functionality right now.

test/storage/storage_util_test.cpp
mbutrovich: I don't agree with this one. The function being tested is BlockLayoutFromSchema in storage_util

test/include/util/storage_test_util.h
mbutrovich: Sure

src/storage/data_table.cpp
mbutrovich: :)

src/include/storage/sql_table.h
```diff
@@ -0,0 +1,207 @@
...
+  /**
+   * Generates an ProjectedColumnsInitializer for the execution layer to use. This performs the translation from col_oid
+   * to col_id for the Initializer's constructor so that the execution layer doesn't need to know anything about col_id.
+   * @param col_oids set of col_oids to be projected
+   * @param max_tuples the maximum number of tuples to store in the ProjectedColumn
+   * @return pair of: initializer to create ProjectedColumns, and a mapping between col_oid and the offset within the
+   * ProjectedColumn
+   * @warning col_oids must be a set (no repeats)
+   */
+  std::pair<ProjectedColumnsInitializer, ProjectionMap> ProjectionInitializer(const std::vector<col_oid_t> &col_oids,
```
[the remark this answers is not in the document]
mbutrovich: Any name suggestions? I never came up with anything reasonable.

test/storage/storage_util_test.cpp
tli2: Yeah I was thinking we move that one too. Low priority, can always get that later.

Issue #179, opened 2018-09-25
Compact logical delete column


tli2:
Right now the logical delete column is an 8-byte column with nothing in the tuple, and we use its null bitmap to denote logical deletion. (Always the second column)

This is because we used the first column (version vector)'s bitmap to denote whether a slot is taken or not. It is probably better to push that bitmap (slot validity) into the block header, and use the version vector's bitmap for logical delete.

PR #187, opened 2018-09-28
column access API

### LINKED ISSUE #144: Column access for DataTable
Implement a column access API for DataTable that returns a raw column

tli2:
First-draft implementation to provide column access to `DataTable`.

Still need to write documentation and tests. But since the current version is a naive call to select repeatedly, it should not be much of an effort. 

We need add more efficient versions later (direct memcpy when possible). Fix #144


src/include/storage/storage_defs.h
```diff
@@ -48,86 +40,6 @@ struct RawBlock {
   // store offsets within a block in ine 8-byte word.
 } __attribute__((aligned(common::Constants::BLOCK_SIZE)));
```
mbutrovich: Since we're poking around in this code right now, can we switch this to the standardized [alignas](https://en.cppreference.com/w/cpp/language/alignas)?

src/include/storage/storage_defs.h
```diff
@@ -189,11 +101,13 @@ class TupleSlot {
     return os << "block: " << slot.GetBlock() << ", offset: " << slot.GetOffset();
   }
 
+  static TupleSlot INVALID;
```
mbutrovich: What's this for?

src/include/storage/tuple_access_strategy.h
```diff
@@ -122,7 +133,14 @@ class TupleAccessStrategy {
    */
   void InitializeRawBlock(RawBlock *raw, layout_version_t layout_version) const;
 
-  /* Vectorized Access */
+  /**
+   * @param slot tuple slot value to check
+   * @return whether the given slot is occupied by a tuple
+   */
+  bool ValidSlot(TupleSlot slot) const {
```
mbutrovich: Naming isn't super important and we can debate this, but I prefer something like Allocated or Present, especially if we're still calling Allocate() to get a slot in the first place. Valid implies to me there's some extra meaning or verification going on.

src/include/storage/write_ahead_log/log_record.h
```diff
@@ -6,7 +6,7 @@ namespace terrier::storage {
 /**
  * Types of LogRecords
  */
-enum class LogRecordType : uint8_t { REDO = 1, COMMIT };
+enum class LogRecordType : uint8_t { REDO = 1, DELETE, COMMIT };
```
mbutrovich: Move to storage_defs where DeltaRecordType lives.

src/storage/block_layout.cpp
```diff
@@ -0,0 +1,48 @@
+#include <lzma.h>
```
mbutrovich: Stray include, I think.

src/storage/materialized_columns.cpp
```diff
@@ -0,0 +1,52 @@
+
+#include <storage/materialized_columns.h>
```
mbutrovich: I think CLion gave you a bracketed auto-include instead of quotes.

src/storage/data_table.cpp
```diff
@@ -198,22 +173,25 @@ bool DataTable::IsVisible(const transaction::TransactionContext &txn, const Tupl
...
-      visible = false;
-    } else if (undo_type == DeltaRecordType::DELETE) {
-      // Applying the undo of a DELETE makes the tuple visible to this txn.
-      visible = true;
+    switch (version_ptr->Type()) {
+      case DeltaRecordType::UPDATE:
+        // Normal delta to be applied. Does not modify the logical delete column.
+        StorageUtil::ApplyDelta(accessor_.GetBlockLayout(), *(version_ptr->Delta()), out_buffer);
+        break;
+      case DeltaRecordType::INSERT:visible = false;
```
mbutrovich: I think the removed comments might be helpful to a new person looking at the code a couple years from now.

test/storage/tuple_access_strategy_test.cpp
```diff
@@ -152,7 +150,7 @@ TEST_F(TupleAccessStrategyTests, MemorySafety) {
     tested.InitializeRawBlock(raw_block_, layout_version_t(0));
 
     // Skip header
-    void *lower_bound = tested.ColumnNullBitmap(raw_block_, PRESENCE_COLUMN_ID);
+    void *lower_bound = tested.ColumnNullBitmap(raw_block_, col_id_t(0));
```
mbutrovich: VERSION_POINTER_COL_ID

tli2:
This PR means we can do a lot more testing with interleaved sequential scanners and writers, and so on. The testing is minimal here. but I want to get this in first and then expand our test suite in a separate PR (maybe have some fresh eyes on the way we do testing too)

tli2:
Also fix #179 now

src/include/storage/data_table.h
```diff
@@ -32,6 +33,76 @@ DEFINE_PERFORMANCE_CLASS(DataTableCounter, DataTableCounterMembers)
...
+
+    /**
+     * pre-fix increment.
+     * @return self-reference after the iterator is advanced
+     */
+    SlotIterator &operator++() {
+      common::SpinLatch::ScopedSpinLatch guard(&table_->blocks_latch_);
+      if (current_slot_.GetOffset() == table_->accessor_.GetBlockLayout().NumSlots()) {
+        ++block_;
+        current_slot_ = {block_ == table_->blocks_.end() ? nullptr : *block_, 0};
```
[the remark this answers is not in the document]
mbutrovich: Nevermind, I misread the ordering on this line!

PR #192, opened 2018-10-04
Thread pool

yangjuns:
1. satisfy llvm thread pool dependency
2. add thread pool tests


test/include/util/test_thread_pool.h
```diff
@@ -34,61 +27,21 @@ class TestThreadPool {
    * @param repeat the number of times this should be done.
    */
   void RunThreadsUntilFinish(uint32_t num_threads, const std::function<void(uint32_t)> &workload, uint32_t repeat = 1) {
```
tli2: Take in an thread pool and make this function a static util under some utility class

src/include/common/worker_pool.h
```diff
@@ -0,0 +1,159 @@
...
+   * it cannot be changed.
+   *
+   * After initialization of the worker pool with a given task queue, worker threads
+   * do NOT start running. Need call StartUp() to start working on tasks.
+   *
+   * @param pool_name the name of the worker pool
+   * @param num_workers the number of workers in this pool
+   * @param task_queue a queue of tasks
+   */
+  WorkerPool(std::string pool_name, uint32_t num_workers, TaskQueue task_queue)
```
tli2: I don't think this has to be fixed size? Granted it provides some immutability, but:

- we want to be able to tune these things at runtime.
- it doesn't seem to be very hard to change the number of threads in the thread pool (in either direction)
- (not really a good reason, but the test thread pool kind of need a flexible number of threads)

src/include/common/worker_pool.h
```diff
@@ -0,0 +1,159 @@
...
+  /**
+   * Get the number of worker threads in this pool
+   *
+   * @return The number of worker threads
+   */
+  uint32_t NumWorkers() const { return num_workers_; }
+
+ private:
+  // The name of this pool
+  std::string pool_name_;
```
tli2: This seems very unnecessary. Either way there's not even an accessor for it so it is unusable anyway. We can remove it for now.

src/include/common/worker_pool.h
```diff
@@ -0,0 +1,159 @@
...
+    }
+    std::unique_lock<std::mutex> lock(task_lock_);  // grab the lock
+    task_queue_.emplace(std::move(func));
+    task_cv_.notify_one();
+  }
+
+  /**
+   * Block until all the tasks in the task queue has been completed
+   */
+  void WaitUtilFinish() {
```
tli2: Additionally, this is not a very good name. Since you would expect to wait until some specific tasks are finished. Whether we want that kind of behavior is debatable, but at the very least call this `WaitUntilAllFinished`

src/include/common/worker_pool.h
```diff
@@ -0,0 +1,159 @@
...
+   * Side effect: If the pool has been shutdown, calling SubmitTask will automatically
+   * re-startup the pool and resume to work.
+   * @param func the new task
+   */
+  template <typename F>
+  inline void SubmitTask(const F &func) {
+    if (!is_running_) {
+      Startup();
+    }
+    std::unique_lock<std::mutex> lock(task_lock_);  // grab the lock
```
tli2: unnecessary comment

src/include/common/worker_pool.h
```diff
@@ -0,0 +1,159 @@
...
+    task_cv_.notify_all();                          // wake up all the threads
+    lock.unlock();                                  // free the lock
+    for (auto &thread : workers_) thread.join();
+  }
+
+  /**
+   * Start the worker pool. If the task queue is empty or we run out of tasks,
+   * workers will be put into sleep.
+   */
+  void Startup() {
```
tli2: Is there a use case where we want a thread pool that is not started? If not, maybe it is better for the constructor to encapsulate this behavior.

src/include/common/worker_pool.h
```diff
@@ -0,0 +1,159 @@
...
+        ++busy_workers_;
+        auto task = std::move(task_queue_.front());
+        task_queue_.pop();
+        // release the lock while we work
+        lock.unlock();
+        task();
+        // we lock again to notify that we're done
+        lock.lock();
+        --busy_workers_;
+        finished_cv_.notify_one();
```
tli2: Why is this `notify_one()`, when multiple threads can be waiting for the tasks to all complete in practice?

src/include/common/worker_pool.h
yangjuns: A thread pool can be shut down. So we would like a way to start it up again.

src/include/common/worker_pool.h
yangjuns: Only the main thread needs to know whether a worker thread has finished the task. Other workers threads don't need this information. At least, that's the design for now.

tli2: Either document this very clearly as a warning or make this `notify_all()`. Otherwise it will be a very unpleasant surprise for people down the line. This is hardly performance critical and I don't see a downside to calling `notify_all()`.

src/include/common/worker_pool.h
tli2: You can keep this method. I was more asking whether it makes sense to always just call it in the constructor body?

test/storage/data_table_concurrent_test.cpp
```diff
@@ -101,8 +100,8 @@ TEST_F(DataTableConcurrentTests, ConcurrentInsert) {
       std::default_random_engine thread_generator(id);
       for (uint32_t i = 0; i < num_inserts / num_threads; i++) fake_txns[id]->InsertRandomTuple(&thread_generator);
     };
-
-    thread_pool.RunThreadsUntilFinish(num_threads, workload);
+    common::WorkerPool thread_pool;
```
mbutrovich: Why did we make it so that a new thread_pool is created every iteration?

test/storage/data_table_concurrent_test.cpp
```diff
@@ -147,8 +145,8 @@ TEST_F(DataTableConcurrentTests, ConcurrentUpdateOneWriterWins) {
       else
         fail++;
     };
-
-    thread_pool.RunThreadsUntilFinish(num_threads, workload);
+    common::WorkerPool thread_pool;
```
mbutrovich: Same question.

test/storage/varlen_pool_test.cpp
```diff
@@ -66,8 +65,8 @@ TEST(VarlenPoolTests, ConcurrentCorrectnessTest) {
 
       RandomTestUtil::InvokeWorkloadWithDistribution({free, allocate}, {0.2, 0.8}, &generator, 100);
     };
-
-    thread_pool.RunThreadsUntilFinish(num_threads, workload);
+    common::WorkerPool thread_pool;
```
mbutrovich: Same question.

src/include/common/worker_pool.h
```diff
@@ -0,0 +1,173 @@
...
+   * Initialize the worker pool. Once the number of worker is set by the constructor
+   * it cannot be changed.
+   *
+   * After initialization, the worker pool automatically starts up.
+   *
+   * @param num_workers the number of workers in this pool
+   * @param task_queue a queue of tasks
+   */
+  // NOLINTNEXTLINE  lint thinks it has only one arguement
+  WorkerPool(uint32_t num_workers = 1, TaskQueue task_queue = TaskQueue())
```
mbutrovich: We're avoiding default arguments in the new codebase. TERRIER_ASSERT that it's > 0 in the body of the constructor.

mbutrovich:
The last run on Travis looks like it might have deadlocked (it timed out).

yangjuns:
Yes. The bug is not fixed by this commit. I haven't pushed the corrected version because I need to fix some other stuff (API, design, documentation).

mbutrovich:
I'll also add that while I think this suffices for our use cases right now, we might want to look at Arrow's thread pool to see if there's anything we should incorporate.

https://github.com/apache/arrow/blob/master/cpp/src/arrow/util/thread-pool.h
https://github.com/apache/arrow/blob/master/cpp/src/arrow/util/thread-pool.cc
https://github.com/apache/arrow/blob/master/cpp/src/arrow/util/thread-pool-test.cc

The use of std::packaged_task and std::future is new to me, but might be worth looking into its intended uses.

mbutrovich:
After looking at the std::future and std::packaged_task stuff a bit more, I think those are the foundations for a more modern C++ thread pool. If the time comes to bolster the functionality of our thread pool implementation, we should look at Arrow's (linked previously).

In the meantime, if @lmwnshn can review this then this can go in.

lmwnshn:
Just started looking at this, sorry. I may try to add some tests and remove dead code if that's alright.

test/common/worker_pool_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+  thread_pool.Shutdown();
+}
+
+// NOLINTNEXTLINE
+TEST(WorkerPoolTests, MoreTest) {
+  common::TaskQueue tasks;
+  common::WorkerPool thread_pool(5, tasks);
+  uint32_t iteration = 10;
+  for (uint32_t it = 0; it < iteration; it++) {
+    // std::cout << "iteration : " << it << std::endl;
```
lmwnshn: Do we want these comments?

test/common/worker_pool_test.cpp
```diff
@@ -0,0 +1,77 @@
...
+    auto workload = [](uint32_t) {
+      // std::cout << "start sleeping" << std::endl;
+      std::this_thread::sleep_for(std::chrono::milliseconds(200));
+      // std::cout << "Done sleeping" << std::endl;
+    };
+
+    MultiThreadTestUtil::RunThreadsUntilFinish(&thread_pool, 12, workload);
```
lmwnshn: MoreTest always uses 12 workers, BasicTest always uses 5. Do we want to randomly generate the number of works?

For what it's worth, I tried changing to a variety of pool sizes and the test still passes.

PR #207, opened 2018-10-11
Log performance fix

tli2:
WAL was limiting transaction throughput even when not waiting on disk flush because of contention at a global callback map. It no longer does that as we now embed callback into a commit record. We achieve comparable performance to when logging is turned off on dev 4.  More detailed charts are coming.

Retardedness in the log IO code is also gone.


mbutrovich:
At some point a flowchart describing the process of Persist(), Flush(), BufferWrite(), etc., who calls them, and under what circumstances would be helpful. It can be a bit hard to piece the whole process together just from code.

PR #218, opened 2018-10-16
Micro bench updates 

pervazea:
Update benchmark configuration

-  Use history from PR #200 and later (= build 28)
- Use all builds, not just successful ones. PR #200 caused builds to fail
   due to the performance changes.   
- Don't use pa-terrier any more, now that we have enough history in terrier-nightly


tli2:
@pervazea Do we need to go back after a while and flip back to using only builds that passed?

pervazea:
Maybe not. Now that we have nightly builds, as long as don't merge in erroneous performance changes, the results from nightly should reflect where we are. We'll have to see, it isn't completely clear how it will work out.

PR #221, opened 2018-10-17
read-only transaction optimization

tli2:
They don't really need to hold latches when committing. This brings our readonly performance from 600 to 770 k txns / s on 8 threads on dev 4 (same benchmark as #220)

Some performance degradation still happens at higher thread counts( > 10), but overall not as severe.


mbutrovich:
`pass 1.269e+06 1.017e+06 30     +24    lax      6    LargeTransactionBenchmark    SingleStatementSelect
`

That's a nice win.

PR #223, opened 2018-10-23
My modifs: Add common files.

amlatyrngom:


src/include/common/synchronization/readwrite_latch.h
```diff
@@ -0,0 +1,53 @@
+//===----------------------------------------------------------------------===//
```
mbutrovich: We can lose this file.

src/CMakeLists.txt
```diff
@@ -27,5 +27,5 @@ ADD_TERRIER_LIB(terrier
 ###############################################
 # Terrier executable
 ###############################################
-add_executable(terrier main/terrier.cpp)
+add_executable(terrier main/terrier.cpp include/common/container_tuple.h include/common/abstract_tuple.h include/common/exception.h include/common/internal_types.h include/common/item_pointer.h include/common/platform.h include/common/singleton.h include/common/timer.h include/common/synchronization/readwrite_latch.h include/common/synchronization/count_down_latch.h)
```
mbutrovich: This shouldn't be necessary.

src/include/common/item_pointer.h
```diff
@@ -0,0 +1,75 @@
+
```
mbutrovich: We can lose this file.

PR #224, opened 2018-10-23
Stop serializing commit record for read-only transactions

### LINKED ISSUE #209: Write-ahead-log code path can be optimized for read-only transactions
As it stands, read-only transactions also generate commit records. Since we use commit records to embed callbacks for group-commit, we cannot remove them. However, there is no reason for them to be actually serialized and written out to disk.

It is possible to structure commit records to somehow convey that a transactions is read-only (commit_time == begin_time, things like that), and check for this in the `LogManager`. We can then process the callback only and not write the record out to disk.

swimj:
Added an IsReadOnly function to CommitRecord to indicate if the transaction generating that commit record was read-only or not.  When the logger is processing it's flush queue, if it encounters a CommitRecord, it uses this function to determine whether it should skip serialization. For read-only transactions, there is no need to write out the commit record to anywhere, since no modifications were made.  In that case, the commit record is only needed for invoking the transaction's callback function.

fix #209.


test/storage/log_test.cpp
```diff
@@ -47,8 +47,10 @@ class WriteAheadLoggingTests : public TerrierTest {
     auto txn_begin = in->ReadValue<timestamp_t>();
     if (record_type == storage::LogRecordType::COMMIT) {
       auto txn_commit = in->ReadValue<timestamp_t>();
-      // Okay to fill in null since nobody will invoke the callback
-      return storage::CommitRecord::Initialize(buf, txn_begin, txn_commit, nullptr, nullptr);
+      // Okay to fill in null since nobody will invoke the callback.
+      // is_read_only argument is set to false, because we do not write out a commit record for a transaction if it is
+      // not read-only.
+      return storage::CommitRecord::Initialize(buf, txn_begin, txn_commit, nullptr, nullptr, false);
```
tli2: Can you add a test to verify that read-only transactions are not logged out?

swimj:
Adjusted LargeLogTest to have a 50-50 update-select ratio (previously was all selects) and properly account for read-only transactions with this feature implemented.

Added ReadOnlyTransactionsGenerateNoLogTest to test that no log records are generated by read only transactions, which is the primary goal of this PR.

tli2:
You need to fix documentation for something according to Travis

PR #228, opened 2018-10-24
Fix object pool exceptions

swimj:
While expanding tests in log_test.cpp, I exceeded the number of RecordBuffers.  This triggered an object pool exception, but no message was printed to my terminal, because the message is returned as a (char *) to a stack-allocated string, which turned into junk as soon as the function returning the message exited.

Rather than overriding std::exception's what(), I switched to derivatives std::length_error and std::runtime_error, which take the message to be output as a constructor argument. Presumably it handles the memory in a safe way.

Tested the fix locally by running my same under-allocated test (e.g. in log_test set update-select ratio to {0.5, 0.5} and the object pool size to be something small, like 10).  Now, when the allocator runs out of objects, an informative message is printed out before the test gets killed.


yangjuns:
Let me try to add a test that captures this issue.

test/common/object_pool_test.cpp
```diff
@@ -34,6 +34,46 @@ TEST(ObjectPoolTests, SimpleReuseTest) {
...
+TEST(ObjectPoolTests, ExceedLimitTest) {
+  const uint32_t repeat = 1;
+  const uint64_t size_limit = 10;
+  const uint64_t reuse_limit = 10;
+  for (uint32_t iter = 0; iter < repeat; iter++) {
+    common::ObjectPool<uint32_t> tested(size_limit, reuse_limit);
+
+    // Get 11 objects
+    std::vector<uint32_t *> objects;
+    for (uint32_t i = 1; i <= 11; i++) {
```
swimj:
```suggestion
    for (uint32_t i = 1; i <= size_limit + 1; i++) {
```

test/common/object_pool_test.cpp
```diff
@@ -34,6 +34,46 @@ TEST(ObjectPoolTests, SimpleReuseTest) {
...
+      uint32_t *cur_ptr = nullptr;
+      try {
+        cur_ptr = tested.Get();
+        if (i == 11) {
```
swimj:
```suggestion
        if (i == size_limit + 1) {
```

test/common/object_pool_test.cpp
```diff
@@ -34,6 +34,46 @@ TEST(ObjectPoolTests, SimpleReuseTest) {
...
+          // free memory before we fail
+          tested.Release(cur_ptr);
+          for (auto &ptr : objects) tested.Release(ptr);
+          FAIL() << "Expect std::length_error.";
```
swimj:
```suggestion
          FAIL() << "Expect std::length_error when attempting to get object from pool with size limit " << size_limit 
          << " and " << i - 1 << " objects already allocated.";
```

test/common/object_pool_test.cpp
```diff
@@ -34,6 +34,46 @@ TEST(ObjectPoolTests, SimpleReuseTest) {
...
+        }
+        objects.push_back(cur_ptr);
+      } catch (const std::length_error &e) {
+        // got length error
+        if (i <= 10) {
+          // free memory before we fail
+          for (auto &ptr : objects) tested.Release(ptr);
+          FAIL() << "Unexpected std::length_error";
```
swimj:
```suggestion
          FAIL() << "Unexpected std::length_error; object pool has not allocated to its size limit yet.";
```

test/common/object_pool_test.cpp
```diff
@@ -34,6 +34,46 @@ TEST(ObjectPoolTests, SimpleReuseTest) {
...
+        }
+      } catch (...) {
+        // free memory before we fail
+        for (auto &ptr : objects) tested.Release(ptr);
+        FAIL() << "Unexpected exceptions";
```
swimj: Is there an easy way to print what exception actually got raised? If not, the test is small enough where this is probably easy to decipher.

test/common/object_pool_test.cpp
```diff
@@ -34,6 +34,46 @@ TEST(ObjectPoolTests, SimpleReuseTest) {
   }
 }
 
+// Allocate more memory space than the object pool and expect exceptions
+// NOLINTNEXTLINE
+TEST(ObjectPoolTests, ExceedLimitTest) {
+  const uint32_t repeat = 1;
+  const uint64_t size_limit = 10;
+  const uint64_t reuse_limit = 10;
```
swimj:
```suggestion
  const uint64_t reuse_limit = size_limit;
```

Nit: maybe makes it slightly clearer that the value of this variable isn't relevant to what is being tested?

test/common/object_pool_test.cpp
yangjuns: To my understanding, c++ allows you to throw any type of exceptions (int, char*, std::exception ...). I can't think of a way to catch any type. 

(Maybe boost::any can do the job, but I don't think that's necessary.)

src/include/common/object_pool.h
```diff
@@ -92,13 +52,15 @@ class ObjectPool {
 
   /**
    * Returns a piece of memory to hold an object of T.
-   * @throw NoMoreObjectException if the object pool fails to fetch memory.
-   * @throw AllocatorFailureException if the allocator fails to return a valid memory address.
+   * @throw std::length_error if the object pool has reached the limit of how many objects it may hand out.
```
tli2: I am not sure this is the right exception to use.

test/common/object_pool_test.cpp
tli2: There should be EXPECT_THROW or something like that in gtest that you can use instead of this

src/include/common/object_pool.h
swimj: Hmm, do you want to go back to our own-defined exception then?  According to [this](https://en.cppreference.com/w/cpp/error/length_error), length_error "reports errors that result from attempts to exceed implementation defined length limits for some object."

swimj: It's easy enough to go back to our custom exceptions so I'll just do that.

yangjuns:
Justin and I were having the conversation about if we should have a test testing whether the exception message contains valid strings. I feel uncomfortable about testing hard coded strings but Justin argued that it can be done with gmock. We kinda need a third opinion. @tli2

tli2:
I don't think you need to test strings. See https://github.com/google/googletest/blob/master/googletest/docs/advanced.md for how to test for errors and exceptions

tli2:
It is okay to not hit diff coverage target here since it is a bug fix on exception handling

PR #233, opened 2018-11-05
Parser

lmwnshn:
Still WIP.

**TODO: More tests for coverage**

As of 11/27: usable, bug and feature (almost) parity with peloton except for varlen/string-related stuff.

Some xyzStatement should have been split up into multiple ones (tagged union kind of thing), maybe adding some asserts to the accessors that check the tag would be a good idea. 

Also C++ quality in here is terrible and should not be used as an example.


pervazea:
I added some test updates to try to improve code coverage, but still short of where we'd like to be. Observations:
1. We are a couple of hundred lines short on desired code coverage. No single easy fix => lots of test cases needed.

2. Some of the tests added exercise function definition SQL. What is supported syntax was not at all clear. Trying to deduce this from the yacc input file was not enlightening. Notably, _lang_name_ of plpgsql is accepted, but SQL is not. 

The result should be documented (on the wiki) as well as supplemented with unit tests.

3. We need varlen support from the "ValueFactory" and related components, in order to exercise things like type casts, as well as anywhere where a string is used, e.g. CREATE TABLE foo.

pervazea:
Had several discussions over the last few days. Summarizing what else we need to do:

1. Would like to address code coverage concerns. 

- Remove code that isn't going to be used in the near future. That means UDF support. I'd ported the tests from Peloton for UDFs, but those are incomplete in several ways. Rather than try to cover something that we can't use for a long time, take out, and bring in UDF support later, when we need it.
- White list the parser in codecov (Matt B), so we can correctly monitor coverage in other components.

2. Add string support (PA). I've got possible code for this. For the parser, we need support for strings. The feeling was to use C strings  (rather than std::string). This will allow us to proceed with some of the transforms and unit tests.

3. Add NULL support, preferably as a null type. This will be different from Peloton, where NULLs were typed and implemented as sentinel values.

4. Fix the places where we throw exceptions. Ok, to simplify those code points and not dynamically construct strings. Constant strings would be ok. 

5. Review and improve docstrings, where better descriptions are warranted.

pervazea:
Issues, possibly needing attention:
1. Strings. Who owns the string? Who frees it? Preventing double frees? 
I think at the point the Value object is created, it "takes ownership" of the string from the postgres parser (rather than copy it). Need to zero out the string ptr in the postgres parser structure.

Issue #237, opened 2018-11-08
clang-tidy shows usage info


tli2:
See travis output for https://api.travis-ci.org/v3/job/449990761/log.txt. This happens locally as well if you run `make check-clang-tidy`. Checks do seem to still happen for the rest of the files.

pervazea:
This is annoying.

Running things manually, it appears that it spits out the usage information, every time we invoke it on libpg_query files, where we've put a clang-tidy configuration that suppresses all checks. Running -list-checks says that "no checks are enabled", which is as intended. As far as I can tell, it is not complaining about the syntax of the clang-tidy configuration.

So, from what I can tell:
- it is correctly checking everything desired
- it complains about all files in suppressed directories (libpg_query). 

Need to find a way to quiet down the spurious complaints.

pervazea:
Confirming that in clang-tidy-6.0, this situation returns 0, i.e. does not result in an error being returned.
In the latest version of the code, this will result in an error (returns 1)

mbutrovich:
We should talk to @lmwnshn about why we went away from Arrow's approach of invoking clang-tidy from a shell script that greps the results against a blacklist. I seem to remember it suppressed the actual error messages and just gave a vague "clang-tidy found errors" message at the end, but we may need to go down that route if we really need to suppress output.

pervazea:
An alternate to wrapping run-clang-tidy.py with a shell script, is to modify the former. It looks easy.  Run clang-tidy, except when the directory or file is black-listed. 

Are there cases where we want to run clang-tidy directly, i.e. without going through the run-clang-tidy.py script?

PR #243, opened 2018-11-16
Fix for #237 - clang-tidy show usage

pervazea:
Suppress invocation of clang-tidy, for selected directories (or files)


build-support/run_clang_tidy_extra.py
```diff
@@ -0,0 +1,46 @@
...
+        self._init_config()
+        # debug prints
+        self.debug = False
+        return
+
+    def _init_config(self):
+        """ Any path matching one of the ignore_pats regular expressions,
+            denotes that we do NOT want to run clang-tidy on that item.
+        """
+        self.ignore_pats = [".*/third_party/libpg_query/.*", ]
```
mbutrovich: Should we make it all of third_party right now, just preemptively?

pervazea: Sure.

PR #244, opened 2018-11-16
Reclaim slots when deleted in GC

tli2:
One of those things we never got around to doing. This is surprisingly easy given the way we do GC and adds minimal overhead.

I am reasonably confident that this is correct because the change is minor, and with basic tests it works. To test this fully though we need to extend the current random testing framework to handle inserts and deletes as well, which would not be easy as the current version is being stretched close to its limit already. This might be a great project. Will write a separate issue with details.

### DIFF
--- a/src/include/storage/data_table.h
+++ b/src/include/storage/data_table.h
@@ -177,6 +177,16 @@ class DataTable {
   }
 
   /**
+   * @warning NOT the number of tuples currently in the data table. Such a value is not well-defined unless referring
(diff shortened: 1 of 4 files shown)


mbutrovich:
We should note that this PR ins't doing full slot recycling across the table because only slots that are reclaimed in the insertion head block can be reused by an Insert. Slots not in the insertion head block are simply being marked as deallocated and can be used in the future by something like a background compaction thread.

PR #246, opened 2018-11-16
Port BwTree from Peloton

mbutrovich:
I ported the BwTree from Peloton as it exists here:
https://github.com/cmu-db/peloton/tree/master/src/index

Added:
third_party/bwtree/atomic_stack.h
third_party/bwtree/bloom_filter.h
third_party/bwtree/bwtree.cpp
third_party/bwtree/bwtree.h
third_party/bwtree/sorted_small_set.h

This required C++17 updates and changes to the debug logging. The BwTree uses its own index_logger so as not to pollute the storage_logger output. I also ran the formatter and applied some basic linter suggestions. It's still not fully compliant with the repository requirements so it will stay in third_party.

I ported all of the correctness tests from @wangziqi2013's repo here:
https://github.com/wangziqi2013/BwTree/tree/master/test

Added:
test/include/util/bwtree_test_util.h
test/storage/bwtree_test.cpp

This required C++17 updates, tidy/lint/formatting compliance, changes to use Google Test semantics, and changes to use our thread pool. I couldn't put the test file in test/third_party because then it would be skipped by lint, tidy, etc. which ignore paths with "third_party" in them. Storage seemed like the closest relative.

I ported some of the benchmarks from @wangziqi2013's repo here:
https://github.com/wangziqi2013/BwTree/tree/master/test

Added:
benchmark/storage/bwtree_benchmark.cpp

I didn't make these part of the CI benchmark script because we don't want the BwTree performance changing to fail us out of CI. When we add our bwtree_index tests and benchmarks we should add those to CI.

Note that this PR _does not_ include the wrapper we'll need for the storage layer to use the BwTree. This just has the plain BwTree class, its dependencies, and targeted tests. The next step after this PR is to write a new wrapper and port Peloton's index tests.


mbutrovich:
Will need @pervazea's help tomorrow to understand why tidy only hates the bwtree after adding benchmarks. tidy was fine when it was just Google Tests.

mbutrovich:
I think that's all of the changes for this PR. CI seems happy. I'll start on the wrapper next, but this can go in first.

PR #249, opened 2018-11-23
Reverse byte order in Block allocation bitmap

### LINKED ISSUE #242: Reverse byte ordering in bitmap for arrow compatibility
Arrow uses least significant bit ordering for its bitmaps in a single byte. (e.g. 0000 0000 0000 0001 means that the 9th element is 1). Our code currently does not do this, but this should be a rather effortless thing to do in order to conform to their standards.

All these accesses should already be hidden behind methods.

amlatyrngom:
Fix #242


src/include/common/container/bitmap.h
```diff
@@ -15,6 +15,8 @@ static_assert(BYTE_SIZE == 8u, "BYTE_SIZE should be set to 8!");
 #define ONE_HOT_MASK(n) (1u << (BYTE_SIZE - (n)-1u))
```
tli2: I think we can remove this at this time if it's no longer used?

src/include/common/container/concurrent_bitmap.h
```diff
@@ -204,12 +206,8 @@ class RawConcurrentBitmap {
...
     if (bits == static_cast<T>(-1)) {
       *byte_pos += static_cast<uint32_t>(sizeof(T));
-      // prevent underflow
-      if (*bits_left < sizeof(T) * BYTE_SIZE) {
-        *bits_left = 0;
-      } else {
-        *bits_left = *bits_left - static_cast<uint32_t>(sizeof(T) * BYTE_SIZE);
-      }
+      // Assumes IsAlignedAndFits<T> is called before, so no need to check for underflow.
+      *bits_left = *bits_left - static_cast<uint32_t>(sizeof(T) * BYTE_SIZE);
```
tli2: Can you document the assumptions and the purpose of this a little more clearly? It is not immediately obvious to me this is true.

src/include/common/container/concurrent_bitmap.h
```diff
@@ -204,12 +206,10 @@ class RawConcurrentBitmap {
     T bits = reinterpret_cast<const std::atomic<T> *>(&bits_[*byte_pos])->load();
     if (bits == static_cast<T>(-1)) {
       *byte_pos += static_cast<uint32_t>(sizeof(T));
-      // prevent underflow
-      if (*bits_left < sizeof(T) * BYTE_SIZE) {
-        *bits_left = 0;
-      } else {
-        *bits_left = *bits_left - static_cast<uint32_t>(sizeof(T) * BYTE_SIZE);
-      }
+      // IsAlignedAndFits<T> is always called before FindUnsetBit.
```
tli2: If you think this assumption should always be true, the right place to document this is as precondition for the method. Basically, assert that this is true at the beginning of the method.

amlatyrngom: Like an assert statement at the beginning of the function? Or just comments?

I can also put the check back in, in case someone might call the function without checking bits_left.

tli2: Asserts at the beginning of the function is preferred as we discussed

Issue #247, opened 2018-11-17
Bump AppleClang/Xcode version to 10


mbutrovich:
#246 requires the addition of ` -Wno-invalid-offsetof` to Clang compiler flags because the way we use `offsetof` in the BwTree is undefined behavior before C++17. We are currently using AppleClang/Xcode 9 on Travis and Jenkins as well as our [System Setup](https://github.com/cmu-db/terrier/wiki/System-Setup) wiki page. AppleClang/Xcode 10 brings better support for C++17, including no longer needing that flag.

I have a branch to handle the Travis changes [here](https://github.com/mbutrovich/terrier/tree/xcode10) which are pretty simple, but will need to talk to @crd477 next week (or after Thanksgiving) about updating the Jenkins Docker images. After those are in place and a PR is merged, we should update the Wiki as well.

crd477:
I'm working on this now.  It shouldn't take too long.

mbutrovich:
Gonna leave this open until the Travis stuff gets merged in.

PR #252, opened 2018-11-27
LLVM_DIR set to 6.0.1_1 due to version change on Homebrew

mbutrovich:
Fixes compiler issues seen on macOS Travis this week (see [this PR](https://travis-ci.org/cmu-db/terrier/jobs/459466255)) I don't think we see this on Jenkins because it has 6.0.1 baked into the image already, but if we do a `brew upgrade` or set up a new image then we'll have to update the LLVM_DIR environment variable on Jenkins as well.

Will discuss with @crd477 if we just want to update the Jenkins image at the same time as resolving #247.


mbutrovich:
Actually, just going to merge this change in with #253

PR #260, opened 2018-12-04
Catalog

yangjuns:
From @pervazea's comment below:

## Overview of the catalog code

For each table in the catalog, there are two classes:

1. Handle, e.g. NamespaceHandle.
The handle provides methods for the overall table. These include:

- Create (used during initialization)
- Doing lookups, returning an Entry (see below)
- Deleting

2. Entry, e.g. NamespaceEntry.
An entry is a single row result from a lookup, with methods to access the columns by name. The contents of an entry do not change once it has been created. 

The catalog class contains higher level methods:

- creating and deleting databases
- creating and deleting tables
- obtaining handles for the catalogs.

The structure of the catalog tables, is taken from Postgres, with some variance. Detailed information is present in the Postgres documentation. In brief:

- A DatabaseHandle provides access to pg_database. pg_database is global, and records databases.
- A NamespaceHandle provides access to pg_namespace. There is a pg_namespace for each database.
- A TablespaceHandle provides access to pg_tablespace. This is implemented but not used.
- An AttributeHandle provides access to pg_attribute. pg_attribute is per database. There is an entry (i.e. row) for every column in every table.
- A TypeHandle provides access to pg_type. pg_type is per database, and records each available type.  pg_attribute has foreign key references to pg_type.
- An AttrDefHandle provides access to pg_attrdef. pg_attrdef is per database and records default values (where defined) for attributes (i.e. columns).
- A ClassHandle provides access to pg_class. pg_class is per database, and has a row for anything that is table like, e.g. tables, indexes etc. 

- A TableHandle provides a view equivalent to pg_tables, a simpler view  of  tables. The API here is different since there are multiple underlying tables.


src/include/catalog/catalog.h
```diff
@@ -0,0 +1,127 @@
...
+ * the entry point for transactions to access any data in any sql table.
+ *
+ * OID assignment:
+ * Note that we do not have a concept of oid_t anymore. Instead, we have
+ *  db_oid_t, nsp_oid_t, table_oid_t, col_oid_t
+ * In addition, for nsp_oid_t, table_oid_t, and col_oid_t, we only guarantee uniqueness inside a database, which means
+ * that the table_oid for pg_attribute in database A could be the same as pg_attribute in database B.
+ *
+ * db_oid_t:
+ *  0          is reserved for default database - terrier
```
pervazea: It is preferable to not use oid value 0. Leave 0 as a "reserved" value, and start the numbering from 1.

src/include/catalog/catalog.h
```diff
@@ -0,0 +1,127 @@
...
+ *  1+         are used for new databases
+ *
+ * nsp_oid_t:
```
pervazea: nsp is not an intuitive abbreviation. Expand all nsp abbreviations to namespace.

src/include/catalog/catalog_defs.h
```diff
@@ -2,6 +2,22 @@
 #include "common/strong_typedef.h"
 
 namespace terrier::catalog {
+
```
pervazea: Would be helpful to add a comment, refering the reader to catalog.h, where the numbering of oids is explained.

src/catalog/catalog.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include "loggers/catalog_logger.h"
+#include "storage/storage_defs.h"
+#include "transaction/transaction_manager.h"
+
+namespace terrier::catalog {
+
+std::shared_ptr<Catalog> terrier_catalog;
+
+Catalog::Catalog(transaction::TransactionManager *txn_manager) : txn_manager_(txn_manager) {
+  CATALOG_LOG_INFO("Creating catalog ...");
```
pervazea: You should go through the LOG calls and adjust them. 
INFO should be used for messages that are output under normal operation. There should be very little.
DEBUG should be used for logging additional information when the code goes through some error path and you want to record what is happening.
TRACE is for tracing execution flow. Most of the current messages are probably TRACE level.

test/catalog/namespace_handle_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+#include <random>
+#include <vector>
+#include "catalog/catalog.h"
+#include "transaction/transaction_manager.h"
+#include "util/test_harness.h"
+namespace terrier {
+
+struct NamespaceHandleTests : public TerrierTest {
+  void SetUp() override {
+    TerrierTest::SetUp();
```
pervazea: You may need to add logger initialization and shutdown calls in your tests.

src/include/catalog/catalog.h
lmwnshn: Arguably, Postgres does it this way: https://www.postgresql.org/docs/11/catalog-pg-namespace.html

But I will expand it nevertheless.

test/catalog/namespace_handle_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+  const catalog::db_oid_t terrier_oid(0);
+  auto db_handle = catalog_->GetDatabaseHandle(terrier_oid);
+  auto nsp_handle = db_handle.GetNamespaceHandle();
+  // get the pg_catalog namespace
+  catalog::nsp_oid_t pg_catalog_oid(0);
+  auto nsp_entry_ptr = nsp_handle.GetNamespaceEntry(txn_, pg_catalog_oid);
+  EXPECT_TRUE(nsp_entry_ptr != nullptr);
+  // test if we are getting the correct value
+  // oid has col_oid_t = 1000
+  EXPECT_TRUE(*reinterpret_cast<uint32_t *>(nsp_entry_ptr->GetValue(catalog::col_oid_t(1000))) == !pg_catalog_oid);
```
lmwnshn: !pg_catalog_oid -- is this intended?

src/include/catalog/database_handle.h
```diff
@@ -0,0 +1,96 @@
...
+     * @param map a map that encodes how to access attributes of the row
+     */
+    DatabaseEntry(db_oid_t oid, storage::ProjectedRow *row, storage::ProjectionMap map)
+        : oid_(oid), row_(row), map_(std::move(map)) {}
+    /**
+     * Get the value of an attribute
+     * @param col the col_oid of the attribute
+     * @return a pointer to the attribute value
+     */
+    byte *GetValue(col_oid_t col) { return row_->AccessWithNullCheck(map_[col]); }
```
lmwnshn: map_[col] may not be the desired behavior. If we access a non-existent column, it will silently insert and return physical offset 0, which in turn maps to ?? (currently the oid, but we sort so it depends what columns there are).

In other words, GetValue(nonexistent column) will always return the smallest column. I have changed this to .at() which errors out.

test/catalog/namespace_handle_test.cpp
lmwnshn: I thought the [test harness](https://github.com/cmu-db/terrier/pull/260/files#diff-a776bc3ea698b0d08a2f6d09c3f25f4cR23) would do this, what should I invoke?

src/catalog/catalog.cpp
lmwnshn: Changed.

src/include/catalog/catalog_defs.h
lmwnshn: Added.

src/include/catalog/catalog.h
lmwnshn: Reserved.

src/catalog/catalog.cpp
```diff
@@ -0,0 +1,162 @@
...
+void Catalog::BootstrapDatabase(transaction::TransactionContext *txn, db_oid_t db_oid) {
+  // refer to OID assignment scheme in catalog.h
+  table_oid_t table_oid(DATABASE_CATALOG_TABLE_START_OID);
+  col_oid_t col_oid(DATABASE_CATALOG_COL_START_OID);
+  namespace_oid_t namespace_oid(PG_CATALOG_OID);
+
+  // create pg_namespace
+  table_oid_t pg_namespace_oid(table_oid++);
+  std::vector<Schema::Column> cols;
+  cols.emplace_back("oid", type::TypeId::INTEGER, false, col_oid++);
```
pervazea: I did not see how col_oid value is updated (preserved) after these increments.
I also think that (*start_col_oid)++ is not safe, e.g when we later do things in parallel, and try to create 2 user tables at the same time.

One solution to both of these, would be to have methods accessed via the database handle, that give you an oid value, e.g. something like GetNextColOid(). The increment needs to be atomic (so it could be lockless atomic or use a lock). You'd need functions to hand out oids for each range.

yeshengm:
`__ptr` in our `pg_class` implementation breaks the numbering with Postgres, needs some workaround to fix that. Otherwise all columns are shifted by 1.

pervazea:
Overview of the catalog code.

For each table in the catalog, there are two classes:

1. Handle, e.g. NamespaceHandle.
The handle provides methods for the overall table. These include:

- Create (used during initialization)
- Doing lookups, returning an Entry (see below)
- Deleting

2. Entry, e.g. NamespaceEntry.
An entry is a single row result from a lookup, with methods to access the columns by name. The contents of an entry do not change once it has been created. 

The catalog class contains higher level methods:

- creating and deleting databases
- creating and deleting tables
- obtaining handles for the catalogs.

The structure of the catalog tables, is taken from Postgres, with some variance. Detailed information is present in the Postgres documentation. In brief:

- A DatabaseHandle provides access to pg_database. pg_database is global, and records databases.
- A NamespaceHandle provides access to pg_namespace. There is a pg_namespace for each database.
- A TablespaceHandle provides access to pg_tablespace. This is implemented but not used.
- An AttributeHandle provides access to pg_attribute. pg_attribute is per database. There is an entry (i.e. row) for every column in every table.
- A TypeHandle provides access to pg_type. pg_type is per database, and records each available type.  pg_attribute has foreign key references to pg_type.
- An AttrDefHandle provides access to pg_attrdef. pg_attrdef is per datbaase and records default values (where defined) for attributes (i.e. columns).
- A ClassHandle provides access to pg_class. pg_class is per database, and has a row for anything that is table like, e.g. tables, indexes etc.

src/include/catalog/attribute_handle.h
```diff
@@ -0,0 +1,122 @@
...
+ * reasons.
+ */
+class AttributeHandle {
+ public:
+  /**
+   * Construct an attribute handle
+   * @param catalog catalog ptr
+   * @param pg_attribute a pointer to pg_attribute sql table rw helper instance
+   */
+  explicit AttributeHandle(Catalog *catalog, std::shared_ptr<catalog::SqlTableRW> pg_attribute)
```
malin1993ml: The `catalog` argument is never used in the constructor. And the constructor below is deprecated, which has an extra `table` argument. I'm confused...

src/include/catalog/attribute_handle.h
```diff
@@ -0,0 +1,122 @@
...
+  // end Debug methods
+
+  /** Used schema columns */
+  static const std::vector<SchemaCol> schema_cols_;
+  /** Unused schema columns */
+  static const std::vector<SchemaCol> unused_schema_cols_;
+
+ private:
+  // Catalog *catalog_;
+  SqlTableRW *table_;
```
malin1993ml: Which `table_` this is?

src/catalog/catalog.cpp
```diff
@@ -0,0 +1,556 @@
...
+  // - pg_class
+  // - pg_type
+  // - pg_attrdef
+
+  map_.erase(oid);
+  name_map_.erase(oid);
+}
+
+table_oid_t Catalog::CreateTable(transaction::TransactionContext *txn, db_oid_t db_oid, const std::string &table_name,
+                                 const Schema &schema) {
```
malin1993ml: Is this method used to create a catalog table or user table, or both? Right now it is only used in the `CreateUserTableTest`. Also, I don't think you can call this method concurrently, because it is modifying `map_` and `name_map_`, which are not concurrent data structures. Is it designed that way?

src/include/catalog/catalog_sql_table.h
```diff
@@ -0,0 +1,718 @@
...
+#include "type/transient_value_factory.h"
+#include "type/transient_value_peeker.h"
+#include "util/storage_test_util.h"
+
+namespace terrier::catalog {
+
+/**
+ * Helper class to simplify operations on a SqlTable
+ */
+class SqlTableRW {
```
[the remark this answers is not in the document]
parsec0re: It should be read/write. I've no idea why it is named like this though.

src/include/catalog/catalog.h
```diff
@@ -0,0 +1,314 @@
...
+namespace terrier::catalog {
+
+class DatabaseHandle;
+class TablespaceHandle;
+class SettingsHandle;
+
+/**
+ * Schema column for used/unused schema rows.
+ */
+struct SchemaCol {
```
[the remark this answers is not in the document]
pervazea: Correct. It is only used when creating catalog tables. Will comment.

Stating probable direction. We discussed in a team meeting adding default values into the schema stored with the sql table. The comment in schema.h implies we might replace the sql table schema with the schema from the catalog. This is possible, but extracting the schema information from the catalog is much more work. So, in my opinion, possible approach is to continue:

- schema stored in the catalog (as expected)
- retain the schema in the sql table layer, with default value extension. This can be viewed as a cached version of the schema, that is easier to interpret.
- need to ensure that such a scheme works correctly with the schema versioning project.

src/include/catalog/catalog_sql_table.h
```diff
@@ -0,0 +1,718 @@
...
+        return strncmp(varchar.data(), reinterpret_cast<const char *>(vc_entry->Content()), size) == 0;
+      } break;
+
+      default:
+        throw NOT_IMPLEMENTED_EXCEPTION("unsupported type in ColEqualsValue");
+    }
+  }
+
+  storage::RecordBufferSegmentPool buffer_pool_{100, 100};
+  transaction::TransactionManager txn_manager_ = {&buffer_pool_, true, LOGGING_DISABLED};
```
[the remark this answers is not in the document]
pervazea: Will remove.

src/include/catalog/catalog_sql_table.h
```diff
@@ -0,0 +1,718 @@
...
+
+  storage::BlockStore block_store_{100, 100};
```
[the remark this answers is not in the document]
pervazea: These were cut/paste from other block_store instances. I don't know what the right values are. These are good enough for current tests, but:

1. Need to determine what the right initial values should be
1. Is the block store sizing static or do we need to make it adjustable?

src/include/catalog/catalog_sql_table.h
```diff
@@ -0,0 +1,718 @@
...
+     */
+    RowIterator(transaction::TransactionContext *txn, SqlTableRW *tblrw, bool begin)
+        : txn_(txn), tblrw_(tblrw), buffer_(nullptr), dtsi_(tblrw->GetSqlTable()->begin()), layout_(nullptr) {
+      if (!begin) {
+        // constructing end
+        proj_col_bufp = nullptr;
+        return;
+      }
+      layout_ = tblrw_->GetLayoutP();
+      all_cols = StorageTestUtil::ProjectionListAllColumns(*layout_);
```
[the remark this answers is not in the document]
pervazea: Needs to be looked at.

src/catalog/catalog.cpp
```diff
@@ -0,0 +1,556 @@
...
+  name_map_[db_oid]["pg_database"] = pg_database_->Oid();
+  name_map_[db_oid]["pg_tablespace"] = pg_tablespace_->Oid();
+  name_map_[db_oid]["pg_settings"] = pg_settings_->Oid();
+
+  // Order: pg_attribute -> pg_namespace -> pg_class
+  CreatePGAttribute(txn, db_oid);
+  CreatePGNamespace(txn, db_oid);
+  CreatePGClass(txn, db_oid);
+  CreatePGType(txn, db_oid);
+  AttrDefHandle::Create(txn, this, db_oid, "pg_attrdef");
```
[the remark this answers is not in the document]
pervazea: It should have a Create function.

src/catalog/catalog.cpp
```diff
@@ -0,0 +1,556 @@
...
+void Catalog::DeleteDatabase(transaction::TransactionContext *txn, const std::string &db_name) {
+  // get database handle
+  auto db_handle = GetDatabaseHandle();
+  auto db_entry = db_handle.GetDatabaseEntry(txn, db_name);
+  auto oid = db_entry->GetOid();
+  // remove entry from pg_database
+  db_handle.DeleteEntry(txn, db_entry);
+
+  // TODO(pakhtar):
+  // - delete all the tables
```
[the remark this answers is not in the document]
pervazea: Yes, but as a follow on PR. I believe this has to be done via the GC. The GC will need support added for object (table) deletion.

src/include/catalog/catalog_sql_table.h
```diff
@@ -0,0 +1,718 @@
...
+      auto col_initer = new storage::ProjectedColumnsInitializer(*layout_, all_cols, 1);
```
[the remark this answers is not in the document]
pervazea: We do return a row. Here we are doing a sequential scan. The projected column API supports scanning populated slots, but the projected row API looks at all slots, not just valid ones. So, we use the projected column API to do the scan, then use the methods in that API to return a row.

src/include/catalog/attribute_handle.h
pervazea: This is to be deprecated and removed, so that the attribute handle can be consistent with the structure of the other handles  and entries.

src/catalog/catalog.cpp
pervazea: It is just for creation of user tables and currently just used in the test. 

Issue #311 notes the concurrency issue with maps.

src/include/catalog/catalog_sql_table.h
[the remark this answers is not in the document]
pervazea: This is something that will need to be dynamic and adjustable by the "self-driving" part.

src/include/catalog/catalog_sql_table.h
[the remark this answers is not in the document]
pervazea: It was created as a class to assist in reading and writing via the sql table class. We should feel free to rename as desired.

src/include/catalog/catalog_sql_table.h
```diff
@@ -0,0 +1,718 @@
...
+      proj_col_bufp = col_initer->Initialize(buffer_);
+
+      while (dtsi_ != tblrw_->GetSqlTable()->end()) {
+        tblrw_->GetSqlTable()->Scan(txn, &dtsi_, proj_col_bufp);
+        if (proj_col_bufp->NumTuples() == 0) {
+          continue;
+        }
+        break;
+      }
+      if (dtsi_ == tblrw_->GetSqlTable()->end()) {
```
amlatyrngom: There seems to be a bug here. We should first check if NumTuples() is 0 before ending prematurely.

pervazea: I'm not clear on where you think that check should be... can you clarify please?

src/include/catalog/catalog.h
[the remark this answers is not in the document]
pervazea: Comment added, clarifying usage.

src/catalog/catalog.cpp
```diff
@@ -0,0 +1,556 @@
...
+    pg_database_->DefineColumn(col.col_name, col.type_id, false, col_oid_t(GetNextOid()));
+  }
+
+  // columns we don't use
+  for (auto col : DatabaseHandle::unused_schema_cols_) {
+    pg_database_->DefineColumn(col.col_name, col.type_id, false, col_oid_t(GetNextOid()));
+  }
+  // create the table
+  pg_database_->Create();
+  db_oid_t terrier_oid = DEFAULT_DATABASE_OID;
```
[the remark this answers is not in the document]
pervazea: renamed to default_db_oid

src/catalog/catalog.cpp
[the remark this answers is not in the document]
pervazea: Fixed.

src/include/catalog/attribute_handle.h
pervazea: AttributeHandle changed to address.

src/include/catalog/attribute_handle.h
pervazea: AttributeHandle changed to address.

src/include/catalog/catalog_sql_table.h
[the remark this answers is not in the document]
pervazea: Now uses a function in StorageUtil.

src/include/catalog/catalog_sql_table.h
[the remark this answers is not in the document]
pervazea: Removed.

src/include/catalog/catalog.h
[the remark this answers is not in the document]
malin1993ml: When you say "the schema from the catalog", do you just mean the information in pg_attribute and pg_attrdef? I think even though we retrieve the schema information for the user tables from the catalog, in our own code base, we still need some class to store the schema information for the catalog tables themselves. It's kind of like a chicken-and-egg problem. Without some sort of schema representation, we cannot even define pg_attribute and pg_attrdef, right?

src/include/catalog/catalog_sql_table.h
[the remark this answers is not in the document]
malin1993ml: Maybe document somewhere, like creating an issue? Also, should this hook up with the settings manager?

src/catalog/catalog.cpp
malin1993ml: Maybe rename it to `CreateUserTable` to make it more explicit?

src/catalog/catalog.cpp
```diff
@@ -0,0 +1,556 @@
...
+
+  // add it to the map
+  map_[terrier_oid] = std::unordered_map<table_oid_t, std::shared_ptr<catalog::SqlTableRW>>();
```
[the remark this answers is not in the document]
malin1993ml: I saw your comment [here](https://github.com/cmu-db/terrier/issues/311#issuecomment-472151679), which is basically talking about the same thing. What's the verdict on that? Especially for my second question above, is this map repopulatable on shutdowns/crashes?

src/include/catalog/catalog_sql_table.h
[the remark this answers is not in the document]
malin1993ml: I see. Just out of curiosity, what are the general trade-offs between using projected row and project column, besides that they have different APIs? What would be the scenarios to use one versus the other?

src/include/catalog/catalog_sql_table.h
amlatyrngom: The check should in the if statement. I come across this error when there is only one tuple in the table. Here is what I see:
1. SqlTableRW::begin() is called. Because there is only one tuple, dsti_ will reach the end of the table after one call to Scan.
2. Without this check, proj_col_bufp will be set to nullptr since dtsi_ is at the end of the table.
3. Now begin() == end() even though the table is not empty.
Am I maybe using this object incorrectly?

src/include/catalog/attribute_handle.h
pervazea: Removed.

src/main/terrier.cpp
```diff
@@ -42,11 +45,18 @@ int main() {
   // initialize stat registry
   auto main_stat_reg = std::make_shared<terrier::common::StatisticsRegistry>();
 
+  // create the global transaction mgr
+  terrier::storage::RecordBufferSegmentPool buffer_pool_(100000, 10000);
```
malin1993ml: How these numbers are set? Do we need to dynamically adjust them? Should they hook up with the settings manager in the future?

test/catalog/catalog_test.cpp
```diff
@@ -0,0 +1,75 @@
...
+  catalog::Catalog *catalog_;
+  storage::RecordBufferSegmentPool buffer_pool_{100, 100};
+
+  transaction::TransactionContext *txn_ = nullptr;
+  transaction::TransactionManager *txn_manager_;
+};
+
+// Tests for higher level catalog API
+// NOLINTNEXTLINE
+TEST_F(CatalogTests, CreateDatabaseTest) {
```
malin1993ml: All the tests in this file do not have explicit checks, but only dump the catalog content to the terminal. What do we expect to see in those dumps? Can we add automated checks for them?

src/include/catalog/table_handle.h
```diff
@@ -0,0 +1,212 @@
...
+ *    1) it needs schemaname from pg_namespace
+ *    2) it needs tablename from pg_class
+ *    3) it needs tablespace from pg_tablespace
+ *
+ * The class uses a vector of size 3 to stores 3 rows in these three tables.
+ *    rows_[0] stores a row in pg_namespace which contains namespace information for this TableEntry
+ *    rows_[1] stores a row in pg_class which contains table information for this TableEntry
+ *    rows_[2] stores a row in pg_tablespace which contains table information for this TableEntry
+ */
+class TableEntry {
```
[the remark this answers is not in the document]
parsec0re: pg_table is a view. So it is implemented in a differently. I can't think of a generic way of handling this.

src/catalog/catalog.cpp
```diff
@@ -0,0 +1,557 @@
...
+
+  // add to maps
+  AddToMaps(db_oid, tbl_rw->Oid(), table_name, tbl_rw);
+
+  // enter attribute information
+  AddColumnsToPGAttribute(txn, db_oid, tbl_rw->GetSqlTable());
+  return tbl_rw->Oid();
+}
+
+void Catalog::DeleteTable(transaction::TransactionContext *txn, db_oid_t db_oid, table_oid_t table_oid) {
```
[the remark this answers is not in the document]
pervazea: Deleted.

src/catalog/catalog.cpp
```diff
@@ -0,0 +1,557 @@
...
+  // TODO(pakhtar): drop table
+}
+
+DatabaseHandle Catalog::GetDatabaseHandle() { return DatabaseHandle(this, pg_database_); }
+
+TablespaceHandle Catalog::GetTablespaceHandle() { return TablespaceHandle(this, pg_tablespace_); }
+
+SettingsHandle Catalog::GetSettingsHandle() { return SettingsHandle(pg_settings_); }
+
+std::shared_ptr<catalog::SqlTableRW> Catalog::GetDatabaseCatalog(db_oid_t db_oid, table_oid_t table_oid) {
```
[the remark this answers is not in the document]
pervazea: Renamed to GetCatalogTable.

src/include/catalog/database_handle.h
```diff
@@ -0,0 +1,152 @@
...
+   * @param entry a row in pg_database that represents this table
+   */
+  DatabaseEntry(db_oid_t oid, catalog::SqlTableRW *sql_table, std::vector<type::TransientValue> &&entry)
+      : CatalogEntry(oid, sql_table, std::move(entry)) {}
+  /**
+   * Delete the data (for this entry) from the storage table.
+   * After this, the entry object must be deleted as no other
+   * operations are possible.
+   */
+  bool Delete(transaction::TransactionContext *txn);
```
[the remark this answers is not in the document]
pervazea: Deleted. This functionality is replaced by DeleteEntry.

src/include/catalog/catalog_sql_table.h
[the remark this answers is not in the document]
pervazea: created issue #346

src/main/terrier.cpp
pervazea: created issue #346

src/include/catalog/catalog_sql_table.h
pervazea: Nice catch. Your logic is sound, but I have not been able to recreate this bug. I've only had a small amount of time to try. Nevertheless, I added a check, so that the iterator will not prematurely set the termination condition.

src/catalog/catalog.cpp
pervazea: Renamed to CreateUserTable

src/include/catalog/catalog_sql_table.h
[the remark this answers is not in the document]
pervazea: The trade-offs are something that Matt B. and/or Tian Yu could answer. We started with the projected row API, but were pointed to the projected column API as the one to use.

src/catalog/catalog.cpp
```diff
@@ -0,0 +1,557 @@
...
+  auto pg_global_ts_oid = !GetTablespaceHandle().GetTablespaceEntry(txn, "pg_global")->GetOid();
+  auto pg_default_ts_oid = !GetTablespaceHandle().GetTablespaceEntry(txn, "pg_default")->GetOid();
+
+  // Insert pg_database
+  // (namespace: catalog, tablespace: global)
+  CATALOG_LOG_TRACE("Inserting pg_database into pg_class ...");
+  auto pg_db_tbl_p = reinterpret_cast<uint64_t>(GetDatabaseCatalog(db_oid, "pg_database").get());
+  auto pg_database_entry_oid = !GetDatabaseCatalog(db_oid, "pg_database")->Oid();
+
+  class_handle.AddEntry(txn, pg_db_tbl_p, pg_database_entry_oid, "pg_database", pg_catalog_namespace_oid,
```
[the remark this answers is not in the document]
pervazea: Yes, table pointers are stored here. The maps provide the table pointers, but the original intention was to get them from what is stored here.

See #347. I created that so we can further discuss, and so the history is  not lost.

src/catalog/database_handle.cpp
```diff
@@ -0,0 +1,97 @@
...
+    {8, "datconnlimit", type::TypeId::INTEGER}};
+
+/**
+ * Handle methods
+ */
+
+DatabaseHandle::DatabaseHandle(Catalog *catalog, std::shared_ptr<catalog::SqlTableRW> pg_database)
+    : catalog_(catalog), pg_database_rw_(std::move(pg_database)) {}
+
+ClassHandle DatabaseHandle::GetClassHandle(transaction::TransactionContext *txn, db_oid_t oid) {
```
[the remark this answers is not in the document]
pervazea: See issue #347

test/catalog/catalog_test.cpp
pervazea: I've added some checks to the test, but more can be done. This was useful, a couple of items related to table creation and deletion were missing. There is more (known) work for deletion to follow this PR.

src/include/catalog/catalog.h
```diff
@@ -0,0 +1,325 @@
...
+  void DeleteDatabase(transaction::TransactionContext *txn, const std::string &db_name);
+
+  /**
+   * Create a table with schema
+   * @param txn transaction to use
+   * @param db_oid oid of the database
+   * @param table_name table name
+   * @param schema schema to use
+   */
+  table_oid_t CreateUserTable(transaction::TransactionContext *txn, db_oid_t db_oid, const std::string &table_name,
```
apavlo: How do I specify the namespace for this table?

src/include/catalog/catalog.h
```diff
@@ -0,0 +1,325 @@
...
+ * OID assignment:
+ * In terrier, each table row is identified by an oid. Further, as strong
+ * typing is used in terrier, each (catalog) table has it's own oid type,
+ * e.g. db_oid_t, namespace_oid_t etc.
+ *
+ * The values for all oid types are generated from a single value space
+ * i.e. every oid regardless of type, is unique.
+ *
+ */
+class Catalog {
```
apavlo: It seems weird that one can use the `Catalog` object to create both databases and user tables. Why not namespaces? See my comments about `CreateUserTable()` and  `DeleteTable()` where you are not specifying a namespace, thereby violating the catalog hierarchy.

src/include/catalog/catalog.h
```diff
@@ -0,0 +1,325 @@
...
+  table_oid_t CreateUserTable(transaction::TransactionContext *txn, db_oid_t db_oid, const std::string &table_name,
+                              const Schema &schema);
+
+  /**
+   * Delete a table
+   * @param txn transaction to use
+   * @param db_oid oid of the database
+   * @param table_name table to delete
+   */
+  void DeleteTable(transaction::TransactionContext *txn, db_oid_t db_oid, const std::string &table_name);
```
apavlo: Does this delete both user tables and catalog tables?

Should we not expose a public API for deleting catalog tables? I feel like only the catalog (or a friend class) should be allowed to do that.

src/include/catalog/catalog.h
```diff
@@ -0,0 +1,325 @@
...
+  /**
+   * Get a pointer to the storage table.
+   * Supports both catalog tables, and user created tables.
+   *
+   * @param db_oid database that owns the table
+   * @param table_oid returns the storage table pointer for this table_oid
+   * @return a pointer to the catalog
+   * @throw out_of_range exception if either oid doesn't exist or the catalog doesn't exist.
+   */
+  std::shared_ptr<catalog::SqlTableRW> GetCatalogTable(db_oid_t db_oid, table_oid_t table_oid);
```
apavlo: This is slightly confusing. The name of the method is `GetCatalogTable()`, but it returns a `catalog::SqlTableRW`. The comment says it returns a "storage table", but I would have thought that it returns a `storage::SqlTable`.

apavlo: Should this method be protected? Does anybody other than the catalog need it?

apavlo: What about `namespace_oid`?

src/include/catalog/catalog.h
```diff
@@ -0,0 +1,325 @@
...
+  /**
+   * TODO(pakhtar): Implement shutdown.
+   */
+
+  /*
+   * Destructor
+   */
+  ~Catalog() = default;
+
+  //  ~Catalog() {
```
apavlo: Remove dead code.

src/include/catalog/catalog.h
```diff
@@ -0,0 +1,325 @@
...
+   * @param cols vector of column types
+   */
+  void SetUnusedColumns(std::vector<type::TransientValue> *vec, const std::vector<SchemaCol> &cols);
+
+  /**
+   * Convert type id to schema type
+   * @param type_id type id
+   * @return schema type
+   */
+  type::TransientValue ValueTypeIdToSchemaType(type::TypeId type_id);
```
apavlo: Put in separate helper class.

src/include/catalog/catalog.h
```diff
@@ -0,0 +1,325 @@
...
+   */
+  void AddEntryToPGDatabase(transaction::TransactionContext *txn, db_oid_t oid, const std::string &name);
+
+  /**
+   * Add columns created for Postgres compatibility, but unused, to the schema
+   * @param db_p - shared_ptr to database
+   * @param cols - vector specifying the columns
+   *
+   */
+  void AddUnusedSchemaColumns(const std::shared_ptr<catalog::SqlTableRW> &db_p, const std::vector<SchemaCol> &cols);
```
apavlo: I don't understand the purpose of specifying which columns are unused? This sounds like a premature optimization it is going to be a pain to maintain in the future.

src/include/catalog/catalog.h
```diff
@@ -0,0 +1,325 @@
...
+   * Bootstrap all the catalog tables so that new coming transactions can
+   * correctly perform SQL queries.
+   * 1) It creates and populates all the global catalogs
+   * 2) It creates a default database named "terrier"
+   * 3) It bootstraps the default database.
+   * @param txn to be used for bootstrapping
+   */
+  void Bootstrap(transaction::TransactionContext *txn);
+
+  void CreatePGDatabase(table_oid_t table_oid);
```
apavlo: Missing documentation.

src/include/catalog/class_handle.h
```diff
@@ -0,0 +1,111 @@
...
+   */
+  void AddEntry(transaction::TransactionContext *txn, int64_t tbl_ptr, int32_t entry_oid, const std::string &name,
+                int32_t ns_oid, int32_t ts_oid);
+
+  /**
+   * Constructor
+   * @param catalog the global catalog object
+   * @param pg_class the pg_class sql table rw helper instance
+   */
+  explicit ClassHandle(Catalog *catalog, std::shared_ptr<catalog::SqlTableRW> pg_class)
```
apavlo: Move constructor to top of class.

src/include/catalog/class_handle.h
```diff
@@ -0,0 +1,111 @@
...
+
+  /**
+   * Create the storage table
+   * @param txn the txn that creates this table
+   * @param catalog ptr to the catalog
+   * @param db_oid db_oid of this handle
+   * @param name catalog name
+   * @return a shared pointer to the catalog table
+   */
+  static std::shared_ptr<catalog::SqlTableRW> Create(transaction::TransactionContext *txn, Catalog *catalog,
```
apavlo: Why does this need to be static? We should probably move into a helper class.

src/include/catalog/class_handle.h
```diff
@@ -0,0 +1,111 @@
...
+   */
+  bool DeleteEntry(transaction::TransactionContext *txn, const std::shared_ptr<ClassEntry> &entry);
+
+  /**
+   * Debug methods
+   */
+  void Dump(transaction::TransactionContext *txn) { pg_class_rw_->Dump(txn); }
+
+  /** Used schema columns */
+  static const std::vector<SchemaCol> schema_cols_;
```
apavlo: Does this need to be public?

src/include/catalog/database_handle.h
```diff
@@ -0,0 +1,146 @@
...
+   */
+  DatabaseEntry(db_oid_t oid, catalog::SqlTableRW *sql_table, std::vector<type::TransientValue> &&entry)
+      : CatalogEntry(oid, sql_table, std::move(entry)) {}
+};
+
+/**
+ * A DatabaseHandle provides access to the (global) system pg_database
+ * catalog.
+ *
+ * This pg_database is a subset of Postgres (v11)  pg_database, and
```
apavlo: We should bring in all of `pg_database` if possible.

src/include/catalog/table_handle.h
```diff
@@ -0,0 +1,212 @@
...
+    rows_.resize(3);
+    rows_[1] = std::move(row);
+  }
+
+  /**
+   * Get the value for a given column number
+   * @param col_num the column number
+   * @return the value
+   */
+  const type::TransientValue &GetColInRow(uint32_t col_num) {
```
apavlo: This method should be private. We don't want to expose any of the internals of how we are implementing the view.

src/include/catalog/table_handle.h
```diff
@@ -0,0 +1,212 @@
...
+   */
+  std::shared_ptr<TableEntry> GetTableEntry(transaction::TransactionContext *txn, const std::string &name);
+
+  /**
+   * Create a SqlTable. The namespace of the table is the same as the TableHandle.
+   * @param txn the transaction context
+   * @param name the table name
+   * @param schema the table schema
+   */
+  SqlTableRW *CreateTable(transaction::TransactionContext *txn, const Schema &schema, const std::string &name);
```
apavlo: Sometimes your Create methods return `SqlTableRW*` and sometimes they return `std::shared_ptr<catalog::SqlTableRW>`. Please be consistent.

src/include/catalog/catalog.h
pervazea: The API is short changing namespaces. Will fix.

src/include/catalog/class_handle.h
pervazea: It is static because:
- The Create function needs to be called before an instance of the class can be instantiated. 
- To keep the code related to the class in one place. The first implementation had creation and column definitions in the catalog class code. This was hard to read and poor for maintainability.

This could be moved to a util class, but IMO clearer to keep all the related code together. Singletons are not being created.

src/include/catalog/table_handle.h
pervazea: We need to return raw pointers and adjust the tests and shutdown code accordingly.

src/include/catalog/catalog.h
pervazea: The reasons why the Unused columns definition was added:

- To make it clearer which columns matter. Prior to this, the catalog code when initializing, sometimes had a few real columns initialized, then many unused columns initialized. This made it hard to see what mattered.
- Unused VARCHAR columns MUST be initialized, otherwise reading from the table generates segmentation faults. This was a recurring problem in earlier instances of the code. Scalar columns can be left unset, but will then contain garbage. Having a specification of which columns are unused allows them to be initialized with a single call.

src/include/catalog/table_handle.h
pervazea: These have been converted to raw pointers, so we can control the lifetime of the tables correctly.

src/include/catalog/catalog.h
pervazea: Returned ptr is now raw ptr to SqlTableHelper

pervazea: Protected now.

src/include/catalog/catalog.h
pervazea: Replaced by DeleteUserTable.

src/include/catalog/catalog.h
pervazea: Used only internal to the catalog. Made private.

src/include/catalog/catalog.h
pervazea: Done.

src/include/catalog/catalog.h
pervazea: Merged used and unused columns.

src/include/catalog/database_handle.h
pervazea: 1. Deleted obsolete portion of comment
2. Added in remaining fields of pg_database.

src/include/catalog/class_handle.h
pervazea: Yes, at the moment this needs to be static.

src/include/catalog/catalog.h
pervazea: Updated namespace support.

src/include/catalog/catalog.h
pervazea: Updated namespace support.

pervazea:
_Should we make the handle objects inherit from a CatalogHandle base class in the same way that all of the entry objects inherit from CatalogEntry?_

In principle, yes, but I don't see any benefit at present. There are small differences in how the different catalogs have to be treated, so it is more involved than the change for CatalogEntry.

parsec0re:
@pervazea Even the current CatalogEntry solution seems not very elegant too. But I can't think of a better approach to tackle this problem.

pervazea:
Status: previous to pull from master, passed Jenkins and Travis checks. Currently Jenkins experiencing network problems with some executors, causing failure to checkout from git.

See issues for future catalog work. This version has a map for internal mapping between catalog tables and the sql table instance.

PR #261, opened 2018-12-04
Add varlen flag to block layout

tli2:
in the data table's layout. We now can ask the block layout whether a value is varlen or not.

We will implement all the logic in GC and logging in later PRs

### DIFF
--- a/src/include/storage/block_layout.h
+++ b/src/include/storage/block_layout.h
@@ -44,7 +44,7 @@ struct BlockLayout {
    * @param col_id the column id to check for
    * @return if the given column id is varlen or not
    */
-  bool IsVarlen(col_id_t col_id) const { return static_cast<int8_t>(AttrSize(col_id)) < 0; }
+  bool IsVarlen(col_id_t col_id) const { return static_cast<int8_t>(attr_sizes_.at(!col_id)) < 0; }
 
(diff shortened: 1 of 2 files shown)


src/include/storage/block_layout.h
```diff
@@ -23,27 +26,41 @@ struct BlockLayout {
...
   const uint16_t NumColumns() const { return static_cast<uint16_t>(attr_sizes_.size()); }
 
   /**
-   * attribute size at given col_id.
+   * @param col_id the column id to check for
+   * @return attribute size at given col_id.
+   */
+  uint8_t AttrSize(col_id_t col_id) const {
+    // mask off the first bit as we use that to check for varlen
+    return static_cast<uint8_t>(INT8_MAX & attr_sizes_.at(!col_id));
```
mbutrovich: Nit: it looks like INT8_MAX is coming from <cstdint> in common/constants.h via storage/storage_defs.h.

Do we want to standardize now on C++'s [numeric_limits (<limits>)](https://en.cppreference.com/w/cpp/types/numeric_limits) going forward rather than the C-style macros? That's what I've been using while working on index wrapper code.

tli2: Probably not, I don't see a very tangible benefit of using the std one other than from a pedantic stand point. There's little ambiguity in using the <cstdint> one and it's much more concise than the c++ one.

Issue #262, opened 2018-12-05
Improve serialization of delta records


swimj:
This is issue is related to both the correctness and performance of logging.  

Currently, we do a direct dump of the in-memory delta record (i.e. a ProjectedRow) to the log file on disk.  This was so we could get some basic logging infrastructure set up in the system, but was never the right thing to do.

Writing out the raw ProjectedRow has at least 2 problems:
- [x] (1) Log file does not store enough information to replay changes made to fields that use a varlen pool for storage.  This is because the value in the ProjectedRow is only a pointer to the varlen entry.  The log manager writes out the pointer instead of the actual contents of the delta. ~~ 

It will take some work to actually support this functionality (not just change a few lines), because by dumping out the entire delta, the log manager currently doesn't get its hands dirty with the details of ProjectedRow.  To differentiate between inline/varlen attributes, the log manager should at least have to iterate over the projection list.

- [x] (2) Writing out the entire in-memory ProjectedRow is inefficient.  For example, a ProjectedRow needs to have space for all attributes in the projection list, whether or not those attributes are actually null or not.  There is no need to write out the data corresponding to the "value" of a column that is actually null, but we currently do this because the size of a ProjectedRow only depends on the BlockLayout and projection list, and not on whether any of the columns happen to be null.  As another example, a ProjectedRow may have padding so that accesses at runtime are well-aligned. There's no reason for the log to keep this padding.

tli2:
Partially fixed by #370

GustavoAngulo:
@tli2 I believe #370 also addressed the second point. You can see in this code we skip over null attributes. We're able to do this because we also serialize out the null bitmap.
https://github.com/cmu-db/terrier/blob/6b1d5aaedaa3d4d8cb17b7e61b8397df2a185c2a/src/storage/write_ahead_log/log_serializer_task.cpp#L121-L126
And in this code, we only write out the value without padding:
https://github.com/cmu-db/terrier/blob/6b1d5aaedaa3d4d8cb17b7e61b8397df2a185c2a/src/storage/write_ahead_log/log_serializer_task.cpp#L143-L146

PR #263, opened 2018-12-05
New serialization format for log redo records.

swimj:
This PR is a significant rewrite of how the log manager serializes the delta record in a RedoRecord.  It goes most of the way towards fixing #262.

The log manager now dives into the guts of a delta record to determine how to serialize it.  The three essential pieces of information that are always written out are: (1) the number of columns, (2) the projection list, and (3) the null bitmap.  

The log manager then uses the null bitmap and block layout to determine whether/how to write out an attribute value.  If the column is null, then nothing is written out. Else, the log manager consults the BlockLayout to determine whether the column is inline or varlen. If inline, it writes out the delta's inline value; else, it follows the pointer to the varlen entry and writes out those contents.

As an example of how the log file can be processed on recovery, see the rewritten implementation of ReadNextRecord() within test/storage/log_test.cpp. We'll assume that the process responsible for recovery will have access to the BlockLayout through some other means (presumably with checkpoint / catalog).

As an initial proof of concept, I've tested this with the existing log tests (which do not have any varlen columns).  I also increased the number of rows in the test table to 10 (from 1), so that things are a little more robust. When we do "bookkeeping" with only 1 row, that means only 1 update ever actually gets written to the log file.  Adding tests for logging of varlen columns will be the next step.

### DIFF
--- a/src/include/storage/write_ahead_log/log_manager.h
+++ b/src/include/storage/write_ahead_log/log_manager.h
@@ -88,5 +88,10 @@ class LogManager {
   void WriteValue(const T &val) {
     out_.BufferWrite(&val, sizeof(T));
   }
+
+  // TODO(Justin): Figure out where the right place to glue this functionality in is. Probably the catalog?
+  SqlTable *LookupSqlTableFromOid(table_oid_t table_oid) {
+    return nullptr;
+  }
 };
 }  // namespace terrier::storage
--- a/src/include/storage/write_ahead_log/log_manager.h
+++ b/src/include/storage/write_ahead_log/log_manager.h
@@ -69,14 +69,14 @@ class LogManager {
 
  private:
   // TODO(Tianyu): This can be changed later to be include things that are not necessarily backed by a disk
-  // (e.g. logs can be streamed out to the network for remote replication)
+  //  (e.g. logs can be streamed out to the network for remote replication)
   BufferedLogWriter out_;
   RecordBufferSegmentPool *buffer_pool_;
 
   // TODO(Tianyu): Might not be necessary, since commit on txn manager is already protected with a latch
   common::SpinLatch flush_queue_latch_;
   // TODO(Tianyu): benchmark for if these should be concurrent data structures, and if we should apply the same
-  // optimization we applied to the GC queue.
+  //  optimization we applied to the GC queue.
   std::queue<RecordBufferSegment *> flush_queue_;
 
   // These do not need to be thread safe since the only thread adding or removing from it is the flushing thread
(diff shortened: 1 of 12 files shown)


swimj:
I will improve coverage on the untested cases (null column, varlen column, Delete) in the next few days, but I think the main logic is all ready for review.

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -69,14 +69,14 @@ class LogManager {
 
  private:
   // TODO(Tianyu): This can be changed later to be include things that are not necessarily backed by a disk
-  // (e.g. logs can be streamed out to the network for remote replication)
+  //  (e.g. logs can be streamed out to the network for remote replication)
```
tli2: nit: these look like unintentional changes

src/include/storage/write_ahead_log/log_record.h
```diff
@@ -142,32 +142,11 @@ class RedoRecord {
...
-    body->tuple_slot_ = tuple_slot;
-    return result;
-  }
-
  private:
   // TODO(Tianyu): We will eventually need to consult the DataTable to determine how to serialize a given column
-  // (varlen? compressed? from an outdated schema?) For now we just assume we can serialize everything out as-is,
-  // and the reader still have access to the layout on recovery and can deserialize. This is why we are not
-  // just taking an oid.
+  //  (varlen? compressed? from an outdated schema?) For now we just assume we can serialize everything out as-is,
```
tli2: same.

src/include/storage/write_ahead_log/log_manager.h
swimj: Oh, I think I may have done this so that CLion highlights multiline todos: https://blog.jetbrains.com/clion/2018/10/clion-2018-3-eap-profiler-multiline-todo/. I don't feel strongly either way (since it might look weird for anyone not using CLion), but it is nice to have the entire TODO highlighted consistently.

apavlo:
@songzhaozhe @codeworm96 @lmy1229 You guys are going to need to merge this PR into your code. It fixes storing VARCHAR values in log records.

swimj:
Oh cool, glad someone is picking back up on this! Let me know if you have
questions about this PR.

songzhaozhe:
How is the progress of tests (null column, varlen column, Delete)? Are you still working on it or we should take over?

apavlo:
@songzhaozhe Justin has graduated and is not working with us anymore. He has a new lover in California.

songzhaozhe:
Since we do not have access to the original branch, shall we fork out a new branch and create another pull request, or directly merge this branch into our code (in this case the GC group will have to branch out from our code)?

songzhaozhe:
The first step would be adding uncovered tests to ensure the details are correct.

swimj:
Sorry for the delayed response -- Andy's right (though I wouldn't quite put
it in those terms).

Did you figure out what to do with branching/merging? If there's anything I
can do to change the permissions on that branch, I'm happy to do so.

Thank you for extending my work / adding test coverage!

songzhaozhe:
Hi Justin, yes we have merged your branch into our PR and are working on it. So probably no further action is needed on your side.

tli2:
Incorporated into #370

PR #270, opened 2019-01-04
Network layer

tanujnay112:
Initial PR of network layer. It only receives messages through a port and sends an acknowledgement.


CMakeLists.txt
```diff
@@ -77,8 +77,8 @@ set(BUILD_SUPPORT_DIR "${CMAKE_SOURCE_DIR}/build-support")
 # CMAKE OPTIONS
 ###################################################
 
-set(TERRIER_CXXFLAGS "" CACHE STRING
```
tli2: Some builds on Travis seems to be failing. I don't think this is your issue. It may be that those failed builds are just running on an outdated pre-built image. Talk to Wan or Pervaze.

src/common/exception.cpp
```diff
@@ -0,0 +1,22 @@
...
+//
+// Copyright (c) 2015-2017, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#include "common/exception.h"
+
+namespace terrier {
+
+/*std::ostream &operator<<(std::ostream &os, const terrier::Exception &e) {
```
tli2: Can we remove this file?

src/common/exception.cpp
```diff
@@ -0,0 +1,22 @@
+//===----------------------------------------------------------------------===//
```
tli2: We can probably safely get rid of these headers. It's our code anyway.

src/common/internal_types.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include "common/internal_types.h"
+
+namespace terrier {
+
+// FileHandle INVALID_FILE_HANDLE;
+//
+
+// For threads
+size_t CONNECTION_THREAD_COUNT = 1;
```
tli2: Maybe move into 'common/constants.h' or start a `network/network_defs.h` as is customary in the new code base,

src/common/notifiable_task.cpp
```diff
@@ -0,0 +1,55 @@
...
+// Copyright (c) 2015-2018, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+#include "common/notifiable_task.h"
+#include <cstring>
+#include "common/event_util.h"
+#include "loggers/main_logger.h"
+
+namespace terrier {
```
tli2: also need `namespace common` in the new code base

src/common/utility.cpp
```diff
@@ -0,0 +1,66 @@
...
+#if __APPLE__
+extern "C" {
+#include <sys/cdefs.h>
+int close$NOCANCEL(int);
+};
+#endif
+
+namespace terrier {
+
+int peloton_close(int fd) {
```
tli2: We should probably retrofit this file a little. For this particular function:

- If the return code is not supposed to be checked, maybe there's no reason to return it in the wrapper
- the name should make clear that the close is final. (maybe `DefinitelyClose` or something like that)
- CamelCase naming convention in the new codebase. We also don't allow loose global functions like this one.

Also I am not convinced that this is deserving of its own file. One place I can think of is to merge this with `PosixIoWrappers` in `storage/write_ahead_log/log_io.h` and promote that to a file under `common`. It doesn't need to be done in this PR though.

src/include/common/container/lock_free_queue.h
```diff
@@ -0,0 +1,64 @@
+//===----------------------------------------------------------------------===//
```
tli2: No need to bring in files that are not used - we want everything in terrier to be used and tested. We can always bring these in when we need them.

src/include/common/exception.h
```diff
@@ -14,11 +14,12 @@ namespace terrier {
 
 #define NOT_IMPLEMENTED_EXCEPTION(msg) NotImplementedException(msg, __FILE__, __LINE__)
```
tli2: We probably want to hold off on porting over the exception system from peloton, since we are still discussing what to do with exceptions in the new system.

src/include/network/connection_handle.h
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  /**
+   * Handles a libevent event. This simply delegates the the state machine.
+   */
+  inline void HandleEvent(int, short) { state_machine_.Accept(Transition::WAKEUP, *this); }  // NOLINT
+
+  /* State Machine Actions */
+  // TODO(Tianyu): Write some documentation when feeling like it
+  inline Transition TryRead() { return io_wrapper_->FillReadBuffer(); }
```
tli2: As a general rule in the new code base `inline` is a bad word. Newer compiler versions pretty much either ignore it or treat it differently from its old semantics. Just remove all of them.

src/include/network/connection_handle_factory.h
```diff
@@ -0,0 +1,55 @@
...
+ * Each ConnectionHandle is associated with read and write buffers that are
+ * expensive to reallocate on the fly. Thus, instead of destroying these wrapper
+ * objects when they are out of scope, we save them until we can transfer their
+ * buffers to other wrappers.
+ */
+// TODO(Tianyu): Additionally, it is hard to make sure the ConnectionHandles
+// don't leak without this factory since they are essentially managed by
+// libevent if nothing in our system holds reference to them, and libevent
+// doesn't cleanup raw pointers.
+class ConnectionHandleFactory {
```
tli2: We will eventually want to replace this with the general purpose object pool from terrier. You don't have to do it in this PR though.

src/include/network/connection_handler_task.h
```diff
@@ -0,0 +1,78 @@
...
+#include <event2/listener.h>
+
+#include <unistd.h>
+
+#include "common/container/lock_free_queue.h"
+#include "common/exception.h"
+#include "common/notifiable_task.h"
+#include "loggers/main_logger.h"
+
+namespace terrier {
```
tli2: nit: in newer c++ you can nest namespaces in declaration, e.g. `namespace terrier::network`

src/include/network/marshal.h
```diff
@@ -0,0 +1,180 @@
...
+//
+// marshal.h
+//
+// Identification: src/include/network/marshal.h
+//
+// Copyright (c) 2015-2018, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+
+// #pragma once
```
tli2: same here: can remove file

src/include/network/network_io_utils.h
```diff
@@ -0,0 +1,426 @@
...
+#include <string>
+#include <vector>
+
+#include "common/exception.h"
+#include "common/internal_types.h"
+#include "util/portable_endian.h"
+
+namespace terrier {
+namespace network {
+#define _CAST(type, val) ((type)(val))
```
tli2: This looks pretty close to the buffer code we had to write for terrier. Again not that you have to do this in this PR, I think it might be interesting to look at the terrier code and explore the possibility of merging the two.

src/include/network/postgres_network_commands.h
```diff
@@ -0,0 +1,86 @@
...
+//                         Terrier
+//
+// postgres_network_commands.h
+//
+// Identification: src/include/network/postgres_network_commands.h
+//
+// Copyright (c) 2015-2018, Carnegie Mellon University Database Group
+//
+//===----------------------------------------------------------------------===//
+// #pragma once
```
tli2: I think we do want to be able to parse packets correctly in the initial port to make sure the state machines and everything is working. Basically all the commands should be stubs that return immediately, but everything above should be fully functional.

tanujnay112: Right now I made sure that the layer can read packets sent via ftp to port 2888 and reply with a generic message. Are you saying I plop this dummy functionality into the network commands?

src/include/common/exception.h
tanujnay112: This is added onto the exceptions stuff that the parser PR had added.

pervazea:
On the simple_query_test failure.

Ran in a Ubuntu 18.04 VM on my laptop. Fails intermittently, roughly 40% of the time. Unable to make it fail if run under gdb. Collecting a core file and examining, I see, failing on connection_handle.h:79

which is 
 Transition TryRead() { return io_wrapper_->FillReadBuffer(); }

io_wrapper_ appears to be 0.

Why does this not fail on Jenkins? With Peloton we saw many cases where failures were seen when run in environments with more cores, or few cores. In this case, it is probably fewer. My VM has 1 processor. The Jenkins machines have many cores, Travis, don't know, but probably few, since we use the free service from Travis.

pervazea:
The failure is occurring after the test completes when the test environment is in tear-down. In other words, shutdown of the networking pieces is not happening correctly.

src/include/network/connection_handle.h
```diff
@@ -0,0 +1,203 @@
...
+    static transition_result Delta_(ConnState state, Transition transition);
+    ConnState current_state_ = ConnState::READ;
+  };
+
+  friend class StateMachine;
+  friend class ConnectionHandleFactory;
+
+  // A raw pointer is used here because references cannot be rebound.
+  ConnectionHandlerTask *conn_handler_;
+  std::unique_ptr<NetworkIoWrapper> io_wrapper_;
```
pervazea: Should this be unique_ptr of type PosixSocketIoWrapper?

src/main/terrier.cpp
```diff
@@ -24,8 +25,10 @@ int main() {
     terrier::storage::init_index_logger();
     terrier::storage::init_storage_logger();
     terrier::transaction::init_transaction_logger();
+    terrier::network::TerrierServer terrier_server;
```
pervazea: I find this confusing to have the declaration in the middle of the logger initialization calls.
Move this to after all the logger initialization calls have been made, and possibly separate from them with a blank line.

src/include/network/connection_handle.h
tanujnay112: I think the point of this was to abstract away the implementation of NetworkIoWrapper in order for ease of switching out different implementations in Peloton?

src/main/terrier.cpp
```diff
@@ -26,7 +28,11 @@ int main() {
     terrier::storage::init_storage_logger();
     terrier::transaction::init_transaction_logger();
     terrier::parser::init_parser_logger();
+    terrier::network::init_network_logger();
 
+    terrier::network::TerrierServer terrier_server;
+
+    terrier_server.SetupServer().ServerLoop();
```
pervazea: These two lines (setting up the TerrierServer), should be after this try...catch block.
The try...catch block should be only for logger initialization.

test/network/network_test.cpp
```diff
@@ -1,3 +1,5 @@
+#pragma GCC diagnostic ignored "-Wconversion"
```
mbutrovich: This shouldn't be necessary. If you were getting warnings about implicit conversation, those are typically resolved with an explicit static_cast.

tanujnay112: GCC still seems to complain about that static_cast from size_t (uint64_t) to int32_t due to information loss though.

src/include/network/connection_handle.h
pervazea: It seems unlikely to me that we'll be implementing other network wrappers. So two options:
1. Change it to Posix... wrapper, and add a comment that this should be changed to be a generic NetworkIoWrapper if switchable wrappers implemented in future.
2. Add a comment stating that this is currently a PosixSocketIOWrapper. 
I'd probably do 1), because it will make looking up classes and definitions in CLion work naturally, rather than having to search and find the actual class used.

src/network/connection_dispatcher_task.cpp
```diff
@@ -0,0 +1,64 @@
...
+  }
+}
+
+void ConnectionDispatcherTask::DispatchConnection(int fd, int16_t) {  // NOLINT
+  struct sockaddr_storage addr;
+  socklen_t addrlen = sizeof(addr);
+
+  int new_conn_fd = accept(fd, reinterpret_cast<struct sockaddr *>(&addr), &addrlen);
+  if (new_conn_fd == -1) {
+    NETWORK_LOG_ERROR("Failed to accept");
```
pervazea: Need error handling here. If the accept failed, we are continuing through the normal path.

src/network/connection_handler_task.cpp
```diff
@@ -0,0 +1,43 @@
...
+#include "network/connection_handle.h"
+#include "network/connection_handle_factory.h"
+
+namespace terrier::network {
+
+ConnectionHandlerTask::ConnectionHandlerTask(const int task_id) : NotifiableTask(task_id) {
+  int fds[2];
+  if (pipe(fds) != 0) {
+    NETWORK_LOG_ERROR("Can't create notify pipe to accept connections");
+    exit(1);
```
pervazea: Is this the right error handling? Are there other options that allow continued operation?

src/network/connection_handler_task.cpp
```diff
@@ -0,0 +1,43 @@
...
+  }
+  new_conn_send_fd_ = fds[1];
+  RegisterEvent(fds[0], EV_READ | EV_PERSIST, METHOD_AS_CALLBACK(ConnectionHandlerTask, HandleDispatch), this);
+}
+
+void ConnectionHandlerTask::Notify(int conn_fd) {
+  int buf[1];
+  buf[0] = conn_fd;
+  if (write(new_conn_send_fd_, buf, sizeof(int)) != sizeof(int)) {
+    NETWORK_LOG_ERROR("Failed to write to thread notify pipe");
```
pervazea: Error recovery? What happens if the write fails?

src/network/connection_handler_task.cpp
```diff
@@ -0,0 +1,43 @@
...
+void ConnectionHandlerTask::HandleDispatch(int new_conn_recv_fd, int16_t) {  // NOLINT as we don't use the flags arg
+  // buffer used to receive messages from the main thread
+  char client_fd[sizeof(int)];
+  size_t bytes_read = 0;
+
+  // read fully
+  while (bytes_read < sizeof(int)) {
+    ssize_t result = read(new_conn_recv_fd, client_fd + bytes_read, sizeof(int) - bytes_read);
+    if (result < 0) {
+      NETWORK_LOG_ERROR("Error when reading from dispatch");
```
pervazea: Error recovery? 
If the read fails, you don't want to -1 from bytes_read. Any other action required?

src/network/connection_handler_task.cpp
```diff
@@ -0,0 +1,44 @@
...
+void ConnectionHandlerTask::HandleDispatch(int new_conn_recv_fd, int16_t) {  // NOLINT as we don't use the flags arg
+  // buffer used to receive messages from the main thread
+  char client_fd[sizeof(int)];
+  size_t bytes_read = 0;
+
+  // read fully
+  while (bytes_read < sizeof(int)) {
+    ssize_t result = read(new_conn_recv_fd, client_fd + bytes_read, sizeof(int) - bytes_read);
+    if (result < 0) {
+      NETWORK_LOG_ERROR("Error when reading from dispatch: errno {}", errno);
```
pervazea: In the event of a read error, is error recovery required? Logs and ignores at present.

test/network/network_test.cpp
```diff
@@ -0,0 +1,366 @@
...
+    EXPECT_TRUE(false);
+  }
+  LOG_DEBUG("[SimpleQueryTest] Client has closed");
+}
+
+ssize_t ReadUntilReadyOrClose(char *in_buffer, size_t max_len, int socket_fd) {
+  ssize_t n;
+  while (true) {
+    n = read(socket_fd, in_buffer, max_len);
+    if (n == 0 || in_buffer[n - 6] == 'Z')  // Ready for request
```
pervazea: If n > 0 but n < 6, this is going to produce an invalid array index.

test/network/network_test.cpp
```diff
@@ -0,0 +1,366 @@
...
+  serv_addr.sin_port = htons(port);
+
+  int64_t ret = connect(socket_fd, reinterpret_cast<sockaddr *>(&serv_addr), sizeof(serv_addr));
+  if (ret < 0) LOG_ERROR("Connection Error");
+
+  // Build the startup message
+  char out_buffer[TEST_BUF_SIZE] = {};
+  char in_buffer[TEST_BUF_SIZE] = {};
+  // 3: protocol version number
+  out_buffer[5] = 3;
```
pervazea: 5?
This should be done differently. Actual implementation can be done several ways, but the general approach should be one of defining the packet structure and then using the fields to get and set values.

Peloton defines in marshal.h, struct OutputPacket for instance. 
Terrier defines in storage/projected_row.h, a class, used for MEM_REINTERPRETATION_ONLY, overlaying structure onto a byte buffer.

Input and output packets should use an approach as above, or equivalent. This comment also should be taken to apply to any other place where packet contents are read and written, in the code or tests.

test/network/network_test.cpp
pervazea: Take another look. It seemed to work for me in suppressing the warnings, but I only did a sample.

tanujnay112: Ya I did it in the end. My bad

test/network/network_test.cpp
tanujnay112: @DarkForte Could you look at this

test/network/network_test.cpp
tanujnay112: @DarkForte

pervazea: Later often never happens :-)

DarkForte: Sorry for not keeping you updated. T_T
I was trying to support it in PostgresPacketWriter by adding more functions since Traffic Cop needs a utility to send the ResultSet back to the client anyway. I think we can just tell PostgresPacketWriter everything and it will write the packet for us. If that succeeded, all the packet writing code can be eliminated.

src/include/network/connection_handle.h
tanujnay112: Alright sounds good.

src/include/network/network_defs.h
```diff
@@ -0,0 +1,164 @@
...
+//===--------------------------------------------------------------------===//
+// Result Types
+//===--------------------------------------------------------------------===//
+
+enum class ResultType {
+  INVALID = INVALID_TYPE_ID,  // invalid result type
+  SUCCESS = 1,
+  FAILURE = 2,
+  ABORTED = 3,   // aborted
+  NOOP = 4,      // no op  // TODO(tanujnay112) Remove this type
```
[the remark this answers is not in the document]
tanujnay112: I'm thinking of keeping all these for future use?

src/include/network/postgres_protocol_utils.h
```diff
@@ -0,0 +1,383 @@
...
+#include <utility>
+#include <vector>
+
+#include "network/network_defs.h"
+#include "network/network_io_utils.h"
+
+#define NULL_CONTENT_SIZE (-1)
+namespace terrier::network {
+
+// TODO(Tianyu): It looks very broken that this never changes.
```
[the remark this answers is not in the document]
tanujnay112: These all seem to have been pulled off the list of parameter values in the network protocol documentation for Postgres. I don't think they matter at this stage of testing right? This is just information that the client may or may not need to know as you said.

PR #274, opened 2019-01-15
Alignment

tli2:
Arrow requires that the start of every contiguous buffer be 8-byte aligned. We aligned up to the alignment of the element it held. This is just a fix to make ours adhere to the arrow principle.

While I was at it, 16-byte varlens now align at 8-byte boundary instead of 16, which would be useless.


mbutrovich:
10% drop in TAS' SimpleInsert benchmark is interesting. Did we anticipate that and is it reproducible?

mbutrovich:
After investigating, that performance dip is because it's comparing before #268 which actually introduced the regression and was known.

PR #276, opened 2019-01-17
More test content in parser test

DarkForte:
Added some test content for the old parser tests.


test/parser/parser_test.cpp
```diff
@@ -1,3 +1,4 @@
+#include <parser/expression/aggregate_expression.h>
```
lmwnshn: Maybe move this to line 7 to match the other includes in style.

DarkForte: Got it. This is auto-generated by CLion.

DarkForte:
One of the tests `DISABLED_OldStringUpdateTest` says "need varchar support", and the others say "need string support". Do we have them now?

apavlo:
We should merge this, right?

lmwnshn:
We should/could. We were thinking about waiting for type::Value to stabilize (@mbutrovich) because of the strings that the parser spits out, but everything else is fine / it is a small change we can make at a future date.

Per yesterday's discussion, type::Value will be immutable so we don't need to do the delete in tests any more @DarkForte.

test/parser/parser_test.cpp
```diff
@@ -1504,50 +1608,91 @@ TEST_F(ParserTestBase, OldCaseTest) {
...
-  for (const auto &query : valid_queries) {
+  {
+    query = "INSERT INTO test_table VALUES (1, 2, '2017-01-01'::DATE);";
     auto stmt_list = pgparser.BuildParseTree(query);
+    auto statement = reinterpret_cast<InsertStatement *>(stmt_list[0].get());
+    auto values = *(statement->GetValues());
+    auto cast_expr = reinterpret_cast<TypeCastExpression *>(values[0][2].get());
+    EXPECT_EQ(type::TypeId::DATE, cast_expr->GetReturnValueType());
+
+    // TODO(Weichen): add back strcmp when mem leak is solved
```
lmwnshn: I think you can test for this now.

test/parser/parser_test.cpp
```diff
@@ -595,41 +598,97 @@ TEST_F(ParserTestBase, CompareTest) {
...
+  {
+    query = "SELECT MIN(*) FROM foo;";
+    auto stmt_list = pgparser.BuildParseTree(query);
+    EXPECT_EQ(1, stmt_list.size());
+    EXPECT_EQ(StatementType::SELECT, stmt_list[0]->GetType());
+
+    auto statement = reinterpret_cast<SelectStatement *>(stmt_list[0].get());
+    EXPECT_EQ("foo", statement->GetSelectTable()->GetTableName());
+    EXPECT_EQ(ExpressionType::AGGREGATE_MIN, statement->GetSelectColumns()[0]->GetExpressionType());
   }
```
lmwnshn: Could you add AVG so that we're hitting it like in Peloton?

test/parser/parser_test.cpp
```diff
@@ -1187,8 +1283,7 @@ TEST_F(ParserTestBase, OldCreateSchemaTest) {
 }
 
 // NOLINTNEXTLINE
-TEST_F(ParserTestBase, DISABLED_OldCreateViewTest) {
-  // TODO(WAN): need strings
+TEST_F(ParserTestBase, OldCreateViewTest) {
   std::string query = "CREATE VIEW comedies AS SELECT * FROM films WHERE kind = 'Comedy';";
   auto stmt_list = pgparser.BuildParseTree(query);
   auto create_stmt = reinterpret_cast<CreateStatement *>(stmt_list[0].get());
```
lmwnshn: Could you check for the Comedy string?

PR #286, opened 2019-02-05
Add flags to VarlenEntry (#285)

amlatyrngom:
This PR refactors VarlenEntry object to support prefix storage and inlining of short varlens. The concept of gathered values is generalized to "reclaimability". This will help us implement a number of optimizations in the execution layer and support Arrow transformation better.


src/include/storage/storage_defs.h
```diff
@@ -176,30 +178,48 @@ class VarlenEntry {
...
    * @return pointer to the varlen entry contents.
    */
   const byte *Content() const { return content_; }
 
  private:
-  // we use the sign bit to denote if
   int32_t size_;
-  // TODO(Tianyu): we can use the extra 4 bytes for something else (storing the prefix?)
+  // flags: stores the gathered bit and the compressed bit (only 2 bits are needed).
+  int32_t flags_;
```
tli2: Can we turn these into uints if we don't check for sign bit?

src/include/storage/storage_defs.h
```diff
@@ -176,30 +178,48 @@ class VarlenEntry {
...
+
+  /**
+   * @return whether the content is stored using dictionary compression.
+   */
+  bool IsCompressed() const { return static_cast<bool>((flags_ >> VARLEN_COMPRESSED) & 1); }
+
+  /**
+   * @return whether the content is inlined or not.
+   */
+  bool IsInlined() const { return static_cast<bool>(Size() <= sizeof(int64_t)); }
```
tli2: It would be ideal if we can also use up the spare bytes from the flags_ field for inlining varlens

src/include/storage/storage_defs.h
```diff
@@ -176,30 +178,48 @@ class VarlenEntry {
...
+
+  /**
+   * Helper method to decide if the content can be GCed separately
+   * @return whether the content can be deallocated by itself
+   */
+  bool IsReclaimable() const { return !(IsGathered() || IsCompressed() || IsInlined()); }
```
tli2: Naming-wise, it really isn't about whether  we 'can' reclaim the varlen, but about whether we 'should.' So maybe change to "ShouldReclaim" or "NeedReclaim"?

src/include/storage/storage_defs.h
```diff
@@ -176,30 +178,48 @@ class VarlenEntry {
    * @param size length of the varlen content, in bytes (no C-style nul-terminator)
    * @param gathered whether the varlen entry's content pointer is part of a large buffer (for arrow-compatibility),
    *                 which means it cannot be deallocated by itself.
+   * @param compressed whether the varlen entry is stored using dictionnary compression
    */
-  VarlenEntry(byte *content, uint32_t size, bool gathered)
-      // the sign bit on size is used to store the "gathered" attribute, so we mask it off on size depending on that.
-      : size_(size | (gathered ? INT32_MIN : 0)), content_(content) {}
+  VarlenEntry(byte *content, uint32_t size, bool gathered, bool compressed)
```
tli2: Two boolean arguments are confusing when reading code. Let's take in an enum for the constructor so it's explicit. e.g. VarlenEntry(bytes, VarlenEntryType::COMPRESSED)

mbutrovich:
Can we update the PR description to give an overview of what this PR's purpose is?

mbutrovich:
Looks like merging in master makes it now fail Doxygen on CI.

PR #288, opened 2019-02-06
Add early stop for sequential scans down a partially filled block

tli2:
Sequential scan can now stop early.


src/include/storage/data_table.h
```diff
@@ -166,11 +166,17 @@ class DataTable {
...
+   * concurrent accesses are happening, as inserts maybe in flight. However, the number given is always transactionally
+   * correct, as any inserts that might have happened is not going to be visible to the calling transaction.
+   *
+   * @return one past the last tuple slot contained in the data table.
    */
   SlotIterator end() const {
     common::SpinLatch::ScopedSpinLatch guard(&blocks_latch_);
-    return {this, blocks_.end(), 0};
+    // TODO(Tianyu): Need to look in detail at how this interacts with compaction when that gets in.
+    return blocks_.empty() ? SlotIterator(this, blocks_.end(), 0)
```
apavlo: This seems like some complex logic here. I would add a comment about what the two branches are doing.

PR #289, opened 2019-02-07
Fix cases where committed transactions may be empty

### LINKED ISSUE #287: GC test failure on 32-core on high abort
C++ exception of vector index access out of bounds reported by student. Because the access was UINT64_MAX into a 0-size array, overflow is the likely cause.

tli2:
fix #287


crd477:
On 2019-02-08 18:14, Matt Butrovich wrote:
> LGTM. Waiting on macOS machine for CI which looks busted at the moment. 
> @crd477 <https://github.com/crd477>
> 

The worker VM was hung again.  I suspect there have been intermittent 
network problems where the system is in WeH.  I'll move the iMac host to 
a more reliable network next week.

-- 
     -Chad

mbutrovich:
Thank you for your help, @crd477! It's always appreciated!

PR #290, opened 2019-02-09
type::TransientValue

mbutrovich:
`TransientValue` class. This an immutable value transport class for bringing C types from the parser into SQL types (`TransientValueFactory`), or alternatively extracting C types from SQL types (`TransientValuePeeker`). The latter might end up not being needed, but it's good for testing and debugging right now. This is a stripped down version of Values from Peloton:

1. removes comparators and arithmetic operators. These should be handled by the future execution engine's type system. Index scans will have their own comparators to reason about ProjectedRows, to avoid having to instantiate Value objects to do comparisons
2. type promotion, see 1)
3. Serialization and deserialization from storage layouts. We don't want to rely on creating `TransientValue` objects for query execution, simply for transport into the system from the parser. Putting it back out on the wire should be doable without going to TransientValues in between

Looking for a few reviews since this touches the parser as well.


src/include/type/transient_value.h
```diff
@@ -0,0 +1,282 @@
...
+  TypeId Type() const {
+    // bitwise AND the TypeId with 01111111 to return TypeId value without the embedded NULL bit
+    return static_cast<TypeId>(static_cast<uint8_t>(type_) & 0x7F);
+  }
+
+  /**
+   * Can't see any reason to have a default constructor. If we were to have one, it would probably just set the TypeId
+   * to boolean and SetNull(true) but I can't come up with a good reason to expose this.
+   */
+  TransientValue() = delete;
```
tli2: nit: Since you already has another constructor it's okay to omit this line

src/include/type/transient_value.h
```diff
@@ -0,0 +1,282 @@
...
+  /**
+   * @param rhs TransientValue to compare this against
+   * @return true if two TransientValues are equal, satisfying the following predicates:
+   * 1) they share the same TypeId
+   * 2) they share the same NULL value
+   * 3) they share the same data_ value if non-VARCHARs, or their VARCHAR contents are the same if VARCHARS
+   * false otherwise
+   */
+  bool operator==(const TransientValue &rhs) const {
+    if (type_ != rhs.type_) return false;  // checks TypeId and NULL at the same time due to stolen MSB of type_ field
```
tli2: I believe in SQL, null == null should either be false or null.

src/include/type/transient_value.h
```diff
@@ -0,0 +1,282 @@
...
+    const char *const rhs_varchar_contents = rhs_varchar + sizeof(uint32_t);
+
+    return std::memcmp(varchar_contents, rhs_varchar_contents, length) == 0;
+  }
+
+  /**
+   * @param rhs TransientValue to compare this against
+   * @return Negation of @see TransientValue::operator==
+   */
+  bool operator!=(const TransientValue &rhs) const { return !(operator==(rhs)); }
```
tli2: For the same reason above this actually is not guaranteed.

src/include/type/transient_value.h
```diff
@@ -0,0 +1,282 @@
...
+    // copy the VARCHAR contents into the VARCHAR buffer
+    char *const varchar_contents = varchar + sizeof(uint32_t);
+    const char *const other_varchar_contents = other_varchar + sizeof(uint32_t);
+    std::memcpy(varchar_contents, other_varchar_contents, length);
+    // copy the pointer to the VARCHAR buffer into the internal buffer
+    data_ = reinterpret_cast<uintptr_t>(varchar);
+  }
+
+  TypeId type_ = TypeId::INVALID;
+  uintptr_t data_;
```
tli2: Not saying we do this in this PR, but this should leave 7 bytes of padding, giving us a total of 15 bytes to actually play with if we ever want to inline small varlens

apavlo:
@mbutrovich I can review after you address @tli2 comments.

src/include/type/transient_value.h
mbutrovich: These are following object equality semantics and this operator wouldn't be used for SQL comparisons (like if the parser needs to check equality of nodes with contained TransientValues). If you think that's incorrect we can talk about it, but the idea would be SQL predicate evaluations are only done in the index comparators which will be handled elsewhere.

src/include/type/transient_value.h
mbutrovich: I originally wrote the `data_` field as a byte array and reinterpreted accordingly. I could just change it back now if you think we want those 15 bytes now.

src/include/type/transient_value.h
tli2: That's reasonable. Let's note that explicitly in the doc though that this is not equivalent to the SQL equal operator

src/include/type/transient_value.h
tli2: I think a TODO would be fine.

src/include/type/transient_value.h
apavlo: Should we even include `==` then? I am thinking about collations. 

When would anybody actually need to compare to `TransientValues` to see whether they are equivalent?

src/include/type/type_id.h
```diff
@@ -22,10 +16,7 @@ enum class TypeId : uint8_t {
   DECIMAL,
   TIMESTAMP,
   DATE,
-  STRING,
   VARCHAR,
-  VARBINARY,
-  ARRAY,
```
apavlo: We should talk about what it will take to add back ARRAY after we merge this PR.

test/type/transient_value_test.cpp
```diff
@@ -0,0 +1,392 @@
...
+ * 3) Assert that the new TransientValue is not NULL
+ * 4) Assert that the C type returned by TransientValuePeeker matches the reference C type data
+ * 5) Flip a coin for reference NULL value
+ * 6) Test SetNull, compare NULL value to reference
+ * 7) Clear SetNull, compare to original reference C type
+ * 8) Test copy constructor
+ * 9) Test copy assignment operator
+ * 10) Test move constructor
+ * 11) Test move assignment operator
+ */
```
apavlo: Does it make sense to include a simple test that checks that we can use `TransientValue` in a std::map? Or is `Hash` enough?

mbutrovich: std::map provides ordering, so I could write a < operator if we want that. std::unordered_map should be fine with our Hash function and the == operator.

mbutrovich:
It's randomly failing CI because of a race in `make`. If it tries to build terrier before the `gflags_ep`, `googletest_ep`, and `gbenchmark_ep` targets are done, it fails due to `gtest_prod.h` missing. I thought it was just an issue with the Coverage build, but I'll poke at all of the CI configs to make sure they all work correctly.

PR #292, opened 2019-02-12
Adding more parser tests for coverage

Ghatage:
Added more unit tests to enhance existing code coverage for postgresparser.cpp

### DIFF
--- a/test/parser/parser_test.cpp
+++ b/test/parser/parser_test.cpp
@@ -318,6 +318,102 @@ TEST_F(ParserTestBase, UpdateTest) {
   EXPECT_EQ(update_stmt->GetUpdateCondition(), nullptr);
 }
 
+// NOLINTNEXTLINE
+TEST_F(ParserTestBase, OperatorTest) {
+  {
+    std::string query = "SELECT 10+10 AS Addition;";
+    auto stmt_list = pgparser.BuildParseTree(query);
+    auto &sql_stmt = stmt_list[0];
+    auto select_stmt = reinterpret_cast<SelectStatement *>(sql_stmt.get());
+    auto expr = select_stmt->GetSelectColumns().at(0).get();
+    EXPECT_EQ(expr->GetExpressionType(), ExpressionType::OPERATOR_PLUS);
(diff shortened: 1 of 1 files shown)


test/parser/parser_test.cpp
```diff
@@ -318,6 +318,165 @@ TEST_F(ParserTestBase, UpdateTest) {
...
+    EXPECT_EQ(expr->GetExpressionType(), ExpressionType::OPERATOR_MOD);
+  }
+
+  {
+    std::string query = "SELECT CAST('100' AS INTEGER) AS Casting;";
+    auto stmt_list = pgparser.BuildParseTree(query);
+    auto &sql_stmt = stmt_list[0];
+    auto select_stmt = reinterpret_cast<SelectStatement *>(sql_stmt.get());
+    auto expr = select_stmt->GetSelectColumns().at(0).get();
+    EXPECT_EQ(expr->GetExpressionType(), ExpressionType::CAST);
```
Ghatage: I understand most of the code is actively being developed, but was not sure what the difference was in `ExpressionType::OPERATOR_CAST` and `ExpressionType::CAST` from expression_defs.h so added this atm.
cc @lmwnshn

apavlo:
@Ghatage Anup! Thanks! We'll look into this tomorrow!

test/parser/parser_test.cpp
lmwnshn: @Ghatage I thought there was a difference, but I've been digging around the old Peloton codebase and libpg_query source and you're right -- OPERATOR_CAST isn't being used anywhere, even though it was introduced in the same commit that CAST was. Could you go ahead and merge the functionality of the two?

Me and @pervazea talked briefly, we think OPERATOR_CAST is more descriptive and consistent. This should only affect your tests and TypecastExpression.

Ghatage: I have another PR coming for more coverage will try and add those changes in that PR.

lmwnshn: Thanks!

Issue #264, opened 2018-12-06
bwtree unit test fails


pervazea:
Working on parser and plan nodes. Pulled from master yesterday and merged it into the branch on which Wan and I are working.

When running unit tests this morning, the bwtree test fails (run time 168.86 sec).

No other information at this time. Just ran once so far.

tli2:
I think it reliably fails on dev4 as of master today

yeshengm:
Here's a screenshot of the failed test case.

![screenshot from 2019-02-12 22-06-56](https://user-images.githubusercontent.com/12269969/52683983-a07d6500-2f12-11e9-9188-0ed6683b1046.png)

mbutrovich:
What machine is this on: specifically how many threads? This has not reproduced on any machines I've tried, but can try dev4 since @tli2 says he saw failures there.

yeshengm:
I'm testing on EC2 c5.9xlarge (36 cores).

mbutrovich:
I can reproduce it in Terrier on dev4 which is just a port of @wangziqi2013's tests. I'm trying the original BwTree repo to see if it reproduces there too.

mbutrovich:
Reproduced on @wangziqi2013's BwTree repo. I'll track him down for help. :)

[bwtree_test_failure.txt](https://github.com/cmu-db/terrier/files/2861256/bwtree_test_failure.txt)

mbutrovich:
I think we're good here.

PR #293, opened 2019-02-13
bwtree: reconcile mapping table discrepancy, add recycling of NodeIds

mbutrovich:
Fix at least one issue in #264. This should fix the problem on systems with large numbers of threads (32+) running bwtree_test. Still unclear about the single-threaded VM case @pervazea has seen.

This PR changes 2 things in the BwTree:
1. The MAPPING_TABLE_SIZE that is used in assertions throughout the code isn't actually the value that was used in mmap to allocate the mapping table. That was hardcoded at 1GB (per @wangziqi2013 this is what Microsoft suggested), but the MAPPING_TABLE_SIZE was never changed to reflect this.
2. NodeIds were never recycled. Made a one line change at @wangziqi2013's suggestion to add NodeId recycling in.

Eventually we may want to add a resizing or reallocating feature for the mapping table to the bwtree.


mbutrovich:
@malin1993ml we should probably make sure this makes it into the `15721-s19-project1` branch so students don't see spurious failures on their large EC2 instances.

PR #295, opened 2019-02-14
Merge CAST() and ExpressionType::OPERATOR_CAST

Ghatage:
As discussed in #292 merging the CAST() operator with the ExpressionType::OPERATOR_CAST.
All `OPERATOR_*` operators are treated as AbstractExpressions which are handled as `T_A_Expr` by `AExprTransform()`, while CAST() was being handled separately as T_TypeCast under TypeCastTransform.

In order to make all cast operations use OPERATOR_CAST() we need to make the casting operation of the AbstractExpression type - which is what is being done here.

So now a casting operator goes through AExprTransform() where it gets triaged to the existing TypeCastTransform() there by reusing the existing code.

### DIFF
--- a/src/parser/postgresparser.cpp
+++ b/src/parser/postgresparser.cpp
@@ -409,6 +409,8 @@ std::unique_ptr<AbstractExpression> PostgresParser::AExprTransform(A_Expr *root)
 
   if (root->kind == AEXPR_DISTINCT) {
     target_type = ExpressionType::COMPARE_IS_DISTINCT_FROM;
+    children.emplace_back(ExprTransform(root->lexpr));
+    children.emplace_back(ExprTransform(root->rexpr));
   } else if (root->kind == AEXPR_OP && root->type == T_TypeCast) {
     target_type = ExpressionType::OPERATOR_CAST;
   } else {
--- a/src/parser/postgresparser.cpp
+++ b/src/parser/postgresparser.cpp
@@ -230,7 +230,7 @@ std::unique_ptr<AbstractExpression> PostgresParser::ExprTransform(Node *node) {
       break;
     }
     case T_TypeCast: {
-      expr = TypeCastTransform(reinterpret_cast<TypeCast *>(node));
+      expr = AExprTransform(reinterpret_cast<A_Expr *>(node));
       break;
     }
     default: {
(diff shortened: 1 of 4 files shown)


src/parser/postgresparser.cpp
```diff
@@ -411,18 +405,19 @@ std::unique_ptr<AbstractExpression> PostgresParser::AExprTransform(A_Expr *root)
...
     auto name = (reinterpret_cast<value *>(root->name->head->data.ptr_value))->val.str;
     target_type = StringToExpressionType(name);
+    children.emplace_back(ExprTransform(root->lexpr));
+    children.emplace_back(ExprTransform(root->rexpr));
   }
 
-  std::vector<std::shared_ptr<AbstractExpression>> children;
-  children.emplace_back(ExprTransform(root->lexpr));
-  children.emplace_back(ExprTransform(root->rexpr));
-
```
lmwnshn: Do we need to look at root-lexpr and root->rexpr when root->kind is AEXPR_DISTINCT?

src/parser/postgresparser.cpp
```diff
@@ -230,7 +230,7 @@ std::unique_ptr<AbstractExpression> PostgresParser::ExprTransform(Node *node) {
       break;
     }
     case T_TypeCast: {
-      expr = TypeCastTransform(reinterpret_cast<TypeCast *>(node));
+      expr = AExprTransform(reinterpret_cast<A_Expr *>(node));
```
lmwnshn: I don't have a strong opinion on this, but is there a reason we're making it go through AExprTransform first?

src/parser/postgresparser.cpp
Ghatage: AEXPR_DISTINCT does seem to use root->lexpr and root->rexpr.
I'll push a change to and fix that.

src/parser/postgresparser.cpp
Ghatage: Yes.. mostly for consistency and a little bit for my pedantic itch.
The original TypeCast expression is T_TypeCast which is handled separately while OPERATOR_CAST is T_A_Expr and those are handled together through AExprTransform.

lmwnshn: Sure, sounds good.

src/parser/postgresparser.cpp
lmwnshn: Okay, thanks. Ready to merge once that's in.

PR #296, opened 2019-02-14
More parser test for coverage

Ghatage:
Enhanced a few existing tests to reach more code / cases.

### DIFF
--- a/test/parser/parser_test.cpp
+++ b/test/parser/parser_test.cpp
@@ -99,6 +99,42 @@ TEST_F(ParserTestBase, CreateFunctionTest) {
       "BEGIN RETURN i + 1; END; $$ "
       "LANGUAGE plpgsql;";
   stmts = pgparser.BuildParseTree(query);
+
+  query =
+      "CREATE OR REPLACE FUNCTION return_varchar ("
+      " i VARCHAR"
+      " )"
+      " RETURNS VARCHAR AS $$ "
+      "BEGIN RETURN 'foo'; END; $$ "
+      "LANGUAGE plpgsql;";
+  stmts = pgparser.BuildParseTree(query);
+
+  query =
+      "CREATE OR REPLACE FUNCTION return_text ("
+      " i TEXT"
+      " )"
+      " RETURNS TEXT AS $$ "
+      "BEGIN RETURN 'foo'; END; $$ "
+      "LANGUAGE plpgsql;";
+  stmts = pgparser.BuildParseTree(query);
+
+  query =
+      "CREATE OR REPLACE FUNCTION return_bool ("
+      " i BOOLEAN"
+      " )"
+      " RETURNS BOOLEAN AS $$ "
+      "BEGIN RETURN false; END; $$ "
+      "LANGUAGE plpgsql;";
+  stmts = pgparser.BuildParseTree(query);
+
+  query =
+      "CREATE OR REPLACE FUNCTION return_real ("
+      " i REAL"
+      " )"
+      " RETURNS REAL AS $$ "
+      "BEGIN RETURN 1.0; END; $$ "
+      "LANGUAGE plpgsql;";
+  stmts = pgparser.BuildParseTree(query);
 }
 
 // NOLINTNEXTLINE
--- a/test/parser/parser_test.cpp
+++ b/test/parser/parser_test.cpp
@@ -81,6 +81,12 @@ TEST_F(ParserTestBase, CreateFunctionTest) {
       " BEGIN RETURN i + 1; END; $$ "
       "LANGUAGE plpgsql;";
   auto stmts = pgparser.BuildParseTree(query);
+  auto stmt = reinterpret_cast<CreateFunctionStatement *>(stmts[0].get());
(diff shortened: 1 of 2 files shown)


test/parser/parser_test.cpp
```diff
@@ -99,6 +99,42 @@ TEST_F(ParserTestBase, CreateFunctionTest) {
       "BEGIN RETURN i + 1; END; $$ "
       "LANGUAGE plpgsql;";
   stmts = pgparser.BuildParseTree(query);
+
+  query =
```
apavlo: We probably don't want to have too many FUNCTION tests since it's going to be a while until we support pl/pgsql again.

test/parser/parser_test.cpp
```diff
@@ -221,6 +257,8 @@ TEST_F(ParserTestBase, GarbageTest) {
 TEST_F(ParserTestBase, InsertTest) {
   auto stmts = pgparser.BuildParseTree("INSERT INTO foo VALUES (1, 2, 3), (4, 5, 6);");
   EXPECT_EQ(stmts.size(), 1);
+  stmts = pgparser.BuildParseTree("INSERT INTO foo (id,bar,entry) VALUES (1, 2, 3);");
+  EXPECT_EQ(stmts.size(), 1);
```
apavlo: Is there anything more substantial that we could be testing here?

test/parser/parser_test.cpp
```diff
@@ -404,6 +448,19 @@ TEST_F(ParserTestBase, OperatorTest) {
...
+  {
+    std::vector<std::string> queries;
+
+    // Coverage for NullNodeTransform
+    queries.emplace_back("SELECT * FROM foo WHERE 0 IS NULL;");
+    queries.emplace_back("SELECT * FROM foo WHERE 0*1 IS NULL;");
+    queries.emplace_back("SELECT * FROM foo WHERE ? IS NULL;");
+
+    for (const auto &query : queries) {
+      auto stmt_list = pgparser.BuildParseTree(query);
```
apavlo: We need to be checking something here. We're building the parse tree so this is a false code coverage result

test/parser/parser_test.cpp
```diff
@@ -1250,7 +1307,7 @@ TEST_F(ParserTestBase, OldCreateTriggerTest) {
       "BEFORE UPDATE OF balance ON accounts "
       "FOR EACH ROW "
       "WHEN (OLD.balance <> NEW.balance) "
-      "EXECUTE PROCEDURE check_account_update();";
+      "EXECUTE PROCEDURE check_account_update(update_date);";
```
apavlo: Why does this matter?

test/parser/parser_test.cpp
```diff
@@ -1379,6 +1438,8 @@ TEST_F(ParserTestBase, OldUDFFuncCallTest) {
 TEST_F(ParserTestBase, OldCaseTest) {
   std::string query = "SELECT id, case when id=100 then 1 else 0 end from tbl;";
   auto stmt_list = pgparser.BuildParseTree(query);
+  query = "SELECT id, case id when 100 then 1 when 200 then 2 end from tbl;";
+  stmt_list = pgparser.BuildParseTree(query);
```
apavlo: Again, missing real checks here.

test/parser/parser_test.cpp
Ghatage: You're right, I could do much better on some of these tests that are missing verification.
I'll put in the checks for the new ones and push an update.

test/parser/parser_test.cpp
```diff
@@ -99,6 +99,42 @@ TEST_F(ParserTestBase, CreateFunctionTest) {
...
+      " i TEXT"
+      " )"
+      " RETURNS TEXT AS $$ "
+      "BEGIN RETURN 'foo'; END; $$ "
+      "LANGUAGE plpgsql;";
+  stmts = pgparser.BuildParseTree(query);
+
+  query =
+      "CREATE OR REPLACE FUNCTION return_bool ("
+      " i BOOLEAN"
```
Ghatage: Side question:
While taking a look again at these tests, I found that [create_function_statement.h](https://github.com/cmu-db/terrier/blob/master/src/include/parser/create_function_statement.h#L38) as two types for boolean. `Parameter::Datatype::BOOL` and `Parameter::Datatype::BOOLEAN`.
At the moment will go with `Parameter::Datatype::BOOLEAN` in the test and will update accordingly in the future.

Ghatage:
Is there any way to restart the build on Travis to get a clean sheet?
After I pushed the changes @apavlo suggested, one of the builds timed out due to what looks like an infrastructure issue.

apavlo:
@Ghatage I restarted the mac build.

@pervazea Are we seeing this issue with other PRs? Or should we just assume that it is a temporary Travis hiccup?

pervazea:
Some of the requested enhancements are still outstanding. @Ghatage, are you expecting to address those? (does not look difficult).

Ghatage:
Sorry guys, been swamped. Might take it further sometime later.
Thanks for merging the PR!

PR #303, opened 2019-02-21
Remove gtest dependency from terrier lib

mbutrovich:
Duplicated the macro definition in macros.h. @apavlo told us to do it.


src/include/common/macros.h
```diff
@@ -143,6 +143,12 @@
 //===----------------------------------------------------------------------===//
 #ifdef NDEBUG
 #define GTEST_DEBUG_ONLY(TestName) DISABLED_##TestName
+
+#define FRIEND_TEST(test_case_name, test_name)
```
lmwnshn: This kills the test (in anything but not debug mode).

PR #306, opened 2019-02-26
Bad words filter as "make check-censored" target

lmwnshn:
This uses a blacklist of regex in build-support/bad_words.txt.
By carefully picking regex, we don't need to // NOLINT innocuous uses too much, e.g. schema.h had a variable called inlined_ that we were picking up.

As a general rule of thumb, if the word is standalone, use the word boundary regex,
\bWORD\b   e.g. \binline\b
if the word is part of a function, include the calling paren in the regex,
func(            e.g. malloc(

Also added this check to CI and fixed up all existing files.

NB: I didn't think it was worth porting over the old validator script. That's a lot of Python, the include-what-you-use and namespace-closing functionality has been subsumed by check-lint, we basically just want to run an intelligent grep. And grep is pretty fast.

It is also much easier to understand how the make checks work compared to the old git-hook system.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,8 @@ before_script:
       docker exec build /bin/sh -c "cd /repo/build && make gbenchmark_ep -j 4" &&
       docker exec build /bin/sh -c "cd /repo/build && make check-clang-tidy" &&
       docker exec build /bin/sh -c "cd /repo/build && make check-format" &&
-      docker exec build /bin/sh -c "cd /repo/build && make check-lint";
+      docker exec build /bin/sh -c "cd /repo/build && make check-lint" &&
+      docker exec build /bin/sh -c "cd /repo/build && make check-censored";
     else
       cd apidoc &&
       touch warnings.txt &&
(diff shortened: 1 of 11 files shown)


lmwnshn:
@tli2 Suggestions on what to do with the printf's and std::cout's (see Travis)?

Options:
1. ignore the test folder and terrier.cpp
2. nolint them out manually, line by line
3. remove them from the list of censored files in the CMake target
4. have a text file of not-to-be-censored files (this might encourage people to add things here, CMake is a little less friendly..)

tli2:
It looks like we already have some errors and it's failing the build. Can we also fix those in this PR? Otherwise LGTM.

apavlo:
@lmwnshn We should remove those printfs. If we leave them in there, then new people are going to think that it's okay and repeat them.

lmwnshn:
@apavlo @tli2 The main offender is a function whose only purpose is to print the row [1]. Should this function be deleted, switched to use a logger, or maybe return a string?

[1] https://github.com/cmu-db/terrier/blob/master/test/include/util/storage_test_util.h#L187

tli2:
I don't think this function is ever used. It's mostly here for debugging purposes if ever we need to print a row. We can probably move it to using the log or returning a string. I can see someone needing this utility in the future.

PR #312, opened 2019-03-13
Port over Peloton plan nodes

wenxuanqiu:
Major changes in this PR:
- An OutputSchema object, specifying the output columns of a plan node and mapping from input columns to output columns, is a now a member variable of the AbstractPlanNode. This provides an unified interface to connect a plan node with its children
- Make plan nodes immutable after construction via Builders
- Break apart CreatePlan and DropPlan based on the object being created/dropped
- Change individual plans to use OIDs instead of strings where possible
- Remove plan nodes that we no longer plan to support
- Because plan nodes will be immutable, we change them to be contained using `shared_ptr` as opposed to `unique_ptr`
- Begin moving all planner code to `src/planner` (#355)


src/include/catalog/catalog_defs.h
```diff
@@ -3,6 +3,8 @@
 #include "common/strong_typedef.h"
 
 namespace terrier::catalog {
+STRONG_TYPEDEF(offset_oid_t, uint32_t);
```
tli2: Not sure you need an oid type for offsets. Postgres defines oid as essentially the primary key for `pg_class` I believe. So unless something can be looked up in `pg_class`, it probably shouldn't be an oid

src/include/parser/expression/abstract_expression.h
```diff
@@ -67,7 +67,7 @@ class AbstractExpression {
   /**
    * Logical inequality check.
    * @param rhs other
-   * @return true if the two expressions are not logically equal
+   * @return true if the two expressions are logically equal
```
tli2: unintentional change?

src/include/parser/expression/abstract_expression.h
```diff
@@ -102,6 +102,19 @@ class AbstractExpression {
     return children_[index];
   }
 
+  /**
+   * Exact equality check, including value equality
+   * @param other
+   * @return true if the two expressions are exactly equal
+   */
+  bool ExactlyEquals(const AbstractExpression &other) const {
```
tli2: It's unclear what `==` is vs. this `ExactlyEquals`. It probably is better to write more detailed comments explaining the semantics. Better still, maybe it is better to not override `==` and instead just write two methods. (e.g. `ShallowEquals` and `DeepEquals` with documentation explaining the difference)

src/include/parser/expression/abstract_expression.h
```diff
@@ -102,6 +102,19 @@ class AbstractExpression {
...
+    if (expression_type_ != other.expression_type_ || children_.size() != other.children_.size()) return false;
+    for (unsigned i = 0; i < children_.size(); i++) {
```
tli2: new code base uses `uint32_t`

src/include/plan_node/abstract_join_plan_node.h
```diff
@@ -0,0 +1,41 @@
...
+// Abstract Join Plan Node
+//===--------------------------------------------------------------------===//
+
+class AbstractJoinPlanNode : public AbstractPlanNode {
+ public:
+  AbstractJoinPlanNode(std::shared_ptr<OutputSchema> output_schema, LogicalJoinType join_type,
+                       parser::AbstractExpression *predicate)
+      : AbstractPlanNode(std::move(output_schema)), join_type_(join_type), predicate_(predicate) {}
+
+  common::hash_t Hash() const override;
```
tli2: can probably get rid of `hash_t` unless you can think of a reason we need it.

src/include/plan_node/abstract_join_plan_node.h
```diff
@@ -0,0 +1,41 @@
...
+  const parser::AbstractExpression *GetPredicate() const { return predicate_; }
+
+ private:
+  LogicalJoinType join_type_;
+
+  const parser::AbstractExpression *predicate_;
+
+  DISALLOW_COPY_AND_MOVE(AbstractJoinPlanNode);
+};
+
```
tli2: The general c++ approach is to write a functor to do hashing instead of a virtual method. It probably doesn't matter that much for plan nodes since they need a vtable anyway, but it might be beneficial to write a `std::hash` specialization so you can use these classes easily with STL. Find example in https://github.com/cmu-db/terrier/blob/master/src/include/common/strong_typedef.h#L333

src/include/plan_node/abstract_join_plan_node.h
```diff
@@ -0,0 +1,41 @@
+#pragma once
+
+#include <memory>
+#include <utility>
+#include "parser/expression/abstract_expression.h"
+#include "plan_node/abstract_plan_node.h"
+
+namespace terrier::plan_node {
```
tli2: I don't think naming convention permits `_` in namespaces?

src/include/plan_node/abstract_join_plan_node.h
```diff
@@ -0,0 +1,41 @@
...
+  LogicalJoinType GetLogicalJoinType() const { return join_type_; }
+
+  const parser::AbstractExpression *GetPredicate() const { return predicate_; }
+
+ private:
+  LogicalJoinType join_type_;
+
+  const parser::AbstractExpression *predicate_;
+
+  DISALLOW_COPY_AND_MOVE(AbstractJoinPlanNode);
```
tli2: make deleted constructors public

src/include/plan_node/abstract_plan_node.h
```diff
@@ -0,0 +1,141 @@
...
+    if (num != rhs.GetChildren().size()) return false;
+    for (unsigned int i = 0; i < num; i++) {
+      if (*GetChild(i) != *const_cast<AbstractPlanNode *>(rhs.GetChild(i))) return false;
+    }
+    return true;
+  }
+
+  virtual bool operator!=(const AbstractPlanNode &rhs) const { return !(*this == rhs); }
+
+  virtual std::unique_ptr<AbstractPlanNode> Copy() const = 0;
```
tli2: Do we use this method anywhere in the code base?

src/include/plan_node/abstract_plan_node.h
```diff
@@ -0,0 +1,141 @@
...
+ private:
+  // A plan node can have multiple children
+  std::vector<std::unique_ptr<AbstractPlanNode>> children_;
+
+  int estimated_cardinality_ = 500000;
+
+  std::shared_ptr<OutputSchema> output_schema_;
+
+ private:
+  DISALLOW_COPY_AND_MOVE(AbstractPlanNode);
```
tli2: make public

src/include/plan_node/abstract_plan_node.h
```diff
@@ -0,0 +1,141 @@
...
+
+class Equal {
+ public:
+  bool operator()(const std::shared_ptr<plan_node::AbstractPlanNode> &a,
+                  const std::shared_ptr<plan_node::AbstractPlanNode> &b) const {
+    return *a == *b;
+  }
+};
+
+class Hash {
```
tli2: can just specialize the template for `std::hash` and `std::equal`

src/include/plan_node/abstract_plan_node.h
```diff
@@ -0,0 +1,141 @@
...
+  //===--------------------------------------------------------------------===//
+  // Accessors
+  //===--------------------------------------------------------------------===//
+
+  // Each sub-class will have to implement this function to return their type
+  // This is better than having to store redundant types in all the objects
+  virtual PlanNodeType GetPlanNodeType() const { return PlanNodeType::ABSTRACTPLAN; }
+
+  // Get a string representation for debugging
+  virtual const std::string GetInfo() const;
```
tli2: This probably goes away as you implement the json version of this

src/include/plan_node/abstract_plan_node.h
```diff
@@ -0,0 +1,141 @@
...
+    return children_[child_index].get();
+  }
+
+  //===--------------------------------------------------------------------===//
+  // Accessors
+  //===--------------------------------------------------------------------===//
+
+  // Each sub-class will have to implement this function to return their type
+  // This is better than having to store redundant types in all the objects
+  virtual PlanNodeType GetPlanNodeType() const { return PlanNodeType::ABSTRACTPLAN; }
```
tli2: You probably want this to be a pure virtual method since I don't think anything will have most specific type of abstract plan. There are pure virtual methods in this class already.

src/include/plan_node/abstract_plan_node.h
```diff
@@ -0,0 +1,141 @@
...
+
+  // Get a string representation for debugging
+  virtual const std::string GetInfo() const;
+
+  // Get the output schema for the plan node. The output schema contains information on columns of the output of
+  // the plan node operator
+  std::shared_ptr<OutputSchema> GetOutputSchema() const { return output_schema_; }
+
+  // Get the estimated cardinality of this plan
+  int GetEstimatedCardinality() const { return estimated_cardinality_; }
```
tli2: use stdint ints. (i.e. `int32_t`)

Also, why is this signed anyway?

src/include/plan_node/abstract_plan_node.h
```diff
@@ -0,0 +1,141 @@
...
+
+  // FOR TESTING ONLY. This function should only be called during construction of plan (ConvertOpExpression) or
+  // for tests.
+  void SetEstimatedCardinality(int cardinality) { estimated_cardinality_ = cardinality; }
```
tli2: It is a good idea to reflect it in the method name if it's test only. e.g. `SetEstimatedCardinalityForUnitTest`

We can talk about general best practices for this kind of things in the group meeting.

src/include/plan_node/abstract_plan_node.h
```diff
@@ -0,0 +1,141 @@
...
+
+  virtual bool operator!=(const AbstractPlanNode &rhs) const { return !(*this == rhs); }
+
+  virtual std::unique_ptr<AbstractPlanNode> Copy() const = 0;
+
+ private:
+  // A plan node can have multiple children
+  std::vector<std::unique_ptr<AbstractPlanNode>> children_;
+
+  int estimated_cardinality_ = 500000;
```
tli2: do we really want this default value here?

src/include/plan_node/abstract_plan_node.h
```diff
@@ -0,0 +1,141 @@
...
+  // For Deserialization and DDL statements
+  AbstractPlanNode() = default;
+
+  virtual ~AbstractPlanNode() = default;
+
+  //===--------------------------------------------------------------------===//
+  // Children Helpers
+  //===--------------------------------------------------------------------===//
+
+  void AddChild(std::unique_ptr<AbstractPlanNode> &&child) { children_.emplace_back(std::move(child)); }
```
tli2: Didn't we say this class is going to be immutable?

src/include/plan_node/abstract_scan_plan_node.h
```diff
@@ -0,0 +1,38 @@
...
+#include <memory>
+#include <utility>
+#include "parser/expression/abstract_expression.h"
+#include "plan_node/abstract_plan_node.h"
+
+namespace terrier::plan_node {
+class AbstractScanPlanNode : public AbstractPlanNode {
+ public:
+  AbstractScanPlanNode(std::shared_ptr<OutputSchema> output_schema, parser::AbstractExpression *predicate,
+                       bool is_for_update = false, bool parallel = false)
```
tli2: probably don't want default values here.

src/include/plan_node/abstract_scan_plan_node.h
```diff
@@ -0,0 +1,38 @@
...
+  std::unique_ptr<parser::AbstractExpression> predicate_;
+
+  // Are the tuples produced by this plan intended for update?
+  bool is_for_update_ = false;
+
+  // Should this scan be performed in parallel?
+  bool parallel_;
+
+ private:
+  DISALLOW_COPY_AND_MOVE(AbstractScanPlanNode);
```
tli2: make public

src/include/plan_node/abstract_plan_node.h
GustavoAngulo: This was a hack in original peloton, I removed it once and it caused the bloomfilter to crash. It should probably be safe to remove in terrier and given that we're going to have a new execution engine.

src/include/plan_node/abstract_plan_node.h
GustavoAngulo: `AddChild` is only used when constructing the plan node. Do you think it would be better to pass in the child plans into the constructor? That should make them immutable.

src/include/parser/expression/abstract_expression.h
wenxuanqiu: This method is not needed for plan nodes. It was required for parts of the optimizer. I think I am going to take it out for now and investigate whether we need it later.

src/include/plan_node/abstract_plan_node.h
GustavoAngulo: Marking as resolved, we're changing the plan nodes to use a builder pattern.

src/include/plan_node/abstract_scan_plan_node.h
GustavoAngulo: Moved the default values to the member variables on the builder template for AbstractScanPlanNode

src/include/plan_node/abstract_plan_node.h
GustavoAngulo: Changing it to unit32_t. I believe -1 was being used in the optimizer as a "i have 0 clue what the cardinality is", but we should default that to `max(0, cardinality)` when we construct the plan.

src/include/plan_node/abstract_plan_node.h
GustavoAngulo: I removed the method as we're using a builder pattern now.

src/include/plan_node/abstract_join_plan_node.h
GustavoAngulo: I figured we were using it as a clarity tool to make it more readable in the code.

src/include/plan_node/abstract_plan_node.h
GustavoAngulo: Done, let me know if thats what you meant. I haven't specialized in c++ before.

src/include/plan_node/abstract_join_plan_node.h
apavlo: Correct.

src/include/plan_node/abstract_join_plan_node.h
```diff
@@ -0,0 +1,99 @@
...
+
+   protected:
+    /**
+     * Logical join type
+     */
+    LogicalJoinType join_type_;
+    /**
+     * Join predicate
+     */
+    std::unique_ptr<const parser::AbstractExpression> predicate_;
```
apavlo: Rename variable to `join_predicate_` to make it more clear what this is.

src/include/plan_node/abstract_join_plan_node.h
```diff
@@ -0,0 +1,99 @@
...
+   * Base constructor for joins. Derived join plans should call this constructor
+   * @param children child plan nodes
+   * @param output_schema Schema representing the structure of the output of this plan node
+   * @param estimated_cardinality estimated cardinality of output of node
+   * @param join_type logical join type
+   * @param predicate join predicate
+   */
+  AbstractJoinPlanNode(std::vector<std::unique_ptr<AbstractPlanNode>> &&children,
+                       std::shared_ptr<OutputSchema> output_schema, uint32_t estimated_cardinality,
+                       LogicalJoinType join_type, std::unique_ptr<const parser::AbstractExpression> &&predicate)
```
apavlo: Do we want the double reference?

`std::unique_ptr<const parser::AbstractExpression> &&predicate`

src/include/plan_node/abstract_join_plan_node.h
apavlo: Agreed. I think it would be important to start doing this more.

src/include/plan_node/abstract_plan_node.h
```diff
@@ -0,0 +1,246 @@
...
+     */
+    std::vector<std::unique_ptr<AbstractPlanNode>> children_;
+    /**
+     * schema describing output of the node
+     */
+    std::shared_ptr<OutputSchema> output_schema_;
+    /**
+     * estimated cardinality of output for node
+     */
+    uint32_t estimated_cardinality_ = 0;
```
apavlo: I don't think we want this in here since it will change over time as the database changes. It should be stored in a context object.

src/include/plan_node/aggregate_plan_node.h
```diff
@@ -0,0 +1,203 @@
...
+  /**
+   * A global aggregate does not aggregate over specific columns, example: count(*)
+   * @return true if aggregation is global
+   */
+  bool IsGlobal() const { return false; }
+
+  /**
+   * @return pointer to predicate for having clause
+   */
+  const parser::AbstractExpression *GetHavingClausePredicate() const { return having_clause_predicate_.get(); }
```
apavlo: What's the point of having a unique pointer if we're just handing out the raw pointer?

src/include/plan_node/aggregate_plan_node.h
```diff
@@ -0,0 +1,203 @@
...
+   */
+  common::hash_t Hash() const override;
+
+  bool operator==(const AbstractPlanNode &rhs) const override;
+
+ private:
+  /**
+   * @return true of two vectors of aggregate terms are equal
+   */
+  bool AreEqual(const std::vector<AggregatePlanNode::AggregateTerm> &A,
```
apavlo: I feel like this should be in a static util class?

src/include/plan_node/analyze_plan_node.h
```diff
@@ -0,0 +1,140 @@
...
+
+  /**
+   * @return the hashed value of this plan node
+   */
+  common::hash_t Hash() const override;
+
+  bool operator==(const AbstractPlanNode &rhs) const override;
+
+ private:
+  catalog::table_oid_t target_table_oid_;  // OID of the target table
```
apavlo: Inconsistent comment convention.

src/include/plan_node/create_function_plan_node.h
```diff
@@ -0,0 +1,257 @@
...
+
+  /**
+   * @return the hashed value of this plan node
+   */
+  common::hash_t Hash() const override;
+
+  bool operator==(const AbstractPlanNode &rhs) const override;
+
+ private:
+  // Indicates the UDF language type
```
apavlo: Inconsistent comment convention.

src/include/plan_node/csv_scan_plan_node.h
```diff
@@ -0,0 +1,181 @@
...
+#include <string>
+#include <utility>
+#include <vector>
+#include "plan_node/abstract_scan_plan_node.h"
+
+namespace terrier::plan_node {
+/**
+ * Plan node for a CSV scan
+ */
+class CSVScanPlanNode : public AbstractScanPlanNode {
```
apavlo: @pmenon Should we rename this `CopyPlanNode`?

src/include/plan_node/delete_plan_node.h
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * @param delete_stmt the SQL DELETE statement
+     * @return builder object
+     */
+    Builder &SetFromDeleteStatement(parser::DeleteStatement *delete_stmt) {
+      table_name_ = delete_stmt->GetDeletionTable()->GetTableName();
+      delete_condition_ = delete_stmt->GetDeleteCondition();
+      // TODO(Gus,Wen) get table OID from catalog
```
apavlo: Why? The binder should do this for you.

src/include/plan_node/export_external_file_plan_node.h
```diff
@@ -0,0 +1,157 @@
...
+
+  /**
+   * @param children child plan nodes
+   * @param file_name string representation of file name
+   * @param delimiter delimiter character
+   * @param quote quote character
+   * @param escape escape character
+   */
+  explicit ExportExternalFilePlanNode(std::vector<std::unique_ptr<AbstractPlanNode>> &&children, std::string file_name,
+                                      char delimiter = ',', char quote = '"', char escape = '\"')
```
apavlo: I say we get rid of default argument values.

@tli2 What does the design style say?

src/include/plan_node/hybrid_scan_plan_node.h
```diff
@@ -0,0 +1,117 @@
...
+#include "catalog/schema.h"
+#include "plan_node/abstract_scan_plan_node.h"
+#include "plan_node/plan_node_defs.h"
+
+namespace terrier::plan_node {
+
+/**
+ * Plan node for a hybryd scan
+ */
+class HybridScanPlanNode : public AbstractScanPlanNode {
```
apavlo: We should drop this class. We're not going to support this in the new system.

src/include/plan_node/insert_plan_node.h
```diff
@@ -0,0 +1,189 @@
...
+#include "parser/insert_statement.h"
+#include "plan_node/abstract_plan_node.h"
+#include "plan_node/abstract_scan_plan_node.h"
+#include "type/transient_value.h"
+#include "type/transient_value_peeker.h"
+
+namespace terrier {
+
+namespace storage {
+class SqlTable;
```
apavlo: Why is this needed?

src/include/plan_node/limit_plan_node.h
```diff
@@ -0,0 +1,118 @@
+#pragma once
+
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+#include "catalog/schema.h"
+#include "plan_node/abstract_plan_node.h"
+
+// TODO(Gus,Wen): I don't think limit really needs an output schema. I'll include it for now, but we can maybe toss it
```
apavlo: I thought we decided that every PlanNode should have one?

src/include/plan_node/materialization_plan_node.h
```diff
@@ -0,0 +1,90 @@
...
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+#include "plan_node/abstract_plan_node.h"
+#include "plan_node/output_schema.h"
+
+namespace terrier::plan_node {
+
+class MaterializationPlanNode : public AbstractPlanNode {
```
apavlo: I think we might be able to remove this one too.

src/include/plan_node/materialization_plan_node.h
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * @return the hashed value of this plan node
+   */
+  common::hash_t Hash() const override;
+
+  bool operator==(const AbstractPlanNode &rhs) const override;
+
+ private:
+  /**
+   * Whether to create a physical tile or just pass through underlying
```
apavlo: Yes, we don't have tiles anymore so I think we might want to remove this.

src/include/plan_node/output_schema.h
```diff
@@ -0,0 +1,228 @@
...
+    const std::string name_;
+    const type::TypeId type_;
+    const bool nullable_;
+    const catalog::col_oid_t oid_;
+  };
+
+  /**
+   * An intermediate column produced by plan nodes
+   */
+  struct DerivedColumn {
```
apavlo: Why is `OutputSchema::Column` a class but `OutputSchema::DerivedColumn` is a struct? I think we should be consistent.

src/include/plan_node/output_schema.h
```diff
@@ -0,0 +1,228 @@
...
+  bool operator!=(const OutputSchema &rhs) const { return !operator==(rhs); }
+
+  /**
+   * Make a copy of this OutputSchema
+   * @return shared pointer to the copy
+   */
+  std::shared_ptr<OutputSchema> Copy() const { return std::make_shared<OutputSchema>(*this); }
+
+ private:
+  /**
```
apavlo: We should put this at the top of the class instead of hidden down here.

src/include/plan_node/plan_node_defs.h
```diff
@@ -0,0 +1,144 @@
...
+  RESULT,
+
+  // Test
+  MOCK
+};
+
+//===--------------------------------------------------------------------===//
+// Aggregate Stragegies
+//===--------------------------------------------------------------------===//
+enum class AggregateStrategy {
```
apavlo: `AggregateStrategy` -> `AggregateStrategyType`

src/include/plan_node/plan_node_defs.h
```diff
@@ -0,0 +1,144 @@
...
+//===--------------------------------------------------------------------===//
+// Hybrid Scan Types
+//===--------------------------------------------------------------------===//
+enum class HybridScanType { INVALID = INVALID_TYPE_ID, SEQUENTIAL = 1, INDEX = 2, HYBRID = 3 };
+
+//===--------------------------------------------------------------------===//
+// Order by Orderings
+//===--------------------------------------------------------------------===//
+
+enum class OrderByOrdering { ASC, DESC };
```
apavlo: Rename to `OrderByOrderingType`

src/include/plan_node/plan_node_defs.h
```diff
@@ -0,0 +1,144 @@
...
+// Set Operation Types
+//===--------------------------------------------------------------------===//
+
+enum class SetOpType { INVALID = INVALID_TYPE_ID, INTERSECT = 1, INTERSECT_ALL = 2, EXCEPT = 3, EXCEPT_ALL = 4 };
+
+// TODO(Gus,Wen) Tuple as a concept does not exist yet, someone need to define it in the storage layer, possibly a
+/**
+ * i hate u doxygen
+ */
+class Tuple {
```
apavlo: Why do you need this here?

src/include/plan_node/populate_index_plan_node.h
```diff
@@ -0,0 +1,135 @@
...
+
+namespace terrier {
+namespace plan_node {
+
+/**
+ * Plan node for populating an index
+ * IMPORTANT All tiles got from child must have the same physical schema.
+ */
+
+class PopulateIndexPlanNode : public AbstractPlanNode {
```
apavlo: I think we can remove this class for now.

src/include/plan_node/projection_plan_node.h
```diff
@@ -0,0 +1,65 @@
...
+#include <utility>
+#include <vector>
+#include "plan_node/abstract_plan_node.h"
+
+namespace terrier::plan_node {
+
+/**
+ * Plan node for projection
+ */
+class ProjectionPlanNode : public AbstractPlanNode {
```
apavlo: I am surprised that `ProjectionPlanNode` does not have any member objects? We should find out what the old code actually did with this plan node.

src/include/plan_node/result_plan_node.h
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * @return the hashed value of this plan node
+   */
+  common::hash_t Hash() const override;
+
+  bool operator==(const AbstractPlanNode &rhs) const override;
+
+ private:
+  // the tuple in the storage layer
+  std::shared_ptr<Tuple> tuple_;
```
apavlo: I don't think we should have a `Tuple` here. I should probably be an expression. For example, `SELECT NOW() + 1` should not be materialized as a tuple since it will change each time it is invoked

src/plan_node/aggregate_plan_node.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+  hash = common::HashUtil::CombineHashes(hash, HashAggregateTerms(GetAggregateTerms()));
+
+  auto agg_strategy = GetAggregateStrategy();
+  hash = common::HashUtil::CombineHashes(hash, common::HashUtil::Hash(&agg_strategy));
+
+  return common::HashUtil::CombineHashes(hash, AbstractPlanNode::Hash());
+}
+
+bool AggregatePlanNode::AreEqual(const std::vector<AggregatePlanNode::AggregateTerm> &A,
```
apavlo: Yes, this should be in a helper class.

src/include/plan_node/abstract_join_plan_node.h
pervazea: Where is the prohibition against _ in namespaces stated? Because I'm not seeing it.

src/include/plan_node/export_external_file_plan_node.h
GustavoAngulo: I forgot to remove these. For default arguments I've been making then default values to the member variables of the respective builder class.

src/include/plan_node/projection_plan_node.h
GustavoAngulo: It used to have [member variables](https://github.com/cmu-db/peloton/blob/484d76df9344cb5c153a2c361c5d5018912d4cf4/src/include/planner/projection_plan.h#L87-L94), but all of these are replaced by `OutputSchema`

src/include/plan_node/aggregate_plan_node.h
GustavoAngulo: Unsure, seems like a disease from the old code. Changed it to return the unique pointer

src/include/plan_node/limit_plan_node.h
GustavoAngulo: Limit will never change the projection from its child plan, but for consistency sake we can include the output schema. I've deleted the comment

pervazea:
How are you folks exercising the plan nodes? The directory test/plan_node doesn't contain much, which is where I've looked so far.

GustavoAngulo:
> How are you folks exercising the plan nodes? The directory test/plan_node doesn't contain much, which is where I've looked so far.

The old test cases in peloton required the optimizer, parser, and other systems. So we were not able to port them over. The most I could think of right now is write test cases that test construction and equality of plan nodes. Is that worth doing?

src/include/plan_node/aggregate_plan_node.h
GustavoAngulo: I decided to implement `operator==` for `AggregateTerm` instead. The method was only being used inside the plan node anyway.

src/include/plan_node/csv_scan_plan_node.h
pmenon: This node normally has an `Insert` parent node to perform an import. While we can probably collapse it into a single `CopyPlanNode`, I think it's better to leave it as is since it's more descriptive, we can later support scanning files of different formats, and also scan files without the intention of importing the data into the database.

src/include/plan_node/create_plan_node.h
```diff
@@ -0,0 +1,970 @@
...
+          create_type_ = CreateType::SCHEMA;
+          database_name_ = std::string(create_stmt->GetDatabaseName());
+          schema_name_ = std::string(create_stmt->GetSchemaName());
+          break;
+        }
+
+        case parser::CreateStatement::CreateType::kTable: {
+          table_name_ = std::string(create_stmt->GetTableName());
+          schema_name_ = std::string(create_stmt->GetSchemaName());
+          database_name_ = std::string(create_stmt->GetDatabaseName());
```
pervazea: 1. General comment on create plan node. Big. Amazingly big. Was consideration given to breaking it up into create sub-types, and if so, what prevented that approach?

2. GetDatabaseName(). This is inconsistent across plan node types.

Somewhere associated with each connection, we will have a database for that connection. We can store the database name, but in addition may wish to store the oid, since the latter is more directly useful. Therefore, in plan nodes, choices:

1) Don't store database at all, retrieve from connection.
2) Store database (consistently) across plan node types. Name possibly, oid preferred, or both. (Except for create database, where it has to be name).

src/include/plan_node/delete_plan_node.h
```diff
@@ -0,0 +1,148 @@
...
+ private:
+  /**
+   * the table to be deleted
+   */
+  catalog::table_oid_t target_table_oid_;
+  /**
+   * name of the table
+   */
+  std::string table_name_;
+  /**
```
pervazea: On naming:
- target_table_oid
- table_name

Which more consistently would be target_table_oid_, target_table_name_ or more concisely table_oid_, table_name_. There is only one table in question, so the target prefix doesn't provide any extra clarity.

GustavoAngulo:
@pervazea An important note regarding testing of the plan node construction. Each test case for serializing plan nodes will involve actually constructing the plan nodes, so the JSON serialization tests will actually test every single plan node constructor.

lmwnshn:
I haven't been able to look fully at this yet, sorry. We're using them in the class project and sort of trying them out that way. Did we want to change from parser/expression/expressions... to just expression/expressions...?

GustavoAngulo:
> I haven't been able to look fully at this yet, sorry. We're using them in the class project and sort of trying them out that way. Did we want to change from parser/expression/expressions... to just expression/expressions...?

I am going to change them in a future PR per #355

src/include/planner/plannodes/aggregate_plan_node.h
```diff
@@ -0,0 +1,213 @@
...
+
+namespace terrier::planner {
+
+/**
+ * Plan node for aggregates
+ */
+class AggregatePlanNode : public AbstractPlanNode {
+ public:
+  /**
+   * Information for each term being aggregated on
```
tli2: An example would help. It's not clear what this is still.

src/include/planner/plannodes/abstract_plan_node.h
```diff
@@ -0,0 +1,225 @@
...
+     */
+    std::shared_ptr<OutputSchema> output_schema_;
+  };
+
+  /**
+   * Constructor for the base AbstractPlanNode. Derived plan nodes should call this constructor to set output_schema
+   * @param children child plan nodes
+   * @param output_schema Schema representing the structure of the output of this plan node
+   */
+  explicit AbstractPlanNode(std::vector<std::shared_ptr<AbstractPlanNode>> &&children,
```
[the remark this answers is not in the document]
GustavoAngulo: Done.

src/include/planner/plannodes/abstract_plan_node.h
```diff
@@ -0,0 +1,225 @@
...
+
+ private:
+  std::vector<std::shared_ptr<AbstractPlanNode>> children_;
+  std::shared_ptr<OutputSchema> output_schema_;
+
+ public:
+  /**
+   * Don't allow plan to be copied or moved
+   */
+  DISALLOW_COPY_AND_MOVE(AbstractPlanNode);
```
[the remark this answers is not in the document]
GustavoAngulo: Sure thing

src/include/planner/plannodes/aggregate_plan_node.h
GustavoAngulo: Added an example. Let me know if this is clear

pervazea:
The hierarchy is _database.namespace.table_  In the Postgres documentation, namespaces are also referred to as schema. (This is confusing of course. This is distinct from the term schema describing the meta data of a table).

So, any place where one refers to a table, it should be specified as a pair of items: namespace, table. Where namespace is not explicitly specified, it defaults to the "public" namespace. The public namespace is created for you as the default namespace.

GustavoAngulo:
I believe we did not want to rely on the connection context, so I will add database, namespace, and table oids to the relevant plan nodes

src/include/planner/plannodes/csv_scan_plan_node.h
```diff
@@ -0,0 +1,181 @@
...
+
+   protected:
+    /**
+     * string representation of file name
+     */
+    std::string file_name_;
+    /**
+     * delimiter character for CSV
+     */
+    char delimiter_ = ',';
```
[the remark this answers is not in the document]
GustavoAngulo: Done.

src/include/planner/plannodes/export_external_file_plan_node.h
```diff
@@ -0,0 +1,156 @@
...
+
+   protected:
+    /**
+     * string representation of file name
+     */
+    std::string file_name_;
+    /**
+     * delimiter character
+     */
+    char delimiter_ = ',';
```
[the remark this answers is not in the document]
GustavoAngulo: Done.

src/include/planner/plannodes/insert_plan_node.h
```diff
@@ -0,0 +1,201 @@
...
+
+  /**
+   * values to insert
+   */
+  std::vector<type::TransientValue> values_;
+
+  /**
+   * parameter information <tuple_index,  tuple_column_index, value_index>
+   */
+  std::vector<std::tuple<uint32_t, uint32_t, uint32_t>> parameter_info_;
```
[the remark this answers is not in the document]
GustavoAngulo: Done, added one.

src/include/planner/plannodes/seq_scan_plan_node.h
```diff
@@ -0,0 +1,103 @@
...
+      return std::shared_ptr<SeqScanPlanNode>(new SeqScanPlanNode(std::move(children_), std::move(output_schema_),
+                                                                  std::move(scan_predicate_), is_for_update_,
+                                                                  is_parallel_, database_oid_, table_oid_));
+    }
+
+   protected:
+    /**
+     * OID for table being scanned
+     */
+    catalog::table_oid_t table_oid_;
```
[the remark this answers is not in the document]
GustavoAngulo: The parent class `AbstractScanPlanNode` already had `database_oid` information. I added the `namespace_oid` to `AbstractScanPlanNode` as well.

src/include/planner/plannodes/create_index_plan_node.h
```diff
@@ -0,0 +1,267 @@
...
+      return std::shared_ptr<CreateIndexPlanNode>(new CreateIndexPlanNode(
+          std::move(children_), std::move(output_schema_), database_oid_, table_oid_, index_type_, unique_index_,
+          std::move(index_name_), std::move(index_attrs_), std::move(key_attrs_)));
+    }
+
+   protected:
+    /**
+     * OID of the database
+     */
+    catalog::db_oid_t database_oid_;
```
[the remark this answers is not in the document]
GustavoAngulo: Added namespace oid to all necessary DDL and Scan plan nodes.

src/include/planner/plannodes/create_index_plan_node.h
```diff
@@ -0,0 +1,267 @@
...
+   * @param index_attrs index attributes
+   * @param key_attrs key attributes
+   */
+  CreateIndexPlanNode(std::vector<std::shared_ptr<AbstractPlanNode>> &&children,
+                      std::shared_ptr<OutputSchema> output_schema, catalog::db_oid_t database_oid,
+                      catalog::table_oid_t table_oid, parser::IndexType index_type, bool unique_index,
+                      std::string index_name, std::vector<std::string> &&index_attrs,
+                      std::vector<std::string> &&key_attrs)
+      : AbstractPlanNode(std::move(children), std::move(output_schema)),
+        database_oid_(database_oid),
```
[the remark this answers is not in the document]
GustavoAngulo: Added namespace oid to all necessary DDL and Scan plan nodes.

src/include/planner/plannodes/create_trigger_plan_node.h
```diff
@@ -0,0 +1,307 @@
...
+#include "parser/expression/abstract_expression.h"
+#include "parser/expression/constant_value_expression.h"
+#include "planner/plannodes/abstract_plan_node.h"
+
+namespace terrier::planner {
+
+/**
+ * Plan node for creating triggers
+ */
+class CreateTriggerPlanNode : public AbstractPlanNode {
```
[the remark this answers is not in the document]
GustavoAngulo: Added namespace oid to all necessary DDL and Scan plan nodes.

PR #314, opened 2019-03-16
Index keys and builder

### LINKED ISSUE #147: Bringing in Index
We want to start bringing in the bw_tree as our index. This will entail:
- [x] adding bwtree.h/.cpp to the third_party folder
- [ ] modify the wrapper bwtree_index.h to not inherit from index.h anymore, and any other adjustments we need to make it work with Terrier
- [ ] discuss any other dependencies you find before bringing them in

lmwnshn:
Fix #147.

CompactIntsKey; GenericKey; BwTree index wrapper which supports inserts, conditional inserts, deletes, point queries.

**Changes from Peloton**
- Our ProjectedRow reshuffles attributes. We precompute some math in the index metadata to figure out what goes where.
- Our value system is less overpowered. This complicated the keys.
- Much more key test coverage relative to Peloton. We generate random key schemas, populate them, tweak an attribute with some probability, and make sure the == and < make sense. We do this for both CompactIntsKey and GenericKey.
- We hammer the keys in tests much more than the index builder/wrapper itself. The BwTree is tested separately, so correctness of keys + correctness of underlying index should be good?

**Ready for review**


src/include/catalog/catalog_defs.h
```diff
@@ -5,5 +5,8 @@
 namespace terrier::catalog {
 STRONG_TYPEDEF(col_oid_t, uint32_t);
 STRONG_TYPEDEF(db_oid_t, uint32_t);
+STRONG_TYPEDEF(index_oid_t, uint32_t);
+STRONG_TYPEDEF(indexkeycol_oid_t, uint32_t);
```
mbutrovich: I have no idea what to call this. Should probably rename it before merging.

src/main/terrier.cpp
```diff
@@ -13,6 +13,7 @@
 #include "loggers/transaction_logger.h"
 #include "loggers/type_logger.h"
 #include "storage/data_table.h"
+#include "storage/index/index_builder.h"
```
mbutrovich: Should take this out.

src/include/catalog/catalog_defs.h
lmwnshn: I too have no idea what to call this..

src/main/terrier.cpp
lmwnshn: Taken out.

src/include/catalog/catalog_defs.h
```diff
@@ -5,5 +5,8 @@
...
+STRONG_TYPEDEF(sqltable_oid_t, uint32_t);
```
lmwnshn: @mbutrovich Are we using this? Code search not working.

src/include/storage/index/index.h
```diff
@@ -0,0 +1,108 @@
...
+#include "catalog/catalog_defs.h"
+#include "common/performance_counter.h"
+#include "storage/index/compact_ints_key.h"
+#include "storage/index/generic_key.h"
+#include "storage/index/index_defs.h"
+#include "storage/index/index_metadata.h"
+#include "storage/storage_defs.h"
+
+// clang-format off
+#define IndexCounterMembers(f) \
```
lmwnshn: We're not really using performance counters right now. Are we leaving them here as an example or should we get rid of them?

src/include/catalog/catalog_defs.h
mbutrovich: It's not really in the scope of this PR. I'll see if it makes sense to revert some.

src/include/storage/index/bwtree_index.h
```diff
@@ -0,0 +1,79 @@
...
+    if (!predicate_satisfied) {
+      TERRIER_ASSERT(ret, "Insertion should always succeed. (Ziqi)");
+    } else {
+      TERRIER_ASSERT(!ret, "Insertion should always fail. (Ziqi)");
+    }
+
+    return ret;
+  }
+
+  void ScanKey(const ProjectedRow &key, std::vector<TupleSlot> *value_list) final {
```
parsec0re: This method name is a bit weird. This is should be a point query right?

parsec0re: The same in `index.h`.

lmwnshn: It used to be ScanKey and ScanAllKey, among others not ported so far. I think we kept the ScanXYZ naming convention. What do you propose calling it instead?

parsec0re: I think it's fine!

src/storage/projected_row.cpp
```diff
@@ -64,4 +60,44 @@ ProjectedRow *ProjectedRowInitializer::InitializeRow(void *const head) const {
...
+  for (auto const &col_id : col_ids) {
+    attr_sizes.emplace_back(layout.AttrSize(col_id));
+  }
+  return ProjectedRowInitializer(attr_sizes, std::move(col_ids));
+}
+
+template <typename AttrType>
+ProjectedRowInitializer ProjectedRowInitializer::CreateProjectedRowInitializerForIndexes(
+    std::vector<AttrType> attr_sizes, const std::vector<uint16_t> &pr_offsets) {
+  // uint8 requires masking off the MSB varlen flag, uint16 are actual sizes
```
amlatyrngom: Why is the MSB flipped for uint8?

test/storage/bwtree_index_test.cpp
```diff
@@ -0,0 +1,1204 @@
...
+  EXPECT_EQ(pr_offsets[2], 2);
+  EXPECT_EQ(pr_offsets[3], 4);
+  EXPECT_EQ(pr_offsets[4], 0);
+}
+
+/**
+ * 1. Generate a reference key schema.
+ * 2. Fill two keys (pr_A, pr_B) with random data (data_A, data_B)
+ * 3. Set CompactIntsKeys (key_A, key_B) from (pr_A, pr_B)
+ * 4. e
```
amlatyrngom: Incomplete comment?

src/include/catalog/catalog_defs.h
parsec0re: Should we name it `indkey_oid_t` as in pg_index? https://www.postgresql.org/docs/current/catalog-pg-index.html

src/include/storage/index/index_defs.h
```diff
@@ -0,0 +1,108 @@
...
+ */
+class IndexKeyColumn {
+ public:
+  /**
+   * Non-varlen constructor for index key columns.
+   * @param oid key column oid
+   * @param nullable whether the column is nullable
+   * @param type_id the non-varlen type of the column
+   */
+  IndexKeyColumn(catalog::indexkeycol_oid_t oid, type::TypeId type_id, bool nullable) : oid_(oid), packed_type_(0) {
```
yangjuns: This class seems a lot like Column in catalog::schema except that it uses indexkeycol_oid_t. 
I do not see how using col_oid_t would break things. Could you explain do we need indexkeycol_oid_t? I feel like the values should be the same.

src/include/storage/index/index_defs.h
```diff
@@ -0,0 +1,108 @@
...
+/**
+ * The type of index.
+ */
+enum class ConstraintType : uint8_t {
+  // invalid index constraint type
+  INVALID = 0,
+  // default type - not used to enforce constraints
+  DEFAULT = 1,
+  // used to enforce primary key constraint
+  PRIMARY_KEY = 2,
```
parsec0re: I feel like PK constraint is a table-level constraint. Why are we having both PK and UNIQUE for an index? Point out if I'm wrong

src/include/storage/index/compact_ints_key.h
```diff
@@ -0,0 +1,331 @@
...
+
+#include "common/hash_util.h"
+#include "portable_endian/portable_endian.h"
+#include "storage/index/index_metadata.h"
+#include "storage/projected_row.h"
+#include "storage/storage_defs.h"
+
+namespace terrier::storage::index {
+
+// This is the maximum number of 8-byte slots to pack into a single CompactIntsKey template. This contraint is not
```
parsec0re: Typo here "contraint"

test/storage/bwtree_index_test.cpp
```diff
@@ -0,0 +1,1204 @@
...
+  }
+
+  /**
+   * Tests:
+   * 1. Scan -> Insert -> Scan -> Delete -> Scan
+   * 2. If primary key or unique index, 2 x Conditional Insert -> Scan -> Delete -> Scan
+   */
+  void BasicOps(Index *const index) {
+    // instantiate projected row and key
+    const auto &metadata = index->metadata_;
```
yangjuns: This requires this test class to be a friend class of Index. Maybe we can pass in metadata object since the test controls the creation of metadata. This way, we can separate the friend class, which is only for testing, from the Index class.

src/include/storage/index/compact_ints_key.h
```diff
@@ -0,0 +1,331 @@
...
+ * bytes. This gives the compiler opportunities to optimize comparison
+ * using advanced techniques such as SIMD or loop unrolling.
+ *
+ * For details of how and why integers must be stored in a big-endian and
+ * sign-magnitude format, please refer to adaptive radix tree's key format
+ *
+ * Note: CompactIntsKey size must always be aligned to 64 bit boundaries; There
+ * are static assertion to enforce this rule
+ *
+ * @tparam KeySize number of 8-byte fields to use. Valid range is 1 through 4.
```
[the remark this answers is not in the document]
mbutrovich: See line 15.

src/include/storage/index/generic_key.h
```diff
@@ -0,0 +1,356 @@
...
+    }
+
+    /**
+     * @param type_id TypeId to interpret both pointers as
+     * @param lhs_attr first value to be compared
+     * @param rhs_attr second value to be compared
+     * @return true if first is equal to second
+     */
+    static bool CompareEquals(const type::TypeId type_id, const byte *const lhs_attr, const byte *const rhs_attr) {
+      COMPARE_FUNC(==)
```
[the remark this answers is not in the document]
mbutrovich: Is it failing on yours?

src/include/storage/index/index.h
lmwnshn: Removed.

src/storage/projected_row.cpp
lmwnshn: Refactored into an additional method GetRealAttrSizes() in IndexMetadata, which we call before invoking CreateProjectedRowInitializerForIndexes(). The MSB was flipped because the uint8_t's use the MSB to indicate whether a column is VARLEN or not. Now, the initializer always expects the real attribute sizes.

test/storage/bwtree_index_test.cpp
lmwnshn: Fixed.

src/include/storage/index/index_defs.h
lmwnshn: We think you're right. Subsuming into UNIQUE.

src/include/storage/index/compact_ints_key.h
lmwnshn: Thanks!

src/include/storage/index/compact_ints_key.h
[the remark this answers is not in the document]
lmwnshn: Also re: namespacing, OIDs are a catalog-ish concept, so we think it makes sense for them to live there. We want OIDs to be unique across the various strong typedefs, and the catalog is in a better position to enforce that.

src/include/storage/index/index_defs.h
lmwnshn: The col_oid_t and indexkeycol_oid_t represent different concepts, indexkeycol is more ephemeral than the columns of a table. There can also be indexkey columns that do not map directly to a column in a table, e.g. index on (2*A + B). So we think maybe separating them is clearer.

Though we agree the name is bad...

test/storage/bwtree_index_test.cpp
lmwnshn: Inside the builder tests, though, we are not constructing the metadata object directly. The metadata concept is supposed to be purely internal and hidden from the world, and therefore there are no accessors either. We would need to add a friend elsewhere, do you think this is worth pursuing?

src/include/storage/index/index_defs.h
yangjuns: I see. Thanks for explaining.

src/include/storage/index/generic_key.h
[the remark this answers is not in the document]
yangjuns: Nope. My bad.

test/storage/bwtree_index_test.cpp
yangjuns: If there is no way around it, then I am ok with the current implementation.

PR #324, opened 2019-03-24
Layout changes for Arrow

tli2:
This PR cherry-picks from the Arrow branch with all the changes to block layout and alignment.

All buffers are now aligned at 8-byte boundaries as opposed to at attribute size. Extra space is reserved for hot-cold access control and metadata for Arrow


mbutrovich:
This PR resolves the segfault I've found in master while writing TPC-C. It looks like the calculations for BlockLayout isn't quite right after introducing varlens to the system. This can be confirmed by looking at the MemorySafety and Alignment TAS tests and noting that their random layouts don't use varlen attributes. This branch addresses those test changes and appears to have the fix. Would like to get this reviewed and merged ASAP.

src/include/storage/storage_defs.h
```diff
@@ -259,6 +269,51 @@ class VarlenEntry {
 // To make sure our explicit padding is not screwing up the layout
 static_assert(sizeof(VarlenEntry) == 16, "size of the class should be 16 bytes");
 
+/**
+ * Equality checker that checks the underlying varlen bytes are equal (deep)
+ */
+struct VarlenContentDeepEqual {
```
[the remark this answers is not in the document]
tli2: I think `std::hash` and `std::equal_to` should only be used for "default" behavior, because they are default template arguments to the STL data structures and people typically don't even pay attention to them when using STL.

I would argue that deep comparison that ignores ownership information is not "default" behavior, and I would rather users be conscious about needing this behavior.

mbutrovich:
I can't reproduce the performance regressions on data_table_benchmark locally. The random read benchmarks are actually faster for me, as opposed to on Jenkins. We may need to revisit how to properly evaluate Data Table since it's so low level in the system, and minor changes seem to wildly swing its benchmarks that don't really manifest when you exercise the system through higher-level interfaces.

src/include/storage/arrow_block_metadata.h
```diff
@@ -0,0 +1,241 @@
+#pragma once
+#include <map>
+#include <unordered_set>
+#include "storage/block_layout.h"
+#include "storage/storage_defs.h"
+#include "storage/storage_util.h"
+
+namespace terrier::storage {
+
+// TODO(Tianyu): Can use to specify cases where we don't concat per-block in the future (e.g. no need to put
```
mbutrovich: Is this mostly a TODO for yourself? I can't imagine anyone else will know what this means.

src/include/storage/arrow_block_metadata.h
```diff
@@ -0,0 +1,241 @@
...
+   * Constructs an empty Arrow VarlenColumn
+   */
+  ArrowVarlenColumn() = default;
+
+  /**
+   * Constructs a new arrow varlen column object
+   * @param values_length size of the values buffer, in bytes
+   * @param offsets_length number of elements in the offsets array
+   */
+  ArrowVarlenColumn(uint32_t values_length, uint32_t offsets_length)
```
mbutrovich: It's confusing to have one "length" arg mean bytes, and another mean "elements". Suggest renaming one or both.

src/include/storage/block_access_controller.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+#include <atomic>
+#include <utility>
+#include "common/macros.h"
+#include "common/strong_typedef.h"
+
+namespace terrier::storage {
+/**
+ * TODO(Tianyu): Placeholder for the actual implementation that has the same number of bytes. We use this to make sure
```
mbutrovich: Can you give a brief description of what this will be, since it's going to confuse people working on course projects to see this in the codebase?

src/include/storage/data_table.h
```diff
@@ -166,6 +166,16 @@ class DataTable {
    */
   SlotIterator end() const;
 
+  /**
+   * @warning NOT the number of tuples currently in the data table. Such a value is not well-defined unless referring
+   * to a transactional snapshot.
+   * @return number of slots allocated for this data table.
+   */
+  uint64_t NumSlots() const {
```
mbutrovich: Is this something that would eventually be replaced by the metrics team?

src/include/storage/storage_defs.h
```diff
@@ -40,10 +45,15 @@ struct alignas(common::Constants::BLOCK_SIZE) RawBlock {
    * scans through blocks and free up slots.
    */
   std::atomic<uint32_t> insert_head_;
+  /**
+   * Access controller of this block
```
mbutrovich: Which means what?

src/storage/data_table.cpp
```diff
@@ -133,31 +133,34 @@ TupleSlot DataTable::Insert(transaction::TransactionContext *const txn, const Pr
...
   }
-
-  data_table_counter_.IncrementNumInsert(1);
-  return result;
 }
 
 bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSlot slot) {
   data_table_counter_.IncrementNumDelete(1);
   // Create a redo
+  // TODO(Tianyu): Is it better to be consistent with the StageWrite behavior where we have caller explicitly
```
mbutrovich: Probably.

test/storage/tuple_access_strategy_test.cpp
```diff
@@ -182,19 +191,20 @@ TEST_F(TupleAccessStrategyTests, MemorySafety) {
 // These properties are necessary to ensure high performance by accessing aligned fields.
 // NOLINTNEXTLINE
 TEST_F(TupleAccessStrategyTests, Alignment) {
-  const uint32_t repeat = 100;
+  const uint32_t repeat = 1;
```
mbutrovich: Why was this changed to 1?

tli2: Reverted. Sorry about this

src/include/storage/arrow_block_metadata.h
tli2: Sorry. Rewritten this to be more clear.

src/include/storage/arrow_block_metadata.h
tli2: These are meant to be the length of these two Arrow arrays. Using Arrow's terminology for consistency. I modified 
the documentation of values to just be the length of the array, instead of the size in bytes, to hopefully make it clearer.

src/include/storage/block_access_controller.h
tli2: Good point. Copied the docs from actual blocks.

src/include/storage/data_table.h
tli2: Removed for now.

PR #325, opened 2019-03-26
Optimize GC

tli2:
The original GC unlinks every record it scans through in the transaction's context. Lin found out that in a highly skewed workload this can be a quadratic operation on a very long version chain. This PR fixes that by only using the records to infer the tuple slot to look at and truncating the version chain once per GC invocation.


mbutrovich:
I'm still struggling to explain the performance regression on the ReclaimTime benchmark. The workload didn't change, you didn't change the code exercised in that benchmark. It's baffling. I've profiled it...nothing explains it.

PR #328, opened 2019-03-26
Traffic cop with sqlite

DarkForte:
A prototype for the traffic cop. Now the traffic cop can only forward Simple Queries. 
It also brings in sqlite for testing the front-end stuff.

Status (2019.05.15):
- [x] Parse
- [x] Bind 
- [x] Execute 
- [x] Describe
- [x] Sync
- [x] TODO: Use each column's format in the result set.
- [ ] TODO: Query return tags
- [ ] TODO: Parameter Type inference
- [x] Junit Tests
- [ ] JDBC Tests (Not supported because it queries postgres catalog)
- [x] No-op benchmark
- [x] New API
- [x] Network Layer Refactor (Command Factory Injection, refactor `GetInstance` From ConnectionHandleFactory )
- [x] Network Layer Refactor (Multiple protocol support)


DarkForte:
Working on extended queries now...

src/network/postgres_network_commands.cpp
```diff
@@ -1,78 +1,211 @@
...
+    for(size_t i = 0; i < num_formats; i++)
+    {
+      auto format = in_.ReadValue<int16_t>();
+      is_binary.push_back(format);
+    }
+  }
+  else
+  {
+    NETWORK_LOG_ERROR("Error: Numbers of parameters don't match. {0} in statement, (1) in format code.", num_params, num_formats);
+    throw NETWORK_PROCESS_EXCEPTION("");
```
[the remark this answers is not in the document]
DarkForte: Oh, this kills the server? 
How should I close the connection?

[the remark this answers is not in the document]
DarkForte: ok, I'm now returning `Transition::TERMINATE`. Seems to be the right way to do it.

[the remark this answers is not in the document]
DarkForte: Thanks @pervazea, will send back an error message.

src/include/network/connection_context.h
```diff
@@ -0,0 +1,37 @@
...
+  /**
+   * The portals in this connection
+   */
+  std::unordered_map<std::string, traffic_cop::Portal> portals;
+
+  /**
+   * Cleans up this ConnectionContext
+   */
+  void Reset() {
+    for (auto pair : statements) pair.second.Finalize();
```
apavlo: Add a comment to explain what this does. It's not obvious.

src/include/network/connection_context.h
```diff
@@ -0,0 +1,37 @@
...
+
+  /**
+   * The portals in this connection
+   */
+  std::unordered_map<std::string, traffic_cop::Portal> portals;
+
+  /**
+   * Cleans up this ConnectionContext
+   */
+  void Reset() {
```
apavlo: Can you explain when should a `ConnectionContext` be reset? Is it when the connection closes or when the query finishes executing?

src/include/network/network_defs.h
```diff
@@ -20,15 +20,21 @@
...
+
 namespace terrier::network {
+class PostgresPacketWriter;
 
 // For epoch
 // static const size_t EPOCH_LENGTH = 40;
 
 // For threads
-#define CONNECTION_THREAD_COUNT 1
+#define CONNECTION_THREAD_COUNT 11
```
apavlo: 11 😐

src/include/network/postgres_protocol_utils.h
```diff
@@ -265,14 +313,72 @@ class PostgresPacketWriter {
...
+    }
+
+    EndPacket();
+  }
+
+  /**
+   * Writes a data row.
+   * @param values a row's values.
+   */
+  void WriteDataRow(const std::vector<std::string> &values) {
```
apavlo: Is the real system going to be passing a vector of strings?

src/include/network/postgres_protocol_utils.h
```diff
@@ -265,14 +313,72 @@ class PostgresPacketWriter {
...
+    for (auto &type : param_types) AppendValue<int32_t>(static_cast<int32_t>(type));
+
+    EndPacket();
+  }
+
+  /**
+   * Writes row description, as the first packet of sending query results
+   * @param columns the column names
+   */
+  void WriteRowDescription(const std::vector<std::string> &columns) {
```
apavlo: Is this information going to be provided by `ResultSet`? We should have the ExecEngine generate the exact object that this methods needs.

src/include/traffic_cop/portal.h
```diff
@@ -0,0 +1,26 @@
...
+#include <memory>
+#include <vector>
+#include "type/transient_value.h"
+
+namespace terrier::traffic_cop {
+
+/**
+ * A portal is a statement with bound parameters and is ready to execute.
+ */
+struct Portal {
```
apavlo: Will the real system have the same concept? Is this not just `ExecutionContext`?

src/network/command_factory.cpp
```diff
@@ -0,0 +1,31 @@
+#include "network/command_factory.h"
+namespace terrier::network {
+
+#define MAKE_COMMAND(type) \
+  std::static_pointer_cast<PostgresNetworkCommand, type>(std::make_shared<type>(packet))
+
+std::shared_ptr<PostgresNetworkCommand> CommandFactory::PacketToCommand(PostgresInputPacket *packet) {
```
apavlo: Should this be called `PostgresPacketToCommand`?

src/network/connection_handle.cpp
```diff
@@ -158,10 +159,11 @@ END_DEF
 }
 
 // TODO(Tianyu): Maybe use a factory to initialize protocol_interpreter here
-ConnectionHandle::ConnectionHandle(int sock_fd, ConnectionHandlerTask *handler)
+ConnectionHandle::ConnectionHandle(int sock_fd, ConnectionHandlerTask *handler, TrafficCop* t_cop, CommandFactory* command_factory)
     : conn_handler_(handler),
       io_wrapper_{new NetworkIoWrapper(sock_fd)},
-      protocol_interpreter_{new PostgresProtocolInterpreter()} {}
+      protocol_interpreter_{new PostgresProtocolInterpreter(command_factory)},
```
apavlo: Postgres specific code

src/include/network/postgres_protocol_utils.h
DarkForte: I don't think so, the `ResultSet` should have actual column schema rather than all strings... I haven't updated the `ResultSet` yet so it is like that right now.

src/include/network/postgres_protocol_utils.h
DarkForte: Yes, because actually this command also needs a lot of other information like table OIDs. They should go to `ResultSet`. Perhaps this function should accept a `ResultSet`? 

The current implementation does not make a `Resultset` for describe command, so the API is like this...

src/include/network/connection_context.h
DarkForte: Thanks for the comment! It's called when its connection handle is reused or destroyed. I have added doc on that.

src/include/network/network_defs.h
mbutrovich: Did we say that this should be 4 for now?

PR #330, opened 2019-03-29
GC contention benchmark

mbutrovich:
This new test supports #325 as a positive change, despite performance regressions on existing GC benchmarks, which strictly evaluate the GC's unlink and reclaim performance in isolation (i.e. no concurrent transactions). While this is useful for spotting regressions in the core GC algorithms, they doesn't properly evaluate how well the GC runs at the same time as the real system, particularly in high contention workloads from transactions. This PR adds a benchmark for that. On master the GC can clean up ~50k txns/second in this update-heavy workload. #325 increases that by an order of magnitude to ~500k txns/second on my laptop.

### DIFF
--- a/benchmark/storage/garbage_collector_benchmark.cpp
+++ b/benchmark/storage/garbage_collector_benchmark.cpp
@@ -127,14 +127,15 @@ BENCHMARK_DEFINE_F(GarbageCollectorBenchmark, ReclaimTime)(benchmark::State &sta
 }
 
 /**
- * Run a high number of statements with lots of updates to try to trigger aborts.
+ * Run a large number of updates on a small table to generate contention with the GC. Measure the number of transactions
+ * that the GC managed to free during the workload by subtracting the number of "lagging" transactions that still
+ * remained to be cleaned up by the GC after the workload was done running.
  */
 // NOLINTNEXTLINE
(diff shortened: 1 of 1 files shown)


benchmark/storage/garbage_collector_benchmark.cpp
```diff
@@ -98,9 +126,37 @@ BENCHMARK_DEFINE_F(GarbageCollectorBenchmark, ReclaimTime)(benchmark::State &sta
...
+ * Run a large number of updates on a small table to generate contention with the GC. Measure the number of transactions
+ * that the GC managed to free during the workload by subtracting the number of "lagging" transactions that still
+ * remained to be cleaned up by the GC after the workload was done running.
+ */
+// NOLINTNEXTLINE
+BENCHMARK_DEFINE_F(GarbageCollectorBenchmark, HighContention)(benchmark::State &state) {
+  uint64_t lag_count = 0;
+  // NOLINTNEXTLINE
+  for (auto _ : state) {
+    LargeTransactionBenchmarkObject tested({8, 8, 8}, 100, txn_length, update_select_ratio, &block_store_,
```
tli2: Not in your PR, but we apparently forgot to switch these to using the builders as well. (Tanuj changed the test versions a couple months back)

PR #334, opened 2019-04-03
Expression JSON (De)serialization

GustavoAngulo:
Changes added in this PR:
* JSON serialization and deserialization for expressions
* Change `Copy` methods of expressions to use `std::shared_ptr`. By convention, we want `AbstractExpression`s to be contained in `shared_ptr`s
* Fix for #343

If you are able to help out, it should take you less than 30min to add support for an expression type. 

**Instructions**
* Clone my branch [`GustavoAngulo/terrier:jesus_christ_its_json_bourne`](https://github.com/GustavoAngulo/terrier/tree/jesus_christ_its_json_bourne)
* Sign up for an expression type [here](https://docs.google.com/spreadsheets/d/1Pjl8LeVmUUOLOPfbuZ8GbBsEUFNZeoEaEAl1wKukDwY/edit?usp=sharing). I would recommend not signing up until you are going to work on it, as they are fairly quick to do, and maybe someone else can get it done sooner.
* Write the serialization and deserialization functions for the expression.  Look at `AggregateExpression` as an example. This should be your starting point for understanding how to do this. 
* Uncomment the cases in the switch statement in `abstract_expression.cpp:DeserializeExpression` that create your specific expression.
* Write a test case. Your test case should create a simple expression of the type you created serialization for, serialize it, deserialize it, and compare to make sure the expression and its fields are equal. Look at `expression_test.cpp` for an example for `AggregateExpression`.
* Send me a patch (gangulo@andrew.cmu.edu) with your change. Instructions on how to do that [here](https://stackoverflow.com/questions/6658313/generate-a-git-patch-for-a-specific-commit)

**Some notes**
* You may need to do (de)serialization for any subtypes that are members of your expression, such as `TransientValue`.
* Code style for json fields is the same as variables: lower_case_with_underscores. Look at `AbstractExpression::ToJson` as an example.
* You may need to remove any const quantifiers on member variables so you can set their values during deserialization. This is okay.
* Make sure to use the `DEFINE_JSON_DECLARATIONS` macro in the **namespace** of your expression, not inside the class. 
* Make sure to call `AbstractExpression::FromJson` and `AbstractExpression::ToJson` from each (de)serializer for expressions. Example of this in `AggregateExpression`
* Look at the test case in `expression_test.cpp` to see how (de)serialization will work in practice. 
* Make sure to add a default constructor if a no parameter constructor does not exist already 
* From the JSON library README:

> In function [FromJson], use function at() to access the object values rather than operator[]. In case a key does not exist, at throws an exception that you can handle, whereas operator[] exhibits undefined behavior.


Chenyao2333:
According to the discussion with Gus, we can use `pgpaser` to construct a `SubqueryExpression` for testing. But we cannot include `pgpaser` header file due to issue https://github.com/cmu-db/terrier/issues/343. I will add the test case since the issue resolved.

PR #336, opened 2019-04-06
DBMain and Settings Manager

DarkForte:
Added the Settings Manager.

Finished:
* Peloton code migration
* Callback framework
* `gflags`
* migrate to TransientValue
* An example of mutable parameter (object pool size)


src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,175 @@
...
+    delete txn;
+}
+
+int32_t SettingsManager::GetInt(Param param) { return GetValue(param).GetIntValue(); }
+
+double SettingsManager::GetDouble(Param param) { return GetValue(param).GetDecimalValue(); }
+
+bool SettingsManager::GetBool(Param param) { return GetValue(param).GetBooleanValue(); }
+
+std::string SettingsManager::GetString(Param param) { return GetValue(param).GetVarcharValue(); }
```
parsec0re: Be careful here. If you are using the new value system, there might be lifetime issues. Consider where to make a copy and where to just "peek" values.

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+type::Value SettingsManager::GetValue(Param param) {
+  auto param_info = param_map_.find(param);
+  return param_info->second.value;
+}
+
+void SettingsManager::SetValue(Param param, const type::Value &value) {
+  ParamInfo& param_info = param_map_.find(param)->second;
+
+  if (!param_info.is_mutable) throw SETTINGS_EXCEPTION((param_info.name + " is not mutable.").c_str());
```
parsec0re: Is it the correct behavior to throw an exception? Is the upper level users responsible for catching these exceptions?

test/settings/settings_test.cpp
```diff
@@ -0,0 +1,66 @@
...
+
+  bufferPoolSize = txn_manager_->GetBufferPoolSizeLimit();
+  EXPECT_EQ(bufferPoolSize, defaultBufferPoolSize);
+
+  // Setting new value should invoke callback.
+  const int64_t newBufferPoolSize = defaultBufferPoolSize + 1;
+  settings_manager_->SetInt(Param::buffer_pool_size, static_cast<int32_t>(newBufferPoolSize));
+  bufferPoolSize = static_cast<int64_t>(settings_manager_->GetInt(Param::buffer_pool_size));
+  EXPECT_EQ(bufferPoolSize, newBufferPoolSize);
+
```
parsec0re: Maybe more tests such as setting immutable config.

src/include/settings/settings_manager.h
```diff
@@ -0,0 +1,132 @@
...
+
+  /**
+   * Migrate values from GFlags to internal map
+   */
+  void InitParams();
+
+ private:
+  catalog::SettingsHandle settings_handle_;
+  transaction::TransactionManager *txn_manager_;
+  std::unordered_map<Param, ParamInfo> param_map_;
```
parsec0re: Just not sure whether these data structures are thread-safe.

src/include/settings/settings_manager.h
```diff
@@ -0,0 +1,132 @@
...
+ public:
+  SettingsManager() = delete;
+  SettingsManager(const SettingsManager &) = delete;
+
+  /**
+   * The constructor of settings manager
+   * @param catalog a shared pointer to the system catalog
+   * @param txn_manager a pointer to the transaction manager
+   */
+  SettingsManager(const std::shared_ptr<catalog::Catalog> &catalog, transaction::TransactionManager *txn_manager);
```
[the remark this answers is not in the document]
parsec0re: Or if you really want the "shared" semantics, pass by value.

src/include/settings/settings_manager.h
YuzeLiao: Thanks for your comments. They are not thread-safe for now, but we will solve it upon the completion of most of the other work . It should be easy and a coarse-grained latch may work because settings should not be read/written very frequently.

src/include/settings/settings_defs.h
```diff
@@ -0,0 +1,184 @@
+//===----------------------------------------------------------------------===//
+// FILE LOCATIONS
+//===----------------------------------------------------------------------===//
+
+//===----------------------------------------------------------------------===//
+// CONNECTIONS
+//===----------------------------------------------------------------------===//
+
+// Peloton port
```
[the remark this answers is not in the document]
YuzeLiao: Thanks for your reminder. Fixed.

src/include/type/value.h
```diff
@@ -245,6 +245,42 @@ class Value {
     return *this;  // return the object itself (by reference)
```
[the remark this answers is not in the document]
YuzeLiao: Thanks for your comments. We've merged the latest catalog and migrated to TransientValue system.

src/settings/settings_manager.cpp
YuzeLiao: Thanks for your reminder. We don't make copies here in the new TransientValue System.

src/include/settings/settings_param.h
```diff
@@ -0,0 +1,43 @@
+#pragma once
+#include <type/value.h>
+#include <string>
+
+namespace terrier::settings {
+
+enum class Param {
+#define __SETTING_ENUM__
```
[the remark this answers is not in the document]
YuzeLiao: Thanks for your comments. Yes, it's a macro expansion so we do need some docs. Added.

src/include/settings/settings_manager.h
```diff
@@ -0,0 +1,139 @@
...
```
Calculuser: Not understand why use `shared_ptr<> &`. I think a reference is better because `shared_ptr<> &` may affect the performance

src/include/loggers/settings_logger.h
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <memory>
+#include "spdlog/sinks/basic_file_sink.h"
+#include "spdlog/spdlog.h"
+
+namespace terrier::settings {
+extern std::shared_ptr<spdlog::logger> settings_logger;
```
Calculuser: Why use `shared_ptr` here? Maybe a raw pointer is better.

src/settings/README.md
```diff
@@ -0,0 +1,90 @@
...
+- Can mutable parameters be correctly changed? If it has a callback function, is that called?
+- Can the Settings Manager check if the new value is between `min_value` and `max_value`?
+- Can the Settings Manager refuse to change immutable parameters?
+
+## Trade-offs and Potential Problems
+
+> Write down any conscious trade-off you made that can be problematic in the future or any problems discovered during the design process that remain unaddressed (technical debts).
+
+By adding the internal map, the system is storing 3 copies of each parameter value: one in its component (each component is very likely to maintain a local copy, like in the Garbage Collector there is very likely to be a `num_threads` member variable), one in the map, and one in the catalog. However, this is necessary as stated above. The real unnecessary copy is actually the `pg_settings` table. In fact, Postgresql does not have this physical table. It also stores the values in an internal map. When someone calls `SELECT * FROM pg_settings`, it generates a view from the internal map. However, Terrier does not support generating a view from functions yet. Therefore, we must maintain the physical table.
```
Calculuser: Since there are 3 copies of each parameter value, will those values come into an inconsistent state? If so, how do you plan to handle this inconsistency?

src/include/loggers/settings_logger.h
Calculuser: Or `unique_ptr`.

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+void SettingsManager::SetString(Param param, const std::string_view &value) {
+  std::string_view old_value = GetString(param);
+  SetValue(param, ValueFactory::GetVarChar(value));
+  callback_fn callback = callback_map_.find(param)->second;
+  std::string_view new_value(value);
+  callback(static_cast<void *>(&old_value), static_cast<void *>(&new_value));
+}
+
+const std::string SettingsManager::GetInfo() {
```
Calculuser: It's better to remove those comment-out code.

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,190 @@
...
+  param_info.value = ValueFactory::GetCopy(value);
+
+  auto txn = txn_manager_->BeginTransaction();
+  auto entry = settings_handle_.GetSettingsEntry(txn, param_info.name);
+  entry->SetColumn(static_cast<int32_t>(Index::SETTING), value);
+  txn_manager_->Commit(txn, EmptyCallback, nullptr);
+  delete txn;
+}
+
+bool SettingsManager::ValidateValue(const type::TransientValue &value, const type::TransientValue &min_value,
```
Calculuser: Maybe overloading `<` and `>` in the class `TransientValue` of those value type is better. It will make this function more concise.

src/include/settings/settings_defs.h
```diff
@@ -0,0 +1,184 @@
...
+    "path to root certificate file",
+    "root.crt",
+    false, false)
+    */
+
+    //===----------------------------------------------------------------------===//
+    // RESOURCE USAGE
+    //===----------------------------------------------------------------------===//
+
+    SETTING_double(bnlj_buffer_size, "The default buffer size to use for blockwise nested loop joins (default: 1 MB)",
```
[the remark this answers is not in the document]
DarkForte: Many of them are not. But I believe many of them may be needed in the future, so it does no harm to leave them there for now. Anyway, they are examples of how to define parameters.

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+    entry[static_cast<int>(Index::OID)] = ValueFactory::GetInteger(!oid);
+    entry[static_cast<int>(Index::NAME)] = ValueFactory::GetVarChar(info.name.c_str());
+    entry[static_cast<int>(Index::SHORT_DESC)] = ValueFactory::GetVarChar(info.desc.c_str());
+
+    settings_handle_.InsertRow(txn, entry);
+  }
+
+  txn_manager_->Commit(txn, EmptyCallback, nullptr);
+  delete txn;
```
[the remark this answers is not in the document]
lagoon0o0: [This comment](https://github.com/cmu-db/terrier/issues/339#issuecomment-481742215) probably explains when and where to delete a transaction context.

src/include/settings/settings_manager.h
lagoon0o0: Agreed.

test/settings/settings_test.cpp
```diff
@@ -0,0 +1,71 @@
...
+  bufferPoolSize = txn_manager_->GetBufferPoolSizeLimit();
+  EXPECT_EQ(bufferPoolSize, newBufferPoolSize);
+}
```
lagoon0o0: A multi-transaction, GC enabled test would be better.

src/main/main_database.cpp
```diff
@@ -0,0 +1,57 @@
...
+  // log init now complete
+  LOG_TRACE("Logger initialization complete");
+  // initialize loggers
+  ::google::SetUsageMessage("Usage Info: \n");
+  ::google::ParseCommandLineFlags(&argc, &argv, true);
+  // initialize stat registry
+  auto main_stat_reg = std::make_shared<terrier::common::StatisticsRegistry>();
+
+  // create the global transaction mgr
+  terrier::storage::RecordBufferSegmentPool buffer_pool_(100000, 10000);
```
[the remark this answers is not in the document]
DarkForte: Will be handled in DBMain PR.

src/settings/settings_manager.cpp
DarkForte: Maybe you are right. It would be better to LOG_ERROR and tell the caller that this `set` failed.

test/settings/settings_test.cpp
DarkForte: There is one: https://github.com/DarkForte/terrier/blob/settings_manager/test/settings/settings_test.cpp#L49

src/settings/settings_manager.cpp
DarkForte: I think you are right.

src/include/loggers/settings_logger.h
DarkForte: Because everyone is using `shared_ptr` in their loggers...

src/settings/settings_manager.cpp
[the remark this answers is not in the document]
DarkForte: ok, we are now running GC in the tests.

src/settings/README.md
DarkForte: No, as long as the component is using our setters and getters to change and load configurations.

src/include/main/db_main.h
```diff
@@ -0,0 +1,102 @@
...
+   * @param old_value old value of buffer pool size
+   * @param new_value new value of buffer pool size
+   * @param action_context action context for changing buffer pool size
+   */
+  void BufferPoolSizeCallback(void *old_value, void *new_value,
+                              const std::shared_ptr<common::ActionContext> &action_context);
+
+ private:
+  friend class settings::SettingsManager;
+  friend class settings::SettingsTests;
```
yeshengm: Use the `FRIEND_TEST` macro?

src/include/main/db_main.h
```diff
@@ -0,0 +1,102 @@
...
+  std::shared_ptr<common::StatisticsRegistry> main_stat_reg_;
+  std::unordered_map<settings::Param, settings::ParamInfo> param_map_;
+  transaction::TransactionManager *terrier_txn_manager_;
```
yeshengm: I think just leave these fields as `txn_manager_` is enough. The additional `terrier` is a bit verbose.

src/include/main/db_main.h
```diff
@@ -0,0 +1,102 @@
...
+
+namespace terrier {
+
+namespace settings {
+class SettingsManager;
+class SettingsTests;
+}  // namespace settings
+
+/**
+ * The DBMain Class holds all the static functions, singleton pointers, etc.
```
yeshengm: I think we should mention that the db main serves as a global context of the whole DBMS in the documentation.

src/include/main/db_main.h
```diff
@@ -0,0 +1,102 @@
...
+ * It has the full knowledge of the whole database systems.
+ * *Only the settings manager should be able to access the DBMain object.*
+ */
+class DBMain {
```
lagoon0o0: If one DBMain only correspond to one settings manager, it might be a good idea to merge these two classes into one.

src/include/main/db_main.h
```diff
@@ -0,0 +1,102 @@
...
+  void ForceShutdown();
+
+  // TODO(Weichen): Use unique ptr is enough.
+  /**
+   * Basic empty callbacks used by settings manager
+   * @param old_value the old value of corresponding setting
+   * @param new_value the new value of corresponding setting
+   * @param action_context action context for empty callback
+   */
+  void EmptyCallback(void *old_value, void *new_value, const std::shared_ptr<common::ActionContext> &action_context);
```
lagoon0o0: Making the callback to be a member function is a little weird.

lagoon0o0: Probably defining a wrapper class for callback function will work.

src/include/main/db_main.h
```diff
@@ -0,0 +1,102 @@
...
+   * @param action_context action context for changing buffer pool size
+   */
+  void BufferPoolSizeCallback(void *old_value, void *new_value,
+                              const std::shared_ptr<common::ActionContext> &action_context);
+
+ private:
+  friend class settings::SettingsManager;
+  friend class settings::SettingsTests;
+  std::shared_ptr<common::StatisticsRegistry> main_stat_reg_;
+  std::unordered_map<settings::Param, settings::ParamInfo> param_map_;
```
lagoon0o0: Is there any potential concurrency issue?

src/settings/README.md
```diff
@@ -0,0 +1,96 @@
...
+- It is much easier to operate on a map than the catalog. 
+- Since the number of parameters is low, an extra map won't introduce much overhead.
+
+- Although this essentially turned the Settings Manager as a cache layer for the `pg_settings` table, its consistency is easy to maintain (just update the catalog in `set` methods).
+
+This value map is an external map actually stored in DBMain. We also have another internal callback map, which maps a setting to its callback. It should be immutable during runtime.
+
+### Synchronous Modification Enforcement
```
Calculuser: How does async call maintain consistency?

test/settings/settings_test.cpp
```diff
@@ -0,0 +1,192 @@
...
+                            setter_callback);
+  bufferPoolSize = static_cast<int64_t>(settings_manager_->GetInt(Param::buffer_pool_size));
+  EXPECT_EQ(bufferPoolSize, newBufferPoolSize);
+
+  bufferPoolSize = txn_manager_->GetBufferPoolSizeLimit();
+  EXPECT_EQ(bufferPoolSize, newBufferPoolSize);
+}
+
+// NOLINTNEXTLINE
+TEST_F(SettingsTests, ConcurrentModifyTest) {
```
Calculuser: What if multiple transactions are modifying the same settings? I think it is better to add this test.

test/settings/settings_test.cpp
```diff
@@ -0,0 +1,192 @@
...
+
+  std::thread t2([&] {
+    std::shared_ptr<common::ActionContext> action_context = std::make_shared<common::ActionContext>(2);
+    settings_manager_->SetBool(Param::parallel_execution, false, action_context, setter_callback);
+    EXPECT_FALSE(settings_manager_->GetBool(Param::parallel_execution));
+  });
+
+  std::thread t3([&] {
+    std::shared_ptr<common::ActionContext> action_context = std::make_shared<common::ActionContext>(3);
+    settings_manager_->SetString(Param::db_name, "PelotonSP", action_context, setter_callback);
```
Calculuser: Change "Peloton" to "terrier".

src/main/terrier.cpp
```diff
@@ -1,53 +1,50 @@
...
 
-  LOG_INFO("Initialization complete");
-
-  terrier::network::TerrierServer terrier_server;
-  terrier_server.SetupServer().ServerLoop();
-
-  // shutdown loggers
-  spdlog::shutdown();
-  main_stat_reg->Shutdown(false);
+  ConstructParamMap(param_map);
```
yeshengm: This piece of code is weird. Is it possible to generate the param map during the database bootstrap phase? I.e. `db.Bootstrap()`, `db.GetParams()`, `db.Init()`, `db.Run()`.

src/main/db_main.cpp
```diff
@@ -0,0 +1,88 @@
...
+}
+
+void DBMain::Init() {
+  InitLoggers();
+
+  // initialize stat registry
+  main_stat_reg_ = std::make_shared<terrier::common::StatisticsRegistry>();
+
+  // create the global transaction mgr
+  auto *buffer_pool = new storage::RecordBufferSegmentPool(100000, 10000);
```
lagoon0o0: Probably better to remove the magic number here.

src/include/main/db_main.h
Calculuser: Will this map become a contention?

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,227 @@
...
+    throw SETTINGS_EXCEPTION("Invalid setting value");
+  }
+}
+
+void SettingsManager::InitializeCatalog() {
+  auto txn = txn_manager_->BeginTransaction();
+  auto column_num = catalog::SettingsHandle::schema_cols_.size();
+
+  for (const auto &pair : db_->param_map_) {
+    const Param param = pair.first;
```
yeshengm: You can use C++17 structured bindings here if you wish. This can simplify handling of data structures such as `std::pair`.

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,227 @@
...
+    const ParamInfo &info = pair.second;
+
+    catalog::settings_oid_t oid(static_cast<uint32_t>(param));
+    std::vector<type::TransientValue> entry;
+    for (auto i = column_num; i > 0; --i) {
+      // NOLINTNEXTLINE
```
yeshengm: NOLINT?

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,227 @@
...
+  return ValuePeeker::PeekBoolean(GetValue(param));
+}
+
+std::string_view SettingsManager::GetString(Param param) {
+  common::SharedLatch::ScopedExclusiveLatch guard(&latch_);
+  return ValuePeeker::PeekVarChar(GetValue(param));
+}
+
+void SettingsManager::SetInt(Param param, int32_t value, std::shared_ptr<ActionContext> action_context,
+                             setter_callback_fn setter_callback) {
```
yeshengm: Why `setter_callback_fn` is snake case whereas all other types and camel case?

src/settings/settings_manager.cpp
lagoon0o0: Why NOLINT?

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,227 @@
...
+void SettingsManager::SetString(Param param, const std::string_view &value,
+                                std::shared_ptr<ActionContext> action_context, setter_callback_fn setter_callback) {
+  common::SharedLatch::ScopedExclusiveLatch guard(&latch_);
+  std::string_view old_value = ValuePeeker::PeekVarChar(GetValue(param));
+  if (!SetValue(param, ValueFactory::GetVarChar(value))) {
+    action_context->SetState(ActionState::FAILURE);
+  } else {
+    std::string_view new_value(value);
+    ActionState action_state =
+        InvokeCallback(param, static_cast<void *>(&old_value), static_cast<void *>(&new_value), action_context);
```
yeshengm: If you are passing a pointer to function with argument type `void *`, IIRC there is no need to use a cast.

src/include/settings/settings_param.h
```diff
@@ -0,0 +1,70 @@
...
+   * @param max_value allowed maximum value
+   * @param callback function invoked when param is changed
+   */
+  ParamInfo(std::string name, type::TransientValue &&value, std::string desc, type::TransientValue &&default_value,
+            bool is_mutable, double min_value, double max_value, callback_fn callback)
+      : name(std::move(name)),
+        value(std::move(value)),
+        desc(std::move(desc)),
+        default_value(std::move(default_value)),
+        is_mutable(is_mutable),
```
yeshengm: I think name this as `mutable` is enough since the type `bool` already tells you what the variable is for.

src/include/main/db_main.h
lagoon0o0: I know there should be only one settings mgr accessing DBmain here, but adding a latch probably will make this data structure safer.

src/include/settings/settings_manager.h
```diff
@@ -0,0 +1,139 @@
...
+   * @return current setting value
+   */
+  bool GetBool(Param param);
+
+  /**
+   * Get the value of a string setting
+   * @param param setting name
+   * @return current setting value
+   */
+  std::string_view GetString(Param param);
```
yeshengm: I'm worried about potential lifetime issues here. `string_view`s themselves do not own underlying memory. But here you are giving out the reference, which is a dangerous thing to do. Our transient value implementation returns `string_view`s mainly because we relies on programmers to manage lifetime issues. I don't know whether the `string_view` here is safe.

src/catalog/settings_handle.cpp
```diff
@@ -0,0 +1,66 @@
...
+  pg_settings = new catalog::SqlTableHelper(pg_settings_oid);
+
+  // columns we use
+  for (auto col : SettingsHandle::schema_cols_) {
+    pg_settings->DefineColumn(col.col_name, col.type_id, false, col_oid_t(catalog->GetNextOid()));
+  }
+
+  // now actually create, with the provided schema
+  pg_settings->Create();
+  // catalog->AddToMaps(db_oid, pg_settings_oid, name, pg_settings);
```
lagoon0o0: dead code

src/main/db_main.cpp
```diff
@@ -0,0 +1,88 @@
+#include "main/db_main.h"
+#include <memory>
```
lagoon0o0: Maybe better to reorder this

src/settings/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Macro based Parameter Definition
+
+We provide some macros for the developers to define their parameters. They include `SETTING_int`, `SETTING_bool`, `SETTING_double` and `SETTING_string`. The definition of the macros is in `settings_macro.h`.
+
+Although macros are confusing and not straightforward, it offers more benefits than disadvantages in the Settings Manager. The biggest benefit is that the users can now specify each parameter only once. Without macros, when a user wants to define a new parameter, he must do it for three times: append it in the `Params` `enum`, register it in `gflags`, and initialize it in the Settings Manager. It is clear that this exposes too many internal details of the Settings Manager to the users. 
+
+In the current design, users only need to define his parameter in `settings.h` only once, and the Settings Manager will do the `enum`, `gflags` and initialization stuff for them by simply `#include settings.h` and interpret `SETTING_xxx` in different ways.
+
+### Internal and External Map
```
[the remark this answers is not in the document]
DarkForte: This map is a hotspot, but the Settings Manager is not in the critical path, so it's okay.

src/include/common/action_context.h
```diff
@@ -0,0 +1,53 @@
...
+  IN_PROGRESS = 1,
+  SUCCESS = 2,
+  FAILURE = 3,
+  DEFERRED = 4,
+};
+
+/**
+ * ActionContext is used to keep track of action state and system behavior.
+ */
+class ActionContext {
```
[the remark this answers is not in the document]
DarkForte: No, because there may also be logs and other information in `ActionContext` later.

src/include/settings/settings_common.h
```diff
@@ -0,0 +1,180 @@
...
+#ifdef SETTING_string
+#undef SETTING_string
+#endif
+
+#define VALIDATOR_int(name, default_value)                                                      \
+  static bool Validate##name(const char *setting_name, int value) {                             \
+    if (FLAGS_##name == static_cast<int>(default_value)) {                                      \
+      return true;                                                                              \
+    }                                                                                           \
+    std::cerr << "Value for \"" << #name << "\" has been set to " << FLAGS_##name << std::endl; \
```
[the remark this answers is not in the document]
DarkForte: No, because when this is called, the loggers are not initialized yet. So we have to use `cerr`.

src/include/main/db_main.h
DarkForte: I think it will make the class too big. I would insist that DBMain and Settings Manager be two classes.

src/include/main/db_main.h
DarkForte: FRIEND_TEST is actually `friend class TheTestClassName`, so in my opinion, it doesn't hurt to keep the current code.

src/include/main/db_main.h
DarkForte: Fixed, thx!

src/include/settings/settings_param.h
```diff
@@ -0,0 +1,70 @@
...
+   * @param name setting name
+   * @param value setting value
+   * @param desc a description of the setting
+   * @param default_value the default value of the setting
+   * @param is_mutable if the setting is mutable or not
+   * @param min_value allowed minimum value
+   * @param max_value allowed maximum value
+   * @param callback function invoked when param is changed
+   */
+  ParamInfo(std::string name, type::TransientValue &&value, std::string desc, type::TransientValue &&default_value,
```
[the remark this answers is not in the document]
DarkForte: clang-tidy suggests "pass by value and use move constructor"...

src/include/settings/settings_param.h
```diff
@@ -0,0 +1,70 @@
...
+        desc(std::move(desc)),
+        default_value(std::move(default_value)),
+        is_mutable(is_mutable),
+        min_value(min_value),
+        max_value(max_value),
+        callback(callback) {}
+
+ private:
+  friend class SettingsManager;
+  std::string name;
```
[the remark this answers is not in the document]
DarkForte: Fixed, thx!

src/main/terrier.cpp
DarkForte: No, because we don't want to expose command line arguments to DBMain.

src/settings/settings_manager.cpp
DarkForte: Removed it. Thanks!

src/settings/settings_manager.cpp
DarkForte: That is news to me. Thank you!

src/main/db_main.cpp
YuzeLiao: Solved. Used specified buffer_pool_size to replace the first number. As long as the other setting values come into use, the other corresponding numbers can then be replaced by specified values, too.

src/include/main/db_main.h
YuzeLiao: We have made it thread-safe. At least in the short term, it will not become a hot-spot bottleneck since not everybody (i.e. thread)  can (will) have frequent interaction with settings manager.

src/catalog/settings_handle.cpp
YuzeLiao: Solved.

src/settings/settings_manager.cpp
YuzeLiao: We were trying to be consistent with transaction_manager.

src/settings/README.md
YuzeLiao: Modification in an asynchronous manner with callback itself will not have consistency issue. Because the observer of any setting value are not changed during the whole process. Namely, consistency issue happens when there are multiple observers (or owners) of a value, which is not what we are explaining here.

If you are asking how to maintain consistency between different observers, which would happen no matter the modification is async or not, you can refer to the sections above :)

src/settings/settings_manager.cpp
YuzeLiao: Thank you for the reminder :)

src/include/settings/settings_param.h
YuzeLiao: Thank you. We are trying to be consistent with the other part of the system like postgresparser and worker_pool. And you may not always remember what the type of a variable is on seeing it :)

src/include/main/db_main.h
YuzeLiao: Probably, but that depends on the global design of the main database class. Our DBMain is the first version (based on a discussion with Andy about a month ago). With the further development of the whole system, this part may be more elegant.

src/include/main/db_main.h
YuzeLiao: Thanks, for now there will be no concurrency issue.

src/include/settings/settings_manager.h
YuzeLiao: Thank you. Solved.

src/include/settings/settings_defs.h
```diff
@@ -0,0 +1,212 @@
...
+// GENERAL
+//===----------------------------------------------------------------------===//
+
+// Display configuration
+SETTING_bool(display_settings, "Display settings (default: false)", false, true, terrier::DBMain::EmptyCallback)
+
+//===----------------------------------------------------------------------===//
+// Used only in Tests
+//===----------------------------------------------------------------------===//
+
```
pervazea: Having test values in the main code is not too nice. 
These should be in the test directory with whatever test is exercising these values.

src/include/catalog/settings_handle.h
```diff
@@ -102,4 +102,25 @@ class SettingsCatalogTable {
   catalog::SqlTableHelper *pg_settings_;
 };
 
+enum class SettingsTableColumn {
```
mbutrovich: Did these come from Peloton? I see most aren't used yet. I'm not clear on which ones we want yet, but would just like to understand the process that got them here.

DarkForte: @pervazea is the author of `settings_handle`, we don't know how it works...

src/include/common/action_context.h
```diff
@@ -0,0 +1,53 @@
...
+  void SetState(ActionState state) { state_.store(state); }
+
+  /**
+   * Get action id.
+   * @return action id.
+   */
+  int32_t GetActionId() { return action_id_; }
+
+ private:
+  int32_t action_id_;
```
mbutrovich: Is there a reason we didn't make these strong typedef'd oids?

src/include/transaction/transaction_manager.h
```diff
@@ -76,6 +76,19 @@ class TransactionManager {
    */
   TransactionQueue CompletedTransactionsForGC();
 
+  /**
+   * Set the current buffer pool size
+   * @param new_size the new size of buffer pool
+   * @return true if the value is set successfully; false otherwise
+   */
+  bool SetBufferPoolSizeLimit(uint64_t new_size) { return buffer_pool_->SetSizeLimit(new_size); }
```
mbutrovich: The transaction manager holds a reference to the buffer segment pool, but I'm not convinced it actually owns it, and thus maybe isn't where these methods should reside.

src/include/catalog/settings_handle.h
pervazea: The columns implemented in settiings_handle from Postgres pg_settings. Settings handle is "just a catalog table". You can set or read columns in the settings table.
Which ones are used is up to the team implementing settings.

The enum here though, seems unnecessary. The catalog code provides methods to access the columns. See jrolli's in progress catalog PR.

src/main/db_main.cpp
```diff
@@ -0,0 +1,90 @@
...
+void DBMain::Init() {
+  InitLoggers();
+
+  // initialize stat registry
+  main_stat_reg_ = std::make_shared<common::StatisticsRegistry>();
+
+  // create the global transaction mgr
+  auto *buffer_pool = new storage::RecordBufferSegmentPool(
+      type::TransientValuePeeker::PeekInteger(param_map_.find(settings::Param::buffer_pool_size)->second.value_),
+      10000);
```
pervazea: The size limit is via a setting, but the re-use limit is hard coded? IMO the latter should also be a settings value.

src/include/main/db_main.h
```diff
@@ -0,0 +1,102 @@
...
+   */
+  void KnobCallback(void *old_value, void *new_value, const std::shared_ptr<common::ActionContext> &action_context);
+
+  /**
+   * Buffer pool size callback used by settings manager
+   * @param old_value old value of buffer pool size
+   * @param new_value new value of buffer pool size
+   * @param action_context action context for changing buffer pool size
+   */
+  void BufferPoolSizeCallback(void *old_value, void *new_value,
```
mbutrovich: DBMain seems like a really weird place to stick callbacks for Settings Manager actions.

DarkForte: But it is the only place where all the components are accessible...

mbutrovich: Is there a hierarchy that makes sense without us having have to define action callbacks in DBMain? I'm envisioning potentially dozens or hundreds some day and having them all in DBMain seems incorrect.

src/settings/settings_manager.cpp
```diff
@@ -0,0 +1,209 @@
...
+    SETTINGS_LOG_ERROR(
+        "Value given for \"{}"
+        "\" is not in its min-max bounds",
+        info.name_);
+    throw SETTINGS_EXCEPTION("Invalid setting value");
+  }
+}
+
+int32_t SettingsManager::GetInt(Param param) {
+  common::SharedLatch::ScopedExclusiveLatch guard(&latch_);
```
mbutrovich: These don't need to be exclusive do they? SharedLatch should suffice since it's only reading.

PR #337, opened 2019-04-06
Add/drop indexes

parsec0re:
- [ ] Add catalog support.
- [ ] Support building index in a very basic way.


src/include/storage/index/index_populator.h
```diff
@@ -0,0 +1,65 @@
...
+    auto table_pr_init = sql_table.InitializerForProjectedRow(col_oids).first;
+    auto *table_pr_buf = common::AllocationUtil::AllocateAligned(table_pr_init.ProjectedRowSize());
+    ProjectedRow *select_pr = table_pr_init.InitializeRow(table_pr_buf);
+
+    // Record the col_id of each column
+    const col_id_t *columns = select_pr->ColumnIds();
+    uint16_t num_cols = select_pr->NumColumns();
+    std::vector<col_id_t> sql_table_cols(columns, columns + num_cols);
+
+    for (const auto &it : sql_table) {
```
wenhaoh2: Are you implementing the blocking approach here? If so, should it acquire lock for the sql table before populating the index?

test/storage/index_populator_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+  auto index_key_schema = IndexKeySchema(key_cols_0);
+
+  // Build the index
+  index_builder.SetOid(index_oid_0);
+  index_builder.SetConstraintType(ConstraintType::DEFAULT);
+  index_builder.SetKeySchema(index_key_schema);
+  auto index_0 = index_builder.Build();
+
+  // Populate the index
+  IndexPopulator::PopulateIndex(txn_, *table->GetSqlTable(), IndexKeySchema(key_cols_0), *index_0);
```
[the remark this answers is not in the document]
parsec0re: We're considering decoupling index & SQL table APIs. Therefore an index builder only builds a skeleton for later population. (BTW, IndexBuilder is not written by us).

src/storage/lock_manager.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+namespace terrier::storage {
+
+common::SharedLatch *LockManager::GetLock(const catalog::table_oid_t oid) {
+  common::SpinLatch::ScopedSpinLatch guard(&latch_);
+  if (lock_table_.find(oid) == lock_table_.end()) {
+    lock_table_[oid] = std::make_unique<common::SharedLatch>();
+  }
+
+  return lock_table_.at(oid).get();
```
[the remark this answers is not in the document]
YuzeLiao: I guess it's probably because a global lock manager may lead to some hot spots, so `shared_ptr` may cause performance degradation.

test/storage/index_populator_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include "util/test_harness.h"
+#include "util/transaction_test_util.h"
+
+namespace terrier::storage::index {
+struct IndexPopulatorTest : public TerrierTest {
+  void SetUp() override {
+    TerrierTest::SetUp();
+    txn_manager_ = new transaction::TransactionManager(&buffer_pool_, true, LOGGING_DISABLED);
+
+    catalog_ = new catalog::Catalog(txn_manager_);
```
YuzeLiao: [Major] The test probably can not be built because the Catalog does not have such constructor. But I think it's totally OK because it just needs some efforts to fix.

src/include/storage/index/index_populator.h
```diff
@@ -0,0 +1,65 @@
...
+  /**
+   * The method populates all tuples in the table visible to current transaction with latest version,
+   * then pack target columns into ProjectedRow's and insert into the given index.
+   *
+   * @param txn current transaction context
+   * @param sql_table the target sql table
+   * @param index target index inserted into
+   */
+  static void PopulateIndex(transaction::TransactionContext *txn,                                         // NOLINT
+                            SqlTable &sql_table, const IndexKeySchema &index_key_schema, Index &index) {  // NOLINT
```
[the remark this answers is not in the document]
Calculuser: Thank you for your comment. Actually, we plan to check the uniqueness constraint when adding record into the catalog, which is before populating index.

src/include/storage/index/index_populator.h
Calculuser: Yes, you are right. Actually, here we just write some necessary methods here and we will write a wrapper function later to show steps of blocking approach. That wrapper function will include acquiring and releasing locks.

test/storage/index_populator_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+  for (const auto &it : *table->GetSqlTable()) {
+    if (table->GetSqlTable()->Select(txn_, it, key)) {
+      for (uint16_t i = 0; i < key->NumColumns(); ++i) {
+        key->ColumnIds()[i] = index_key->ColumnIds()[i];
+      }
+      std::vector<TupleSlot> tup_slots;
+      index_0->ScanKey(*key, &tup_slots);
+      bool flag = false;
+      for (const auto &tup_slot : tup_slots) {
+        if (it == tup_slot) {
```
YuzeLiao: [Major] Is it better to also expect that there is only one tuple in returned `tup_slots`?

test/storage/index_manager_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+  }
+  storage::RecordBufferSegmentPool buffer_pool_{100, 100};
+  transaction::TransactionManager *txn_manager_;
+  IndexManager *index_manager_;
+};
+
+// Check the correctness of drop index
+// NOLINTNEXTLINE
+TEST_F(IndexManagerTest, DropIndexCorrectnessTest) {
+  auto txn0 = txn_manager_->BeginTransaction();
```
YuzeLiao: [Major] Since the current implementation follows the non-blocking manner, a concurrent correctness test may be necessary.

src/storage/index/index_manager.cpp
```diff
@@ -0,0 +1,189 @@
...
+  // initialize the building flag to false
+  auto index_id = make_index_id(db_oid, ns_oid, index_oid);
+  SetIndexBuildingFlag(index_id, false);
+
+  // Commit first transaction
+  transaction::timestamp_t commit_time = txn_mgr->Commit(txn1, TestCallbacks::EmptyCallback, nullptr);
+
+  // Wait for all transactions older than the timestamp of previous transaction commit
+  // TODO(jiaqizuo): use more efficient way to wait for all previous transactions to complete.
+  while (txn_mgr->OldestTransactionStartTime() < commit_time) {
```
DarkForte: For a more efficient solution, I can only come up with using a condition variable that is signaled every time this condition is true. However, I think this will add a lot of complexity, and it requires modification to the transaction manager to know the `commit_time`. I'm not sure if it is worth doing that, or the busy waiting is actually acceptable. 

================

I noticed that `OldestTransactionStartTime` requires grabbing a latch. If so, perhaps using a conditional variable is the way to go.

test/storage/index_manager_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+
+  // Commit the setting transaction
+  txn_manager_->Commit(txn0, TestCallbacks::EmptyCallback, nullptr);
+
+  // Set up index attributes and key attributes
+  std::vector<std::string> index_attrs{"sex", "id", "name"};
+  std::vector<std::string> key_attrs{"id", "name"};
+
+  // Create the index
+  auto index_oid = index_manager_->CreateConcurrently(terrier_oid, ns_oid, table_oid, parser::IndexType::BWTREE, false,
```
DarkForte: When you are calling `CreateConcurrently`, there is no active txns running, so you didn't actually test whether `CreateConcurrently` successfully build an index on its visible version. At least, I would suggest running `CreateConcurrently` when there are some active txns.

test/storage/index_manager_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+  EXPECT_GT(!index_oid, 0);
+  auto index_handle = db_handle.GetIndexHandle(txn1, terrier_oid);
+  auto index_entry = index_handle.GetIndexEntry(txn1, index_oid);
+  EXPECT_NE(index_entry, nullptr);
+  auto ret = index_entry->GetIntegerColumn("indexrelid");
+  EXPECT_EQ(ret, !index_oid);
+  EXPECT_EQ(index_entry->GetIntegerColumn("indrelid"), !table_oid);
+  EXPECT_EQ(index_entry->GetBooleanColumn("indisready"), false);
+  EXPECT_EQ(index_entry->GetBooleanColumn("indisvalid"), true);
+  txn_manager_->Commit(txn1, TestCallbacks::EmptyCallback, nullptr);
```
DarkForte: There is no check on whether the index itself is correct. As said above, I think a better test case is to start some active txns, then run `CreateConcurrently` and check if it builds the correct index without blocking the active txns. The current test case does not show the feature of `CreateConcurrently`.

test/storage/index_builder_test.cpp
```diff
@@ -0,0 +1,179 @@
...
+  // Insert a few rows into the table
+  auto ptr = table_handle.GetTable(txn_, "test_table");
+  EXPECT_EQ(ptr, table);
+
+  for (int i = 0; i < 100; ++i) {
+    std::vector<type::TransientValue> row;
+    std::ostringstream stringStream;
+    stringStream << "name_" << i;
+    std::string name_string = stringStream.str();
+    stringStream << "address_" << i;
```
DarkForte: As said in the previous code review, I suggest using `fmt::format` instead of `stringstream`. since it is more straightforward.
And there are some other unaddressed code review comments in this file too. Please have a look at them.

src/storage/index/index_manager.cpp
```diff
@@ -0,0 +1,189 @@
...
+  IndexFactory index_factory;
+  ConstraintType constraint = (unique_index) ? ConstraintType::UNIQUE : ConstraintType::DEFAULT;
+  index_factory.SetOid(index_oid);
+  index_factory.SetConstraintType(constraint);
+
+  // Setup the key schema for the index
+  IndexKeySchema key_schema;
+  for (const auto &key_name : key_attrs) {
+    // Get the corresponding column
+    auto col = sql_table->GetSchema().GetColumn(key_name);
```
DarkForte: I totally agree with Yuze that someone should catch the out-of-range exception thrown by `GetColumn` because a wrong column name specified by a user should not kill the entire server.

src/storage/index/index_manager.cpp
```diff
@@ -0,0 +1,189 @@
...
+  txn_mgr->Commit(txn2, TestCallbacks::EmptyCallback, nullptr);
+  // FIXME(xueyuanz): Delete the txns to pass the test, since the GC is disabled.
+  delete txn1;
+  delete txn2;
+  return index_oid;
+}
+
+void IndexManager::Drop(catalog::db_oid_t db_oid, catalog::namespace_oid_t ns_oid, catalog::table_oid_t table_oid,
+                        catalog::index_oid_t index_oid, const std::string &index_name,
+                        transaction::TransactionManager *txn_mgr, catalog::Catalog *catalog) {
```
DarkForte: `index_name` is not used in this function, consider removing it?

src/include/storage/index/design_doc.md
```diff
@@ -0,0 +1,69 @@
...
+
+### Non-Blocking manner
+
+The idea of creating index without blocking modifications on the table comes from [Postgres](https://www.postgresql.org/docs/11/sql-createindex.html). We use the similar idea here with small modifications . The creation of an index will be completed in two transactions:
+* The first transaction adds a record corresponding to the index into the system catalogs and sets the status of index to be ```READY```(allow insert/update but does not allow deletes nor queries)
+* The second transaction takes the scan of the whole target table and builds the index by adding all entries of its snapshot into the index. After that the transaction sets the status of the index in system catalogs to be ```VALID``` and commits.
+
+Note that before the start of the second transaction, the building process should wait for all running transactions to terminate, whose timestamps are smaller than the commit timestamp of the first transaction in the building process (the transaction adding a new entry on index into catalog).
+
```
DarkForte: Why do you need to separate the two phases into two transactions? An alternative way to do it is to start one transaction, which first inserts into the catalog, and then builds the index on its visible version. What is the problem of this simpler method?

test/storage/index_builder_test.cpp
```diff
@@ -0,0 +1,179 @@
...
+        key->ColumnIds()[i] = sql_table_cols[i];
+      }
+    }
+  }
+  delete[] key_buf_index;
+  delete[] key_buf;
+  delete index_0;
+  delete table;
+}
+}  // namespace terrier::storage::index
```
DarkForte: You mentioned an edge case in your design document where a key could be inserted twice into the index. Therefore, there is supposed to be a test case on that in your unit test.

test/storage/index_manager_test.cpp
```diff
@@ -0,0 +1,117 @@
...
+  EXPECT_EQ(index_entry, nullptr);
+  txn_manager_->Commit(txn2, TestCallbacks::EmptyCallback, nullptr);
+
+  delete table;
+  delete catalog_;
+  delete txn0;
+  delete txn1;
+  delete txn2;
+}
+}  // namespace terrier::storage::index
```
DarkForte: You mentioned an edge case in your design document where a key could be inserted twice into the index. Therefore, there is supposed to be a test case on that in your unit test.

test/catalog/index_handle_test.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+  catalog::Catalog *catalog_;
+  storage::RecordBufferSegmentPool buffer_pool_{100, 100};
+
+  transaction::TransactionContext *txn_;
+  transaction::TransactionManager *txn_manager_;
+};
+
+// Tests that we can correctly add new index entry to the pg_index and correctly get the index entry by the index oid.
+// NOLINTNEXTLINE
```
YuzeLiao: [Major] Probably need another test to test `DeleteEntry()`

src/include/catalog/index_handle.h
```diff
@@ -0,0 +1,100 @@
...
+   * @param txn the transaction context
+   * @param entry the target entry
+   * @return true if successfull otherwise false
+   */
+  bool DeleteEntry(transaction::TransactionContext *txn, const std::shared_ptr<IndexEntry> &entry);
+
+  /**
+   * Create storage table
+   */
+  static catalog::SqlTableHelper *Create(transaction::TransactionContext *txn, Catalog *catalog, db_oid_t db_oid,
```
YuzeLiao: [Major] It's probably better to document the parameters and return value of this function in detail.

src/storage/index/index_factory.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+  // figure out if we can use CompactIntsKey
+  bool use_compact_ints = true;
+  uint32_t key_size = 0;
+
+  for (uint16_t i = 0; use_compact_ints && i < key_schema_.size(); i++) {
+    const auto &attr = key_schema_[i];
+    use_compact_ints = use_compact_ints && !attr.IsNullable() && CompactIntsOk(attr.GetType());  // key type ok?
+    key_size += type::TypeUtil::GetTypeSize(attr.GetType());
+    use_compact_ints = use_compact_ints && key_size <= sizeof(uint64_t) * INTSKEY_MAX_SLOTS;  // key size fits?
```
YuzeLiao: [Minor] Hope you don't mind that I feel like these three lines are a little bit cumbersome. But it's a really tiny problem :)

test/storage/index_builder_test.cpp
```diff
@@ -0,0 +1,179 @@
...
+  catalog::Catalog *catalog_;
+  storage::RecordBufferSegmentPool buffer_pool_{100, 100};
+
+  transaction::TransactionContext *txn_;
+  transaction::TransactionManager *txn_manager_;
+  IndexManager *index_manager_;
+};
+
+// Check the basic correctness of index populator
+// NOLINTNEXTLINE
```
YuzeLiao: [Major] Now that `index_builder` has been changed to `index_manager`, this test, which tests `PopulateIndex()`, ought to be integrated into `index_manager_test`.

src/storage/index/index_manager.cpp
```diff
@@ -0,0 +1,189 @@
...
+      key_schema.emplace_back(IndexKeyColumn(col_oid, col_type, col.GetNullable()));
+  }
+  index_factory.SetKeySchema(key_schema);
+
+  // Build an empty index
+  return index_factory.Build();
+}
+
+catalog::index_oid_t IndexManager::CreateConcurrently(catalog::db_oid_t db_oid, catalog::namespace_oid_t ns_oid,
+                                                      catalog::table_oid_t table_oid, parser::IndexType index_type,
```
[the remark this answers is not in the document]
yeshengm: Fixed, thanks!

src/storage/index/index_factory.cpp
```diff
@@ -0,0 +1,77 @@
+#include "storage/index/index_factory.h"
+#include <utility>
+
+namespace terrier::storage::index {
+Index *IndexFactory::Build() const {
+  TERRIER_ASSERT(!key_schema_.empty(), "Cannot build an index without a KeySchema.");
+  TERRIER_ASSERT(constraint_type_ != ConstraintType::INVALID, "Cannot build an index without a ConstraintType.");
```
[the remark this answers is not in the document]
yeshengm: Actually this is Wan's code. I think this code is fine, which uses builder pattern (though not used that often in C++ code).

src/storage/index/index_manager.cpp
yeshengm: This is to be compatible with plan node arguments. Ideally this oid should be known at binding phase, but I think we can use this name as a sanity check here.

src/include/storage/index/design_doc.md
yeshengm: Cuz during this period of time, other txns might also be working on the same table, but they cannot see that invisible index catalog entry. The index's content has to catch up with the most recent txns, but it is never going to happen if there's only one txn for building indexes.

test/storage/index_manager_test.cpp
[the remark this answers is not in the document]
Calculuser: Thanks for your comments. We will add them later.

test/storage/index_manager_test.cpp
Calculuser: Thanks for your comments. We will add it later.

test/storage/index_manager_test.cpp
Calculuser: That is actually the concurrent test. We will add it later.

src/storage/index/index_factory.cpp
Calculuser: Actually, it is Wen's code.

src/storage/index/index_manager.cpp
```diff
@@ -0,0 +1,189 @@
...
+  // Add IndexEntry
+  index_handle.AddEntry(txn1, index, index_oid, table_oid, indnatts, indnkeyatts, indisunique, indisprimary, indisvalid,
+                        indisready, indislive);
+
+  // initialize the building flag to false
+  auto index_id = make_index_id(db_oid, ns_oid, index_oid);
+  SetIndexBuildingFlag(index_id, false);
+
+  // Commit first transaction
+  transaction::timestamp_t commit_time = txn_mgr->Commit(txn1, TestCallbacks::EmptyCallback, nullptr);
```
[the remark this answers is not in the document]
yeshengm: Fixed.

src/include/storage/index/index_manager.h
```diff
@@ -0,0 +1,150 @@
...
+  void SetIndexBuildingFlag(const index_id_t &key, bool value) {
+    common::SpinLatch::ScopedSpinLatch guard(&index_building_map_latch_);
+    index_building_map_[key] = value;
+  }
+
+  /**
+   * Get the status of index as building in the flag
+   *
+   * @param key the index_id representing the index
+   * @return true if the index is in the process of building otherwise false
```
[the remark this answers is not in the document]
yeshengm: Good point! Resolved.

src/storage/index/index_factory.cpp
yeshengm: "Wan"

test/catalog/index_handle_test.cpp
yeshengm: Added.

src/storage/index/index_manager.cpp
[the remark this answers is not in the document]
yeshengm: Fixed by catching in `CreateConcurrently`.

src/storage/index/index_manager.cpp
```diff
@@ -0,0 +1,189 @@
...
+  auto indnatts = static_cast<int32_t>(index_attrs.size());
+  auto indnkeyatts = static_cast<int32_t>(key_attrs.size());
+  bool indisunique = unique_index;
+  bool indisprimary = false;
+  bool indisvalid = false;
+  bool indisready = true;
+  bool indislive = false;
+
+  // Intialize the index
+  Index *index = GetEmptyIndex(txn1, db_oid, index_oid, sql_table.get(), indisunique, key_attrs, catalog);
```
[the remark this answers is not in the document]
yeshengm: Fixed.

src/storage/index/index_manager.cpp
Calculuser: Fixed! Thank you.

test/storage/index_manager_test.cpp
Calculuser: Fixed! Thank you.

test/storage/index_builder_test.cpp
Calculuser: We decide to keep the test for populating index. The reason is that if someone in the future changes the implementation of populating index, they can reuse our test to check the correctness of their implementation.

src/storage/index/index_manager.cpp
```diff
@@ -0,0 +1,189 @@
...
+  // Initializing the index fails
+  if (index == nullptr) {
```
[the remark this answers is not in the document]
Calculuser: Yes, you are right. Thank you.

src/include/catalog/index_handle.h
```diff
@@ -0,0 +1,98 @@
...
+
+#include <type/transient_value.h>
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "catalog/catalog_defs.h"
+#include "catalog/catalog_sql_table.h"
+#include "catalog/namespace_handle.h"
```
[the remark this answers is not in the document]
yeshengm: Fixed.

src/include/catalog/index_handle.h
```diff
@@ -0,0 +1,98 @@
...
+
+namespace terrier::catalog {
+
+class Catalog;
+/**
+ * An IndexHandle contains the information about indexes.
+ *
+ * pg_index:
+ *  indexrelid | indrelid | indnatts | indnkeyatts | indisunique | indisprimary |  indisvalid | indisready | indislive
+ */
```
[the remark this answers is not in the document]
yeshengm: Added.

src/storage/index/index_manager.cpp
```diff
@@ -0,0 +1,193 @@
...
+
+Index *IndexManager::GetEmptyIndex(catalog::index_oid_t index_oid, SqlTable *sql_table, bool unique_index,
+                                   const std::vector<std::string> &key_attrs) {
+  // Setup the oid and constraint type for the index
+  IndexFactory index_factory;
+  ConstraintType constraint = (unique_index) ? ConstraintType::UNIQUE : ConstraintType::DEFAULT;
+  index_factory.SetOid(index_oid);
+  index_factory.SetConstraintType(constraint);
+
+  // Setup the key schema for the index
```
mbutrovich: This logic assumes that you're indexing on columns, but you can build an index on expressions as well. It's probably too late to change this implementation now, but I wanted to point it out.

yeshengm: Do we have infrastructure to evaluate expressions now? We can definitely implement this if time permits.

src/include/storage/index/index_populator.h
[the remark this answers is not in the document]
lagoon0o0: We have enforced the uniqueness constraint now.

test/storage/index_populator_test.cpp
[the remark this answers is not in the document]
lagoon0o0: Refactored.

src/storage/lock_manager.cpp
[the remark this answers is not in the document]
lagoon0o0: The lock manager is deprecated.

src/catalog/index_handle.cpp
```diff
@@ -0,0 +1,82 @@
...
+                                                                    index_oid_t oid) {
+  std::vector<type::TransientValue> search_vec, ret_row;
+  search_vec.push_back(type::TransientValueFactory::GetInteger(!oid));
+  ret_row = pg_index_rw_->FindRow(txn, search_vec);
+  return std::make_shared<IndexEntry>(oid, std::move(ret_row));
+}
+
+void IndexHandle::AddEntry(transaction::TransactionContext *txn, index_oid_t indexrelid, table_oid_t indrelid,
+                           int32_t indnatts, int32_t indnkeyatts, bool indisunique, bool indisprimary, bool indisvalid,
+                           bool indisready, bool indislive) {
```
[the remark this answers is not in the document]
lagoon0o0: Thanks for your advice!

test/storage/index_populator_test.cpp
```diff
@@ -0,0 +1,177 @@
...
+  auto ptr = table_handle.GetTable(txn_, "test_table");
+  EXPECT_EQ(ptr, table);
+
+  for (int i = 0; i < 100; ++i) {
+    std::vector<type::TransientValue> row;
+    std::ostringstream stringStream;
+    stringStream << "name_" << i;
+    std::string name_string = stringStream.str();
+    stringStream << "address_" << i;
+    std::string address_string = stringStream.str();
```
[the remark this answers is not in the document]
lagoon0o0: Thanks!

test/storage/index_populator_test.cpp
lagoon0o0: No, since the index is not multi-versioned.

src/include/storage/index/index_manager.h
```diff
@@ -0,0 +1,150 @@
...
+   * Record the oid of db, the oid of namespace and the oid of index.
+   */
+  // FIXME(jiaqizuo): The data struture is not good and should use more clear data structure later
+  using index_id_t = std::pair<std::pair<catalog::db_oid_t, catalog::namespace_oid_t>, catalog::index_oid_t>;
+
+  /**
+   * A map from index_id to the status of the index (in the process of building or not)
+   */
+  std::map<index_id_t, bool> index_building_map_;
+  // FIXME(xueyuanz): This latch might not be necessary, the index_builing_map_ is also guarded by the commit_latch.
```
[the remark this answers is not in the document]
Calculuser: FIxed. Thank you for your comment.

PR #338, opened 2019-04-06
Metric Collection Infrastructure

wenxuanqiu:
# Metric Collection Infrastructure

## Overview

>We plan to build an infrastructure for collecting internal statistics during execution in the CMU DBMS. This infrastructure allows data to be collected at multiple levels including database, index, table, and tuple levels. Examples of data that can be collected are CPU usage, memory usage, and transaction latency. The collected statistics are stored in SQL tables, which can be accessed from the catalog and persisted like other application defined tables. Statistics collected are expected to be used to achieve autonomous self-driving.

>To realize above functions, we construct a metric collection framework that is extensible and relatively independent of the database management system. Specifically, our framework has multi-level data structures: different event types such as transaction commit, different metric classes such as metric handler (class Metric) and the actual data holders (class RawData) in metric classes. As the following figure demonstrates: 1) Metrics are registered to corresponding events in initialization. 2) When an event is triggered at the running time, related metric handlers are called to apply for latch to access the data. 3) The actual data holders update the values of the variables. 

>At the same time, a background aggregator regularly contacts worker threads to swap out their collected data in all data holders and then aggregate them. The aggregated data will be updated and persisted in metric catalog tables.

data_flow

> We have considered other implementations. For example, the most direct way is to construct several data holders for different statistics and insert update code in database management system where some statistics are changed. Though the method is simple, it is difficult to maintain and extend.


src/include/stats/abstract_metric.h
```diff
@@ -0,0 +1,310 @@
...
+
+  /**
+   * @param txn context of the transaction committing
+   * @param database_oid OID fo the database where the txn happens.
+   */
+  virtual void OnTransactionCommit(const transaction::TransactionContext *txn, catalog::db_oid_t database_oid) {}
+
+  /**
+   * @param txn context of the transaction aborting
+   * @param database_oid OID fo the database where the txn happens.
```
thepulkitagarwal: Typo in `for`

src/include/stats/abstract_metric.h
```diff
@@ -0,0 +1,310 @@
...
+  virtual ~Metric() = default;
+
+  /**
+   * @param txn context of the transaction beginning
+   */
+  virtual void OnTransactionBegin(const transaction::TransactionContext *txn) {}
+
+  /**
+   * @param txn context of the transaction committing
+   * @param database_oid OID fo the database where the txn happens.
```
thepulkitagarwal: Typo. `fo` -> `of`

src/include/stats/abstract_metric.h
```diff
@@ -0,0 +1,310 @@
...
+    return {raw_data_.load(), &safe_};
+  }
+
+ private:
+  /**
+   * Pointer to raw data
+   */
+  std::atomic<DataType *> raw_data_;
+  /**
+   * Indicate whether it is safe ti read the raw data, similar to a latch
```
thepulkitagarwal: Typo.
```suggestion
   * Indicate whether it is safe to read the raw data, similar to a latch
```

src/include/stats/abstract_metric.h
```diff
@@ -0,0 +1,310 @@
...
+ * by the database. @see StatsEventType for a list of available events.
+ * It is guaranteed that the appropriate callback method (identified by the
+ * naming pattern On[event name]) is invoked and the args filled out with
+ * relevant information. To enable safe and efficient collection of data,
+ * it is required that all data be collected be written to a RawData
+ * object, @see AbstractRawData.
+ *
+ * While you could write your own metric directly extending from this class,
+ * it is recommended that you use @see AbstractMetric class, which takes in
+ * an AbstractRawData class a a template argument and implements the tricky
```
huzaifaabbasi: Typo
```suggestion
 * an AbstractRawData class as a template argument and implements the tricky
```

src/include/stats/thread_level_stats_collector.h
```diff
@@ -0,0 +1,327 @@
...
+   * @param txn context of the transaction beginning
+   */
+  void CollectTransactionBegin(const transaction::TransactionContext *txn) {
+    for (auto &metric : metric_dispatch_[StatsEventType::TXN_BEGIN]) metric->OnTransactionBegin(txn);
+  }
+
+  /**
+   * Collector action on transaction commit
+   * @param txn context of the transaction committing
+   * @param database_oid OID fo the database where the txn happens.
```
thepulkitagarwal: Typo.
```suggestion
   * @param database_oid OID of the database where the txn happens.
```

src/include/stats/thread_level_stats_collector.h
```diff
@@ -0,0 +1,327 @@
...
+   */
+  void CollectTransactionCommit(const transaction::TransactionContext *txn, catalog::db_oid_t database_oid) {
+    for (auto &metric : metric_dispatch_[StatsEventType::TXN_COMMIT]) metric->OnTransactionCommit(txn, database_oid);
+  }
+
+  /**
+   * Collector action on transaction abort
+   * @param txn context of the transaction aborting
+   * @param database_oid OID fo the database where the txn happens.
```
thepulkitagarwal: Typo.
```suggestion
   * @param database_oid OID of the database where the txn happens.
```

src/include/stats/abstract_metric.h
```diff
@@ -0,0 +1,310 @@
...
+   * @param database_oid OID fo the database where the txn happens.
+   */
+  virtual void OnTransactionAbort(const transaction::TransactionContext *txn, catalog::db_oid_t database_oid) {}
+
+  /**
+   * @param txn context of the transaction performing read
+   * @param src database and table id pair that the tuple read happens
+   */
+  virtual void OnTupleRead(const transaction::TransactionContext *txn,
+                           std::pair<catalog::db_oid_t, catalog::table_oid_t> src) {}
```
utkarsh39: I think `pair` is an unnecessary level of indirection. As whenever this function is used in 
_thread_level_stats_collector.h_, an explicit pair has to be constructed. All functions in _thread_level_stats_collector.h_ are anyways getting the arguments separately.

src/include/stats/thread_level_stats_collector.h
```diff
@@ -0,0 +1,327 @@
...
+   * @param types A list of event types to receive updates about.
+   */
+  template <typename metric>
+  void RegisterMetric(std::vector<StatsEventType> types) {
+    auto m = std::make_shared<metric>();
+    metrics_.push_back(m);
+    for (StatsEventType type : types) metric_dispatch_[type].push_back(m);
+  }
+
+  using MetricList = std::vector<std::shared_ptr<Metric>>;
```
thepulkitagarwal: `List` could mean something else in C++ [http://www.cplusplus.com/reference/list/list/](http://www.cplusplus.com/reference/list/list/), so might be better to use something like `MetricVector`?

src/include/stats/statistic_defs.h
```diff
@@ -0,0 +1,75 @@
...
+  INDEX_MEMORY_USAGE,
+  INDEX_MEMORY_RECLAIM,
+  QUERY_BEGIN,
+  QUERY_END,
+  TEST  // Testing event
+};
+
+/**
+ * Hasher for enums
+ * @tparam E the enum class to be hashed
```
utkarsh39:
```suggestion
 * @param E the enum class to be hashed
```

src/include/stats/stats_aggregator.h
```diff
@@ -0,0 +1,52 @@
...
+   * combine with what was previously persisted in internal SQL tables
+   * and insert new total into SQLtable
+   */
+  void Aggregate();
+
+  /**
+   * Worker method for Aggregate() that performs stats collection
+   * @return raw data collected from all threads
+   */
+  std::vector<std::shared_ptr<AbstractRawData>> AggregateRawData();
```
utkarsh39: This function can be static

test/stats/stats_framework_test.cpp
```diff
@@ -0,0 +1,24 @@
...
+
+namespace terrier {
+
+/**
+ * @brief Test the overall correctness of the stats framework
+ */
+class StatsFrameworkTests : public TerrierTest {};
+
+/**
+ * Basic test for testing metric registrction and stats collection
```
utkarsh39:
```suggestion
 * Basic test for testing metric registration and stats collection
```

src/include/stats/database_metric.h
```diff
@@ -0,0 +1,94 @@
...
+
+#include "catalog/catalog_defs.h"
+#include "stats/abstract_metric.h"
+
+namespace terrier {
+
+namespace transaction {
+class TransactionContext;
+}  // namespace transaction
+
```
utkarsh39: Namespace transaction can be done away with.

src/include/stats/thread_level_stats_collector.h
```diff
@@ -0,0 +1,327 @@
...
+
+ private:
+  /**
+   * Registers a Metric so that its callbacks are invoked.
+   * Use this only in the constructor.
+   * @tparam metric type of Metric to register
+   * @param types A list of event types to receive updates about.
+   */
+  template <typename metric>
+  void RegisterMetric(std::vector<StatsEventType> types) {
```
huzaifaabbasi: types can be made a const reference

src/include/stats/abstract_metric.h
```diff
@@ -0,0 +1,310 @@
...
+#include "catalog/catalog_defs.h"
+#include "stats/abstract_raw_data.h"
+#include "stats/statistic_defs.h"
+
+namespace terrier {
+
+namespace transaction {
+class TransactionContext;
+}  // namespace transaction
+
```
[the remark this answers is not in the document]
wenxuanqiu: This is a forward declaration for methods of AbstractMetric, which have TransactionContext * as parameters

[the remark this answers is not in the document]
utkarsh39: Since you are using `transaction::TransactionContext *` as parameter types, I think this is not necessary. I tried compiling the code without this declaration and it compiled successfully

thepulkitagarwal:
This is an implementation question. 

There is a `Metric` class which has a few functions like `OnTupleRead`, `OnTupleDelete` etc. There is an enum `StatsEventType`, which has all events. Each function in the `Metric` class corresponds to a named value in the enum. Also, there is a class `ThreadLevelStatsCollector` with a collect method for every `StatsEventType` like `CollectTupleRead`, `CollectTupleDelete` etc.

Why was the choice taken to create a function for each event and not a common function like `OnEvent(StatsEventType event_type, ..)` in the `Metric` class, and `CollectEvent(StatsEventType event_type, ..)` in `ThreadLevelStatsCollector`? I guess one of the reasons could be speed, but I feel that a switch statement would cause a negligible overhead.

wenxuanqiu:
> This is an implementation question.
> 
> There is a `Metric` class which has a few functions like `OnTupleRead`, `OnTupleDelete` etc. There is an enum `StatsEventType`, which has all events. Each function in the `Metric` class corresponds to a named value in the enum. Also, there is a class `ThreadLevelStatsCollector` with a collect method for every `StatsEventType` like `CollectTupleRead`, `CollectTupleDelete` etc.
> 
> Why was the choice taken to create a function for each event and not a common function like `OnEvent(StatsEventType event_type, ..)` in the `Metric` class, and `CollectEvent(StatsEventType event_type, ..)` in `ThreadLevelStatsCollector`? I guess one of the reasons could be speed, but I feel that a switch statement would cause a negligible overhead.

Hi! Thank you very much for the suggestion. This design is more for maintainability than performance. Suppose a user want to make use of the stats collection framework and they simply call `OnEvent(StatsEventType event_type, ...)`, what exact arguments do they pass in? They have to supply all the different possible arguments for all stats event. It might be possible that they miss match the parameters and pass in a database_oid_t as index_oid_t (underneath they are all uint32_t). They have to rely on a switch statement to perform a type check. This can cause a runtime error when it could have been avoided at compile time with simpler function signatures. Moreover, with a switch case model, legacy code can remain in the system undetected. For example, if tuple level statistics are no longer needed, deleting the corresponding method at any level would force the developer to delete all traces of the related methods, otherwise a compilation error would occur. With switch case statements, legacy events can fall into the default branch and remain undetected.

I certainly agree that the collector methods  have potential to be combined and look more succinct and elegant, but they bring larger maintenance burden. I hope I addressed your concern.

thepulkitagarwal:
> Hi! Thank you very much for the suggestion. This design is more for maintainability than performance. Suppose a user want to make use of the stats collection framework and they simply call `OnEvent(StatsEventType event_type, ...)`, what exact arguments do they pass in? They have to supply all the different possible arguments for all stats event. It might be possible that they miss match the parameters and pass in a database_oid_t as index_oid_t (underneath they are all uint32_t). They have to rely on a switch statement to perform a type check. This can cause a runtime error when it could have been avoided at compile time with simpler function signatures. Moreover, with a switch case model, legacy code can remain in the system undetected. For example, if tuple level statistics are no longer needed, deleting the corresponding method at any level would force the developer to delete all traces of the related methods, otherwise a compilation error would occur. With switch case statements, legacy events can fall into the default branch and remain undetected.
> 
> I certainly agree that the collector methods have potential to be combined and look more succinct and elegant, but they bring larger maintenance burden. I hope I addressed your concern.

I believe that the `database_oid_t`and the `index_oid_t` issue should be taken care of by `STRONG_TYPEDEF` which should give a compile time error in case of mismatch. Even with the switch statement, deleting the corresponding event from the `StatsEventType` should force the developer to delete the corresponding code from the places that particular event was referenced.

But maybe there would be a larger maintenance burden because of this, so it makes more sense to have the current approach of having a method for every event.

benchmark/storage/metric_benchmark.cpp
```diff
@@ -0,0 +1,165 @@
...
+#include "transaction/transaction_defs.h"
+#include "transaction/transaction_manager.h"
+#include "util/metric_test_util.h"
+#include "util/test_harness.h"
+#include "util/transaction_test_util.h"
+
+namespace terrier {
+
+// This benchmark measures the throughput of the Aggregator and the Collecter in the metric collection pipeline.
+
```
huzaifaabbasi: nit: extra newline before class declaration

benchmark/storage/metric_benchmark.cpp
```diff
@@ -0,0 +1,165 @@
...
+    auto table = table_handle.GetTable(txn_, "database_metric_table");
+    delete table;
+    db_handle = catalog_->GetDatabaseHandle();
+    table_handle = db_handle.GetNamespaceHandle(txn_, terrier_oid).GetTableHandle(txn_, "public");
+    table = table_handle.GetTable(txn_, "txn_metric_table");
+    delete table;
+    delete catalog_;  // need to delete catalog_first
+    delete txn_manager_;
+  }
+  settings::SettingsManager *settings_manager_;
```
thepulkitagarwal: nit: There should be a newline after the function definition ends

benchmark/storage/metric_benchmark.cpp
```diff
@@ -0,0 +1,165 @@
...
+    const catalog::db_oid_t terrier_oid(catalog::DEFAULT_DATABASE_OID);
+    auto db_handle = catalog_->GetDatabaseHandle();
+    auto table_handle = db_handle.GetNamespaceHandle(txn_, terrier_oid).GetTableHandle(txn_, "public");
+    auto table = table_handle.GetTable(txn_, "database_metric_table");
+    delete table;
+    db_handle = catalog_->GetDatabaseHandle();
+    table_handle = db_handle.GetNamespaceHandle(txn_, terrier_oid).GetTableHandle(txn_, "public");
+    table = table_handle.GetTable(txn_, "txn_metric_table");
+    delete table;
+    delete catalog_;  // need to delete catalog_first
```
thepulkitagarwal: nit: spacing
```suggestion
    delete catalog_;  // need to delete catalog_ first
```

utkarsh39:
There was a discussion about merging a few functions earlier:
1. [thepulkitagarwal's comment](https://github.com/cmu-db/terrier/pull/338#issuecomment-483296495)
2. [wenxuanqiu's reply](https://github.com/cmu-db/terrier/pull/338#issuecomment-483307109)
3. [thepulkitagarwal's comment](https://github.com/cmu-db/terrier/pull/338#issuecomment-483312042)

I don't quite agree with the reasoning that it shouldn't be done from a maintainability stand point. I would argue that less code is more maintainable code, and so we should merge the functions wherever possible. To put it in exact terms, I believe that merging the functions in the `Metric` and the `ThreadLevelStatsCollector` classes with a switch statement would be a good idea.

src/include/storage/metric/abstract_metric.h
```diff
@@ -0,0 +1,354 @@
...
+
+#include "catalog/catalog_defs.h"
+#include "storage/metric/abstract_raw_data.h"
+#include "storage/metric/metric_defs.h"
+
+namespace terrier {
+
+namespace transaction {
+class TransactionContext;
+}  // namespace transaction
```
huzaifaabbasi: Is there a reason that you did this declaration here? I believe that this is unnecessary and it should compile without this.

src/include/storage/metric/thread_level_stats_collector.h
```diff
@@ -0,0 +1,372 @@
...
+  using CollectorsMap = common::ConcurrentMap<std::thread::id, ThreadLevelStatsCollector *, std::hash<std::thread::id>>;
+
+  /**
+   * @return the Collector for the calling thread
+   */
+  static ThreadLevelStatsCollector *GetCollectorForThread() {
+    std::thread::id tid = std::this_thread::get_id();
+    auto const iter = collector_map_.Find(tid);
+    if (iter == collector_map_.End()) {
+      return new ThreadLevelStatsCollector();
```
thepulkitagarwal: `new ThreadLevelStatsCollector()` has the side effect of adding the created `ThreadLevelStatsCollector` to the `collector_map_`. This is not clear from the usage here. Would it be better to remove it as side effect and explicitly add it to the `collector_map_` here? Or to note this as a comment?

src/include/storage/metric/thread_level_stats_collector.h
```diff
@@ -0,0 +1,372 @@
...
+  /**
+   * Collection action on table memory reclaim
+   * @param database_oid OID of the database that the table memory reclaim happens
+   * @param namespace_oid OID of the namespace that the table memory reclaim happens
+   * @param table_oid OID of the table that the table memory reclaim happens
+   * @param size number of bytes being reclaim
+   */
+  void CollectTableMemoryReclaim(catalog::db_oid_t database_oid, catalog::namespace_oid_t namespace_oid,
+                                 catalog::table_oid_t table_oid, size_t size) {
+    for (auto &metric : metric_dispatch_[StatsEventType::TABLE_MEMORY_FREE])
```
huzaifaabbasi: Is `TABLE_MEMORY_FREE` used here for a reason? There is no `TABLE_MEMORY_RECLAIM` in the `StatsEventType` enum

src/include/storage/metric/thread_level_stats_collector.h
```diff
@@ -0,0 +1,372 @@
...
+  /**
+   * Collector action on table memory usage
+   * @param database_oid OID of the database that the table memory usage happens
+   * @param namespace_oid OID of the namespace that the table memory usage happens
+   * @param table_oid OID of the table that the table memory usage happens
+   * @param size number of bytes being used
+   */
+  void CollectTableMemoryUsage(catalog::db_oid_t database_oid, catalog::namespace_oid_t namespace_oid,
+                               catalog::table_oid_t table_oid, size_t size) {
+    for (auto &metric : metric_dispatch_[StatsEventType::TABLE_MEMORY_ALLOC])
```
huzaifaabbasi: Is `TABLE_MEMORY_ALLOC` used for a reason here? In the `StatsEventType` enum there's an `INDEX_MEMORY_USAGE` but no `TABLE_MEMORY_USAGE`

thepulkitagarwal:
> There was a discussion about merging a few functions earlier:
> 
> 1. [thepulkitagarwal's comment](https://github.com/cmu-db/terrier/pull/338#issuecomment-483296495)
> 2. [wenxuanqiu's reply](https://github.com/cmu-db/terrier/pull/338#issuecomment-483307109)
> 3. [thepulkitagarwal's comment](https://github.com/cmu-db/terrier/pull/338#issuecomment-483312042)
> 
> I don't quite agree with the reasoning that it shouldn't be done from a maintainability stand point. I would argue that less code is more maintainable code, and so we should merge the functions wherever possible. To put it in exact terms, I believe that merging the functions in the `Metric` and the `ThreadLevelStatsCollector` classes with a switch statement would be a good idea.

To add to this, @huzaifaabbasi found some functions which seem like they have a mismatch (like [this](https://github.com/cmu-db/terrier/pull/338#discussion_r283079819)) - which if true, furthers the point that the functions should be merged.

src/include/storage/metric/transaction_metric.h
```diff
@@ -0,0 +1,236 @@
...
+    (void)namespace_oid;
+    (void)table_oid;
+    GetRawData()->IncrementTupleRead(txn);
+  }
+
+  /**
+   * @param txn context of the transaction performing update
+   * @param database_oid OID of the database that the tuple delete happens
+   * @param namespace_oid OID of the namespace that the tuple delete happens
+   * @param table_oid OID of the table that the tuple delete happens
```
utkarsh39: This and the above parameter comments should say `update` 
```suggestion
   * @param table_oid OID of the table that the tuple update happens
```

I believe it would be better to just have this:
``` suggestion
   * @param table_oid OID of the table
```
And before  any `@param`'s add a comment saying that this function performs actions on tuple updates

src/include/storage/metric/transaction_metric.h
```diff
@@ -0,0 +1,236 @@
...
+    (void)namespace_oid;
+    (void)table_oid;
+    GetRawData()->IncrementTupleUpdate(txn);
+  }
+
+  /**
+   * @param txn context of the transaction performing insert
+   * @param database_oid OID of the database that the tuple delete happens
+   * @param namespace_oid OID of the namespace that the tuple delete happens
+   * @param table_oid OID of the table that the tuple delete happens
```
utkarsh39: Same as [previous comment](https://github.com/cmu-db/terrier/pull/338/files#r283080140). Should be `insert`

src/storage/metric/transaction_metric.cpp
```diff
@@ -0,0 +1,68 @@
...
+  auto table = GetStatsTable(txn_manager, catalog, txn);
+  TERRIER_ASSERT(table != nullptr, "Stats table cannot be nullptr.");
+  std::vector<type::TransientValue> row;
+
+  for (auto &entry : data_) {
+    // one iteration per transaction
+    auto txn_oid = static_cast<uint64_t>(entry.first);
+    auto &counter = entry.second;
+    auto latency = counter.latency_;
+    auto tuple_read = counter.tuple_read_;
```
thepulkitagarwal: This (and similarly for the other `counter` attributes) is a temporary variable only being used in one place in an `if` block ([here](https://github.com/cmu-db/terrier/pull/338/files#diff-e6e2930861c783ee7a094b005eaa7531R56)). Wouldn't it be better to use these directly in the `if` block with scope only in that block?

test/storage/metric_test.cpp
```diff
@@ -0,0 +1,516 @@
...
+    const catalog::db_oid_t terrier_oid(catalog::DEFAULT_DATABASE_OID);
+    auto db_handle = catalog_->GetDatabaseHandle();
+    auto table_handle = db_handle.GetNamespaceHandle(txn_, terrier_oid).GetTableHandle(txn_, default_namespace_);
+    auto table = table_handle.GetTable(txn_, database_metric_table_);
+    delete table;
+    db_handle = catalog_->GetDatabaseHandle();
+    table_handle = db_handle.GetNamespaceHandle(txn_, terrier_oid).GetTableHandle(txn_, default_namespace_);
+    table = table_handle.GetTable(txn_, txn_metric_table_);
+    delete table;
+    delete catalog_;  // need to delete catalog_first
```
thepulkitagarwal: nit: spacing
```suggestion
    delete catalog_;  // need to delete catalog_ first
```

test/storage/metric_test.cpp
```diff
@@ -0,0 +1,516 @@
...
+    auto num_insert = static_cast<uint8_t>(std::uniform_int_distribution<uint8_t>(1, UINT8_MAX)(generator_));
+    auto num_delete = static_cast<uint8_t>(std::uniform_int_distribution<uint8_t>(1, UINT8_MAX)(generator_));
+
+    auto workload = [&](uint32_t id) {
+      // NOTICE: thread level collector must be alive while aggregating
+      if (id == 0) {  // aggregator thread
+        std::this_thread::sleep_for(aggr_period_);
+        aggregator.Aggregate(txn_);
+        auto iter = collectors.Begin();
+        for (; iter != collectors.End(); iter++) {
```
huzaifaabbasi: `auto iter = collectors.Begin();` can be a part of the for loop. It is not being used anywhere else.

benchmark/storage/metric_benchmark.cpp
```diff
@@ -0,0 +1,165 @@
+#include <memory>
+#include <string>
+#include <thread>  //NOLINT
```
utkarsh39: Don't include `thread`. It is included implicitly. Check `logging_benchmark.cpp`

wenxuanqiu:
> There was a discussion about merging a few functions earlier:
> 
> 1. [thepulkitagarwal's comment](https://github.com/cmu-db/terrier/pull/338#issuecomment-483296495)
> 2. [wenxuanqiu's reply](https://github.com/cmu-db/terrier/pull/338#issuecomment-483307109)
> 3. [thepulkitagarwal's comment](https://github.com/cmu-db/terrier/pull/338#issuecomment-483312042)
> 
> I don't quite agree with the reasoning that it shouldn't be done from a maintainability stand point. I would argue that less code is more maintainable code, and so we should merge the functions wherever possible. To put it in exact terms, I believe that merging the functions in the `Metric` and the `ThreadLevelStatsCollector` classes with a switch statement would be a good idea.



> > There was a discussion about merging a few functions earlier:
> > 
> > 1. [thepulkitagarwal's comment](https://github.com/cmu-db/terrier/pull/338#issuecomment-483296495)
> > 2. [wenxuanqiu's reply](https://github.com/cmu-db/terrier/pull/338#issuecomment-483307109)
> > 3. [thepulkitagarwal's comment](https://github.com/cmu-db/terrier/pull/338#issuecomment-483312042)
> > 
> > I don't quite agree with the reasoning that it shouldn't be done from a maintainability stand point. I would argue that less code is more maintainable code, and so we should merge the functions wherever possible. To put it in exact terms, I believe that merging the functions in the `Metric` and the `ThreadLevelStatsCollector` classes with a switch statement would be a good idea.
> 
> To add to this, @huzaifaabbasi found some functions which seem like they have a mismatch (like [this](https://github.com/cmu-db/terrier/pull/338#discussion_r283079819)) - which if true, furthers the point that the functions should be merged.

Hi, thank you for your suggestion. I understand your concern and I agree as well that less code is easier to maintain. However, different collector actions might need different parameters. The current parameters look very similarly because they are what we think that might be useful for each event. With a single method, I am not sure how we can provide the flexibility of different parameters. 
We can certainly pack all the OIDs into a single struct, but even that can have different variants in tabl)oid_t/index_oid_t/view_oid_t (in the future). What about parameters other than OIDs? One way is to pass a huge list of all possible parameters or have some kind of configuration object that contains all possible parameters and construct that object according to the StatsEvent being triggered. Yet if we do that, we are back to as if the methods are separated and more burden is on the user to construct that configuration object than providing a short list of relevant arguments.
I feel the case at hand is similar to plan nodes which are located under planner. The plan nodes have separate classes for Creating tables, indices, UDFs, and views. In Peloton, they used to be a gigantic plan node called CreatePlan. The work required to create such a giant plan work unhealthily massive. The similar argument applies to StatsEvents. Collector actions for tables, indices, UDFs, and views might be similar in nature and in parameters, but they in fact are fundamentally different.
As the needs of the system changes in the future, a single method might be too narrow of an interface as the actions for StatsEvents are bound to diverge.

src/include/storage/metric/database_metric.h
```diff
@@ -0,0 +1,121 @@
...
+
+#include "catalog/catalog.h"
+#include "catalog/catalog_defs.h"
+#include "storage/metric/abstract_metric.h"
+#include "type/transient_value_factory.h"
+#include "type/transient_value_peeker.h"
+
+namespace terrier {
+
+namespace storage::metric {
```
utkarsh39: can do `namespace terrier::storage::metric` instead

src/include/storage/metric/stats_aggregator.h
```diff
@@ -0,0 +1,95 @@
...
+#include "common/macros.h"
+#include "storage/metric/abstract_raw_data.h"
+#include "storage/metric/thread_level_stats_collector.h"
+#include "util/transaction_test_util.h"
+
+namespace terrier {
+
+namespace settings {
+class SettingsManager;
+}
```
utkarsh39: I think you can remove this declaration from here and can combine `namespace terrier::storage::metric`

src/include/storage/metric/thread_level_stats_collector.h
```diff
@@ -0,0 +1,368 @@
...
+   * Collector action for a test event
+   */
+  void CollectTestNum(int number) {
+    for (auto &metric : metric_dispatch_[StatsEventType::TEST]) metric->OnTest(number);
+  }
+
+  /**
+   * @return A vector of raw data, for each registered metric. Each piece of
+   * data is guaranteed to be safe to read and remove, and the same type of
+   * metric is guaranteed to be in the same positopn in the returned vector
```
utkarsh39:
```suggestion
   * metric is guaranteed to be in the same position in the returned vector
```

test/storage/metric_test.cpp
```diff
@@ -0,0 +1,515 @@
...
+#include "transaction/transaction_manager.h"
+#include "util/metric_test_util.h"
+#include "util/test_harness.h"
+#include "util/transaction_test_util.h"
+
+namespace terrier {
+
+namespace settings {
+class SettingsManager;
+}
```
utkarsh39: I think you can remove this namespace declaration from here.

test/catalog/tablespace_handle_test.cpp
```diff
@@ -0,0 +1,44 @@
...
+    catalog_ = new catalog::Catalog(txn_manager_, txn_);
+  }
+
+  void TearDown() override {
+    txn_manager_->Commit(txn_, TestCallbacks::EmptyCallback, nullptr);
+
+    TerrierTest::TearDown();
+    delete catalog_;  // delete catalog first
+    delete txn_manager_;
+    delete txn_;
```
huzaifaabbasi: Don't do this. Turn on the GC instead.

src/include/storage/metric/thread_level_stats_collector.h
wenxuanqiu: Corrected

src/include/storage/metric/thread_level_stats_collector.h
wenxuanqiu: Corrected

src/include/storage/metric/transaction_metric.h
wenxuanqiu: Corrected

src/include/storage/metric/transaction_metric.h
wenxuanqiu: Corrected

test/catalog/tablespace_handle_test.cpp
wenxuanqiu: This is part of the catalog PR we merged

test/storage/metric_test.cpp
wenxuanqiu: Settings manager is not yet merged, so there is not a def file we can include.

src/include/storage/metric/stats_aggregator.h
wenxuanqiu: Settings manager is not yet merged, so there is not a def file we can include.

src/include/storage/metric/database_metric.h
wenxuanqiu: Changed

benchmark/storage/metric_benchmark.cpp
wenxuanqiu: Changed

test/storage/metric_test.cpp
wenxuanqiu: Changed

benchmark/storage/metric_benchmark.cpp
```diff
@@ -0,0 +1,161 @@
...
+
+  std::thread gc_thread_;
+  storage::GarbageCollector *gc_ = nullptr;
+  volatile bool run_gc_ = false;
+  const std::chrono::milliseconds gc_period_{10};
+
+  const std::string default_namespace_{"public"};
+};
+
+//
```
thepulkitagarwal: nit: empty comment

src/include/storage/metric/abstract_metric.h
huzaifaabbasi: I ran `MetricTests` on my pc after removing this, and it compiled and the tests passed. I think you can remove `namespace transaction` and do `namespace terrier::storage::metric` instead.

src/include/storage/metric/test_metric.h
```diff
@@ -0,0 +1,81 @@
...
+  int value_;
+};
+
+/**
+ * @brief Interface around test raw data
+ */
+class TestMetric : public AbstractMetric<TestMetricRawData> {
+ public:
+  /**
+   *
```
huzaifaabbasi: nit: Empty Comment

src/include/storage/metric/thread_level_stats_collector.h
```diff
@@ -0,0 +1,368 @@
...
+#include "catalog/catalog_defs.h"
+#include "common/container/concurrent_map.h"
+#include "storage/metric/abstract_metric.h"
+#include "storage/metric/abstract_raw_data.h"
+#include "storage/metric/metric_defs.h"
+#include "transaction/transaction_manager.h"
+
+namespace terrier {
+
+namespace storage::metric {
```
huzaifaabbasi: Can do `namespace terrier::storage::metric` here

src/include/storage/metric/transaction_metric.h
```diff
@@ -0,0 +1,247 @@
...
+
+#include "catalog/catalog.h"
+#include "catalog/catalog_defs.h"
+#include "common/scoped_timer.h"
+#include "storage/metric/abstract_metric.h"
+#include "transaction/transaction_defs.h"
+
+namespace terrier {
+
+namespace storage::metric {
```
huzaifaabbasi: Can do `namespace terrier::storage::metric` here

src/include/storage/metric/transaction_metric.h
```diff
@@ -0,0 +1,247 @@
...
+   * @param txn transaction context of the beginning transaction
+   */
+  void OnTransactionBegin(const transaction::TransactionContext *txn) override { GetRawData()->SetTxnStart(txn); }
+
+  /**
+   * @param txn transaction context of the committing transaction
+   * @param database_oid OID of the database the transaction is running in
+   */
+  void OnTransactionCommit(const transaction::TransactionContext *txn, catalog::db_oid_t database_oid) override {
+    (void)database_oid;
```
utkarsh39: Can you add a comment stating why this is done?

src/include/storage/metric/abstract_metric.h
```diff
@@ -0,0 +1,354 @@
...
+
+namespace storage::metric {
```
tli2: Should this be under the storage namespace at all? Somehow I feel like you can just have a metric namespace

src/include/storage/metric/abstract_metric.h
```diff
@@ -0,0 +1,354 @@
...
+  virtual void OnIndexDelete(catalog::db_oid_t database_oid, catalog::namespace_oid_t namespace_oid,
+                             catalog::index_oid_t index_oid) {}
+
+  /**
+   * @param database_oid OID of the database that the table memory allocation happens
+   * @param namespace_oid OID of the namespace that the table memory allocation happens
+   * @param table_oid OID of the table that the table memory allocation happens
+   * @param size number of bytes being allocated
+   */
+  virtual void OnTableMemoryAlloc(catalog::db_oid_t database_oid, catalog::namespace_oid_t namespace_oid,
```
tli2: nit: Just say Allocation

src/include/storage/metric/abstract_metric.h
```diff
@@ -0,0 +1,354 @@
...
+   */
+  virtual void OnTransactionAbort(const transaction::TransactionContext *txn, catalog::db_oid_t database_oid) {}
+
+  /**
+   * @param txn context of the transaction performing read
+   * @param database_oid OID of the database that the tuple read happens
+   * @param namespace_oid OID of the namespace that the tuple read happens
+   * @param table_oid OID of the table that the tuple read happens
+   */
+  virtual void OnTupleRead(const transaction::TransactionContext *txn, catalog::db_oid_t database_oid,
```
tli2: Is it intentional that this method does not take the tuple (or a representation of it)?

src/include/storage/metric/abstract_raw_data.h
```diff
@@ -0,0 +1,58 @@
...
+
+  /**
+   * Make necessary updates to the metric raw data and persist the content of
+   * this RawData into internal SqlTables. Expect this object
+   * to be garbage-collected after this method is called.
+   * @param txn_manager transaction manager of the system
+   * @param catalog catalog of the system
+   * @param txn transaction context used for accessing SqlTables
+   */
+  virtual void UpdateAndPersist(transaction::TransactionManager *txn_manager, catalog::Catalog *catalog,
```
tli2: This should be decoupled from the catalog

mbutrovich:
Closing since the work has diverged enough and there are 600+ commits mostly resulting from the old catalog being merged and un-merged in this branch.

PR #340, opened 2019-04-08
Interval GC

thepulkitagarwal:
We mostly made changes in the `garbage_collector.h` and `garbage_collector.cpp` files, specifically the `ProcessUnlinkQueue` function. 

We implemented interval compaction on top of the earlier garbage collection (called "regular gc"). While performing the interval compaction, we unlink the undo records that are being compacted, and then push it into the deallocate queue. The deallocation is basically the same as the regular gc (with a minor change for handling varlens).

Slides from Project Update Presentation: [Google Docs Link](https://docs.google.com/presentation/d/1JDZ1yZ2ZTGQx5yxqLpDENALZ1r4oMIinLFY99_hwFTw/edit?usp=sharing)
Design Docs: [https://github.com/utkarsh39/terrier/wiki](https://github.com/utkarsh39/terrier/wiki)


src/include/transaction/transaction_manager.h
```diff
@@ -51,6 +52,11 @@ class TransactionManager {
    */
   void Abort(TransactionContext *txn);
 
+  /**
+   * @return unordered_set of all active txns
```
wenxuanqiu: Update comment unordered_set -> vector

test/storage/garbage_collector_test.cpp
```diff
@@ -83,7 +83,7 @@ struct GarbageCollectorTests : public ::terrier::TerrierTest {
   storage::BlockStore block_store_{100, 100};
   storage::RecordBufferSegmentPool buffer_pool_{10000, 10000};
   std::default_random_engine generator_;
-  const uint32_t num_iterations_ = 100;
+  const uint32_t num_iterations_ = 1;
```
wenxuanqiu: Unintentional change?

src/storage/garbage_collector.cpp
```diff
@@ -86,80 +117,284 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
...
+  const TupleAccessStrategy &accessor = table->accessor_;
+  const BlockLayout &layout = accessor.GetBlockLayout();
+
+  // Initialize the projected row with the set of columns we have seen in the first pass
+  std::vector<col_id_t> col_id_list(col_set_.begin(), col_set_.end());
+  auto init = storage::ProjectedRowInitializer::CreateProjectedRowInitializer(layout, col_id_list);
+
+  // Get new entry for the undo record from the buffer
+  uint32_t size = static_cast<uint32_t>(sizeof(UndoRecord)) + init.ProjectedRowSize();
+  byte *head = delta_record_compaction_buffer_->NewEntry(size);
```
wenxuanqiu: Might be good to assert head here

src/include/storage/garbage_collector.h
```diff
@@ -51,19 +56,145 @@ class GarbageCollector {
...
+   * Given a UndoRecord that has been deemed safe to unlink by the GC, attempts to remove it from the version chain.
+   * It's possible that this process will fail because the GC is conservative with conflicts. If the UndoRecord in the
+   * version chain to be updated in order to unlink the target UndoRecord is not yet committed, we will fail and
+   * expect this txn to be requeued and we'll try again on the next GC invocation, hopefully after the conflicting txn
+   * is either committed or aborted.
+   * @param undo_record UndoRecord to be unlinked
+   * @param active_txns list of timestamps of running transactions
+   * @return true, if the undo record was unlinked
+   */
+  bool ProcessUndoRecord(UndoRecord *undo_record, std::vector<transaction::timestamp_t> *active_txns);
```
wenxuanqiu: It seems that active_txns is not to be modified (in the function definition, it becomes a pointer to const values anyways), maybe consider const std::vector<transaction::timestamp_t> &active_txns

thepulkitagarwal: `clang-tidy` recommends not to put `const` in the header files. The `const` declaration is present in the .cpp file.

src/storage/garbage_collector.cpp
thepulkitagarwal: Can you clarify what you mean?
Do you mean something like `ASSERT(head != nullptr)`?

wenxuanqiu: Yes, or have a control flow for handling the nullptr case

thepulkitagarwal: Added check in the next line.

src/include/storage/garbage_collector.h
```diff
@@ -51,19 +56,151 @@ class GarbageCollector {
...
   // queue of txns that need to be unlinked
   transaction::TransactionQueue txns_to_unlink_;
+  // Undo buffer to hold compacted undo records
+  storage::UndoBuffer *delta_record_compaction_buffer_;
+  // Variable to mark that undo buffer to hold compacted undo records is empty so that it can be deallocated without
+  // unlinking
+  bool compaction_buffer_empty;
+  // queue of undo buffers containing compacted undo records which are pending unlinking
+  std::forward_list<storage::UndoBuffer *> buffers_to_unlink_;
+  // queue of undo buffers containing compacted undo records which ahve been unlinked abd are pending deallocation
```
wenxuanqiu:
```suggestion
  // queue of undo buffers containing compacted undo records which have been unlinked and are pending deallocation
```

src/include/storage/garbage_collector.h
```diff
@@ -51,19 +56,151 @@ class GarbageCollector {
...
+  std::forward_list<storage::UndoBuffer *> buffers_to_deallocate_;
```
qdHe: According to my understanding, the memory of compacted records is allocaed by interval GC, which needs to be reclaimed through buffers_to_unlink/buffers_to_deallocate_. If I am wrong, please correct me. 
The difference between undorecords in buffers_to_unlink/buffers_to_deallocate_ and undorecords in txns_to_unlink_/txns_to_deallocate_  is a little confusing, it is helpful if you could provide detailed explaination in design document or comments.

src/include/storage/garbage_collector.h
```diff
@@ -51,19 +56,151 @@ class GarbageCollector {
...
+  // set of all column ids in the compacted interval
+  std::unordered_set<col_id_t> col_set_;
+  // list of varlen entries per undo record which need to be reclaimed
+  std::unordered_map<storage::UndoRecord *, std::forward_list<const byte *> > reclaim_varlen_map_;
+  // set of tuple slots which have already been visited in this GC run
+  std::unordered_set<TupleSlot> visited_slots_;
 };
```
qdHe: I think that those "storage::" before UndoRecord and UndoBuffer are unnecessary since it is already in namespace terrier::storage.

src/storage/garbage_collector.cpp
```diff
@@ -86,80 +120,331 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
...
+  DataTable *&table = undo_record->Table();
+  TERRIER_ASSERT(table != nullptr, "Table should not be NULL here");
+  ReclaimSlotIfDeleted(undo_record);
+  // Add Varlen of the Undo record to the reclaim_varlen_map_ where it can be reclaimed later
+  MarkVarlenReclaimable(undo_record);
+  // Mark the record as fully processed
+  table = nullptr;
 }
 
-void GarbageCollector::ReclaimSlotIfDeleted(UndoRecord *const undo_record) const {
```
qdHe: Is there any reason to delete the key word const here?

src/storage/garbage_collector.cpp
```diff
@@ -86,80 +120,331 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
...
+    LinkCompactedUndoRecord(start_record, &curr, next, compacted_undo_record);
+  } else if (interval_length == 1) {
+    if (active_txns_iter == active_txns->end()) {
+      // Exited out of previous loop such that the last record is not visible to any transaction
+      UnlinkUndoRecordVersion(start_record->Next());
+      start_record->Next().store(nullptr);
+    }
+  }
+  // Set interval length to 0
+  EndCompaction(&interval_length);
```
qdHe: Is the function EndCompaction() here useless?

src/storage/garbage_collector.cpp
```diff
@@ -86,80 +120,331 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
...
+
+  // active_trans_iter ends but there are still elements in the version chain. Can GC everything below
   curr->Next().store(nullptr);
```
qdHe: Sorry, I am not sure whether the function UnlinkUndoRecordVersion() will reclaim the variable UndoRecord itself. If the answer is YES, then the pointer curr here may be an empty pointer. If the answer is NO, where will it be reclaimed?

src/include/storage/garbage_collector.h
```diff
@@ -51,19 +56,151 @@ class GarbageCollector {
...
+   * @param undo_record pointer to the head of the chain
+   * @param active_txns vector containing all active transactions
+   */
+  void ProcessTupleVersionChain(UndoRecord *undo_record, std::vector<transaction::timestamp_t> *active_txns);
+
+  /**
+   * Straight up unlink the undo_record and reclaim its space
+   * @param txn
+   * @param undo_record
+   */
```
qdHe: The comment here is incompatible with the definition.

src/storage/garbage_collector.cpp
```diff
@@ -86,80 +120,331 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
...
+        // We didn't unlink all of the UndoRecords (UnlinkUndoRecord returned false due to a write-write conflict),
+        // requeue txn for next GC run. Unlinked UndoRecords will be skipped on the next time around since we use the
+        // table pointer of an UndoRecord as the internal marker of being unlinked or not
+        requeue.push_front(txn);
+      }
+    }
+  }
+
+  storage::UndoBuffer *buf = nullptr;
+  std::forward_list<storage::UndoBuffer *> buf_requeue;
```
yangdsh: I would suggest that all variables and data structures should be declared at the begining of the function.

src/storage/garbage_collector.cpp
```diff
@@ -86,80 +120,331 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
...
+      }
     }
   }
 
-  // Requeue any txns that we were still visible to running transactions
-  txns_to_unlink_ = transaction::TransactionQueue(std::move(requeue));
+  // Requeue any txns that have an undo record still visible to some running transaction
+  if (!requeue.empty()) {
+    txns_to_unlink_ = transaction::TransactionQueue(std::move(requeue));
+  }
```
yangdsh: The code flow could be more coherent if you move the above four lines to line 142.

src/storage/garbage_collector.cpp
```diff
@@ -66,12 +97,15 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
...
   uint32_t txns_processed = 0;
-  // Certain transactions might not be yet safe to gc. Need to requeue them
   transaction::TransactionQueue requeue;
-  // It is sufficient to truncate each version chain once in a GC invocation because we only read the maximal safe
-  // timestamp once, and the version chain is sorted by timestamp. Here we keep a set of slots to truncate to avoid
-  // wasteful traversals of the version chain.
-  std::unordered_set<TupleSlot> visited_slots;
+
+  // Get active_txns in descending sorted order
+  std::vector<transaction::timestamp_t> active_txns = txn_manager_->GetActiveTxns();
```
yangdsh: I would suggest that you declare 'active_txns' at line 90, so that readers can clearly see that 'active_txns' plays a similar role with the 'oldest_txn' in the regular GC. You can still assign the value of 'active_txns' here.

src/include/storage/garbage_collector.h
```diff
@@ -51,19 +56,151 @@ class GarbageCollector {
    */
   uint32_t ProcessUnlinkQueue();
```
yangdsh: You probably want to order all the functions alphabetically as the original file did. 
I notice that the order of the functions are the same in .h and .cpp file, which is very nice.

test/storage/garbage_collector_test.cpp
```diff
@@ -672,4 +672,244 @@ TEST_F(GarbageCollectorTests, InsertUpdate1) {
   }
 }
 
-}  // namespace terrier
+// NOLINTNEXTLINE
```
yangdsh: In my opinion, your tests are comprehensive and convincible enough. But it seems that Andy prefers tests without hand-coded answers. For your implementation, I still think it is fine to test in the way you do.

test/storage/garbage_collector_test.cpp
apavlo: @wenxuanqiu is correct. Should we set this back to 100?

thepulkitagarwal: We already set it back to 100. This was committed by mistake.

test/storage/garbage_collector_test.cpp
apavlo: @yangdsh I think these hardcoded values are unavoidable in this particular example.

test/storage/garbage_collector_test.cpp
```diff
@@ -672,4 +672,244 @@ TEST_F(GarbageCollectorTests, InsertUpdate1) {
...
+TEST_F(GarbageCollectorTests, SingleOLAP) {
+  for (uint32_t iteration = 0; iteration < num_iterations_; ++iteration) {
```
apavlo: You need to document what this test case is doing.

benchmark/storage/garbage_collector_benchmark.cpp
```diff
@@ -150,6 +150,33 @@ BENCHMARK_DEFINE_F(GarbageCollectorBenchmark, HighContention)(benchmark::State &
...
+  // NOLINTNEXTLINE
+  for (auto _ : state) {
+    LargeTransactionBenchmarkObject tested({8, 8, 8}, initial_table_size, 100, update_select_ratio, &block_store_,
+                                           &buffer_pool_, &generator_, true);
+    StartGC(tested.GetTxnManager());
+    auto *olap_txn = tested.GetTxnManager()->BeginTransaction();
+    uint64_t elapsed_ms;
+    {
+      common::ScopedTimer timer(&elapsed_ms);
+      tested.SimulateOltp(num_txns, num_concurrent_txns);
```
yangdsh: Why use SimulateOltp() function for olap?

thepulkitagarwal: There is one OLAP transaction `olap_txn` which starts before all the other OLTP transactions, and would run concurrently with all of them. The `SimulateOltp(..)` function simulates the OLTP transactions.

test/storage/large_garbage_collector_test.cpp
```diff
@@ -304,4 +304,76 @@ TEST_F(LargeGCTests, TPCCishHighThreadWithGC) {
...
+                                            .SetGcOn(true)
+                                            .SetBookkeeping(true)
+                                            .SetVarlenAllowed(true)
+                                            .build();
+    StartGC(tested.GetTxnManager());
+    // OLAP Transaction starts here
+    auto *olap_txn = tested.GetTxnManager()->BeginTransaction();
+    for (uint32_t batch = 0; batch * batch_size < num_txns; batch++) {
+      auto result = tested.SimulateOltp(batch_size, num_concurrent_txns);
+      gc_paused_ = true;
```
yangdsh: Why you pause GC here?

test/storage/garbage_collector_test.cpp
```diff
@@ -671,4 +671,894 @@ TEST_F(GarbageCollectorTests, InsertUpdate1) {
...
+
+    // U1, U2 will be unlinked. Can't unlink U3 as it is version chain head and T0 is still active
+    // Can't unlink INSERT as it is an Insert Undo Record
+    EXPECT_EQ(std::make_pair(0u, 2u), gc.PerformGarbageCollection());
+
+    tested.SelectIntoBuffer(txn0, slot);
+    EXPECT_FALSE(tested.select_result_);
+
+    txn_manager.Commit(txn0, TestCallbacks::EmptyCallback, nullptr);
+    // Unlink U4 as T0 committed and unlink read-only txn T0 and unlink INSERT as no active txn
```
yangdsh: In the comment, U4 should be U3.

test/storage/garbage_collector_test.cpp
```diff
@@ -671,4 +671,894 @@ TEST_F(GarbageCollectorTests, InsertUpdate1) {
...
+    // Deallocate U1, U2
+    EXPECT_EQ(std::make_pair(2u, 3u), gc.PerformGarbageCollection());
+    // Deallocate INSERT, U3
+    EXPECT_EQ(std::make_pair(2u, 0u), gc.PerformGarbageCollection());
+    // Nothing should be deallocated
+    EXPECT_EQ(std::make_pair(0u, 0u), gc.PerformGarbageCollection());
```
yangdsh: Sorry, I am a little confused. There are five transactions and only four are deallocated. T0 is not deallocated according to your comment, why?

yangdsh: Is it because T0 is read-only?

test/storage/garbage_collector_test.cpp
```diff
@@ -671,4 +671,894 @@ TEST_F(GarbageCollectorTests, InsertUpdate1) {
...
+    txn_manager.Commit(txn_delete, TestCallbacks::EmptyCallback, nullptr);
+
+    // Active txn should see correct version
+    auto *select_tuple = tested.SelectIntoBuffer(txn1, slot);
+    EXPECT_TRUE(tested.select_result_);
+    EXPECT_TRUE(StorageTestUtil::ProjectionListEqualShallow(tested.Layout(), select_tuple, insert_tuple));
+
+    // INSERT should be unlinked
+    EXPECT_EQ(std::make_pair(0u, 1u), gc.PerformGarbageCollection());
+
```
yangdsh: I suggest you add an ``EXPECT_EQ(std::make_pair(0u, 0u), gc.PerformGarbageCollection());`` here to test if INSERT is deallocated.

src/storage/garbage_collector.cpp
```diff
@@ -13,25 +17,43 @@
...
   STORAGE_LOG_TRACE("GarbageCollector::PerformGarbageCollection(): txns_deallocated: {}", txns_deallocated);
   uint32_t txns_unlinked = ProcessUnlinkQueue();
   STORAGE_LOG_TRACE("GarbageCollector::PerformGarbageCollection(): txns_unlinked: {}", txns_unlinked);
-  if (txns_unlinked > 0) {
-    // Only update this field if we actually unlinked anything, otherwise we're being too conservative about when it's
-    // safe to deallocate the transactions in our queue.
-    last_unlinked_ = txn_manager_->GetTimestamp();
-  }
+  // Update the last unlinked timestamp every GC run conservatively whether or not any transaction/compacted
+  // undo record is unlinked.
```
wenxuanqiu: Style nit: maybe capitalize first character to be consistent with other comments

src/include/storage/garbage_collector.h
```diff
@@ -39,38 +44,177 @@ class GarbageCollector {
...
   transaction::timestamp_t last_unlinked_;
   // queue of txns that have been unlinked, and should possible be deleted on next GC run
   transaction::TransactionQueue txns_to_deallocate_;
   // queue of txns that need to be unlinked
   transaction::TransactionQueue txns_to_unlink_;
+  // Undo buffer to hold compacted undo records
+  storage::UndoBuffer *delta_record_compaction_buffer_;
+  // Variable to mark that undo buffer to hold compacted undo records is empty so that it can be deallocated without
+  // unlinking
+  bool compaction_buffer_empty;
```
wenxuanqiu: Style nit: compaction_buffer_empty -> compaction_buffer_empty_

src/include/storage/garbage_collector.h
```diff
@@ -39,38 +44,177 @@ class GarbageCollector {
...
+   * Given the data table and the tuple slot, try unlinking the version chain head for that tuple slot
+   * @param table data table
+   * @param slot tuple slot
+   * @param active_txns list of currently running active transactions
+   */
+  void ProcessTupleVersionChainHead(DataTable *table, TupleSlot slot,
+                                    std::vector<transaction::timestamp_t> *active_txns);
+
+  /**
+   * Given a UndoRecord that has been deemed safe to unlink by the GC, attempts to remove it from the version chain.
```
wenxuanqiu: Style nit: a UndoRecord -> an UndoRecord

src/include/storage/garbage_collector.h
```diff
@@ -39,38 +44,177 @@ class GarbageCollector {
...
+   * @param slot tuple slot
+   * @param active_txns vector containing all active transactions
+   */
+  void ProcessTupleVersionChain(DataTable *table, TupleSlot slot, std::vector<transaction::timestamp_t> *active_txns);
+
+  /**
+   * Given the data table and the tuple slot, try unlinking the version chain head for that tuple slot
+   * @param table data table
+   * @param slot tuple slot
+   * @param active_txns list of currently running active transactions
```
wenxuanqiu: Style nit: list -> vector to be consistent with other function description and the actual container type

src/include/storage/garbage_collector.h
```diff
@@ -1,7 +1,11 @@
 #pragma once
 
 #include <queue>
+#include <unordered_map>
```
wenxuanqiu: It seems that unordered_map is not used in this file, consider removing it

src/storage/garbage_collector.cpp
```diff
@@ -13,25 +17,43 @@
 namespace terrier::storage {
 
 std::pair<uint32_t, uint32_t> GarbageCollector::PerformGarbageCollection() {
+  // Create the UndoBuffer for this GC run
+  delta_record_compaction_buffer_ = new UndoBuffer(txn_manager_->buffer_pool_);
+  // The compaction buffer is empty
+  compaction_buffer_empty = true;
```
wenxuanqiu: Style nit: compaction_buffer_empty -> compaction_buffer_empty_

src/storage/garbage_collector.cpp
```diff
@@ -13,25 +17,43 @@
...
-    last_unlinked_ = txn_manager_->GetTimestamp();
-  }
+  // Update the last unlinked timestamp every GC run conservatively whether or not any transaction/compacted
+  // undo record is unlinked.
+  last_unlinked_ = txn_manager_->GetTimestamp();
   STORAGE_LOG_TRACE("GarbageCollector::PerformGarbageCollection(): last_unlinked_: {}",
                     static_cast<uint64_t>(last_unlinked_));
+
+  // Handover compacted buffer for GC
+  if (compaction_buffer_empty) {
```
wenxuanqiu: Same here compaction_buffer_empty -> compaction_buffer_empty_

src/storage/garbage_collector.cpp
```diff
@@ -87,32 +122,111 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
...
+  if (undo_record->txnptr_.IsNull()) return true;
+  const TupleSlot slot = undo_record->Slot();
+  // Process this tuple only
+  if (visited_slots->insert(slot).second) {
+    // Perform interval gc for the entire version chain excluding the head of the chain
+    ProcessTupleVersionChain(table, slot, active_txns);
+    ProcessTupleVersionChainHead(table, slot, active_txns);
+  }
+
+  return undo_record->txnptr_.IsNull();
```
wenxuanqiu: Question: Is this line just going to be false, or `ProcessTupleVersionChain` and `ProcessTupleVersionChainHead` can modified undo_record to change that fact? If it is the former, might as well just return false.

src/storage/garbage_collector.cpp
```diff
@@ -87,32 +122,111 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
...
+    // This version chain is empty, so we can skip it
+    return;
+  }
+  // Perform gc for head of the chain
+  // Assuming can garbage collect any version greater than the oldest timestamp
+  transaction::timestamp_t version_ptr_timestamp = version_chain_head->Timestamp().load();
+  // If there are no active transactions, or if the version pointer is older than the oldest active transaction,
+  // Collect the head of the chain using compare and swap
+  // Note that active_txns is sorted in descending order, so its tail should have the oldest txn's timestamp
+  if (version_ptr_timestamp < active_txns->back()) {
```
wenxuanqiu: Calling `back` on empty vector can cause undefined behaviour, need to check with `ative_txns->empty()` first

src/storage/garbage_collector.cpp
```diff
@@ -129,54 +243,279 @@ void GarbageCollector::ProcessDeferredActions() {
...
+
+  if (interval_length > 1) {
+    // Create the undo record by a second traversal through the records to be compacted.
+    UndoRecord *compacted_undo_record = CreateUndoRecord(start_record, next);
+    // Copy the varlen attributes of the compacted record
+    CopyVarlen(compacted_undo_record);
+    // Link the compacted record to the version chain.
+    LinkCompactedUndoRecord(start_record, &curr, next, compacted_undo_record);
+  } else if (interval_length == 1) {
+    if (transaction::TransactionUtil::NewerThan(active_txns->back(), curr->Timestamp().load())) {
```
wenxuanqiu: `active_txns->back()` has undefined behaviour if `active_txns` is empty, you might want to check for that.

src/storage/garbage_collector.cpp
```diff
@@ -129,54 +243,279 @@ void GarbageCollector::ProcessDeferredActions() {
...
+    // Move next pointer ahead
+    next = next->Next();
+  }
+}
 
-  // If the head of the version chain was not committed, it could have been aborted and requires a retry.
-  if (curr == version_ptr && !transaction::TransactionUtil::Committed(version_ptr->Timestamp().load()) &&
-      table->AtomicallyReadVersionPtr(slot, accessor) != version_ptr)
-    TruncateVersionChain(table, slot, oldest);
+void GarbageCollector::UnlinkUndoRecordVersion(UndoRecord *const undo_record) {
```
wenxuanqiu: It might be a good idea to check whether `undo_record` is a `nullptr` here, because It seems that not all callers of this method do that check

src/storage/garbage_collector.cpp
```diff
@@ -129,54 +243,279 @@ void GarbageCollector::ProcessDeferredActions() {
...
+                                       uint32_t *interval_length_ptr) {
+  col_set_.clear();
+  // Case 1: next is of type UPDATE
+  //         Standard undo record compaction begins
+  // Case 2: next is of type INSERT
+  //         This can only happen when a transaction is sandwiched between INSERT and another Delta record
+  //         INSERT is the last record in the version chain
+  // Case 3: next is of type DELETE
+  //         DELETE can only be the newest record in the version chain
+  TERRIER_ASSERT(next->Type() != DeltaRecordType::DELETE, "Delete cannot be compacted");
```
wenxuanqiu: It is great that all callers of this method guarantee that `next` is not `nullptr`. Maybe document this as an assumption in the header or check again here explicitly? This will certainly help future students.

test/storage/garbage_collector_test.cpp
utkarsh39: Yeah read-only transactions don't get deallocated

src/storage/garbage_collector.cpp
utkarsh39: The comment is a continuation of the previous sentence

src/storage/garbage_collector.cpp
utkarsh39: The code logic which calls this function guarantees it cannot be NULL. However, it might be nice to add an assertion.

src/storage/garbage_collector.cpp
utkarsh39: Transaction Manager ensures that if the active_txns is not empty, it inserts a timestamp.

src/storage/garbage_collector.cpp
utkarsh39: undo records have an extra field containing the pointer to the corresponding txn. This txn pointer is set to NULL when the undo record is unlinked. So this line is not necessarily false.

src/storage/garbage_collector.cpp
utkarsh39: Transaction Manager ensures that if the active_txns is not empty, it inserts a timestamp.

test/storage/garbage_collector_test.cpp
yangdsh: I saw you comment this in the later tests, but I think you should also comment it here.

test/storage/garbage_collector_test.cpp
```diff
@@ -671,4 +671,894 @@ TEST_F(GarbageCollectorTests, InsertUpdate1) {
...
+    auto *txn_insert = txn_manager.BeginTransaction();
+    auto *insert_tuple = tested.GenerateRandomTuple(&generator_);
+    storage::TupleSlot slot = tested.table_.Insert(txn_insert, *insert_tuple);
+    txn_manager.Commit(txn_insert, TestCallbacks::EmptyCallback, nullptr);
+
+    tested.SelectIntoBuffer(txn1, slot);
+    EXPECT_FALSE(tested.select_result_);
+
+    // U1
+    auto *txn_e = txn_manager.BeginTransaction();
```
qdHe: What is the law of transaction naming here? Those suffixes seem confusing.

test/storage/garbage_collector_test.cpp
qdHe: Is it better to run a random number of updates here? For example, run n updates and n-1 should be unlinked (except the newest one).

src/include/storage/undo_record.h
```diff
@@ -165,30 +215,35 @@ class UndoRecord {
...
     result->slot_ = slot;
+    result->txnptr_.Put(txn);
 
     ProjectedRow::CopyProjectedRowLayout(result->varlen_contents_, redo);
 
     return result;
   }
 
  private:
+  friend class GarbageCollector;
```
qdHe: Is it necessary to add GarbageCollector as friend class? Add a public function to provide txnptr_ may be better?

src/storage/garbage_collector.cpp
thepulkitagarwal: `GetActiveTxns` is based on the function `OldestTransactionStartTime`, which was used by the original GC. `OldestTransactionStartTime` returns the oldest active transaction's start time. If there are no active transactions, it returns the current time (that is, it returns `time_`, the internal counter). Similar to `OldestTransactionStartTime`, `GetActiveTxns` returns the active transactions' timestamps. And if there are none, it returns the current time.

test/storage/garbage_collector_test.cpp
thepulkitagarwal: I think that [the next `EXPECT_EQ`](https://github.com/cmu-db/terrier/pull/340/files/e301b972d160510ff76929af06ef02088b12a170#diff-8da770a88e7232c376d2a3a1ec56f9dbR1265) does test that

test/storage/large_garbage_collector_test.cpp
```diff
@@ -304,4 +304,76 @@ TEST_F(LargeGCTests, TPCCishHighThreadWithGC) {
     EndGC();
   }
 }
+
+// This test attempts to simulate a long running read-only OLAP query, with OLTP
+// queries in a TPC-C-like scenario.
+// NOLINTNEXTLINE
+TEST_F(LargeGCTests, OLAPAndTPCCishWithGC) {
+  const uint32_t txn_length = 5;
+  const std::vector<double> update_select_ratio = {0.4, 0.6};
```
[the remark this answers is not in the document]
thepulkitagarwal: It is a common select ratio. Since the update to select ratio is low, the expectation is that the number of aborts will be low, but there will still be a considerable number of updates happening thus increasing the version chains' lengths.

test/storage/large_garbage_collector_test.cpp
```diff
@@ -304,4 +304,76 @@ TEST_F(LargeGCTests, TPCCishHighThreadWithGC) {
...
+                                            .SetGenerator(&generator_)
+                                            .SetGcOn(true)
+                                            .SetBookkeeping(true)
+                                            .SetVarlenAllowed(true)
+                                            .build();
+    StartGC(tested.GetTxnManager());
+    // OLAP Transaction starts here
+    auto *olap_txn = tested.GetTxnManager()->BeginTransaction();
+    for (uint32_t batch = 0; batch * batch_size < num_txns; batch++) {
+      auto result = tested.SimulateOltp(batch_size, num_concurrent_txns);
```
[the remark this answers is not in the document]
utkarsh39: The transaction is replicating the behaviour of OLAP in the sense that it's long-running. That's why we named it OLAP. You are right it doesn't need to do analytical queries, it just needs to sit there for Interval GC to kick in as opposed to regular GC which won't start until the OLAP txn is dead.

src/storage/garbage_collector.cpp
```diff
@@ -13,25 +17,43 @@
 namespace terrier::storage {
 
 std::pair<uint32_t, uint32_t> GarbageCollector::PerformGarbageCollection() {
+  // Create the UndoBuffer for this GC run
+  delta_record_compaction_buffer_ = new UndoBuffer(txn_manager_->buffer_pool_);
+  // The compaction buffer is empty
+  compaction_buffer_empty_ = true;
```
tli2: I don't think you need a separate boolean to track this, right? You can directly ask the buffer if it is empty.

src/storage/garbage_collector.cpp
```diff
@@ -13,25 +17,43 @@
 namespace terrier::storage {
 
 std::pair<uint32_t, uint32_t> GarbageCollector::PerformGarbageCollection() {
+  // Create the UndoBuffer for this GC run
+  delta_record_compaction_buffer_ = new UndoBuffer(txn_manager_->buffer_pool_);
```
tli2: What is the point of storing this on a heap with a pointer member? Can you inline it?

src/storage/garbage_collector.cpp
```diff
@@ -50,13 +72,21 @@ uint32_t GarbageCollector::ProcessDeallocateQueue() {
       }
     }
     txns_to_deallocate_ = std::move(requeue);
+
+    // All of the undo records in my buffers were unlinked before the oldest running txn in the system.
+    // We are now safe to deallocate these buffers because no running transaction should hold a reference to them
+    // anymore
+    while (!buffers_to_deallocate_.empty()) {
```
tli2: Not a problem with your PR, but we will eventually want to rewrite these using the deferred action framework introduced in #358

src/storage/garbage_collector.cpp
```diff
@@ -129,54 +243,281 @@ void GarbageCollector::ProcessDeferredActions() {
...
+  version_chain_head = table->AtomicallyReadVersionPtr(slot, accessor);
+
+  // If no chain is passed, nothing is collected
+  if (version_chain_head == nullptr) {
     return;
   }
 
+  // Otherwise collect as much as possible
+  // Don't collect version_chain_head though
+  auto active_txns_iter = active_txns->begin();
```
tli2: I believe this overshadows the changes made in #325?

We want those changes for performance issues on skewed workloads.

src/storage/garbage_collector.cpp
```diff
@@ -87,32 +122,111 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
       // to safely free the txn
       txns_to_deallocate_.push_front(txn);
       txns_processed++;
-    } else if (transaction::TransactionUtil::NewerThan(oldest_txn, txn->TxnId().load())) {
-      // Safe to garbage collect.
+    } else {
+      // This is a txn that may or may not be visible to any running txns. Proceed with unlinking its UndoRecords
+      // with an Interval GC approach
+      bool all_unlinked = true;
```
tli2: `all_unlinked` looks suspiciously like a legacy from our previous implementations. Is there a real failure scenario where you need this?

tli2:
Overall this PR looks pretty good. One issue I have with this is that this makes interval GC mandatory for everything. Two clear candidates for generalization here is a threshold number of records to compact (i.e., only compact groups of more than n delta records), and a threshold window for transactions to be considered for compaction (i.e., only compact for transactions that are more than 100 timestamps in the past).

What parameters are optimal needs to be determined with further benchmarking, but this implementation makes me somewhat worried about degenerate cases, where the timings of the transactions are such that the GC always does maximal wasteful work.

src/include/storage/undo_record.h
utkarsh39: We would need getters and setters for all 7 fields of undo record class. Moreover, adding get and set would allow everyone to modify the class.

src/include/storage/garbage_collector.h
```diff
@@ -39,38 +43,177 @@ class GarbageCollector {
   std::pair<uint32_t, uint32_t> PerformGarbageCollection();
 
  private:
+  /**
```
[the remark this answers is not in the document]
utkarsh39: In other files as well, doxygen is used for private functions. Should all the private functions use `//`?

src/include/storage/undo_record.h
```diff
@@ -99,16 +142,18 @@ class UndoRecord {
    * @param timestamp timestamp of the transaction that generated this UndoRecord
    * @param slot the TupleSlot this UndoRecord points to
    * @param table the DataTable this UndoRecord points to
+   * @param txn the transaction this UndoRecord belongs to
    * @return pointer to the initialized UndoRecord
    */
   static UndoRecord *InitializeInsert(byte *const head, const transaction::timestamp_t timestamp, const TupleSlot slot,
-                                      DataTable *const table) {
+                                      DataTable *const table, transaction::TransactionContext *const txn) {
```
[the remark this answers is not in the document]
utkarsh39: To handle this, we would need a .cpp file for undo_record as there is a circular dependency. Forward declaration of TransactionContext in undo_record.h won't work.

src/storage/garbage_collector.cpp
utkarsh39: We still need this logic to know when all of the transaction's undo records have been unlinked. #325 eliminates this by only processing a transaction for unlinking when it is older than all the active txns. Here we do not and can not make this assumption (Interval GC).

utkarsh39:
> Overall this PR looks pretty good. One issue I have with this is that this makes interval GC mandatory for everything. Two clear candidates for generalization here is a threshold number of records to compact (i.e., only compact groups of more than n delta records), and a threshold window for transactions to be considered for compaction (i.e., only compact for transactions that are more than 100 timestamps in the past).
> 
> What parameters are optimal needs to be determined with further benchmarking, but this implementation makes me somewhat worried about degenerate cases, where the timings of the transactions are such that the GC always does maximal wasteful work.

We discussed this and we think this could be done as a separate PR later.

src/include/storage/undo_record.h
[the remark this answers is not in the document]
tli2: If you need a `cpp` file that's fine

tli2:
This is good stuff. But due to changes to the version chain in #404, and ongoing effort to rewrite the GC with the generalized deferred event framework, this can no longer be cleanly merged.

Issue #329, opened 2019-03-29
Duplex serialization and buffer write in LogManager


malin1993ml:
Currently the LogManager is single-threaded, and it serializes log records (CPU-intensive) and writes out the log buffer to disk (IO-intensive) iteratively. This can limit the capacity (maximum throughput) of the log manager, and a future optimization is to use two threads to do serialization and buffer write concurrently.

tli2:
fixed by #370

Issue #304, opened 2019-02-23
Checkpoint and Recovery


tli2:
We left checkpoint and recovery unimplemented in the Fall because of the lack of catalog for recovery.

Our current system is essentially physical logging of all the redo records. The original plan was to implement consistent checkpoints by taking advantage of MVCC. On crash, the system reboots to a consistent point and replays the log in one pass, applying all the other committed changes.

One possible optimization to this is to allow checkpoints to be somewhat fuzzy. Specifically, the checkpoint does not need to be a consistent snapshot so long as all the values written out are committed, and their transactions have their log records on disk. The recovery pass can also be optimized in this case by keeping per-block metadata to skip applying certain regions of the log.

pervazea:
There was discussion at team meeting on 30 Apr. of recovery and catalog interactions.

When recovery needs to be done, the sequence is:
- system starts up
- system catalog tables created, but not populated
- recovery will read the logs and populate the catalog tables. To insert data into the catalog tables, recovery needs to take the table oid and find the corresponding catalog table and schema.

Options:
1) Create a way for an oid encode the catalog table type. Something like:
If database oid = 1, then database local tables would be:
- oid 2 = pg_namespace
- oid 3 = pg_class
...

Possible, but not very clean.

2) If recovery has the option of recording not just the oid, but also the table type, by name or by enumerated value, then the schema  can easily be determined from the (oid, table type) pair. No changes would be needed to oid generation.

PR #341, opened 2019-04-08
Checkpoint manager

lmy1229:
# Project Design: Checkpoint & Recovery

## Overview

This PR implements checkpointing, logging and recovery. Checkpointing and recovery are new functionalities. There is also a major re-write of log manager because the master branch's logging is too simple to function right. Since catalog support is under implementation, we reverted all commits related to catalog to keep this PR clean. Logging infrastructure still uses data tables to cater to existing APIs, and should be migrated to SqlTable in the future.
To summarize, the main functionalities implemented are:
- Making a consistent checkpoint of a table
- Logging
- Recover from latest checkpoints and logs
- Varlens supported in all parts
- Async writer of checkpoints
- Handle arbitrarily huge rows when overflowing a checkpoint block
- Handle crashes during checkpoint or recovery phase
- SqlTable Transaction Test object for testing

##### Parts that our project will rely on:

>- storage::sql_table: the checkpoint manager will reply on the Scan interface provided by the sql_table to scan and persist the content of a table.
>- storage::log_manager: the log recovery will rely on the current implementation of the write-ahead-log. We do not intend to modify the current logging scheme. 
>- settings: the checkpoint manager will get the settings from this component and set itself up accordingly.
>- catalog: We need to persist and recover the catalog first, during the checkpoint and recovery respectively, so that we can restore the schema of other tables.

##### Parts that our project will modify or create:

>- **ADD**: storage::checkpoint_manager: a new class added to support checkpointing and recovery.
>- **ADD**: storage::checkpoint_io: defines the BufferedTupleWriter class, which is used to buffer tuples and write data to disk one page at a time.
>- **ADD**: checkpoint_test: checkpoint testing.
>- **Modify**: storage::log_manager: recovery capability will be added to the log_manager.
>- **Modify**: storage_test_util: add functions for checkpoint testing

## Architectural Design
The scanning and writing process is illustrated in the diagram [here](https://drive.google.com/file/d/1jrkZnJybYxG_NnLyWa9ELN0lFStpIY2P/view?usp=sharing)
>The checkpoint_manager scans the table and passes each row to the BufferedTupleWriter class to serialize. The BufferedTupleWriter will buffer the tuples into pages and asynchronously write out pages into the file with AsyncBlockWriter. An async block writer will use a background thread to write blocks into checkpoint file asynchronously. Its design is very similar to that of the logger threads in SiloR. An async block writer will have some pre-allocated block buffers. Workers must first ask for a free block buffer. If there is no free block buffer now, it will be blocked until some free block buffer is available. In this way, the async block writer can give some back pressure on the worker to let it slow down if it is generating blocks too quickly. The worker then fills the block buffer and hand it back to the writer thread. The writer thread writes it into the file and make the buffer available again.

The layout of a checkpoint block is as [here](https://drive.google.com/file/d/16bLiW8dKNqBdfS7nP2iAUIGgGKKBESPp/view?usp=sharing)
>When the current block cannot fit the next tuple, the page will be persisted to file and that tuple will be written to next block.

### checkpoint:

Handle crashes during checkpointing
> - Use another metadata file to mark completed: Would need another file.
> - Rename the checkpoint file once finished: Easy to handle. We adopted this option.

save tuples as Columns v.s. Rows: 
>  - Rows: Easy to implement recovery, using existing SqlTable API. Slower to checkpoint because we need to convert columns into rows.
>  - Columns: Easy to checkpoint, by directly writing ProjectedColumns into a file. Because of the varlens, it is hard to find the correct number of tuples so that it perfectly fit into one page. It is also hard to recovery, because currently the SqlTable doesn't support bulk insert.

Where to store varlen in checkpoints: 
>  - Store in a separate file: Easy to checkpoint and recovery. However, writing checkpoint to 2 files cannot guarantee atomicity.
>  - Inlined in each row: Save most storage space. Hard to checkpoint and recovery, because we cannot write the in-memory representation of ProjectedRow directly into the file. 
>  - Appended after each row: Easy to checkpoint and recover. No disadvantages.

Tuning AsyncBlockWriter:
> - Which block size should we use?
> - If there is no free block buffer, should the BufferedTupleWriter be blocked or just spin?

We did benchmark for these choices.

|             | Throughput (M rows/s) |
|-------------|-----------------------|
| 4K, spin    | 4.77                  |
| 512K, spin  | 4.99                  |
| 4K, block   | 3.85                  |
| 512K, block | 5.00                  |

We pick the 512K + block combination at last.

## Testing Plan
We created a new SqlTableTestObject class to build tables and run transactions on sql table level to test our code. We open GC thread, logging thread along with transactions to ensure our implementation works well under real scenarios.

* Correctness is tested by comparing the scan result of a table before and after recovery.

## Trade-offs and Potential Problems
The main trade-off in this project is between performance during checkpointing and during recovery. We prefer performance during checkpointing over during recovery. So our design decisions are inclined towards faster runtime.

## Future Work
Two major parts that are blocked by other components are:
1. After SqlTables are used in transactions, change logs to log operations on SqlTables.
2. Implement catalog checkpointing and recovery. Moreover, catalog is need to recover schema changes, identified by replaying operations in catalog tables.
3. After the above two parts are finished, we would then be able to recover all tables with their oids, instead of a table of oid fixed to 0 currently.

There are some redundancies in testing code, and can be further compacted.
1. The RandomSqlTableTestObject is almost a subset of SqlLargeTransactionTestObject, so we can consider compacting them together.
2. We used PrintAllRows() to compare tables, and this functionality is similar to StorageTestUtil::ProjectionListEqualShallow/StorageTestUtil::ProjectionListEqualDeep. 

Possible performance optimization in the future include:
1. morsel-based multi-thread checkpointing
2. bulk checkpoint & recovery without converting column store into a row view.
3. checkpointer & logger thread dynamic allocation.
4. Newest-to-oldest log replay.


src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,254 @@
...
+  BufferedTupleWriter() = default;
+
+  /**
+   * Instantiate a new BufferedTupleWriter, open a new checkpoint file to write into.
+   * @param log_file_path path to the checkpoint file.
+   */
+  explicit BufferedTupleWriter(const char *log_file_path)
+      : out_(PosixIoWrappers::Open(log_file_path, O_WRONLY | O_CREAT, S_IRUSR | S_IWUSR)),
+        block_size_(CHECKPOINT_BLOCK_SIZE),
+        buffer_(new byte[block_size_]()) {
```
yash620: You have code replicated here, why can't just use the open function you defined below?

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,254 @@
...
+  // TODO(Zhaozhe): checksum
+ public:
+  BufferedTupleWriter() = default;
+
+  /**
+   * Instantiate a new BufferedTupleWriter, open a new checkpoint file to write into.
+   * @param log_file_path path to the checkpoint file.
+   */
+  explicit BufferedTupleWriter(const char *log_file_path)
+      : out_(PosixIoWrappers::Open(log_file_path, O_WRONLY | O_CREAT, S_IRUSR | S_IWUSR)),
```
yash620: You have code replicated here, why can't you just use the open function you defined below?

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,254 @@
...
+  CheckpointFilePage *GetPage() { return reinterpret_cast<CheckpointFilePage *>(buffer_); }
+
+ private:
+  int out_;  // fd of the output files
+  uint32_t block_size_;
+  uint32_t cur_buffer_size_ = 0;
+  byte *buffer_ = nullptr;
+
+  void ResetBuffer() {
+    memset(buffer_, 0, block_size_);
```
jrolli: Why do you need to zero the buffer?  Nothing in your data structures are null-terminated so "uninitialized" data shouldn't be a problem.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,254 @@
...
+    CheckpointFilePage::Initialize(reinterpret_cast<CheckpointFilePage *>(buffer_));
+    cur_buffer_size_ = sizeof(CheckpointFilePage);
```
yash620: Is there a need for this to be reset every time? Isn't the CheckpointFilePage size consistent the entire time so the cur_buffer_size_ shouldn't change.

src/storage/checkpoint_io.cpp
```diff
@@ -0,0 +1,60 @@
...
+  }
+  AppendTupleToBuffer(row_buffer, varlen_offset, varlen_entries);
+}
+
+void BufferedTupleWriter::AppendTupleToBuffer(ProjectedRow *row_buffer, int32_t total_varlen,
+                                              const std::vector<const VarlenEntry *> &varlen_entries) {
+  int32_t row_size = row_buffer->Size();
+  int32_t tot_size = total_varlen + row_size;
+  // Flush buffer first if the tuple cannot fit in buffer
+  if (cur_buffer_size_ + tot_size > block_size_) {
```
jrolli: There is a major logic problem here: a table can have up to 65,535 columns which can have.  Ignoring varlens, a tuple could 512 kb (2^16 columns at 8 bytes a piece).  With varlens you cannot make any guarantee that a materialized tuple is smaller than your block.

src/storage/checkpoint_io.cpp
```diff
@@ -0,0 +1,60 @@
...
+    PersistBuffer();
+  }
+  // ASSUME that the row can always fit in the block
+  std::memcpy(buffer_ + cur_buffer_size_, row_buffer, row_size);
```
jrolli: Buffer overflow vulnerability!  (See comment on line 42)

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,254 @@
...
+  }
+
+  void PersistBuffer() {
+    // TODO(zhaozhe): calculate CHECKSUM. Currently using default 0 as checksum
+    if (cur_buffer_size_ == sizeof(CheckpointFilePage)) {
+      // If the buffer has no contents, just return
+      return;
+    }
+    PosixIoWrappers::WriteFully(out_, buffer_, block_size_);
```
jrolli: This is likely going to have the same performance problems as discussed in #329.  Can you split this into an asynchronous write and pipelined buffers?  (i.e. a buffer being prepared and a buffer being persisted)

src/storage/checkpoint_io.cpp
```diff
@@ -0,0 +1,60 @@
+#include "storage/checkpoint_io.h"
+#include <vector>
+namespace terrier::storage {
+
+void BufferedTupleWriter::SerializeTuple(ProjectedColumns::RowView *row, ProjectedRow *row_buffer,
```
yash620: Here you guys are trying to materialize every single tuple in the table. It is basically reimplementing the SqlTable::Select code. So at this point why not just grab the SlotIterator::SqlTable and dereference that to get a tuple slot and then call SqlTable::Select on that tuple slot. The SqlTable::Select will materialize the ProjectedRow you want and you can use that here, instead of having to reconstruct the entire ProjectedRow from scratch after already calling Scan.

src/storage/checkpoint_io.cpp
```diff
@@ -0,0 +1,60 @@
...
+                                         const storage::BlockLayout &layout) {
+  // First construct tuple in row_buffer, and calculate total size required
+  int32_t varlen_offset = 0;
+  std::vector<const VarlenEntry *> varlen_entries;
+  for (uint16_t projection_list_idx = 0; projection_list_idx < row->NumColumns(); projection_list_idx++) {
+    if (row->IsNull(projection_list_idx)) {
```
jrolli: `AccessForceNotNull` modifies the underlying data structure (it forces the bitmask to not null) and returns a pointer.  You should really be using `AccessWithNullCheck` for `row` and comparing it to `nullptr` here to condition on when the underlying data is null.

src/storage/checkpoint_io.cpp
```diff
@@ -0,0 +1,60 @@
...
+      row_buffer->SetNull(projection_list_idx);
+    } else {
+      row_buffer->SetNotNull(projection_list_idx);
```
yash620: AccessForceNotNull already sets not null

src/storage/checkpoint_io.cpp
```diff
@@ -0,0 +1,60 @@
...
+  AppendTupleToBuffer(row_buffer, varlen_offset, varlen_entries);
+}
+
+void BufferedTupleWriter::AppendTupleToBuffer(ProjectedRow *row_buffer, int32_t total_varlen,
+                                              const std::vector<const VarlenEntry *> &varlen_entries) {
+  int32_t row_size = row_buffer->Size();
+  int32_t tot_size = total_varlen + row_size;
+  // Flush buffer first if the tuple cannot fit in buffer
+  if (cur_buffer_size_ + tot_size > block_size_) {
+    PersistBuffer();
```
yash620: This is blocking, can't you get a new buffer and add to that while this is asynchronously persisting in the background

src/include/storage/checkpoint_io.h
codeworm96: We have already planned async writing for checkpoint as part of our 100% goal. The synchronized write is only for our 75% goal, and it is only a temporary one. We will have an async implementation soon.

src/include/storage/checkpoint_manager.h
```diff
@@ -0,0 +1,113 @@
...
+#include "transaction/transaction_context.h"
+#include "transaction/transaction_defs.h"
+
+namespace terrier::storage {
+
+/**
+ * A CheckpointManager is responsible for serializing tuples of all tables in the database out and
+ * divide time into epochs by doing checkpoint.
+ */
+class CheckpointManager {
```
jrolli: I'm not sure I understand how you envision the API being used.  The class is called `CheckpointManager` but it appears that all of the management (scheduling of next checkpoint, starting and stopping checkpoint-related transactions, tracking of tables to persist, etc.) is expected to occur somewhere else.  Semantically, I would expect the checkpoint manager to have a single entry point that manages the lifecycle of taking a checkpoint and getter/setter functions for adjusting its parameters (tables being persisted, checkpoint frequency, etc.).  Specifically, it should be expecting access to threads and the worker pool to manage these tasks internally rather than expecting another component to do this for you.

src/include/storage/checkpoint_manager.h
```diff
@@ -0,0 +1,113 @@
...
+   */
+  explicit CheckpointManager(const char *log_file_path_prefix) : log_file_path_prefix_(log_file_path_prefix) {}
+
+  /**
+   * Start a new checkpoint with the given transaxtion context.
+   * @param txn transaction context this checkpoint will be running under.
+   */
+  void StartCheckpoint(transaction::TransactionContext *txn) {
+    txn_ = txn;
+    out_.Open((GetCheckpointFilePath(txn)).c_str());
```
yash620: This might be a good place to call Checkpoint on each SqlTable, since this is where the checkpoint it starting.

src/storage/checkpoint_io.cpp
yash620: Also this isn't the purpose for how RowView was envisioned to be used. It is for use inside of the storage layer as a way of templating between ProjectedRows and ProjectedColumns. Ideally you guys should be needing to access the BlockLayout (that shouldn't be exposed outside of the storage layer) to get the RowView.

src/include/storage/checkpoint_manager.h
```diff
@@ -0,0 +1,113 @@
...
+   * Begin a recovery. This will clear all registered tables and layouts.
+   */
+  void StartRecovery(transaction::TransactionContext *txn) { txn_ = txn; }
+
+  /**
+   * Register a table in the checkpoint manager, so that its content can be restored during the recovery.
+   * @param table to be recovered.
+   * @param layout of the table.
+   */
+  void RegisterTable(SqlTable *table, BlockLayout *layout) {
```
yash620: Ideally the CheckpointManager shouldn't be accessing the BlockLayout. This is only for use within the storage layer, at least that was the initial design choice with BlockLayout.

src/storage/checkpoint_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+  // 0 is version vector so we skip it
+  for (uint16_t col = NUM_RESERVED_COLUMNS; col < layout.NumColumns(); col++) {
+    all_col[col - NUM_RESERVED_COLUMNS] = storage::col_id_t(col);
+  }
+
+  // TODO(Mengyang): should calculate number of tuple to fit in buffer_size_
+  // now I only use a magic number here
+  uint32_t max_tuples = 10;
+
+  ProjectedColumnsInitializer column_initializer(layout, all_col, max_tuples);
```
yash620: Instead of using the block layout that is passed in, you guys should use SqlTable::InitializerForProjectedColumns or SqlTable::InitializerForProjectedRow

src/storage/checkpoint_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+  auto *scan_buffer = common::AllocationUtil::AllocateAligned(column_initializer.ProjectedColumnsSize());
+  ProjectedColumns *columns = column_initializer.Initialize(scan_buffer);
+
+  ProjectedRowInitializer row_initializer = ProjectedRowInitializer::CreateProjectedRowInitializer(layout, all_col);
```
yash620: Should use SqlTable::InitializerForProjectedRow

src/storage/checkpoint_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+  auto *redo_buffer = common::AllocationUtil::AllocateAligned(row_initializer.ProjectedRowSize());
+  ProjectedRow *row_buffer = row_initializer.InitializeRow(redo_buffer);
+
+  auto it = table.begin();
+  auto end = table.end();
+  while (it != end) {
+    table.Scan(txn_, &it, columns);
```
yash620: Using Select here instead of Scan can let you skip the entire part of trying to create a ProjectedRow in serialize tuple, you will just get a ProjectedRow back

jrolli:
What is your plan for ensuring the catalog is persisted in such a way that it is the first set of tables recovered during a recovery?  Specifically, your logic for `Recover` seems to assume that you already have all of the SqlTable OIDs and associated data loaded into the checkpoint manager, but `Recover` is also responsible for reading and reconstructing this from the checkpoint file.

src/storage/checkpoint_manager.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+void CheckpointManager::Recover(const char *log_file_path) {
+  BufferedTupleReader reader(log_file_path);
+
+  while (reader.ReadNextBlock()) {
+    CheckpointFilePage *page = reader.GetPage();
+    // TODO(zhaozhe): check checksum here
+    catalog::table_oid_t oid = page->GetTableOid();
+    SqlTable *table = GetTable(oid);
+    BlockLayout *layout = GetLayout(oid);
```
yash620: How are you getting this? If you are recovering and you aren't persisting this data, where is it coming from?

test/include/util/transaction_test_util.h
```diff
@@ -242,6 +242,11 @@ class LargeTransactionTestObject {
    */
   transaction::TransactionManager *GetTxnManager() { return &txn_manager_; }
 
+  /**
+   * @return the data table used by this test
+   */
+  storage::DataTable *GetTable() { return &table_; }
```
jrolli: You don't appear to use this, and you should not use this even in testing.  If there is something at the DataTable level that you need to interact with, even if it is for testing, then there is an API problem with SqlTable.  Recommend removing.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,254 @@
...
+    return true;
+  }
+
+  /**
+   * Read the next row from the buffer.
+   * @return pointer to the row if the next row is available.
+   *         nullptr if an error happens or there is no other row in the page.
+   */
+  ProjectedRow *ReadNextRow() {
+    if (block_size_ - page_offset_ < sizeof(uint32_t)) {
```
yash620: I don't see any resetting page_offset_. So what happens in the case you have already called ReadNextBlock() a few times and you are on some n'th block? In that case wouldn't page_offset_ be greater than the block_size so this check would always be true even if that block has data in it.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,254 @@
...
+namespace terrier::storage {
+// TODO(Zhaozhe): Should fetch block_size. Use magic number first.
+// The block size to output to disk every time
+#define CHECKPOINT_BLOCK_SIZE 4096
+
+/**
+ * The header of a page in the checkpoint file.
+ */
+// TODO(Zhaozhe, Mengyang): More fields can be added to header
+class CheckpointFilePage {
```
jrolli: If you're going to count on the data getting laid out as expected you may need to use the `PACKED` directive like is used for `ProjectedRow` (see include/storage/projected_row.h).

test/storage/checkpoint_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+#define CHECKPOINT_FILE_PREFIX "checkpoint_"
+
+namespace terrier {
+
+struct CheckpointTests : public TerrierTest {};
+
+// NOLINTNEXTLINE
+TEST_F(CheckpointTests, SimpleCheckpointNoVarlen) {
+  const uint32_t num_rows = 100;
+  const uint32_t num_columns = 3;
```
jrolli: You need a significantly larger number of tests.  Specifically, you should test both 1000s of columns and 1000s of rows because your current test fits completely inside one 4 kb "page".  @yash620 identified that you are probably not resetting your page offset between "blocks", but your tests miss this because you never advance to a new block/page.

test/include/util/storage_test_util.h
```diff
@@ -268,4 +275,173 @@ struct StorageTestUtil {
...
+    std::uniform_int_distribution<uint8_t> char_dist(0, UINT8_MAX);
+    std::uniform_int_distribution<int32_t> int_dist(INT32_MIN, INT32_MAX);
+
+    auto insert_buffer = common::AllocationUtil::AllocateAligned(pri_->ProjectedRowSize());
+    auto insert = pri_->InitializeRow(insert_buffer);
+
+    for (int i = 0; i < static_cast<int>(cols_.size()); i++) {
+      if (coin(*generator)) {  // not null
+        uint16_t offset = pr_map_->at(col_oids_[i]);
+        insert->SetNotNull(offset);
```
yash620: You don't need to SetNotNull AccessForceNotNull you do below already does that

test/storage/checkpoint_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+#include "util/transaction_test_util.h"
+
+#define CHECKPOINT_FILE_PREFIX "checkpoint_"
+
+namespace terrier {
+
+struct CheckpointTests : public TerrierTest {};
+
+// NOLINTNEXTLINE
+TEST_F(CheckpointTests, SimpleCheckpointNoVarlen) {
```
jrolli: This test (and the next) are not actually testing anything other than that your code doesn't segfault.   To unit test checkpoint you should be verifying that your serialized format looks like what your commented spec says it should.

test/storage/checkpoint_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+  std::string checkpoint_path = manager.GetCheckpointFilePath(checkpoint_txn);
+  // read first run
+  transaction::TransactionContext *scan_txn = txn_manager->BeginTransaction();
+  std::vector<std::string> original_rows;
+  tested.PrintAllRows(scan_txn, table, &layout, &original_rows);
+  txn_manager->Commit(scan_txn, StorageTestUtil::EmptyCallback, nullptr);
+  // recovery to another table
+  transaction::TransactionContext *recovery_txn = txn_manager->BeginTransaction();
+  storage::BlockStore block_store_{10000, 10000};
+  storage::SqlTable *recovered_table = new storage::SqlTable(&block_store_, *schema, catalog::table_oid_t(1));
```
yash620: In an actual recovery where will you get the schema from?

src/storage/checkpoint_io.cpp
codeworm96: Same as https://github.com/cmu-db/terrier/pull/341/files#r274158232

src/storage/checkpoint_io.cpp
yash620: After some discussion it looks like other people are using ProjectedColumns for purposes like you guys as well (https://github.com/cmu-db/terrier/pull/260#discussion_r274444563). So by using Scan you wouldn't have to reason about the visibility of the tuples since the SqlTable would handle that so it would be an easier process. To ensure that you guys don't need the BlockLayout we will try to work on being able to generate the RowView from a ProjectedColumn without needing BlockLayout

src/include/storage/checkpoint_io.h
songzhaozhe: Yes. We removed PACKED temporarily because of compilation error due to catalog::table_oid_t is a non-POD type. We might have to cast catalog::table_oid_t to POD type first to allow using PACKED.

src/include/storage/checkpoint_io.h
songzhaozhe: Thanks. We will fix it.

src/storage/checkpoint_io.cpp
lmy1229: Currently, we assume the size of a row is less than 4k bytes. We talked about this with @apavlo and he said we can start with this assumption. For the future extension, we plan to split the data of a row into several pages and use a flag bit in the page header to identify split rows.

src/storage/checkpoint_io.cpp
lmy1229: see https://github.com/cmu-db/terrier/pull/341#discussion_r275523714. Currently, we need to add an assertion to test the size of a row. This won't be a problem once row splitting is supported.

src/include/storage/checkpoint_io.h
songzhaozhe: Thanks. I think just zeroing the checksum bits will be fine.

test/storage/checkpoint_test.cpp
lmy1229: Normally, we should get the schema of a table from the catalog, but the catalog is not yet merged.

Another way to get the schema is to get it from the SqlTable itself. Do you consider adding an API for retrieving the schema from a SqlTable reasonable? Especially with the multi-version schema you guys are working on.

test/storage/checkpoint_test.cpp
```diff
@@ -0,0 +1,194 @@
...
+  tested.InsertRandomRows(num_rows, 0.2, &random_generator);
+
+  storage::CheckpointManager manager(CHECKPOINT_FILE_PREFIX);
+  storage::SqlTable *table = tested.GetTable();
+  storage::BlockLayout layout = tested.GetLayout();
+  catalog::Schema *schema = tested.GetSchema();
+  transaction::TransactionManager *txn_manager = tested.GetTxnManager();
+
+  // checkpoint
+  transaction::TransactionContext *checkpoint_txn = txn_manager->BeginTransaction();
```
saikiriti93: What is the life-cycle of the CheckpointManager object? Since it obviously doesn't live through a crash, shouldn't you be using a new instance of the CheckPointManager to simulate the actual recovery process?

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,254 @@
+#pragma once
+#include <vector>
+#include "storage/projected_columns.h"
+#include "storage/projected_row.h"
+#include "storage/write_ahead_log/log_io.h"
+
+namespace terrier::storage {
+// TODO(Zhaozhe): Should fetch block_size. Use magic number first.
```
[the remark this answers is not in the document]
songzhaozhe: No, it does not change over time. Ideally we hope to set it as the size of one disk read/write. The comment here means that maybe we should fetch its size from settings manager.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * Get the pointer to the first row in this page.
+   * @return pointer to the first row.
+   */
+  byte *GetPayload() { return varlen_contents_; }
+
+ private:
+  uint32_t checksum_;
+  catalog::table_oid_t table_oid_;
+  byte varlen_contents_[0];
```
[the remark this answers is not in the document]
songzhaozhe: Thanks. This should be a naming mistake.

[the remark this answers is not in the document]
lmy1229: We name this varlen_contents_ because all classes with a similar structure do this. For example ProjectedColumns and ProjectedRow.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,254 @@
...
+  /**
+   * Close current file and release the buffer.
+   */
+  void Close() {
+    PosixIoWrappers::Close(out_);
+    delete[] buffer_;
+  }
+
+  /**
+   * Serialize the tuple into the buffer, and write to disk when the buffer is full.
```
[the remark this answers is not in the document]
lmy1229: You are correct. This is a mistake in the comment. We will edit it later.

src/include/storage/checkpoint_io.h
lmy1229: Fixed.

src/include/storage/checkpoint_io.h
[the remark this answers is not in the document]
lmy1229: Fixed in the latest version

src/include/storage/checkpoint_io.h
lmy1229: Yes. In the latest version. the only place we need to set to zeros is after the last row in the page to suggest no more rows in this page. We do not need to reset the whole buffer anymore.

src/include/storage/checkpoint_io.h
lmy1229: resolved in the latest version

src/include/storage/checkpoint_io.h
lmy1229: This is correct.

src/include/storage/checkpoint_manager.h
lmy1229: You are correct. The checkpoint manager should 'manage' the process of checkpointing. However, we decided to first design the basic APIs for checkpoint and recovery, simply because there is no working catalog in the system. When there is a working catalog, a single call to StartCheckpoint() should first persist the catalog to the file. Then, for each user table, it will fetch its schema from the catalog and persist the table automatically.

src/include/storage/checkpoint_manager.h
lmy1229: Yes, this is the ultimate goal. However, we do not know anything about existing tables, because there is no working catalog yet.

src/include/storage/checkpoint_manager.h
lmy1229: Resolved in the latest version. We switched to catalog::Schema instead of BlockLayout.

src/include/storage/storage_defs.h
```diff
@@ -210,6 +210,20 @@ class VarlenEntry {
     return result;
   }
 
+  /**
+   * Constructs a new varlen entry for a checkpointed version. The offset if the offset of the varlen value in the
+   * checkpoint file, and it is stored in the prefix_ field of the VarlenEntry. The content_ field has no meaning.
```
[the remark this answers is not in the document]
lmy1229: This function is no longer used in our code, and was removed in the latest version

src/storage/checkpoint_io.cpp
lmy1229: We abandoned Scan and switched to Select in the latest version, and we do not require BlockLayout anymore. The problem becomes how can we get the schema of a table because we need the Oids of all its columns. Since there are no catalogs yet, is it reasonable to add an API in the SqlTable to return its schema?

src/storage/checkpoint_io.cpp
lmy1229: Thanks for the advice. We changed our code in the latest version.

src/storage/checkpoint_io.cpp
lmy1229: resolved in the latest version

src/storage/checkpoint_manager.cpp
lmy1229: Thanks for the advice. We have changed it.

src/storage/checkpoint_manager.cpp
lmy1229: Already changed.

src/storage/checkpoint_manager.cpp
lmy1229: We do not use BlockLayouts anymore, but we still need to get the schema of a table anyway. This information should come from the catalog, which we will recover first in the recovery process.

test/include/util/storage_test_util.h
lmy1229: resolved.

test/storage/checkpoint_test.cpp
lmy1229: You are right. These 2 tests were written when there was no recovery. Already removed.

src/storage/checkpoint_io.cpp
jrolli: I would recommend against this because it would lead to more work in the end.  Instead, I would recommend either cherry-picking the API/files you need from the Catalog PR or laying out the API you need from the catalog and mocking it out in your test fixture so that it always provides some hard-coded or programmatically generated schema.

pervazea:
See comments in issue #304 regarding interaction of the catalog and checkpoint / recovery.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,500 @@
...
+   * Get the pointer to the first row in this page.
+   * @return pointer to the first row.
+   */
+  byte *GetPayload() { return payload_; }
+
+  /**
+   * Set all the flags at the same time.
+   * @param flags to be set.
+   */
+  void SetFlags(uint32_t flags) { flags_ = flags; }
```
yash620: There is not mention of flags in your design doc images (https://drive.google.com/file/d/16bLiW8dKNqBdfS7nP2iAUIGgGKKBESPp/view). It might help to edit them to include the flags.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,500 @@
...
+    TERRIER_ASSERT(buf != nullptr, "Buffer pointer should not be nullptr");
+    pending_.Enqueue(buf);
+  }
+
+  /**
+   * Return a buffer to the writer, make it available again
+   * without writing it
+   * @param buf the pointer to the buffer to be returned
+   */
+  void ReturnBuffer(byte *buf) {
```
yash620: Should this be exposed in the interface? In what situations would someone need to GetBuffer and return it without calling WriteBuffer. I only see it being used in BufferedTupleWriter.close, which also calls AsyncBlockWriter.close(), so why no just clean up the free queue in that function?

src/include/storage/checkpoint_io.h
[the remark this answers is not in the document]
yash620: Isn't it possible that this value, since it's set by the SettingsManager, is changed when the database is shut down and restarted? So would you have to account for having to recover from a checkpoint is written in a different block_size than the block_size SettingsManager gave you?

src/storage/checkpoint_io.cpp
```diff
@@ -0,0 +1,96 @@
+#include "storage/checkpoint_io.h"
+#include <catalog/schema.h>
+#include <vector>
+
+namespace terrier::storage {
+
+void AsyncBlockWriter::Open(const char *log_file_path, int buffer_num) {
+  out_ = PosixIoWrappers::Open(log_file_path, O_WRONLY | O_CREAT, S_IRUSR | S_IWUSR);
```
yash620: Do you guys needs read permissions? I don't think AsyncBlockWriter reads anything from the file so you would only need write permissions.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,500 @@
...
+  }
+
+  void PersistBuffer() {
+    // TODO(zhaozhe): calculate CHECKSUM. Currently using default 0 as checksum
+    if (page_offset_ == sizeof(CheckpointFilePage)) {
+      // If the buffer has no contents, just return
+      return;
+    }
+    AlignBufferOffset<uint64_t>();
+    if (block_size_ - page_offset_ > sizeof(uint32_t)) {
```
yash620: What happens in the case where (block_size_ - page_offset_) <= sizeof(uint32_t). In BufferedTupleWriter::SerializeTuple you guys only check (page_offset_ + checkpoint_record_size > block_size_) which doesn't take into account these extra 32 bits that need to be zeroed. So it's possible that for some block that this zeroing out won't be happening at the end of some blocks.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,500 @@
+#pragma once
```
yash620: There's a ton of stuff that's happening in this file, many classes. Ideally this should be split apart into multiple files to be maintainable and readable.

src/storage/checkpoint_manager.cpp
```diff
@@ -0,0 +1,232 @@
...
+    auto txn_commit = in->ReadValue<transaction::timestamp_t>();
+    // Okay to fill in null since nobody will invoke the callback.
+    // is_read_only argument is set to false, because we do not write out a commit record for a transaction if it is
+    // not read-only.
+    return storage::CommitRecord::Initialize(buf, txn_begin, txn_commit, nullptr, nullptr, false, nullptr);
+  }
+  auto table_oid = in->ReadValue<catalog::table_oid_t>();
+  auto tuple_slot = in->ReadValue<storage::TupleSlot>();
+  if (record_type == storage::LogRecordType::DELETE) {
+    // TODO(Justin): set a pointer to the correct data table? Will this even be useful for recovery?
```
jrolli: Isn't this for recovery?  Is this comment still current?

src/storage/checkpoint_manager.cpp
```diff
@@ -0,0 +1,232 @@
...
+  //  probably faster, but stick with the more naive way for now. We need a vector, not just a col_id_t[], because
+  //  that is what ProjectedRowInitializer needs.
+  std::vector<catalog::col_oid_t> col_oids(num_cols);
+  for (uint16_t i = 0; i < num_cols; i++) {
+    const auto col_id = in->ReadValue<storage::col_id_t>();
+    col_oids[i] = table->ColOidForId(col_id);
+  }
+  // Initialize the redo record.
+  auto row_pair = table->InitializerForProjectedRow(col_oids);
+  // TODO(Justin): set a pointer to the correct data table? Will this even be useful for recovery?
```
jrolli: Should this comment be here?

src/storage/data_table.cpp
```diff
@@ -1,4 +1,5 @@
 #include "storage/data_table.h"
+#include <loggers/storage_logger.h>
```
jrolli: Shouldn't this be quotes?

src/storage/sql_table.cpp
```diff
@@ -6,11 +6,11 @@
 
 namespace terrier::storage {
 
-SqlTable::SqlTable(BlockStore *const store, const catalog::Schema &schema, const catalog::table_oid_t oid)
-    : block_store_(store), oid_(oid) {
+SqlTable::SqlTable(BlockStore *const store, catalog::Schema schema, const catalog::table_oid_t oid)
+    : block_store_(store), oid_(oid), schema_(std::move(schema)) {
```
jrolli: This is a very significant API change.  If you're going to pass it in what is the justification for changing it from a reference to a move rather than a copy?

src/include/storage/sql_table.h
```diff
@@ -170,10 +177,37 @@ class SqlTable {
...
+  storage::DataTable *get_data_table() { return table_.data_table; }
+
+  /**
+   * This is extremely hacky, just for test purposes only. Because the current transaction and logging infrastructure
+   * are all based on data tables, but checkpointing is implemented on sql tables. Migrating existing API from data
+   * tables to sql tables would require much effort that would need to modify many tests, and this is left for future
+   * work. The current workaround is to get the underlying data table and do not modify existing infrastructure.
+   * @return the data table underlying sqltable
+   */
+  storage::BlockLayout get_layout() { return table_.layout; }
```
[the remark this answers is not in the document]
songzhaozhe: Since existing logging and transaction infrastructure uses data table in APIs, this hack here is required to test our existing code. Because we have to get a data_table and layout for logging purposes.
Migrating logging and transaction infrastructure to SqlTable seems out of our project's scope (involving many transaction layer APIs as well) and we think is dangerous because it breaks many tests. So we think there should better be a separate PR addressing that in the future. 
In this sense this PR would indeed not be "merge ready" before the infrastructure has changed, but we are wondering if the workaround is acceptable here and be fixed in the future.

src/storage/checkpoint_manager.cpp
```diff
@@ -0,0 +1,232 @@
...
+          TERRIER_ASSERT(0, "Update failed during log recovery");
+        }
+      }
+    }
+    delete[] reinterpret_cast<byte *>(log_record);
+  }
+}
+
+storage::LogRecord *CheckpointManager::ReadNextLogRecord(storage::BufferedLogReader *in) {
+  // TODO(Justin): Fit this to new serialization format after it is complete.
```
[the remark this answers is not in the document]
codeworm96: He is the author of #263. We took over his branch since he has graduated and we need the new log serialization format for varlens.

src/storage/checkpoint_manager.cpp
```diff
@@ -0,0 +1,232 @@
...
+    SqlTable *table = GetTable(oid);
+
+    ProjectedRow *row = nullptr;
+    while ((row = reader.ReadNextRow()) != nullptr) {
+      TupleSlot *slot UNUSED_ATTRIBUTE = reader.ReadNextTupleSlot();
+      // loop through columns to deal with non-inlined varlens.
+      for (uint16_t projection_list_idx = 0; projection_list_idx < row->NumColumns(); projection_list_idx++) {
+        if (!row->IsNull(projection_list_idx)) {
+          catalog::col_oid_t col = table->ColOidForId(row->ColumnIds()[projection_list_idx]);
+          if (table->GetSchema().GetColumn(col).IsVarlen()) {
```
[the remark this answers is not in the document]
lmy1229: The point is that even if we recover the catalog first, we wouldn't be able to get the schema, because the catalog is not ready. It lacks fundamental APIs.

src/include/storage/sql_table.h
```diff
@@ -170,10 +177,37 @@ class SqlTable {
...
+
+  /**
+   * Given a col_id, return the corresponding col_oid
+   * @param col_id a col_id, must be in the table's InverseColumnMap
+   * @return col_oid for the col_id
+   */
+  catalog::col_oid_t ColOidForId(const col_id_t col_id) const { return table_.inverse_column_map.at(col_id); }
```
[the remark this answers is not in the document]
lmy1229: This one is added by the catalog group for some reason unknown

src/include/storage/sql_table.h
```diff
@@ -38,13 +39,19 @@ class SqlTable {
...
   /**
    * Destructs a SqlTable, frees all its members.
    */
   ~SqlTable() { delete table_.data_table; }
 
+  /**
+   * Get the schema use of the SQL table
+   * @return the schema
+   */
+  catalog::Schema &GetSchema() { return schema_; }
```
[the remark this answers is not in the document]
lmy1229: This one is added by the catalog group, although it does solve many of our problems

src/include/storage/sql_table.h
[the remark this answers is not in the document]
codeworm96: From my point of view, it is not a hack to make our code pass CI checks. It is only a hack that enables us to use the existing test infrastructure to test our code. Since the existing test infrastructure is still based on DataTable and Layout, we need some hacky way to get the Layout to use it. However, our code is still get tested and it does not change the result of the test itself. We never use that API in our implementation code.

src/storage/checkpoint_manager.cpp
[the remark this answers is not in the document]
jrolli: Then update these TODO's to tag someone who is still here with the responsibility to fix/answer the concern.

src/include/storage/sql_table.h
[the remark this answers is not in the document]
codeworm96: Unfortunately, the DeltaRecords given by transactions contain only IDs. So we can only log IDs, when recovering we need some way to convert IDs to OIDs. I agree that ideally, users should work with SqlTable. However, unfortunately, almost everyone is using DataTable directly and violating the abstraction now. Changing this fact needs cooperation. Our team alone is unable to fix everything.

src/include/storage/data_table.h
```diff
@@ -207,6 +207,12 @@ class DataTable {
    */
   DataTableCounter *GetDataTableCounter() { return &data_table_counter_; }
 
+  /**
+   * Returns a read-only view of this DataTable's BlockLayout.
+   * @return this DataTable's BlockLayout.
+   */
+  const BlockLayout &GetBlockLayout() const { return accessor_.GetBlockLayout(); }
```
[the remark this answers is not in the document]
codeworm96: Similar reason as https://github.com/cmu-db/terrier/pull/341#discussion_r282733370. Currently, when transactions want to log, they only provide DeltaRecords and the DataTable. And most transactions have no idea about SqlTable. (In fact, except the tests written by us, almost all tests directly work on DataTables. They even do not create a SqlTable object.) Therefore, we currently are unable to avoid the DataTable level things.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,500 @@
...
+
+  /**
+   * The writer thread
+   */
+  void RunWriter();
+};
+
+/**
+ * A buffered tuple writer collects tuples into blocks, and output a block once a time.
+ * The block size is fetched from database setting, and should be the page size.
```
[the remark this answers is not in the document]
codeworm96: Based on SiloR paper and our benchmark results, 512K checkpoint page size is better. We already switched to that.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,500 @@
...
+
+  const common::RawBitmap &Bitmap() const { return *reinterpret_cast<const common::RawBitmap *>(&flags_); }
+};
+
+/**
+ * An async block writer will use a background thread to write blocks into
+ * checkpoint file asynchronously. Its design is very similar with that of
+ * the logger threads in SiloR. An async block writer will have some
+ * pre-allocated block buffers. Workers must first ask for a free block
+ * buffer. If there is no free block buffer now, it will be blocked until
```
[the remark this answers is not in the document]
codeworm96: If the system changes to use a different method to allocate threads, the best design may also get changed. But currently, based on our benchmark, 512K page size + blocking is the best combination.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,500 @@
...
+    async_writer_.Open(log_file_path, CHECKPOINT_BUF_NUM);
+    block_size_ = CHECKPOINT_BLOCK_SIZE;
+    buffer_ = async_writer_.GetBuffer();
+    ResetBuffer();
+  }
+
+  /**
+   * Write the content of the buffer into disk.
+   */
+  void Persist() { PersistBuffer(); }
```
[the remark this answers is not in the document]
songzhaozhe: Removed.

src/storage/checkpoint_io.cpp
songzhaozhe: Fixed

src/storage/sql_table.cpp
codeworm96: This change is prompted by clang-tidy. It thinks, for c++17, since we have std::move, passing a value and then move it is better than passing a reference and then copy it inside the function body. For callers, this change has no impact. They do not need to change anything.

src/storage/checkpoint_io.cpp
codeworm96: I guess you miss understood this permission. The `S_IRUSR` permission is stored in the filesystem metadata. For the file descriptor itself, it is indeed write only (`O_WRONLY`). If the permission of the file itself is write-only, nobody could read the content of that file. So how can someone recover it later?

src/include/storage/checkpoint_manager.h
```diff
@@ -0,0 +1,202 @@
...
+   * inside a transaction.
+   * Table and schema are temporary, for test purposes only. They should be fetched from catalogs.
+   * @param txn
+   * @param table
+   * @param schema
+   */
+
+  void Process(transaction::TransactionContext *txn, const SqlTable &table, const catalog::Schema &schema) {
+    StartCheckpoint(txn);
+    // TODO(zhaozhes): This should actually iterate through all tables, using catalog information
```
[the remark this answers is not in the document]
songzhaozhe: As comments implied, table and schema passed in are temporary backdoors for testing. After catalog is finished, they should be obtained from catalog.

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -3,6 +3,7 @@
 #include <sys/stat.h>
 #include <sys/types.h>
 #include <sys/uio.h>
+#include <transaction/transaction_context.h>
```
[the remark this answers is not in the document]
songzhaozhe: Removed.

src/storage/data_table.cpp
songzhaozhe: Removed.

src/include/storage/sql_table.h
```diff
@@ -170,10 +177,37 @@ class SqlTable {
...
+   * @param col_id a col_id, must be in the table's InverseColumnMap
+   * @return col_oid for the col_id
+   */
+  catalog::col_oid_t ColOidForId(const col_id_t col_id) const { return table_.inverse_column_map.at(col_id); }
+
  private:
   BlockStore *const block_store_;
   const catalog::table_oid_t oid_;
 
+  catalog::Schema schema_;
```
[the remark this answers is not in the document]
codeworm96: This is the current catalog implementation, which has been merged into the master. It is out of our control.

src/include/storage/sql_table.h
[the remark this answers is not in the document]
codeworm96: This is the current catalog implementation, which has been merged into the master. It is out of our control.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,500 @@
...
+        memcpy(tmp_buffer + already_read, buffer_ + page_offset_, left_to_read);
+        page_offset_ += left_to_read;
+        loose_ptrs_.emplace_back(tmp_buffer);
+        return tmp_buffer;
+      }
+      // remaining buffer is not enough
+      memcpy(tmp_buffer + already_read, buffer_ + page_offset_, remaining_buffer);
+      if (!ReadNextBlock()) {
+        // should not happen, unless the checkpoint file is corrupted.
+        // TODO(Mengyang): figure out what kind of exception to throw here.
```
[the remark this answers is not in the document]
codeworm96: I don't think `TERRIER_ASSERT` is the right solution. Under non-debug environment, it just does nothing.

src/include/storage/checkpoint_io.h
```diff
@@ -0,0 +1,500 @@
...
+
+  uint32_t GetVersion() { return version_; }
+
+ private:
+  uint32_t checksum_;
+  // Note: use uint32_t instead of table_oid_t, because table_oid_t is not POD so that
+  // we cannot use PACKED in this class.
+  uint32_t table_oid_;
+  uint32_t version_;  // version of the schema(useful with schema change), or, for catalog tables,
+                      // serves as the type of the catalog table.
```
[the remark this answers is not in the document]
lmy1229: This is not an optimization, but to put all the information needed in the header. There are mainly 2 reasons I didn't overload `flags_` field. 
1. Currently, there is only one bit in the `flags_` used, but maybe in the future, there will be more flags needed.
2. It may be more semantically correct to overload the table schema version field as the catalog table type, assuming catalog schemas are never changed.

src/storage/sql_table.cpp
codeworm96: We updated this constructor according to the latest master and clang-tidy stopped prompting this change. So we reverted this API change.

src/include/storage/checkpoint_io.h
codeworm96: AsyncBlockWriter allocates these buffers. Therefore, these buffers had better be freed by the AsyncBlockWriter itself. Requiring BufferedTupleWriter to free these buffers leaks the detail of buffer management out of the AsyncBlockWriter abstraction, which makes it harder to ensure the lifetime of these buffers is correct. I think it is not the right thing to do.

src/storage/checkpoint_manager.cpp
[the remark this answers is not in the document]
codeworm96: According to the project meeting, we plan to implement catalog recovery in this way:

Start DBMS for the first time:
1. Create catalog tables.
2. Fill catalog tables with initial content.
3. Initialize catalog data structure according to contents of catalog tables.

Recovering DBMS:
1. Create catalog tables.
2. Recovery code fills the catalog tables with contents in the checkpoint.
3. Recovery code notifies catalog to initialize its data structure according to contents of catalog tables (including creating user tables).
4. Recovery code fills the user tables with contents in the checkpoint.
5. Recovery code recovers the log. During the process, we may encounter records that modify catalog tables. We will notify catalog to update correspondingly after redoing them.

Currently, the catalog does not meet our requirement:
1) It always fills catalog tables with initial content when creating catalog tables.
2) No API to notify that we have updated catalog tables and catalog needs to update correspondingly.

PR #342, opened 2019-04-08
Lazy Schema Change

yash620:
For the design overview please take a look at the DESIGN.md file. Most of our modifications are localized to sql_table.cpp and sql_table.h . We have added tests in sql_table_test.cpp and catalog_test_util.h, and we have benchmarks in sql_table_benchmark.cpp


DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### SqlTable::Insert
+Insert will always insert the passed in tuple into the DataTable of the schema version number passed in
+
+#### SqlTable::Update
+- Update has three cases
+    1. The expected schema version matches the actual schema version
+        - The update will happen in place on the DataTable of the actual schema version
+    2. The expected schema version doesn’t match the actual schema version but the update doesn’t touch any columns not in the actual schema version
+        - The update will happen in place on the DataTable of the expected schema version
```
codeworm96: It is a bit confusing here. You have the following definition:
> Below we will refer to expected version as the version of the tuple the user expects to see and actual version as the version the tuple is currently in.

How can you update in place on the DataTable of expected schema version while the tuple is currently in a different version?

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+- The iterator for Scan must always begin on the latest version
+    - In cases where the user is interchanging scan call and update calls, if the scan iterator were to start on an older version then it is possible for a tuple retrieved from the scan to be updated, which could move it into the latest schema version
+    - This would mean that when the iterator gets to the latest schema version it will read all the tuples in the DataTable for that version and this would cause that tuple to have been read twice within the scan.
+
+#### SqlTable::Select
+- There are two cases
+    1. Expected schema version matches actual schema version
+        - The tuple is directly selected from the DataTable for that schema version
+    2. Expected schema version differs from the actual schema version
+        - The tuple is selected from the DataTable for that schema version and during this process it is transformed to the expected schema version
```
codeworm96: Will the transformation also happen in the underlying storage? Or it is only a temporary result and then be thrown away. I think it is better to clarify that in the document.

DESIGN.md
jrolli: Thanks!  This is a typo.  The in-place update occurs inside of the DataTable of the actual version.

DESIGN.md
jrolli: The transformation here is temporary.  The only time we persist the transformation is in case 3 for update where the tuple is moved between DataTables via a delete and insert.

src/include/storage/sql_table.h
```diff
@@ -27,6 +31,110 @@ class SqlTable {
...
+     */
+    void AdvanceOnEndOfDatatable_() {
+      if (current_it_ == tables_->Find(curr_version_)->second.data_table->end()) {
+        // layout_version_t is uint32_t so we need to protect against underflow.
+        if (!curr_version_ == 0) {
+          is_end_ = true;
+          return;
+        }
+        curr_version_--;
+        TERRIER_ASSERT(curr_version_ < txn_version_, "Current version must be older than transaction");
```
codeworm96: Why put the assertion here? We just decremented the curr_version_. It is unlikely that the current version will become greater or equal to the transaction version unless underflow happens. But you have checked that before.

src/include/storage/sql_table.h
```diff
@@ -27,6 +31,110 @@ class SqlTable {
...
+     * Checks if DataTable::SlotIterator is at the end of the table and advances
+     * the iterator to the next table or sets is_end flag as appropriate.  This
+     * refactor is necessary to keep iterator in correct state when used in two
+     * distinct ways:  increment called directly on this object or implicitly
+     * through advancement of the DataTable iterator in SqlTable::Scan.
+     */
+    void AdvanceOnEndOfDatatable_() {
+      if (current_it_ == tables_->Find(curr_version_)->second.data_table->end()) {
+        // layout_version_t is uint32_t so we need to protect against underflow.
+        if (!curr_version_ == 0) {
```
codeworm96: `!curr_version_ == 0` will be true when curr_version_ is not 0. The condition is confusing. It seems to have nothing to do with the underflow.

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+- Update will be required to take in a schema version number that indicates the expected schema version number of the tuple being updated
+- In this case the actual version number will differ from the expected version in cases where the tuple being updated has not been shifted to the expected version.
+- If the update modifies columns that are not in the actual version then the tuple will be shifted to the expected version before applying the update.
+
+### SqlTable::Scan
+- Scan will be required to take in a schema version number that indicates the schema version number that the current transaction sees.
+
+### SqlTable::Delete
+- Delete will be required to take in the schema version number that indicates the expected schema version number of the tuple being deleted
+
```
codeworm96: You also modified `InitializerForProjectedColumns`, `InitializerForProjectedRows` .etc. You maybe also need to document them here.

src/storage/sql_table.cpp
```diff
@@ -1,34 +1,324 @@
...
+    auto new_buffer = common::AllocationUtil::AllocateAligned(new_pair.first.ProjectedRowSize());
+    ProjectedRow *new_pr = new_pair.first.InitializeRow(new_buffer);
+    bool valid = Select(txn, slot, new_pr, new_pair.second, version_num);
+    if (!valid) {
+      delete[] new_buffer;
+      return {false, slot};
+    }
+    // 3. Delete the old row
+    bool succ = tables_.Find(old_version)->second.data_table->Delete(txn, slot);
+
```
codeworm96: Consider an interleaving: here another thread begins a scan and runs to finish. The row being updated will be missing in the scan result. I haven't seen any mechanism that prevents this interleaving from happening.

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+        - The update will happen in place on the DataTable of the expected schema version
+    3. The expected schema version doesn’t match the actual schema version and the update touches not in the actual schema version. The following steps occur
+        - Retrieve the tuple from the actual version DataTable
+        - Transform the tuple to the expected version
+        - Delete the tuple from the actual version DataTable
+        - Insert the tuple into the expected version DataTable
+
+#### SqlTable::Scan
+- SqlTable will maintain its own slot iterator which will be used to iterate across all the schema version. The iterator interface exposed to the user will not change
+- The iterator for Scan must always begin on the latest version
```
codeworm96: I am doubting if this method really solves every problem. See the trace in https://github.com/cmu-db/terrier/pull/342/files#r275173692

src/include/storage/sql_table.h
jrolli: `!curr_version_` is not actually a negation.  For brevity, they overloaded the `!` operator to make it static cast their strong typedefs (such as `storage::layout_version_t`) to a standard integer form.  This is logically equivalent to `curr_version_ == 0` but you need the type cast to allow the comparison.

src/include/storage/sql_table.h
jrolli: The assertion is here because we had an underflow in the first draft.  I agree that it should probably move to an assertion of `curr_version_ <= txn_version_` at the start of the function.

src/storage/sql_table.cpp
jrolli: MVCC prevents this interleaving.  Either the delete and insert are uncommitted when the scan starts and it reads the version that is to be "deleted", or the scan started after the commit and sees the updated version.

src/include/storage/sql_table.h
codeworm96: I see. I haven't noticed that the `!` is for casting while reading the code.

src/storage/sql_table.cpp
codeworm96: Is it possible that the scan and the update belongs to the same transactions so that MVCC won't help?

jrolli: This behavior would be undefined in the current version since only installing the undo record is atomic.  Any concurrent read/update under the same transaction’s context would have the same problem in the non-versioned schema setting.

DESIGN.md
codeworm96: As the discussion in https://github.com/cmu-db/terrier/pull/342/files#r275175240, it seems it is not the actual solution you used to solve the scan/update conflict. You rely on MVCC to solve the conflict. Maybe the document also need to change accordingly.

test/storage/sql_table_concurrent_test.cpp
```diff
@@ -0,0 +1,379 @@
...
+    for (auto &version : versioned_col_oids) delete version;
+    versioned_col_oids.clear();
+    // End concurrent section
+    // delete init_txn;
+  }
+}
+
+// NOLINTNEXTLINE
+TEST_F(SqlTableConcurrentTests, ConcurrentQueriesWithSchemaChange) {
+  // const uint32_t num_iterations = 5;
```
codeworm96: This file has many commented out lines, even including two tests, which does not look good.

src/include/storage/sql_table.h
```diff
@@ -27,6 +31,110 @@ class SqlTable {
...
+     */
+    const TupleSlot *operator->() const { return &(*current_it_); }
+
+    /**
+     * pre-fix increment.
+     * @return self-reference after the iterator is advanced
+     */
+    SlotIterator &operator++() {
+      TERRIER_ASSERT(!is_end_, "Cannot increment an end iterator");
+      current_it_++;
```
lmy1229: Is it possible that a dataTable has no data so that its iterator starts at its end? In this case, you will increase the iterator beyond its boundary and not aware of that, because AdvanceOnEndOfDatatable_() has not yet been called.

src/include/storage/sql_table.h
```diff
@@ -27,6 +31,110 @@ class SqlTable {
...
+     */
+    bool operator!=(const SlotIterator &other) const { return !this->operator==(other); }
+
+    DataTable::SlotIterator *GetDataTableSlotIterator() { return &current_it_; }
+
+   private:
+    friend class SqlTable;
+
+    SlotIterator(const common::ConcurrentMap<layout_version_t, DataTableVersion> *tables,
+                 const layout_version_t txn_version, bool is_end)
```
lmy1229: How can a running transaction know which 'txn_version' to use in its scans? Your implementation in the AdvanceOnEndOfDatatable_() suggests that layout versions are incremented sequentially, so the txn cannot simply pass its txn_id here.

src/storage/sql_table.cpp
```diff
@@ -1,34 +1,324 @@
...
+  while (tables_.CBegin() != tables_.CEnd()) {
+    auto pair = *(tables_.CBegin());
+    delete (pair.second.data_table);  // Delete the data_table object on the heap
+    tables_.UnsafeErase(pair.first);
+  }
+}
+
+void SqlTable::UpdateSchema(const catalog::Schema &schema) {
+  STORAGE_LOG_DEBUG("Update schema version: {}", uint32_t(schema.GetVersion()));
+  TERRIER_ASSERT(tables_.Find(schema.GetVersion()) == tables_.End(), "schema versions for an SQL table must be unique");
```
lmy1229: How do you handle concurrent updates? Also, how do you enforce that layout versions are incremented sequentially? If there are gaps in the schema versions, it may cause your iterator to stop prematurely.

benchmark/storage/sql_table_benchmark.cpp
```diff
@@ -0,0 +1,676 @@
...
+  catalog::col_oid_t col_oid(column_num_);
+  std::vector<catalog::Schema::Column> new_columns(columns_.begin(), columns_.end() - 1);
+  new_columns.emplace_back("", type::TypeId::BIGINT, false, col_oid);
+  catalog::Schema new_schema(new_columns, storage::layout_version_t(1));
+  table_->UpdateSchema(new_schema);
+
+  // create a new read buffer
+  std::vector<catalog::col_oid_t> all_col_oids(new_columns.size());
+  for (size_t i = 0; i < new_columns.size(); i++) all_col_oids[i] = new_columns[i].GetOid();
+  auto pair = table_->InitializerForProjectedRow(all_col_oids, storage::layout_version_t(1));
```
codeworm96: In tests, you used hardcoded layout version for test. How can the user of the SQL table know, according to its transaction id, which layout version (schema) should be used? I haven't found any related API. `Schema::GetVersion()` has no parameter. It seems not the right one for this purpose.

src/storage/sql_table.cpp
jrolli: Concurrent updates are prevented because the updating must get a write-lock on the table's schema version in the catalog and therefore any subsequent writers will abort because of a write-write conflict.  This approach also guarantees sequential versions because an update will increment their visible version.  Aborted updates won't be visible and pending versions will cause a write-write conflict.

DESIGN.md
jrolli: This requirement is solving a different problem than what was posed there.  MVCC solves the problem of a race across transactions during the insert/delete process, but this addresses the risk of processing the same logical tuple twice.  Specifically, if we are using a Hyper-style execution model of process a group of tuples as far as we can and this pipeline involves an update that forces migration to the latest version, then we **must** process the latest version first.

For example, if we add a column and then systematically set the value of this column (via update and not a default value at the time we changed the schema), then we will migrate every tuple to the current version.  If we iterate over older versions first, then when we reach the current version they will be visible because a transaction can see its own writes.  While we could handle this by tracking migrated tuples explicitly, it is cleaner to iterate over the current version first so that when we trigger migrations they are inserted behind the `SlotIterator` and won't be visited again.

src/include/storage/sql_table.h
jrolli: I'll double check the semantics of the iterator, but I think C++ generally demands that incrementing `end` becomes a no-op.  However, this is an undocumented assumption that we either need to confirm and document or correct.

test/storage/sql_table_concurrent_test.cpp
jrolli: You are correct.  These are placeholders for two more tests that we need to write.

src/include/storage/sql_table.h
jrolli: A transaction has a single `txn_version` that is constant across its entire lifetime and is determined by what is visible in the catalog when the binder executes.  `AdvanceOnEndOfDatabase_` detects when we've iterated over all of the tuples in a data table and advances to the next data table (newest to oldest) when necessary.

benchmark/storage/sql_table_benchmark.cpp
jrolli: The visible schema (layout) version must be provided to the query by the binder through a query to the catalog.

DESIGN.md
codeworm96: I see. So the user must guarantee that each update/scan call does not interleave with each other. But they can call scan, then update, then scan, etc, which is solved by this method.

DESIGN.md
codeworm96: In the document, you said that your benchmark showed that reading a tuple which version doesn't match has a penalty to a factor of 5x. Consider the case that a hot tuple is in an older version. Each read needs to pay a 5x penalty. If the tuple is hot, we will transform the tuple again and again, and the overhead could be large. I think you may need some policy to migrate the tuple in such case, even we are only reading the tuple, to reduce the overhead.

src/include/storage/sql_table.h
codeworm96: The increment of the end iterator is undefined in STL. They don't say that it must be a no-op.

DESIGN.md
songzhaozhe: Does it mean we are only allowed to update the scanned tuples in the same transaction? For example, if we updated an unscanned tuple and it was migrated to an already scanned data table, then we would never be able to see it in the scan.

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+    - In cases where the user is interchanging scan call and update calls, if the scan iterator were to start on an older version then it is possible for a tuple retrieved from the scan to be updated, which could move it into the latest schema version
+    - This would mean that when the iterator gets to the latest schema version it will read all the tuples in the DataTable for that version and this would cause that tuple to have been read twice within the scan.
+
```
songzhaozhe: According to the code, under case "Expected schema version differs from the actual schema version", Scan() also returns the temporarily transformed tuples, similar to Select() right? Maybe the doc can make this clearer here.

src/include/storage/sql_table.h
```diff
@@ -180,14 +319,16 @@ class SqlTable {
   const catalog::table_oid_t oid_;
 
   // Eventually we'll support adding more tables when schema changes. For now we'll always access the one DataTable.
```
songzhaozhe: Is this comment out of date?

src/storage/sql_table.cpp
```diff
@@ -1,34 +1,324 @@
...
+  BlockLayout layout = storage::BlockLayout(attr_sizes);
+
+  auto dt = new DataTable(block_store_, layout, schema.GetVersion());
+  // clang's memory analysis has a false positive on this allocation.  The TERRIER_ASSERT on the second line of this
+  // function prevents the insert below from failing (can only fail when key is not unique).  The write-lock on the
+  // catalog prevents any other transaction from being in a race condition with this one.  The corresponding delete
+  // for this allocation is in the destructor for SqlTable.  clang-analyzer-cplusplus.NewDeleteLeaks identifies this
+  // as a potential leak and throws an error incorrectly.
+  // NOLINTNEXTLINE
+  tables_.Insert(schema.GetVersion(), {dt, layout, col_map, inv_col_map});
```
songzhaozhe: Consider a user frequently updates the schema. Is there any chance for this map to shrink? Can some entries in the map be deleted when a table becomes empty and would never be used again?
Do we have to worry about this? Is GC responsible for this?

jrolli: At the moment there is no way to shrink the table (short of leveraging checkpoint/recovery to restart the process).  The stretch/longterm goal is to implement a compactor that would work on reducing the number of versions that exist.

DESIGN.md
jrolli: If the update occurred in a different transaction, then the scan will see the old "deleted" tuple because the other transactions writes are not visible to the scan (snapshot isolation).

src/include/storage/sql_table.h
jrolli: Fixed by making `AdvanceOnEndOfDatatable_()` loop until it finds a non-empty DataTable.

src/include/storage/storage_defs.h
```diff
@@ -176,6 +177,12 @@ using ColumnMap = std::unordered_map<catalog::col_oid_t, col_id_t>;
  * Used by execution and storage layers to map between col_oids and offsets within a ProjectedRow
  */
 using ProjectionMap = std::unordered_map<catalog::col_oid_t, uint16_t>;
+using InverseColumnMap = std::unordered_map<col_id_t, catalog::col_oid_t>;
+/**
+ * Used by SqlTable to map between col_oids in Schema and their {default_value, attribute_size}
+ */
+// using DefaultValueMap = std::unordered_map<catalog::col_oid_t, std::pair<byte *, uint8_t>>;
```
songzhaozhe: Redundant line?

src/include/storage/sql_table.h
```diff
@@ -1,9 +1,14 @@
 #pragma once
 #include <list>
+#include <map>
```
songzhaozhe: Is this include still needed?

test/storage/sql_table_test.cpp
```diff
@@ -38,62 +39,134 @@ class SqlTableRW {
...
+    cols_.emplace_back(name, type, nullable, oid, default_value);
+    delete schema_;
+    schema_ = new catalog::Schema(cols_, next_version_++);
+
+    table_->UpdateSchema(*schema_);
+
+    col_oids_.clear();
+    for (const auto &c : cols_) {
+      col_oids_.emplace_back(c.GetOid());
+      // LOG_INFO("{}", !c.GetOid());
```
songzhaozhe: These commented lines may be removed.

test/storage/sql_table_test.cpp
```diff
@@ -246,105 +316,677 @@ class SqlTableRW {
...
+  auto result = table.EndUpdateRow(txn, row2_slot);
+
+  EXPECT_TRUE(result.first);
+  EXPECT_EQ(result.second.GetBlock(), row2_slot.GetBlock());
+  EXPECT_EQ(result.second.GetOffset(), row2_slot.GetOffset());
+  uint32_t new_val = table.GetIntColInRow(txn, catalog::col_oid_t(1), row2_slot);
+  EXPECT_EQ(new_val, 11001);
+
+  // update (400, 10003, 42) -> (400, 11003, 420)
+  // LOG_INFO("----------------------------")
```
songzhaozhe: Same.

src/include/catalog/schema.h
```diff
@@ -134,15 +158,19 @@ class Schema {
     uint16_t max_varlen_size_;
     bool nullable_;
     col_oid_t oid_;
-    // TODO(Matt): default value would go here
-    // Value default_;
+    // TODO(Sai): Consider having a DefaultValueObject containing isNull, 16-byte variable and attribute size
+    // This avoids handling memory explicitly for default values
+    byte default_[16];
```
songzhaozhe: Why is 16 chosen? Do you plan to change it to byte* in the future?

src/storage/sql_table.cpp
```diff
@@ -19,64 +43,329 @@ SqlTable::SqlTable(BlockStore *const store, const catalog::Schema &schema, const
...
+  // NOLINTNEXTLINE
+  for (const auto &column : schema.GetColumns()) {
+    auto col_oid = column.GetOid();
+    auto *default_value = column.GetDefault();
+    // Only populate the default values of the columns which are new and have a default value
+    if (default_value_map_.Find(col_oid) == default_value_map_.End()) {
+      uint8_t attr_size = column.GetAttrSize();
+      byte *temp = nullptr;
+      if (default_value != nullptr) {
+        // clang's memory analyis has a false positive on this allocation.  All memory
```
songzhaozhe: Nitpick: analysis

src/include/storage/sql_table.h
```diff
@@ -192,6 +343,39 @@ class SqlTable {
...
-  ProjectionMap ProjectionMapForInitializer(const ProjectionInitializerType &initializer) const;
+  ProjectionMap ProjectionMapForInitializer(const ProjectionInitializerType &initializer,
+                                            layout_version_t version) const;
+
+  /**
+   * Given a projected row/col translates the column id of each column to the column id of the version passed in
+   * If a column doesn't exist in that version sets the column id to VERSION_POINTER_COLUMN_ID
+   * @param out_buffer - projected row/col whose header to modify
+   * @param curr_dt_version - schema version of the passed in projected row/col
+   * @param old_dt_version - schema version that is desired
```
songzhaozhe: I think the comments are misleading here. These two are not schema versions, but DataTable meta data for that version. (In the above comments you are using 'schema version' to describe version_num instead)

src/include/storage/sql_table.h
```diff
@@ -174,15 +320,20 @@ class SqlTable {
...
-  // Eventually we'll support adding more tables when schema changes. For now we'll always access the one DataTable.
-  DataTableVersion table_;
+  common::ConcurrentMap<layout_version_t, DataTableVersion> tables_;
+  // NOTE: This map only keeps track of the default values specified at column creation
+  // For columns which don't have default value or added later, just set to null
+  // Populating default values into the ProjectedRow inserted later is taken care of by the execution engine
+  DefaultValueMap default_value_map_;
 
   /**
+   * Given a set of col_oids, return a vector of corresponding col_ids to use for ProjectionInitialization
```
songzhaozhe: Duplicate line.

src/storage/sql_table.cpp
```diff
@@ -19,64 +43,329 @@ SqlTable::SqlTable(BlockStore *const store, const catalog::Schema &schema, const
...
+    // 5. Insert the row into new table
+    storage::TupleSlot new_slot;
+    if (succ) {
+      new_slot = tables_.Find(version_num)->second.data_table->Insert(txn, *new_pr);
+    } else {
+      // someone else deleted the old row, write-write conflict
+      delete[] new_buffer;
+      return {false, slot};
+    }
+
```
songzhaozhe: I think there might be a logic problem here regarding varlen contents.
For the new row, you only did a shallow copy. For varlen contents, if you deleted the old-version row, GC can eventually reclaim all its contents. Then for the updated row, its varlen_content will point to already freed memory.

test/storage/sql_table_test.cpp
```diff
@@ -246,105 +316,677 @@ class SqlTableRW {
...
+
+  for (auto iter : seen_map) EXPECT_TRUE(iter.second);
+
+  delete[] buffer;
+  txn_manager_.Commit(txn, TestCallbacks::EmptyCallback, nullptr);
+  delete txn;
 }
 
 // NOLINTNEXTLINE
 TEST_F(SqlTableTests, VarlenInsertTest) {
```
songzhaozhe: I think you might want to open GC in tests to see whether your implementation works well with GC, to ensure you are
(1) not double-reclaiming memory (2) retaining useful contents(specially varlen contents) from reclaiming by GC.
One varlen insert seems too simple here. You can test updating an old-of-date row with varlen contents.
This comment comes from the doubt I mentioned in https://github.com/cmu-db/terrier/pull/342/files?file-filters%5B%5D=.cpp&file-filters%5B%5D=.h&file-filters%5B%5D=.py#r283081606

src/storage/sql_table.cpp
```diff
@@ -19,64 +43,329 @@ SqlTable::SqlTable(BlockStore *const store, const catalog::Schema &schema, const
...
+    bool valid = Select(txn, slot, new_pr, new_pair.second, version_num);
+    if (!valid) {
+      delete[] new_buffer;
+      return {false, slot};
+    }
+    // 3. Delete the old row
+    bool succ = tables_.Find(old_version)->second.data_table->Delete(txn, slot);
+
+    // 4. Update the new row before insert
+    StorageUtil::CopyProjectionIntoProjection(*redo, map, tables_.Find(version_num)->second.layout, new_pr,
```
songzhaozhe: Do you need to populate default values for Update too?
Also in CopyProjectionIntoProjection you called here, there is a TODO listing for default values, and I think you can remove it and add something here outside.

test/storage/sql_table_test.cpp
```diff
@@ -193,24 +260,29 @@ class SqlTableRW {
...
+   * @param col_oid col_oid of the column in the schema
+   */
+  void SetNullInRow(catalog::col_oid_t col_oid) { pr_->SetNull(pr_map_->at(col_oid)); }
+
   /**
    * Read a string from a row
    * @param col_num column number in the schema
    * @param slot - tuple to read from
    * @return malloc'ed C string (with null terminator). Caller must
    *   free.
```
songzhaozhe: Style nit: unnecessary line wrap

test/storage/sql_table_test.cpp
```diff
@@ -246,105 +316,677 @@ class SqlTableRW {
...
+  EXPECT_TRUE(result.first);
+  EXPECT_EQ(result.second.GetBlock(), row4_slot.GetBlock());
+  EXPECT_EQ(result.second.GetOffset(), row4_slot.GetOffset());
+  new_val = table.GetIntColInRow(txn, catalog::col_oid_t(1), result.second);
+  EXPECT_EQ(new_val, 11003);
+  new_val = table.GetIntColInRow(txn, catalog::col_oid_t(2), result.second);
+  EXPECT_EQ(new_val, 420);
+
+  // update (100, 10000, null) -> (100, 11000, 420)
+  // LOG_INFO("----------------------------")
```
songzhaozhe: Style nit: remove this line

test/storage/sql_table_concurrent_test.cpp
```diff
@@ -0,0 +1,471 @@
...
+
+    MultiThreadTestUtil::RunThreadsUntilFinish(&thread_pool, num_threads, workload);
+    // End concurrent section
+
+    for (auto &version : versioned_col_oids) delete version;
+    versioned_col_oids.clear();
+
+    gc_.PerformGarbageCollection();
+    gc_.PerformGarbageCollection();
+    gc_.PerformGarbageCollection();
```
songzhaozhe: calling 2 times is enough

test/storage/sql_table_test.cpp
```diff
@@ -246,105 +316,677 @@ class SqlTableRW {
...
+  EXPECT_EQ(100, id);
+  uint32_t datname = table.GetIntColInRow(txn, catalog::col_oid_t(1), row1_slot);
+  EXPECT_EQ(10000, datname);
 
-  table.StartRow();
-  table.SetIntColInRow(0, 200);
-  table.SetIntColInRow(1, 25721);
-  storage::TupleSlot row2_slot = table.EndRowAndInsert();
+  // manually set the version of the transaction to be 1
+  table.version_ = storage::layout_version_t(1);
```
codeworm96: Directly modifying this field is confusing and seems unsafe (may violate your invariant). A method for setting the version of the transaction may better express your intention.

test/storage/sql_table_test.cpp
```diff
@@ -246,105 +316,677 @@ class SqlTableRW {
...
   table.DefineColumn("datname", type::TypeId::VARCHAR, false, catalog::col_oid_t(1));
   table.Create();
 
-  table.StartRow();
-  table.SetIntColInRow(0, 100);
-  table.SetVarcharColInRow(1, "name");
-  storage::TupleSlot row_slot = table.EndRowAndInsert();
+  table.StartInsertRow();
+  table.SetIntColInRow(catalog::col_oid_t(0), 100);
+  table.SetVarcharColInRow(catalog::col_oid_t(1), "name");
```
codeworm96: You should also use longer varlen values to test for varlen entries which are not inlined.

src/include/catalog/schema.h
```diff
@@ -52,17 +59,20 @@ class Schema {
...
         : name_(std::move(name)),
           type_(type),
           attr_size_(type::TypeUtil::GetTypeSize(type_)),
           max_varlen_size_(max_varlen_size),
           nullable_(nullable),
           oid_(oid) {
+      // TODO(Sai): How to handle default values for VARLEN?
       TERRIER_ASSERT(attr_size_ == VARLEN_COLUMN, "This constructor is meant for VARLEN columns.");
       TERRIER_ASSERT(type_ != type::TypeId::INVALID, "Attribute type cannot be INVALID.");
+      SetDefault(default_value);
```
codeworm96: This line is problematic. The most significant bit of VARLEN_COLUMN is set to 1. You should not use that directly as the size for `memcpy` in `SetDefault`, which will cause scary memory issues.

test/storage/sql_table_test.cpp
```diff
@@ -38,62 +39,134 @@ class SqlTableRW {
       cols_.emplace_back(name, type, 255, nullable, oid);
   }
 
+  void AddColumn(transaction::TransactionContext *txn, std::string name, type::TypeId type, bool nullable,
```
codeworm96: `AddColumn` does not support adding varlen columns. You should also test adding varlen columns to make sure your code is handling varlens correctly.

test/storage/sql_table_concurrent_test.cpp
```diff
@@ -0,0 +1,471 @@
...
+// all of the data is consistent with what we expect in the end by embedding an id
+// and insertion time version into the tuple.
+// NOLINTNEXTLINE
+TEST_F(SqlTableConcurrentTests, ConcurrentInsertsWithSchemaChanges) {
+  const uint32_t num_iterations = 100;
+  const uint32_t txns_per_thread = 10;
+  const uint32_t num_threads = MultiThreadTestUtil::HardwareConcurrency();
+  common::WorkerPool thread_pool(num_threads, {});
+
+  for (uint32_t iteration = 0; iteration < num_iterations; iteration++) {
```
codeworm96: Maybe you do not need to repeat the test by yourself. `googletest` already provides the `--gtest_repeat` flag and the `GTEST_REPEAT` environment variable.

test/storage/sql_table_concurrent_test.cpp
codeworm96: You'd better make these code a function to prevent problems like this.

test/include/util/catalog_test_util.h
```diff
@@ -33,5 +34,102 @@ struct CatalogTestUtil {
...
+        // flip a coin
+        if (*RandomTestUtil::UniformRandomElement(&include, generator)) {
+          pr->SetNull(pr_pair.second.at(col.GetOid()));
+        }
+      }
+    }
+    return pr;
+  }
+
+  // It populates a projected row with random bytes. The schema must not contain VarChar
```
codeworm96: You'd better have some assertions to ensure the schema does not contain VarChar.

test/include/util/catalog_test_util.h
```diff
@@ -33,5 +34,102 @@ struct CatalogTestUtil {
...
+  // The projected row must contain all the columns
+  template <typename Random>
+  static void PopulateRandomRow(storage::ProjectedRow *row, const catalog::Schema &schema,
+                                const storage::ProjectionMap &pr_map, Random *const generator) {
+    for (auto &it : pr_map) {
```
codeworm96: This algorithm has low efficiency. Why not let the outer loop iterate over the columns and then look up the `pr_map`?

src/storage/sql_table.cpp
```diff
@@ -19,64 +43,329 @@ SqlTable::SqlTable(BlockStore *const store, const catalog::Schema &schema, const
...
-  table_ = {new DataTable(block_store_, layout, layout_version_t(0)), layout, col_oid_to_id};
+  // Populate the default value map
+  //
+  // clang's memory analysis gets confused by the iterator here.  This is tied
+  // to the same false positive identified inside of the loop and is caused.
+  // NOLINTNEXTLINE
+  for (const auto &column : schema.GetColumns()) {
+    auto col_oid = column.GetOid();
+    auto *default_value = column.GetDefault();
+    // Only populate the default values of the columns which are new and have a default value
```
lmy1229: Is it possible for the user to change the default value of a column? Or else it is treated as a new column (with a different OID)

src/storage/sql_table.cpp
```diff
@@ -19,64 +43,329 @@ SqlTable::SqlTable(BlockStore *const store, const catalog::Schema &schema, const
...
+  if (old_version_num == version_num) {
+    return tables_.Find(version_num)->second.data_table->Select(txn, slot, out_buffer);
+  }
+
+  auto old_dt_version = tables_.Find(old_version_num)->second;
+  auto curr_dt_version = tables_.Find(version_num)->second;
+
+  // The slot version is not the same as the version_num
+  col_id_t original_column_ids[out_buffer->NumColumns()];
+  ModifyProjectionHeaderForVersion(out_buffer, tables_.Find(version_num)->second, old_dt_version, original_column_ids);
```
lmy1229: tables_.Find(version_num)->second is repeated several times here, and curr_dt_version is unused. Are you doing this for some concurrent issues?

src/storage/sql_table.cpp
yash620: It should be possible for a user to change the default value of a column. But what is happening here is that if a new column is added and a default value is set for that column on creation then it all tuples within the table should receive that default value for that column. But if a default value is set on an already created column then that default value is only applied to tuples inserted from that point on.

src/storage/sql_table.cpp
jrolli: The full materialization of the existing data between line 256 and line 265 will handle default values because we use the SqlTable `Select` to materialize it into the version we will insert.

test/storage/sql_table_test.cpp
jrolli: The comment is incorrect.  This is not manually setting the transaction's version but rather simulating the catalog update that occurs along with a schema update.  I'll update the comment to reflect this.

test/storage/sql_table_concurrent_test.cpp
jrolli: We're using the same setup that was already used for the `DataTable` tests.  The repetitions here are useful for increasing the likelihood that CI will error out on race conditions in the concurrent tests without having to make deep modifications to the CMake build scripts.

src/storage/sql_table.cpp
jrolli: The storage engine only needs to reason about default values that are applied at the time the column is created as these are retroactive (applied to all existing data).  Default value changes at any other time would only affect new inserts and have to be handled by the catalog, execution engine, and binder because we are not in a position at this level to reason about that granularity of visibility.

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+## Trade-offs and Potential Problems
+**Trade-off:** TBB Concurrent Unordered Map for storing DataTableVersion objects.  This decision gives a simple and easy to integrate solution for supporting concurrent insertions (new schemas) and lookups (all DML transactions) on the data tables.  However, this will limit options when we start to implement compaction of obsolete schema versions because the data structure does not support concurrent erasures.  This likely means we will have to take a heavy-weight approach for compaction such as duplicating the structure without the data to be erased and then use an unlink-delete staged method similar to how the GC already works on undo records.
+
+**Trade-off:**  Our decision to manually mangle the headers for ProjectedRow and ProjectedColumn greatly increases the code complexity (manual recreation of the headers) in order to significantly improve performance for reading data across schema versions.  Specifically, we avoid an unnecessary allocation and deallocation for temporary projections by allowing old schema versions to write directly into the final projection.
+
+## Future Work
+### Pending tasks
+#### Default values
```
codeworm96: Outdated? You have done some work about default values.

src/include/catalog/schema.h
codeworm96: Maybe 16 is chosen for a reason, but you should not simply use it as a magic number.

src/include/storage/sql_table.h
```diff
@@ -27,6 +31,110 @@ class SqlTable {
...
+        // layout_version_t is uint32_t so we need to protect against underflow.
+        if ((!curr_version_) == 0) {
+          is_end_ = true;
+          return;
+        }
+        curr_version_--;
+        auto next_table = tables_->Find(curr_version_);
+        if (next_table == tables_->CEnd()) {  // next_table does not exist (at end)
+          is_end_ = true;
+          break;
```
codeworm96: Break the loop with `is_end_` set to be true. But immediately after that the assertion checks `!is_end_`, which is confusing.

src/include/storage/sql_table.h
```diff
@@ -102,45 +231,59 @@ class SqlTable {
...
+  void Scan(transaction::TransactionContext *txn, SqlTable::SlotIterator *start_pos, ProjectedColumns *out_buffer,
+            const ProjectionMap &pr_map, layout_version_t version_num) const;
 
   /**
-   * @return table's unique identifier
+   * @return the first tuple slot contained in the data table
    */
-  catalog::table_oid_t Oid() const { return oid_; }
+  SlotIterator begin(layout_version_t txn_version) const {
+    // common::SpinLatch::ScopedSpinLatch guard(&tables_latch_);
```
codeworm96: Deadcode. Should be removed.

src/include/storage/sql_table.h
```diff
@@ -102,45 +231,59 @@ class SqlTable {
...
+   * Returns one past the last tuple slot contained in the last data table. Note that this is not an accurate number
+   * when concurrent accesses are happening, as inserts maybe in flight. However, the number given is always
+   * transactionally correct, as any inserts that might have happened is not going to be visible to the calling
+   * transaction.
+   *
+   * @return one past the last tuple slot contained in the data table.
    */
-  DataTable::SlotIterator begin() const { return table_.data_table->begin(); }
+  SlotIterator end() const {
+    // common::SpinLatch::ScopedSpinLatch guard(&tables_latch_);
```
codeworm96: Deadcode. Should be removed.

src/include/storage/sql_table.h
```diff
@@ -102,45 +231,59 @@ class SqlTable {
...
+  SlotIterator begin(layout_version_t txn_version) const {
+    // common::SpinLatch::ScopedSpinLatch guard(&tables_latch_);
+    auto ret = SlotIterator(&tables_, txn_version, false);
+    ret.AdvanceOnEndOfDatatable_();
+    return ret;
+  }
 
   /**
-   * @return the first tuple slot contained in the underlying DataTable
+   * Returns one past the last tuple slot contained in the last data table. Note that this is not an accurate number
```
codeworm96: Confusing document: What is the `number` it is talking about? Tuple slot or something else?

src/include/storage/sql_table.h
```diff
@@ -150,21 +293,23 @@ class SqlTable {
...
   std::pair<ProjectedRowInitializer, ProjectionMap> InitializerForProjectedRow(
-      const std::vector<catalog::col_oid_t> &col_oids) const {
+      const std::vector<catalog::col_oid_t> &col_oids, layout_version_t version_num) const {
     TERRIER_ASSERT((std::set<catalog::col_oid_t>(col_oids.cbegin(), col_oids.cend())).size() == col_oids.size(),
                    "There should not be any duplicated in the col_ids!");
-    auto col_ids = ColIdsForOids(col_oids);
+    auto col_ids = ColIdsForOids(col_oids, version_num);
     TERRIER_ASSERT(col_ids.size() == col_oids.size(),
                    "Projection should be the same number of columns as requested col_oids.");
+    TERRIER_ASSERT(tables_.Find(version_num) != tables_.CEnd(), "Table version must exist before insert");
```
codeworm96: The message seems incorrect. `InitializerForProjectedRow` does not insert any row.

src/include/storage/sql_table.h
```diff
@@ -192,6 +341,39 @@ class SqlTable {
...
+   * @param pr_map ProjectionMap of the ProjectedRow passed into Select/Scan
+   * @param old_dt_version old version of the datatable
+   * @return Unordered set of missing column oids
+   */
+  std::unordered_set<catalog::col_oid_t> GetMissingColumnOidsForVersion(const ProjectionMap &pr_map,
+                                                                        const DataTableVersion &old_dt_version) const;
+
+  /**
+   * Fill in the default value for the given column i.e. col_oid. The default value being filled in can be null
+   * @tparam RowType ProjectedRow or ProjectedColumns::RowView
```
codeworm96: typo: tparam

test/include/util/catalog_test_util.h
```diff
@@ -33,5 +34,99 @@ struct CatalogTestUtil {
     }
     return catalog::Schema(columns);
   }
+
+  // The same as RandomSchem but it won't generate Varchar Column
```
codeworm96: typo: RandomSchema?

test/include/util/catalog_test_util.h
```diff
@@ -33,5 +34,99 @@ struct CatalogTestUtil {
...
+      StorageTestUtil::FillWithRandomBytes(size, addr, generator);
+    }
+  }
+
+  // Check if two rows have the same content. It assumes they have the same ProjectionMap.
+  template <class RowType1, class RowType2>
+  static bool ProjectionListEqual(const catalog::Schema &schema, const RowType1 *const one, const RowType2 *const other,
+                                  const storage::ProjectionMap &map) {
+    if (one->NumColumns() != other->NumColumns()) return false;
+    for (auto &it : map) {
```
codeworm96: Similar to https://github.com/cmu-db/terrier/pull/342#discussion_r283112195.

src/storage/sql_table.cpp
```diff
@@ -19,64 +43,330 @@ SqlTable::SqlTable(BlockStore *const store, const catalog::Schema &schema, const
...
+    col_id_t original_column_ids[redo->NumColumns()];
+    ModifyProjectionHeaderForVersion(redo, tables_.Find(version_num)->second, old_dt_version, original_column_ids);
+
+    // Get the result and copy back the old header
+    bool result = old_dt_version.data_table->Update(txn, slot, *redo);
+
+    // We should create a buffer of old Projected Row and update in place. We can't just
+    // directly erase the data without creating a redo and update the chain.
+
+    // auto old_pair = InitializerForProjectedRow(redo_col_oids, old_version);
```
codeworm96: What is these commented out code?

src/storage/sql_table.cpp
codeworm96: I think that part deserves better documentation. The user of SqlTable needs to know for default values which part is handled by SqlTable and which part they should take care of.

src/storage/sql_table.cpp
```diff
@@ -1,13 +1,37 @@
...
 
 SqlTable::SqlTable(BlockStore *const store, const catalog::Schema &schema, const catalog::table_oid_t oid)
     : block_store_(store), oid_(oid) {
+  UpdateSchema(schema);
+}
+
+SqlTable::~SqlTable() {
+  while (default_value_map_.CBegin() != default_value_map_.CEnd()) {
+    auto pair = *(default_value_map_.CBegin());
+    delete[] pair.second.first;
```
codeworm96: For varlens that are not inlined, simply delete `pair.second.first` will leak memory.

src/storage/sql_table.cpp
```diff
@@ -86,8 +376,75 @@ ProjectionMap SqlTable::ProjectionMapForInitializer(const ProjectionInitializerT
...
 
 template ProjectionMap SqlTable::ProjectionMapForInitializer<ProjectedColumnsInitializer>(
-    const ProjectedColumnsInitializer &initializer) const;
+    const ProjectedColumnsInitializer &initializer, layout_version_t version) const;
 template ProjectionMap SqlTable::ProjectionMapForInitializer<ProjectedRowInitializer>(
-    const ProjectedRowInitializer &initializer) const;
+    const ProjectedRowInitializer &initializer, layout_version_t version) const;
+
+// TODO(Yashwanth): don't copy the entire header, no need for template only take in ColumnIds() and then just modify
+// that when resetting header only have memc py ColumnIds()
```
codeworm96: typo: `memc py`

src/storage/sql_table.cpp
```diff
@@ -19,64 +43,330 @@ SqlTable::SqlTable(BlockStore *const store, const catalog::Schema &schema, const
...
+  //
+  // clang's memory analysis gets confused by the iterator here.  This is tied
+  // to the same false positive identified inside of the loop and is caused.
+  // NOLINTNEXTLINE
+  for (const auto &column : schema.GetColumns()) {
+    auto col_oid = column.GetOid();
+    auto *default_value = column.GetDefault();
+    // Only populate the default values of the columns which are new and have a default value
+    if (default_value_map_.Find(col_oid) == default_value_map_.End()) {
+      uint8_t attr_size = column.GetAttrSize();
```
codeworm96: Note that the results of `GetAttrSize()` of varlen attributes have the sign bit set. You will allocate a large chunk of memory.

src/storage/sql_table.cpp
```diff
@@ -86,8 +376,75 @@ ProjectionMap SqlTable::ProjectionMapForInitializer(const ProjectionInitializerT
...
+template <class RowType>
+void SqlTable::FillDefaultValue(RowType *out_buffer, const catalog::col_oid_t col_oid,
+                                const ProjectionMap &pr_map) const {
+  TERRIER_ASSERT(default_value_map_.Find(col_oid) != default_value_map_.CEnd(),
+                 "Every column in schema must exist in default_value_map");
+  auto pair = default_value_map_.Find(col_oid)->second;
+  auto default_value = pair.first;
+  auto attr_size = pair.second;
+  // TODO(Sai): If this becomes a performance bottleneck, we can move this logic to ModifyProjectionHeaderForVersion
+  storage::StorageUtil::CopyWithNullCheck(default_value, out_buffer, attr_size, pr_map.at(col_oid));
```
codeworm96: `CopyWithNullCheck` only does a shallow copy and `VarlenEntry` has no reference counting. You may need careful reasoning about how this will interact with the GC.

src/include/storage/sql_table.h
```diff
@@ -174,15 +319,19 @@ class SqlTable {
   BlockStore *const block_store_;
   const catalog::table_oid_t oid_;
 
-  // Eventually we'll support adding more tables when schema changes. For now we'll always access the one DataTable.
-  DataTableVersion table_;
+  common::ConcurrentMap<layout_version_t, DataTableVersion> tables_;
+  // NOTE: This map only keeps track of the default values specified at column creation
+  // For columns which don't have default value or added later, just set to null
+  // Populating default values into the ProjectedRow inserted later is taken care of by the execution engine
+  DefaultValueMap default_value_map_;
```
codeworm96: Users can drop columns. So some default values may be no longer needed in the future. You may want to have some way to shrink the `DefaultValueMap`.

jrolli:
Blocked on #392 which will solidify the correct API for default values.

PR #344, opened 2019-04-09
DB Main

DarkForte:
A simple DB main object. 
No Settings Manager yet. It will add itself after this is merged.


src/include/main/db_main.h
```diff
@@ -0,0 +1,61 @@
+#pragma once
+
+#include <fstream>
+#include <iostream>
+#include <memory>
+#include <vector>
+#include "bwtree/bwtree.h"
```
tli2: Do you need all these includes in the .h?

src/main/terrier.cpp
```diff
@@ -1,53 +1,9 @@
...
-  LOG_INFO("Initialization complete");
-
-  terrier::network::TerrierServer terrier_server;
-  terrier_server.SetupServer().ServerLoop();
-
-  // shutdown loggers
-  spdlog::shutdown();
-  main_stat_reg->Shutdown(false);
+int main(int argc, char **argv) {
+  auto db_main = std::make_shared<terrier::DBMain>();
```
tli2: I don't see a reason why db_main has to live on the heap and can't just be a stack variable.

src/main/terrier.cpp
```diff
@@ -1,53 +1,9 @@
...
+  db_main->Init(argc, argv);
+  db_main->Start();
```
tli2: It should be made clear that `Start()` never returns in the normal case. When and how `Shutdown()` runs need to be documented.

src/main/db_main.cpp
```diff
@@ -0,0 +1,52 @@
+#include "main/db_main.h"
+#include <memory>
+
+namespace terrier {
+
+int DBMain::Init(int argc, char **argv) {
```
tli2: I am not convinced these are the right arguments. At the very least, GFlags also need to process these raw command args. I think a better idea would be to separate out the interpretation of these command line args and the initialization of `SettingsManager` into their own function or module.

apavlo: @tli2 Is correct here. `DBMain` should **not** have to process commandline arguments. If we did it this way, then every test case would need to hardcode the arguments. This will make our code brittle.

DarkForte: Then we need to bring up `SettingsManager` before bringing up `DBMain`?

src/include/main/db_main.h
```diff
@@ -0,0 +1,61 @@
...
+   * This function boots traffic cop and networking layer
+   * @return 0 if started successfully; 1 otherwise
+   */
+  int Start();
+
+  /**
+   * This function provides the shutdown API
+   */
+  void Shutdown();
+
```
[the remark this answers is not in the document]
DarkForte: Who needs to access them? I can't come up with one...

src/main/db_main.cpp
```diff
@@ -0,0 +1,52 @@
...
+    storage::init_index_logger();
+    storage::init_storage_logger();
+    transaction::init_transaction_logger();
+    parser::init_parser_logger();
+    network::init_network_logger();
+    // Flush all *registered* loggers using a worker thread.
+    // Registered loggers must be thread safe for this to work correctly
+    spdlog::flush_every(std::chrono::seconds(DEBUG_LOG_FLUSH_INTERVAL));
+  } catch (const spdlog::spdlog_ex &ex) {
+    std::cout << "debug log init failed " << ex.what() << std::endl;  // NOLINT
```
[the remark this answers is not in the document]
DarkForte: ok. because we need some way to display an error message before the loggers are up.

src/main/db_main.cpp
DarkForte: We have a map passed to the DB Main object now.

mbutrovich:
Closing in favor of #336, which according to @DarkForte is a superset of these changes. We'll work off of that PR to merge the new main.

PR #348, opened 2019-04-12
Embed attribute sizes into ProjectedColumns

jrolli:
This PR embeds the attribute sizes of each column in a `ProjectedColumns` instance to allow users of `SqlTable::Scan` to obtain a `RowView` without having to be aware of the underlying `BlockLayout` associated with a schema.

Originally, we had assumed that `RowView` would only be utilized internally by the storage layer.  However, PRs #260 and #341 show that `RowView` is useful outside of the storage layer.  Specifically, it provides a means to iterate over all of the _visible_ tuples in an SQL table without having to reproduce the logic to verify visibility (as would be the case using `SqlTable::SlotIterator` and `SqlTable::Select` directly).

In order to enforce the intended separation between the execution and storage layers, this PR removes the `BlockLayoutFromSchema` function from `StorageUtil` and inlines it into the constructor for `SqlTable`.


src/include/storage/projected_columns.h
```diff
@@ -204,12 +201,19 @@ class PACKED ProjectedColumns {
...
+   */
+  const uint32_t AttrSizeForColumn(uint16_t col_id);
+
  private:
   friend class ProjectedColumnsInitializer;
   uint32_t size_;
   uint32_t max_tuples_;
   uint32_t num_tuples_;
   uint16_t num_cols_;
+  uint16_t attr_ends_[4];
```
tli2: Why is this 4? I think we need a comment explaining what each of these mean and why it has size 4. It's not immediately obvious.

jrolli: Replaced all uses of the magic 4 with a defined constant, and added comments there and in `type_util.h` about the power of two assumption on attribute sizes (@mbutrovich recommended this).

jrolli: Replaced the magic number with a constant definition and added comments to explain how we arrived at that comment.  I also put a warning in the type utils where attribute sizes are defined that this constant exists in the storage layer.

PR #351, opened 2019-04-14
Rollback callback

yash620:
This adds in a framework for having callbacks when a transaction rollsback.

### DIFF
--- a/src/include/transaction/transaction_context.h
+++ b/src/include/transaction/transaction_context.h
@@ -1,4 +1,5 @@
 #pragma once
+#include <map>
 #include <vector>
 #include "common/object_pool.h"
 #include "common/strong_typedef.h"
@@ -148,7 +149,6 @@ class TransactionContext {
   bool log_processed_ = false;
 
   // Functions to be invoked on rollback
-  // std::vector<transaction::rollback_fn> rollback_functions_;
   std::map<transaction::rollback_fn, void *> rollback_functions_;
 };
 }  // namespace terrier::transaction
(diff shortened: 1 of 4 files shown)


src/include/transaction/transaction_context.h
```diff
@@ -129,5 +147,8 @@ class TransactionContext {
   // log manager will set this to be true when log records are processed (not necessarily flushed, but will not be read
   // again in the future), so it can be garbage-collected safely.
   bool log_processed_ = false;
+
+  // Functions to be invoked on rollback
+  std::map<transaction::rollback_fn, void *> rollback_functions_;
```
mbutrovich: When @jrolli mentioned this to me the other day it sounded like a stack was what we were going to use. Do we need ordered and random element access that a map provides?

apavlo: We don't need random access. AFAIK all constraint checks are commutative, so the order does not matter.

mbutrovich: Suggest replacing with std::forward_list then.

mbutrovich:
Once this is in, we should discuss if this is the way we want to approach handling index rollbacks as well, and if so modify the index wrapper as needed.

src/include/transaction/transaction_context.h
yash620: Alright, I'll go ahead and do that

yash620:
John's framework of callbacks will take care of this. No need for this anymore.

PR #352, opened 2019-04-14
BwTreeIndex::Scan and PRI accessor.

mbutrovich:
**ProjectedRow Initializer accessor:**
We need a way to get the index's PR initializer without having access to the metadata object (similar to how it works DataTable).

**Index range scan:**
I added a basic range scan to the BwTree wrapper and a simple functionality test. We need this for some of the remaining TPC-C queries. Peloton's Scan function took in a (complicated to me) ConjugateScanPredicate object that looks like it was constructed in the Planner. Peloton's Scan method just extracted the low and high keys from that object, so as long as those are defined we can do range scans. I changed the Scan arguments to just take the low and high keys directly. I'm punting on understanding the 2000 lines of code of index scan plan nodes, index scan executors, and the scan optimizer stuff (CSP) and will worry about that once we need to actually write the index scanner for the new execution engine since I expect a lot of that code will need to change anyway.

### DIFF
--- a/src/include/storage/index/index.h
+++ b/src/include/storage/index/index.h
@@ -104,7 +104,7 @@ class Index {
   /**
    * @return mapping from key oid to projected row offset
    */
-  const std::unordered_map<catalog::indexkeycol_oid_t, uint32_t> &GetKeyOidToOffsetMap() const {
+  const std::unordered_map<catalog::indexkeycol_oid_t, uint16_t> &GetKeyOidToOffsetMap() const {
     return metadata_.GetKeyOidToOffsetMap();
   }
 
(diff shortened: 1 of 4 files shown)


src/include/storage/index/index.h
```diff
@@ -91,6 +100,18 @@ class Index {
    * @return oid of this indes
    */
   catalog::index_oid_t GetOid() const { return oid_; }
+
+  /**
+   * @return mapping from key oid to projected row offset
+   */
+  const std::unordered_map<catalog::indexkeycol_oid_t, uint32_t> &GetKeyOidToOffsetMap() const {
```
lmwnshn: I wonder if it makes more sense for the map to be uint16_t so that it matches the input expected of ProjectedRow::AccessX(.) methods.

Would have saved a fair bit of stupid static_casts.

Issue #335, opened 2019-04-06
TPC-C benchmark


mbutrovich:
We have indexes, we have logging, we have garbage collection. Now do something with them. Implement TPC-C exercising the concurrency control and storage layers (including indexes, logging, and garbage collection). This PR will track our implementation progress. WIP branch is [here](https://github.com/mbutrovich/terrier/tree/tpcc2).

**Table creation (Schema and SqlTable):**
- [x] Item
- [x] Warehouse
- [x] District
- [x] Customer
- [x] History
- [x] New Order
- [x] Order
- [x] Order Line
- [x] Stock

**Table population:**
- [x] Item
- [x] Warehouse
- [x] District
- [x] Customer
- [x] History
- [x] New Order
- [x] Order
- [x] Order Line
- [x] Stock

**Index creation (IndexKeySchema and Index):**
- [x] Item
- [x] Warehouse
- [x] District
- [x] Customer
- [x] Customer secondary (Payment and Order Status)
- [x] New Order
- [x] Order
- [x] Order secondary (Order Status)
- [x] Order Line
- [x] Stock

**Index population:**
- [x] Item
- [x] Warehouse
- [x] District
- [x] Customer
- [x] Customer secondary (Payment and Order Status)
- [x] New Order
- [x] Order
- [x] Order secondary (Order Status)
- [x] Order Line
- [x] Stock

**Workload generation:**
- [x] New Order
- [x] Payment
- [x] Order Status
- [x] Delivery
- [x] Stock Level

**Workload execution:**
- [x] New Order
- [x] Payment
- [x] Order Status
- [x] Delivery
- [x] Stock Level

**Test Framework:**
- [x] Builder
- [ ] Stats
- [x] Parallelized transaction execution
- [x] Documentation

**Nice to have:**
- [ ] [done but don't want it right now] Parallelized table and index population
- [x] Minimize type sizes in table schemas

mbutrovich:
With 5 of the 9 tables down I think I've implemented all of the boilerplate needed for the various types of random attributes by now and can crank out tables pretty quickly now. I expect all 9 will be done tomorrow, and I can sync up with @lmwnshn to see how the indexes are looking.

mbutrovich:
All of the tables have been instantiated and loaded per TPCC spec. I just finished defining all of the index key schemas and instantiating the indexes. Just need to modify the loader to insert into the indexes as well.

mbutrovich:
[...]

PR #356, opened 2019-04-16
BwTreeIndex: Add visibility logic, abort and deferred delete logic, descending and limit scans

mbutrovich:
This PR addresses #357. It also adds scans that I found useful in writing #335.

Changes:
- `DataTable` pointer is now embedded in each Block. This allows indexes to access the `DataTable` via only a TupleSlot. Consider removing `layout_version` if the schema change team doesn't need it.
- BwTree scan results are checked for MVCC visibility by consulting the `DataTable` using `TransactionContext` before returning results.
- `index->InsertUnique()` checks for MVCC visibility of conflicting keys by consulting the `DataTable` using `TransactionContext`.
- `index->Delete()` registers a deferred commit action that then registers a deferred GC action to call delete on the key. This guarantees that the index key remains in the index as long as necessary for any other running transactions.
- `ScanDescending` uses the BwTree's backwards iteration. It's unclear yet if this will prove to be too slow, but for now it's correct and works.
- Limit scans will stop early if you want to restrict the size of your results. For example, `ScanLimitAscending()` with limit 1 is effectively an argmin of a range, while `ScanLimitDescending()` with limit 1 is effectively an argmax of a range.
- Moved tests primarily concerned with `CompactIntsKey` and `GenericKey` to their own test suite.
- Ported the MVCC tests that made sense to exercise the visibility logic of the `BwTreeIndex`.


mbutrovich:
Possibly deferred until we talk about #357.

mbutrovich:
I've updated the scans to take a TransactionContext and perform visibility checks with the DataTable. I want to write the predicate for unique key constraints while I'm poking at these files, but currently still thinking about all the criteria for that since it extends beyond visibility (need first writer wins as well).

mbutrovich:
Done with a first pass on implementation. It'll need a review at some point, but for now I'm ready to hand this off to Masha to port MVCC Data Table tests to verify that the BwTreeIndex enforces the same semantics. I brought one test over as an example (and a sanity check for myself), and we can meet up in person to discuss the implementation details a bit more.

src/include/storage/storage_defs.h
```diff
@@ -55,7 +67,8 @@ struct alignas(common::Constants::BLOCK_SIZE) RawBlock {
   /**
    * Contents of the raw block.
    */
-  byte content_[common::Constants::BLOCK_SIZE - 2 * sizeof(uint32_t) - sizeof(BlockAccessController)];
+  byte content_[common::Constants::BLOCK_SIZE - sizeof(uintptr_t) - sizeof(layout_version_t) - sizeof(uint32_t) -
```
tli2: I think this under-calculates the header by 16 bits? Since you redefined layout version to be 16, but didn't explicitly add the padding here?

mbutrovich: You're right, we still want the old calculation so we can use the extra 16 bits for flags and such.

mbutrovich:
What did I learn at school today: HyPer treats updates on indexed attributes as a delete and an insert at the table level, which solves my problem with visibility checking.

mbutrovich:
The scope of this keeps growing. I kept having to fake in tests what would eventually be the abort logic and deferred delete (handled in GC) and just rolled it into this PR using the new deferred action framework. So far it seems to work well. I'm going to wrap up the basic tests and tag this ready for review soon.

parsec0re:
@mbutrovich When will this be merge? Our add/drop indexes project relies on this PR. I can do a review if necessary.

src/include/storage/index/bwtree_index.h
```diff
@@ -31,61 +28,178 @@ class BwTreeIndex final : public Index {
...
+
+    // Perform lookup in BwTree
+    auto scan_itr = bwtree_->Begin(index_high_key);
+    // Back up one element if we didn't match the high key
+    if (scan_itr.IsEnd() || bwtree_->KeyCmpGreater(scan_itr->first, index_high_key)) scan_itr--;
+
+    while (!scan_itr.IsREnd() && (bwtree_->KeyCmpGreaterEqual(scan_itr->first, index_low_key))) {
+      // Perform visibility check on result
+      if (IsVisible(txn, scan_itr->second)) value_list->emplace_back(scan_itr->second);
+      scan_itr--;
```
jrolli: Have you checked whether reverse iteration is faster than just wrapping `ScanAscending` with an explicit reversing of the vector?  I seem to recall discussion that this particular call is expensive.

src/include/storage/data_table.h
```diff
@@ -246,7 +256,10 @@ class DataTable {
   void AtomicallyWriteVersionPtr(TupleSlot slot, const TupleAccessStrategy &accessor, UndoRecord *desired);
 
   // Checks for Snapshot Isolation conflicts, used by Update
-  bool HasConflict(UndoRecord *version_ptr, const transaction::TransactionContext *txn) const;
+  bool HasConflict(const transaction::TransactionContext &txn, UndoRecord *version_ptr) const;
+
+  // Wrapper around the other HasConflict for indexes to call (they only have tuple slot, not the version pointer)
+  bool HasConflict(const transaction::TransactionContext &txn, TupleSlot slot) const;
```
jrolli: Do we ever call `IsVisible` or `HasConflict` without calling the other?  I'm wondering if the indexes only need a single function that answers both of these questions at the same time.  Would solve question posed below.

src/include/storage/index/index.h
```diff
@@ -50,46 +63,84 @@ class Index {
...
+   * @return true if the value was inserted, false otherwise
+   *         (either because value exists, or predicate returns true for one of the existing values)
    */
-  virtual bool Delete(const ProjectedRow &tuple, TupleSlot location) = 0;
+  virtual bool InsertUnique(transaction::TransactionContext *txn, const ProjectedRow &tuple, TupleSlot location) = 0;
 
   /**
-   * Inserts a key-value pair only if the predicate fails on all existing values.
+   * Doesn't immediately call delete on the index. Registers a commit action in the txn that will eventually register a
+   * deferred action for the GC to safely call delete on the index when no more transactions need to access the key.
```
jrolli: Should this be a warning because of the implications on when it would be safe to deallocate the index?  Specifically, the deferral of the delete is not atomic with the commit itself and therefore may trigger later than when you expect.

src/include/storage/storage_defs.h
```diff
@@ -45,6 +47,16 @@ STRONG_TYPEDEF(layout_version_t, uint32_t);
  * @warning If you change the layout please also change the way header sizes are computed in block layout!
  */
 struct alignas(common::Constants::BLOCK_SIZE) RawBlock {
+  /**
+   * Data Table for this RawBlock. This is used by indexes and GC to get back to the DataTable given only a TupleSlot
+   */
+  DataTable *data_table_;
+
+  /**
+   * Padding for flags or whatever we may want in the future.
```
jrolli: Probably should explicitly state that padding size is determined by `sizeof(layout_version_t)` below.  The typedef makes this logic not immediately obvious.

src/include/storage/tuple_access_strategy.h
```diff
@@ -94,6 +97,19 @@ class TupleAccessStrategy {
...
+   * headers and divide up the blocks into mini blocks(each mini block contains
+   * a column). The raw block needs to be 0-initialized (by default when given out
+   * from a block store), otherwise it will cause undefined behavior.
+   *
+   * @param data_table pointer to the DataTable to reference from this block
+   * @param raw pointer to the raw block to initialize
+   * @param layout_version the layout version of this block
+   */
+  void InitializeRawBlockForDataTable(storage::DataTable *data_table, RawBlock *raw,
+                                      layout_version_t layout_version) const;
```
jrolli: Why are we creating a new function rather than modifying/replacing the existing function.  Is there a nuance here where some blocks need to be initialized with a DataTable pointer and others do not?

src/storage/data_table.cpp
```diff
@@ -300,7 +300,7 @@ void DataTable::NewBlock(RawBlock *expected_val) {
   // Want to stop early if another thread is already getting a new block
   if (expected_val != insertion_head_) return;
   RawBlock *new_block = block_store_->Get();
-  accessor_.InitializeRawBlock(new_block, layout_version_);
+  accessor_.InitializeRawBlockForDataTable(this, new_block, layout_version_);
```
jrolli: This is literally the only reference to `IntializeRawBlock(ForDataTable)` which means one of the two functions will never be used.  Choose what the function name should be and get rid of the other function.

src/storage/data_table.cpp
```diff
@@ -318,4 +318,53 @@ void DataTable::DeallocateVarlensOnShutdown(RawBlock *block) {
...
+
+bool DataTable::IsVisible(const transaction::TransactionContext &txn, const TupleSlot slot) const {
+  UndoRecord *version_ptr;
+  bool visible;
+  do {
+    version_ptr = AtomicallyReadVersionPtr(slot, accessor_);
+    // Here we will need to check that the version pointer did not change during our read. If it did, the content
+    // we have read might have been rolled back and an abort has already unlinked the associated undo-record,
+    // we will have to loop around to avoid a dirty read.
+    // TODO(Matt): might not need to read visible in the loop (move after?) but not confident without large random tests
```
jrolli: You need this because x86 does not guarantee that 16-bytes can be read atomically (only that 16-byte CAS can be executed atomically).

src/storage/data_table.cpp
```diff
@@ -318,4 +318,53 @@ void DataTable::DeallocateVarlensOnShutdown(RawBlock *block) {
...
+    visible = Visible(slot, accessor_);
+  } while (version_ptr != AtomicallyReadVersionPtr(slot, accessor_));
+
+  // Nullptr in version chain means no version visible to any transaction alive at this point.
```
jrolli: Do you mean "no version" or "no other versions"?  I think only the latter is true.

src/storage/data_table.cpp
```diff
@@ -318,4 +318,53 @@ void DataTable::DeallocateVarlensOnShutdown(RawBlock *block) {
...
+  // Alternatively, if the current transaction holds the write lock, it should be able to read its own updates.
+  if (version_ptr == nullptr || version_ptr->Timestamp().load() == txn.TxnId().load()) {
+    return visible;
+  }
+
+  // Apply deltas until we determine a version safe for us to read
+  // If the version chain becomes null, this tuple does not exist for this version, and the last delta
+  // record would be an undo for insert that sets the primary key to null, which is intended behavior.
```
jrolli: I don't think this comment is accurate anymore.  With GC in place we cannot infer anything about visibility simply by the absence of undo records and the last delta is not guaranteed to be an insert.

The comment should also be updated at line 225 as well.

src/storage/data_table.cpp
```diff
@@ -318,4 +318,53 @@ void DataTable::DeallocateVarlensOnShutdown(RawBlock *block) {
...
+  while (version_ptr != nullptr &&
+         transaction::TransactionUtil::NewerThan(version_ptr->Timestamp().load(), txn.StartTime())) {
+    // TODO(Matt): It's possible that if we make some guarantees about where in the version chain INSERTs (last position
+    // in version chain) and DELETEs (first position in version chain) can appear that we can optimize this check
```
jrolli: Is this comment still valid with GC?  If not remove it both here and at line 230.

test/storage/bwtree_index_test.cpp
```diff
@@ -2,1269 +2,1872 @@
...
+  const auto tuple_slot = sql_table_->Insert(txn0, *insert_tuple);
 
-      delete[] data_A;
-    }
+  // txn 0 inserts into index
+  auto *insert_key = unique_index_->GetProjectedRowInitializer().InitializeRow(insert_buffer_);
+  *reinterpret_cast<int32_t *>(insert_key->AccessForceNotNull(0)) = 15721;
+  EXPECT_TRUE(unique_index_->InsertUnique(txn0, *insert_key, tuple_slot));
 
-    delete[]
```
jrolli: Your test description and your code is a mismatch here.  You are not testing an update transaction.  This is significant because updates will have the same value under two different keys and we need to check that behavior is correct under these conditions.  The more I think about it the less convinced that we have the logic correct (see comment on `IsVisible`).

jrolli: This is not an update in the table.

jrolli: Not an update.

jrolli: Not an update.

jrolli: We should have another test for

```
        T1        |        T2        |        T3
BEGIN             |  BEGIN           |
                  |  U(X)            |
                  |  COMMIT          |
                  |                  |  BEGIN
                  |                  |  U(X)\
                  |                  |  COMMIT
COMMIT            |                  |
```

Where T2 changes an indexed attribute from VAL1 to VAL2 and T3 changes the indexed attribute from VAL2 back to VAL1.  I'm pretty sure we currently fail on this.

test/storage/bwtree_key_test.cpp
```diff
@@ -0,0 +1,1130 @@
...
+  key2.SetFromProjectedRow(*pr, metadata);
+
+  EXPECT_FALSE(std::equal_to<CompactIntsKey<KeySize>>()(key1, key2));
+  EXPECT_NE(std::hash<CompactIntsKey<KeySize>>()(key1), std::hash<CompactIntsKey<KeySize>>()(key2));
+  EXPECT_TRUE(std::less<CompactIntsKey<KeySize>>()(key1, key2));
+
+  delete[] pr_buffer;
+}
+
+// NOLINTNEXTLINE
```
jrolli: Test description?

src/include/storage/index/index.h
```diff
@@ -37,6 +39,17 @@ class Index {
...
 
+  /**
+   * Determine if a tuple is visible by asking the DataTable associated with the TupleSlot. Used for scans.
+   * @param txn the calling transaction
+   * @param slot the slot of the tuple to check visibility on
+   * @return true if tuple is visible to this txn, false otherwise
+   */
+  static bool IsVisible(const transaction::TransactionContext &txn, const TupleSlot slot) {
+    const auto *const data_table = slot.GetBlock()->data_table_;
+    return data_table->IsVisible(txn, slot);
```
jrolli: We need more logic here.  This just answers is the *tuple* visible, but we actually need to ask is the *tuple with the indexed attribute I'm looking for* visible.  Specifically, updates on indexed attributes will break this logic.

test/storage/bwtree_index_test.cpp
mbutrovich: See line 1197 for description. It's a Delete and Insert because it's an update on an indexed attribute.

src/include/storage/index/bwtree_index.h
```diff
@@ -31,61 +28,178 @@ class BwTreeIndex final : public Index {
...
 
-    // predicate is set to nullptr if the predicate returns true for some value
-    const bool ret = bwtree_->ConditionalInsert(index_key, location, predicate, &predicate_satisfied);
+    // The predicate checks if any matching keys have write-write conflicts or are still visible to the calling txn.
+    auto predicate = [&](const TupleSlot slot) -> bool {
+      const auto *const data_table = slot.GetBlock()->data_table_;
+      return data_table->HasConflict(*txn, slot) || data_table->IsVisible(*txn, slot);
+    };
+
+    const bool result = bwtree_->ConditionalInsert(index_key, location, predicate, &predicate_satisfied);
```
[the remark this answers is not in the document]
jrolli: This is not a concern because we are requiring a table-level delete and insert.

test/storage/bwtree_index_test.cpp
jrolli: We'd pass this, but its probably still good to have this race documented somewhere.

src/include/storage/index/index.h
jrolli: This should be good because of the table-level insert.

src/include/storage/data_table.h
```diff
@@ -15,6 +15,12 @@ class TransactionManager;
 
 namespace terrier::storage {
 
+namespace index {
+class Index;
+template <typename KeyType>
+class BwTreeIndex;
```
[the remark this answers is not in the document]
mbutrovich: I can see about moving all of the necessary logic up to the Index level so the underlying wrappers don't need to talk to the DataTable.

src/include/storage/data_table.h
mbutrovich: Reads (scans) only care about IsVisible. Writes (InsertUnique, Delete) care about IsVisible and HasConflict.

src/include/storage/index/bwtree_index.h
```diff
@@ -1,15 +1,12 @@
...
 #include "bwtree/bwtree.h"
-#include "catalog/catalog_defs.h"
-#include "storage/index/compact_ints_key.h"
 #include "storage/index/index.h"
-#include "storage/index/index_defs.h"
-#include "storage/index/index_metadata.h"
-#include "storage/sql_table.h"
-#include "storage/storage_defs.h"
+#include "transaction/transaction_context.h"
+#include "transaction/transaction_manager.h"
```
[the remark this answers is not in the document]
mbutrovich: Deferring an action on line 89.

src/include/storage/index/bwtree_index.h
```diff
@@ -31,61 +28,178 @@ class BwTreeIndex final : public Index {
...
+    // Build search key
     KeyType index_key;
     index_key.SetFromProjectedRow(key, metadata_);
-    bwtree_->GetValue(index_key, *value_list);
+
+    // Perform lookup in BwTree
+    bwtree_->GetValue(index_key, results);
+
+    // Avoid resizing our value_list, even if it means over-provisioning
+    value_list->reserve(results.size());
```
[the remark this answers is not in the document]
mbutrovich: 2 different vectors. value_list comes in as an argument where I don't know the reserved size a priori, just that it should be empty. I resize it to the bwtree scan result's size to guarantee it has enough size to store all of the results (some may be filtered out in the process).

src/include/storage/data_table.h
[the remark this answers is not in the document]
mbutrovich: I'm not seeing a clean way to do this without modifying the underlying bwtree's predicate checking lambda to take variadic arguments (so I can pass the txn context). Right now it's captured implicitly by reference from the calling scope, but if I move the predicate function up to the wrapper level, then I need to add it as an argument and can't pass that function directly to the bwtree because it only expects 1 argument (the value type).

src/include/storage/data_table.h
```diff
@@ -263,5 +276,15 @@ class DataTable {
   void NewBlock(RawBlock *expected_val);
 
   void DeallocateVarlensOnShutdown(RawBlock *block);
+
+  /**
+   * Determine if a Tuple is visible (present and not deleted) to the given transaction. It's effectively Select's logic
+   * (follow a version chain if present) without the materialization. If the logic of Select changes, this should change
```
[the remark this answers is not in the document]
mbutrovich: We've avoided using nullptr as a sentinel value in arguments to this point, I think.

test/storage/bwtree_key_test.cpp
mbutrovich: This isn't new code but sure I can come up with one.

test/storage/bwtree_index_test.cpp
mbutrovich: That would actually be enforced at the Data Table level, but I can try to figure out a good place to better describe the index semantics.

src/include/storage/data_table.h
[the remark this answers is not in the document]
jrolli: Ok.  Maybe put a TODO or some other note here in case we end up replacing the BwTree with something else?

test/storage/bwtree_index_test.cpp
jrolli: I think your new comment at the top of `Index.h` is fine.

mbutrovich:
I was trying to avoid the API change to `InitializeRawBlock`, but I guess we'll live with it.

src/include/storage/data_table.h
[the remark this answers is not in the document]
mbutrovich: I don't think there is there right place for it. I'll add more to the top of index.h describing how to add future indexes.

src/include/storage/index/bwtree_index.h
```diff
@@ -31,61 +28,182 @@ class BwTreeIndex final : public Index {
...
+        TERRIER_ASSERT(result, "Deferred delete on the index failed.");
+      });
+    });
   }
 
-  void ScanKey(const ProjectedRow &key, std::vector<TupleSlot> *value_list) final {
-    TERRIER_ASSERT(
-        value_list->empty(),
-        "Result set should begin empty. This can be changed in the future if index scan behavior requires it.");
+  void ScanKey(const transaction::TransactionContext &txn, const ProjectedRow &key,
```
yeshengm: Since we are assuming the `value_list` is non-null, why not make it a reference?

src/include/storage/index/index.h
```diff
@@ -50,46 +69,84 @@ class Index {
...
   /**
-   * Inserts a new key-value pair into the index.
+   * Inserts a new key-value pair into the index, used for non-unique key indexes.
+   * @param txn txn context for the calling txn, used to register abort actions
    * @param tuple key
    * @param location value
    * @return false if the value already exists, true otherwise
    */
-  virtual bool Insert(const ProjectedRow &tuple, TupleSlot location) = 0;
+  virtual bool Insert(transaction::TransactionContext *txn, const ProjectedRow &tuple, TupleSlot location) = 0;
```
yeshengm: From our experience with this method, there is a usability issue. For example, when we are building the index, we scan through the whole table, getting a bunch of `TupleSlot`s and related `ProjectedRow`s.

However, that projected row in the table can not be directly used in this `Insert` method (as well as others) since the internal metadata representation of "index projected rows" and "selected projected row from table" are different. I think the major reason is that indexes and tables are maintaining their own `col_oid`, `col_id`, actual layout id mappings.

yeshengm: I think it could be nice to unify them or provide some handy functions.

src/include/storage/data_table.h
```diff
@@ -246,7 +256,10 @@ class DataTable {
   void AtomicallyWriteVersionPtr(TupleSlot slot, const TupleAccessStrategy &accessor, UndoRecord *desired);
 
   // Checks for Snapshot Isolation conflicts, used by Update
-  bool HasConflict(UndoRecord *version_ptr, const transaction::TransactionContext *txn) const;
+  bool HasConflict(const transaction::TransactionContext &txn, UndoRecord *version_ptr) const;
```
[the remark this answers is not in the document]
mbutrovich: We use pointers for arguments when the object needs modifying within the function, and const reference otherwise.

src/include/storage/data_table.h
```diff
@@ -263,5 +276,15 @@ class DataTable {
...
+
+  /**
+   * Determine if a Tuple is visible (present and not deleted) to the given transaction. It's effectively Select's logic
+   * (follow a version chain if present) without the materialization. If the logic of Select changes, this should change
+   * with it and vice versa.
+   * @param txn the calling transaction
+   * @param slot the slot of the tuple to check visibility on
+   * @return true if tuple is visible to this txn, false otherwise
+   */
+  bool IsVisible(const transaction::TransactionContext &txn, TupleSlot slot) const;
```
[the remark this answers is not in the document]
mbutrovich: We use pointers for arguments when the object needs modifying within the function, and const reference otherwise.

src/include/storage/index/bwtree_index.h
mbutrovich: We use pointers for arguments when the object needs modifying within the function, and const reference otherwise.

src/include/storage/index/index.h
mbutrovich: It's the execution layer's job to reason about passing well-formed buffers to the storage layer. Indexes don't know anything about tables or how to transform from a table's schema to the index's schema. In the case of indexes on expressions, you'd be pushing even more interpretation logic into the index wrapper. I don't think that's the right call. Let that be compiled code in the execution layer.

yeshengm: Sounds reasonable.

PR #358, opened 2019-04-17
Add action framework

jrolli:
This is an implementation of the epoch-based callback framework discussed during the meeting today and is a superset of the abort callbacks that #351 implements.  The framework allows the database to register callbacks into a transaction context to execute at either abort or commit (commit callbacks are currently disallowed in read-only transactions).  Additionally, deferred callbacks can be registered with the TransactionManager which will execute once all currently active transactions have exited (same logic as cleaning up undo records).

In order to allow the specific use case of cleaning up arbitrary side-effects within the system (temporary constraints, staged-then-aborted schema versions, etc.), I had to expose  the transaction manager to the storage layer through the transaction context.  This was the only way that I could figure out of how to get the correct function handle visible when building the lambdas for the callback chain.

### DIFF
--- a/benchmark/storage/data_table_benchmark.cpp
+++ b/benchmark/storage/data_table_benchmark.cpp
@@ -88,7 +88,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, SimpleInsert)(benchmark::State &state) {
     storage::DataTable table(&block_store_, layout_, storage::layout_version_t(0));
     // We can use dummy timestamps here since we're not invoking concurrency control
     transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
-                                        LOGGING_DISABLED, nullptr);
+                                        LOGGING_DISABLED, ACTION_FRAMEWORK_DISABLED);
     for (uint32_t i = 0; i < num_inserts_; ++i) {
       table.Insert(&txn, *redo_);
     }
@@ -106,7 +106,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentInsert)(benchmark::State &state
     auto workload = [&](uint32_t id) {
       // We can use dummy timestamps here since we're not invoking concurrency control
       transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
-                                          LOGGING_DISABLED, nullptr);
+                                          LOGGING_DISABLED, ACTION_FRAMEWORK_DISABLED);
       for (uint32_t i = 0; i < num_inserts_ / num_threads_; i++) table.Insert(&txn, *redo_);
     };
     common::WorkerPool thread_pool(num_threads_, {});
@@ -123,7 +123,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, SequentialRead)(benchmark::State &state) 
   // Populate read_table by inserting tuples
   // We can use dummy timestamps here since we're not invoking concurrency control
   transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
-                                      LOGGING_DISABLED, nullptr);
+                                      LOGGING_DISABLED, ACTION_FRAMEWORK_DISABLED);
   std::vector<storage::TupleSlot> read_order;
   for (uint32_t i = 0; i < num_reads_; ++i) {
     read_order.emplace_back(read_table.Insert(&txn, *redo_));
@@ -145,7 +145,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, RandomRead)(benchmark::State &state) {
   // Populate read_table_ by inserting tuples
   // We can use dummy timestamps here since we're not invoking concurrency control
   transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
-                                      LOGGING_DISABLED, nullptr);
+                                      LOGGING_DISABLED, ACTION_FRAMEWORK_DISABLED);
   std::vector<storage::TupleSlot> read_order;
   for (uint32_t i = 0; i < num_reads_; ++i) {
     read_order.emplace_back(read_table.Insert(&txn, *redo_));
@@ -170,7 +170,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentRandomRead)(benchmark::State &s
   // populate read_table_ by inserting tuples
   // We can use dummy timestamps here since we're not invoking concurrency control
   transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
-                                      LOGGING_DISABLED, nullptr);
+                                      LOGGING_DISABLED, ACTION_FRAMEWORK_DISABLED);
   std::vector<storage::TupleSlot> read_order;
   for (uint32_t i = 0; i < num_reads_; ++i) {
     read_order.emplace_back(read_table.Insert(&txn, *redo_));
@@ -188,7 +188,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentRandomRead)(benchmark::State &s
     auto workload = [&](uint32_t id) {
       // We can use dummy timestamps here since we're not invoking concurrency control
       transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
-                                          LOGGING_DISABLED, nullptr);
+                                          LOGGING_DISABLED, ACTION_FRAMEWORK_DISABLED);
       for (uint32_t i = 0; i < num_reads_ / num_threads_; i++)
         read_table.Select(&txn, read_order[(rand_read_offsets[id] + i) % read_order.size()], reads_[id]);
     };
--- a/benchmark/storage/data_table_benchmark.cpp
+++ b/benchmark/storage/data_table_benchmark.cpp
@@ -88,7 +88,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, SimpleInsert)(benchmark::State &state) {
     storage::DataTable table(&block_store_, layout_, storage::layout_version_t(0));
     // We can use dummy timestamps here since we're not invoking concurrency control
     transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
-                                        LOGGING_DISABLED);
+                                        LOGGING_DISABLED, nullptr);
     for (uint32_t i = 0; i < num_inserts_; ++i) {
       table.Insert(&txn, *redo_);
     }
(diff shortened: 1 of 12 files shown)


benchmark/storage/data_table_benchmark.cpp
```diff
@@ -88,7 +88,7 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, SimpleInsert)(benchmark::State &state) {
     storage::DataTable table(&block_store_, layout_, storage::layout_version_t(0));
     // We can use dummy timestamps here since we're not invoking concurrency control
     transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
-                                        LOGGING_DISABLED);
+                                        LOGGING_DISABLED, nullptr);
```
tli2: nit: Probably best to be consistent and use a MACRO that is more meaningful

src/include/storage/garbage_collector.h
```diff
@@ -64,6 +69,8 @@ class GarbageCollector {
   transaction::TransactionQueue txns_to_deallocate_;
   // queue of txns that need to be unlinked
   transaction::TransactionQueue txns_to_unlink_;
+  // queue of unexecuted deferred actions
+  transaction::ActionQueue deferred_actions_;
```
tli2: Can we switch to a templated typedef instead?

src/include/transaction/transaction_context.h
```diff
@@ -30,10 +30,14 @@ class TransactionContext {
    * @param txn_id the id of the transaction, should be larger than all start time and commit time
    * @param buffer_pool the buffer pool to draw this transaction's undo buffer from
    * @param log_manager pointer to log manager in the system, or nullptr, if logging is disabled
+   * @param txn_mgr pointer to transaction manager int the system (used for action framework)
```
tli2: nit: for consistency's sake, should probably not abbreviate

src/include/transaction/transaction_manager.h
```diff
@@ -75,6 +75,21 @@ class TransactionManager {
    */
   TransactionQueue CompletedTransactionsForGC();
 
+  /**
+   * Adds the action to a buffered list of deferred actions.  This action will
+   * be triggered no sooner than when the epoch (timestamp of oldest running
+   * transaction) is more recent than the time this function was called.
+   * @param a functional implementation of the action that is deferred
+   */
+  void DeferAction(Action a);
```
tli2: Should this be a collection of actions? Transactions can have lots of actions and they can be safely triggered at once using one timestamp. Calling this function in a loop is unnecessary.

src/include/transaction/transaction_context.h
```diff
@@ -116,6 +120,27 @@ class TransactionContext {
     storage::DeleteRecord::Initialize(redo_buffer_.NewEntry(size), start_time_, table, slot);
   }
 
+  /**
+   * Defers an action to be called if and only if the transaction aborts.  Actions executed LIFO.
+   * @param a the action to be executed
+   */
+  void RegisterAbortAction(Action const &a) { abort_actions_.push_front(a); }
```
tli2: nit: our code base do `const ClassName &`

src/transaction/transaction_manager.cpp
```diff
@@ -45,6 +45,7 @@ void TransactionManager::LogCommit(TransactionContext *const txn, const timestam
 
 timestamp_t TransactionManager::ReadOnlyCommitCriticalSection(TransactionContext *const txn, const callback_fn callback,
                                                               void *const callback_arg) {
+  TERRIER_ASSERT(txn->commit_actions_.empty(), "Only updating transactions can have deferred actions");
```
tli2: This might not be true in the future. (For example, callback from logging can actually be replaced with this deferred action as well, eliminating work on the logging thread) We can either leave it here with a comment that this may change in the future or remove it. Up to you.

test/transaction/deferred_actions_test.cpp
```diff
@@ -0,0 +1,299 @@
...
+    TerrierTest::TearDown();
+  }
+
+  storage::RecordBufferSegmentPool buffer_pool_ = {100, 100};
+  transaction::TransactionManager txn_mgr_ = {&buffer_pool_, true, LOGGING_DISABLED};
+  storage::GarbageCollector gc_;
+};
+
+// NOLINTNEXTLINE
+TEST_F(DeferredActionsTest, AbortAction) {
```
tli2: nit: We usually have a comment block at the beginning of each test explaining what it does at a high level.

tli2:
Otherwise mostly LGTM. Thanks for the quick PR!

src/transaction/transaction_manager.cpp
jrolli: If we allow read-only transactions to trigger actions, then we need to reason about how much of this function needs to be atomic.  The current code assumes that read-only transactions have no impact on the actual database and therefore do not need the commit latch.

My inclination is to comment/document the current assumption, unless someone else weighs in.

src/include/transaction/transaction_manager.h
jrolli: I think this is the best interface for epoch deferred actions (abort & commit actions don't go through this call).  While I think transactions can have lots of actions, I think the majority will either occur in bulk at abort or commit or be added piecemeal during a transaction's lifetime and be hard to aggregate into a group deferral without more coupling.

Is there a specific use case you have in mind?

src/include/storage/garbage_collector.h
jrolli: Can you elaborate?  Would you prefer that we no have the `ActionQueue` type and instead invoke the underlying template inline here?

benchmark/storage/data_table_benchmark.cpp
jrolli: Fixed

test/transaction/deferred_actions_test.cpp
jrolli: I've added comments.  Please let me know if you think I need to add more depth on any of them.

src/include/storage/garbage_collector.h
tli2: I meant that maybe we should have:

```
template <class Item>
using Queue = std::forward_list<Item>
...
transaction::Queue<Action *>
```

On second thought, it would be okay to simply explicitly say `forward_list` for all these. I think the typedef is leftover from back when we used `std::queue<transaction::TransactionContext *, std::list<transaction::TransactionContext *>>`

src/include/transaction/transaction_manager.h
tli2: Sure. We can add group of actions later if needed.

src/transaction/transaction_manager.cpp
tli2: Yeah that sounds like you should write a TODO

src/include/storage/garbage_collector.h
jrolli: Unfortunately, `forward_list` works for the GC containers because ordering doesn't really matter, but it would be painful for the `ActionQueue` because it is linear time to append (instead of constant time for `queue` (backed by either `deque` or `list`).

tli2: Oh sorry. Misread that one.

We can probably just lose the typedef for your use case.

jrolli: The only reason I lean towards the typedef is that it makes it easier to transparently swap at the underlying data structure if it becomes a problem.

tli2: We have 3 usages of this in our code base right now. I personally like to avoid typedefs unless the name really is just too long.

Typedef-ing does not provide information hiding. Code can still call whatever specialized function on the object they want. Wrappers are much better at doing that.

src/transaction/transaction_manager.cpp
```diff
@@ -167,6 +185,18 @@ TransactionQueue TransactionManager::CompletedTransactionsForGC() {
...
+void TransactionManager::DeferAction(Action a) {
+  common::SpinLatch::ScopedSpinLatch guard(&deferred_actions_latch_);
+  deferred_actions_.push({time_.load(), a});
+}
+
+ActionQueue TransactionManager::DeferredActionsForGC() {
+  common::SpinLatch::ScopedSpinLatch guard(&deferred_actions_latch_);
+  ActionQueue hand_to_gc(std::move(deferred_actions_));
+  TERRIER_ASSERT(deferred_actions_.empty(), "TransactionManager's queue should now be empty.");
+  return hand_to_gc;
```
mbutrovich: I think we just want to `return std::move(deferred_actions_)` and lose the `TERRIER_ASSERT`. We can update the `CompletedTransactionsForGC()` method at the same time, since it uses the same logic. The return type will probably have to be changed to r-value reference. If the goal was to avoid copies, we failed. :)

src/include/transaction/transaction_context.h
```diff
@@ -116,6 +120,27 @@ class TransactionContext {
...
+  void RegisterCommitAction(const Action &a) { commit_actions_.push_front(a); }
+
+  /**
+   * Get the transaction manager responsible for this context (should be singleton).
+   * @warning This should only be used to support dynamically generating deferred actions
+   *          at abort or commit.  We need to expose the transaction manager because that
+   *          is where the deferred actions queue exists.
+   * @return the transaction manager
+   */
+  TransactionManager *GetTransactionManager() { return txn_mgr_; }
```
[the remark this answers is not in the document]
jrolli: I'm specifically thinking of something like aborting a schema change after having staged the new schema in `SqlTable`'s map.  Nothing above `SqlTable` can see this map so the lambda needs to be generated inside of `SqlTable` but that doesn't have a reference to the transaction manager.  We could add it as a property, but then we're committing to reworking constructors every time we have a new place for side-effects.

jrolli:
> Just so I understand, every action needs a unique timestamp? Is there a reason we can't coalesce actions from a txn under a single timestamp?

Not quite.  Every action needs a timestamp, but that timestamp is not unique.  Because it just needs to ensure no *currently running* transaction is still running when it is executed, we use the transaction manager's internal timestamp as the epoch mark.  Once the oldest transaction is greater than or equal to this (because it hadn't been handed out when we recorded it), we are safe to execute.

Additionally, coalescing actions would be extremely difficult since we're not handing a queue of actions from the transaction to the transaction manager.  The transaction context contains event-oriented (commit or abort) actions that may contain code that generates and defers time-based actions.

src/include/transaction/transaction_context.h
```diff
@@ -116,6 +120,27 @@ class TransactionContext {
     storage::DeleteRecord::Initialize(redo_buffer_.NewEntry(size), start_time_, table, slot);
   }
 
+  /**
+   * Defers an action to be called if and only if the transaction aborts.  Actions executed LIFO.
+   * @param a the action to be executed
+   */
+  void RegisterAbortAction(const Action &a) { abort_actions_.push_front(a); }
```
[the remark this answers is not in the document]
jrolli: As far as I know we have not.  I think the only reliable rule is that code/objects should only be registering actions that read/write/modify/delete data it owns.

src/include/storage/garbage_collector.h
jrolli: Correct it doesn't provide information hiding, but it provides one place to change the underlying data structure with an API compatible replacement, instead of each and every usage point.

At this point I'm not sure what you think it should look like.  Would you prefer this be converted to a wrapper?

src/transaction/transaction_manager.cpp
jrolli: After talking with @mbutrovich, it made more sense to move the commit actions to after the critical section in commit.  Making this change negated the need for the constraint.

src/include/storage/garbage_collector.h
tli2: I think for now just use the c++ typename directly.

PR #363, opened 2019-04-23
Reduce contention on commit latch

jrolli:
This commit makes two fundamental changes:
1. It inverts the two latches present in `TransactionManager::BeginTransaction` so that the latch for the current running transaction data structure prevents the race condition between a beginning transaction, committing transaction, and the garbage collector.
2. It replaces the shared commit latch with a spin latch (previously the exclusive acquire) and an atomic boolean used as a barrier.  New transactions spin on a read-only loop when the atomic boolean is false, and committing transactions (with writes) acquire the commit latch, set the atomic boolean to true, and only then grab their commit timestamp.  This ordering ensures that a new transaction with a more recent timestamp than the committer does not return from `BeginTransaction` until the committer has finished making all of their writes visible.


src/transaction/transaction_manager.cpp
```diff
@@ -5,21 +5,40 @@
...
+  {
+    // There is a three-way race that needs to be prevented.  Specifically, we
+    // cannot allow both a transaction to commit and the GC to poll for the
+    // oldest running transaction in between this transaction acquiring its
+    // begin timestamp and getting inserted into the current running
+    // transactions list.  Using the current running transactions latch
+    // prevents the GC from polling and stops the race.  This allows us to
+    // replace acquiring a shared instance of the commit latch with a
+    // read-only spin-latch and move the allocation out of a critical section.
+    common::SpinLatch::ScopedSpinLatch running_guard(&curr_running_txns_latch_);
```
tli2: How do you plan to replace this with thread-local implementation of `curr_running_txns` in the future?

src/transaction/transaction_manager.cpp
```diff
@@ -5,21 +5,40 @@
...
-  TERRIER_ASSERT(ret.second, "commit start time should be globally unique");
+
+  // This transaction cannot proceed if there is an ongoing write-commit by a
+  // transaction whose commit time is less than this transaction's begin time
+  //
+  // TODO(John):  This check could be replaced with a timestamp comparison to
+  // potentially allow some transactions who may be blocked unnecessarily to
+  // proceed.
+  while (blocking_commit_.load()) {
+  }
```
tli2: Should you use `_mm_pause()` or whatever the fancy new Intel thing is (`__asm__ __volatile__("pause;")` in tbb? check their source code)?

src/transaction/transaction_manager.cpp
```diff
@@ -57,7 +76,8 @@ timestamp_t TransactionManager::ReadOnlyCommitCriticalSection(TransactionContext
 
 timestamp_t TransactionManager::UpdatingCommitCriticalSection(TransactionContext *const txn, const callback_fn callback,
                                                               void *const callback_arg) {
-  common::SharedLatch::ScopedExclusiveLatch guard(&commit_latch_);
+  common::SpinLatch::ScopedSpinLatch guard(&commit_latch_);
+  blocking_commit_ = true;
```
tli2: Looking at this, it's not immediately obvious what this does. I am thinking whether there is some abstraction (`XXLatch`) that captures this thing we are trying to do here. I would much prefer the spin latch and the boolean
flag be incorporated into a class so any future users don't have to mess with the internals. I don't have a good name for it though. @mbutrovich any ideas?

src/transaction/transaction_manager.cpp
```diff
@@ -80,7 +100,7 @@ timestamp_t TransactionManager::UpdatingCommitCriticalSection(TransactionContext
   LogCommit(txn, commit_time, callback, callback_arg);
   // flip all timestamps to be committed
   for (auto &it : txn->undo_buffer_) it.Timestamp().store(commit_time);
-
+  blocking_commit_ = false;
```
tli2: Just curious, do we get a speed up if you pad this and the spin latch to be on different cache lines (insert `byte padding_[64]` in between and try it I guess)

src/transaction/transaction_manager.cpp
jrolli: Assuming the thread's context is passed in as a parameter (same as project 1), then you just replace this latch and the corresponding container with the latch and container in the thread context.

src/transaction/transaction_manager.cpp
jrolli: So I kind of want to call it something along the lines of a "Nand Gate" since it essentially ensures threads can proceed through one of two code points, but not both.  As best as I can tell, there's not really a good synchronization term for what this does.

src/transaction/transaction_manager.cpp
jrolli: This is worth doing but not because of interference with the spin latch.  It improves the select benchmark by about 10% (which has no use of the spin latch).  My best guess is that it ensures that current running transactions is in a different cache line.  When we pull this out into a separate class/object, we should make sure that it occupies an entire cache line by itself.

src/transaction/transaction_manager.cpp
jrolli: A quick test of this slowed down the benchmarks by 2-3%.

jrolli:
Based off discussions with @mbutrovich, closing this request and only proceed with the more complete optimization in #364.

PR #364, opened 2019-04-24
Remove the commit latch

jrolli:
This is an experimental version of the precision commit (#363) that removes the commit latch altogether and allows multiple committers to commit concurrently.  This should be correct, but its unclear whether we have sufficient testing (particularly on logging & recovery side) to ensure this still produces correct results.


src/include/transaction/transaction_manager.h
```diff
@@ -99,7 +99,7 @@ class TransactionManager {
   std::atomic<timestamp_t> time_{timestamp_t(0)};
 
   // TODO(Tianyu): This is the famed HyPer Latch. We will need to re-evaluate performance later.
-  common::SharedLatch commit_latch_;
+  common::Gate txn_gate_;
```
mbutrovich: We can remove the todo from Tianyu. Also should it be "commit_gate_" if that's the only place it's used (or do you foresee that changing?)

mbutrovich:
Also just to make these more visible:
```
pass 9.246e+05 8.222e+05 10     +12.   LargeTransactionBenchmark    SingleStatementUpdate
pass 1.391e+06 1.244e+06 10     +11    LargeTransactionBenchmark    SingleStatementSelect
pass 9.045e+05 7.908e+05 10     +14    LargeTransactionBenchmark    SingleStatementInsert
```

src/include/transaction/transaction_manager.h
jrolli: I removed this comment and updated the comment in `TransactionManager::UpdatingCommitCriticalSection` to reflect the new state for the code.

With regard to the name of the gate, I think its a policy question of whether we name it for where the "gate" is (BeginTransaction) or where it gets locked/unlocked (Commit).  I'm fine either way, but it made more sense to me to label it by where the effect occurs.

PR #370, opened 2019-05-05
Logging with Two Threads

thepulkitagarwal:
Before this PR, our logging was single threaded. Buffers were pushed into a queue by transactions, the buffers were then popped from the queue, serialized, and persisted in disk. This is slow as we can perform the serialization and persisting in parallel.

This PR performs a large reformatting on the log manager. How transactions interact with the log manager remains the same. Transactions hand off buffers to the log manager, which are added to an internal queue. This minimizes the amount of time transactions spend interacting with the log manager.

The main change is the addition of two tasks running on dedicated threads (using the DedicatedThreadRegistry):

1. `LogSerializerTask`: This task pops unserialized buffers from the internal unserialized buffer queue, serializes them, and adds them to a consumer queue to be consumed by the consumer task.
2. `DiskLogConsumerTask`: This task pops serialized buffers from the consumer queue and writes them to the log file. It will also persist the log file under certain conditions (periodically, we have written a significant amount of data since the last persist, or someone forces a flush). The task will also call any corresponding commit callbacks when it persists commit records.

https://www.youtube.com/watch?v=oiY_iKSpWLM


src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -7,11 +7,15 @@
 #include <vector>
 #include "common/spin_latch.h"
 #include "common/strong_typedef.h"
+#include "spdlog/details/mpmc_blocking_q.h"
```
tli2: The use of this seems like a bad idea. Debug logging is supposed to be for debug logging only.

If you want to use this data structure, write a replaceable wrapper (see what we did for concurrent hash table and spin latch)

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -7,11 +7,15 @@
...
 #include "storage/record_buffer.h"
 #include "storage/write_ahead_log/log_io.h"
 #include "storage/write_ahead_log/log_record.h"
 #include "transaction/transaction_defs.h"
 
+#define MAX_BUF 2
```
tli2: Add a TODO to push these to settings manager later

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -28,15 +32,31 @@ class LogManager {
...
+        filled_buffer_queue_(MAX_BUF),
+        run_log_writer_thread_(true),
+        do_persist_(true) {
+    for (int i = 0; i < MAX_BUF; i++) {
+      buffers_.emplace_back(BufferedLogWriter(log_file_path));
+    }
+    for (int i = 0; i < MAX_BUF; i++) {
+      BlockingEnqueueBuffer(&buffers_[i], &empty_buffer_queue_);
+    }
+    log_writer_thread_ = std::thread([this] { WriteToDiskLoop(); });
```
tli2: We don't want to couple log thread start with log manager object creation. Eventually we want to link every thread to a central registry for self-driving stuff. Can you provide start methods instead? (See GC for example)

src/storage/write_ahead_log/log_manager.cpp
```diff
@@ -12,50 +12,118 @@ void LogManager::Process() {
...
+  do_persist_ = true;
+  // Wait for the log writer thread to persist the logs
+  while (do_persist_) {
+  }
+  // Execute the callbacks for the transactions that have been persisted
   for (auto &callback : commits_in_buffer_) callback.first(callback.second);
   commits_in_buffer_.clear();
 }
 
+void LogManager::SerializeTaskBuffer(IterableBufferSegment<LogRecord> *const task_buffer) {
```
tli2: You need some comments explaining the various names you are using (e.g. task buffer)

src/storage/write_ahead_log/log_manager.cpp
```diff
@@ -64,4 +132,59 @@ void LogManager::SerializeRecord(const terrier::storage::LogRecord &record) {
   }
 }
 
+void LogManager::WriteToDiskLoop() {
```
tli2: I suspect you want to decouple the disk writer from the serializer into a separate object for better understandability

src/include/common/container/producer_consumer_queue.h
```diff
@@ -0,0 +1,67 @@
+#pragma once
+
+#include <utility>
+#include "common/macros.h"
+#include "spdlog/details/mpmc_blocking_q.h"
```
codeworm96: spdlog is a logging library. This MPMC queue is its own implementation detail. For me, it seems not a good choice. TBB already provides two kinds of concurrent queues. The wrapper of one is already provided in https://github.com/cmu-db/terrier/blob/master/src/include/common/container/concurrent_queue.h. And our checkpoint group already implemented the wrapper for the other one in https://github.com/cmu-db/terrier/blob/dd70de38f0d649ab38937750aae5c735869f0215/src/include/common/container/concurrent_blocking_queue.h. I hope these will be helpful.

utkarsh39: We need both blocking and non-blocking dequeue. That's why we resorted to MPMC

pervazea: I don't think this is a good approach.

I do not believe that the internals (i.e. the details directory) of spdlog is intended to be public. The documentation and examples use the spdlog.h header file along with  the various loggers as the public API. There is no promise that some future version of spdlog won't replace the internals.

So, possible choices that come to mind:
- Use one of the queue libraries e.g. the TBB or similar. This would be preferred.
- If you must use MPMC, then place a copy in a separate third party directory and use it from there. Add some basic units tests. You do not want to be dependent on the internal components of packages.

src/include/storage/data_table.h
```diff
@@ -207,6 +207,12 @@ class DataTable {
    */
   DataTableCounter *GetDataTableCounter() { return &data_table_counter_; }
 
+  /**
+   * Returns a read-only view of this DataTable's BlockLayout.
+   * @return this DataTable's BlockLayout.
```
tli2: Just write the @return for trivial comments like this

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -98,41 +98,48 @@ class BufferedLogWriter {
    * update is only written out when the BufferedLogWriter is persisted.
    * @param data memory location of the bytes to write
    * @param size number of bytes to write
+   * @return number of bytes written
```
tli2: You need to additionally explain why this is different from `size` param. How one should use this return field, etc.

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -28,15 +33,51 @@ class LogManager {
...
+    }
+    run_log_writer_thread_ = true;
+    log_writer_ = new LogWriter(this);
+  }
 
   /**
    * Must be called when no other threads are doing work
    */
   void Shutdown() {
+    // Process the remaining redo buffers
```
tli2: I don't think the LogManager should own the log writer thread. As we have discussed,  we want a centralized registry for all these threads for self-driving purposes.

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -68,25 +109,97 @@ class LogManager {
...
+   * Serialize out the record to the log
+   * @param record the redo record to serialise
+   */
   void SerializeRecord(const LogRecord &record);
 
+  /**
+   * Serialize out the task buffer to the log
+   * @param task_buffer the iterator to the redo buffer to be serialized
+   */
+  void SerializeTaskBuffer(IterableBufferSegment<LogRecord> *task_buffer);
```
tli2: Just call this `SerializeBuffer` and the parameter `buffer_to_serialize`?

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -68,25 +109,97 @@ class LogManager {
...
+
+  /**
+   * Used by the serializer thread to get a buffer to serialize data to
+   * @return buffer to write to
+   */
+  BufferedLogWriter *GetBufferToWrite() {
```
tli2: `GetCurrentWriteBuffer`?

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -68,25 +109,97 @@ class LogManager {
...
+   * Serialize the data pointed to by val to a buffer
+   * @param val the value
+   * @param size size of the value to serialize
+   */
+  void WriteValue(const void *val, uint32_t size);
+
+  /**
+   * Mark the current buffer that the serializer thread is writing to as filled
+   */
+  void MarkBufferFull() {
```
tli2: You are not merely marking the buffer as full, but also doing stuff with it, right? The name should probably reflect that. (e.g. `HandFilledBufferToWriter`?)

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -68,25 +109,97 @@ class LogManager {
...
+    filled_buffer_queue_.Enqueue(buffer_to_write_);
+    // Signal writer thread that a buffer is ready to be flushed to the disk
+    {
+      std::unique_lock<std::mutex> lock(persist_lock_);
+      wake_writer_thread_cv_.notify_one();
+    }
+    // Mark that serializer thread doesn't have a buffer in its possession to which it can write to
+    buffer_to_write_ = nullptr;
```
tli2: rename in accordance with function name above

src/include/storage/write_ahead_log/log_writer.h
```diff
@@ -0,0 +1,43 @@
...
+
+namespace terrier::storage {
+
+// Forward declaration for class LogManager
+class LogManager;
+/**
+ * A LogWriter is responsible for writing serialized log records out to disk.
+ * @param log_manager pointer to the LogManager
+ */
+class LogWriter {
```
tli2: `LogWriter` is not a good name, because we already have `BufferedLogWriter` in the system. The name should reflect the fact that we are taking serialized log and persisting to a log file on disk. Presumably in the future there will be different types of this (e.g. streaming log to the network.)

Following this line of thought, maybe `LogConsumer` is a good name for the interface. Then we can prefix this with the actual use case.

src/storage/write_ahead_log/log_writer.cpp
```diff
@@ -0,0 +1,59 @@
...
+  // Persist the buffers
+  TERRIER_ASSERT(!(log_manager_->buffers_.empty()), "Buffers vector should not be empty until Shutdown");
+  // Force the buffers to be written to disk
+  log_manager_->buffers_.front().Persist();
+}
+
+void LogWriter::WriteToDisk() {
+  // Log writer thread spins in this loop
+  // It dequeues a filled buffer and flushes it to disk
+  while (log_manager_->run_log_writer_thread_) {
```
tli2: As before, strip the wake-up logic and place elsewhere.

benchmark/storage/logging_benchmark.cpp
```diff
@@ -86,7 +87,10 @@ BENCHMARK_DEFINE_F(LoggingBenchmark, TPCCish)(benchmark::State &state) {
     log_manager_ = new storage::LogManager(LOG_FILE_NAME, &buffer_pool_);
     LargeTransactionBenchmarkObject tested(attr_sizes, initial_table_size, txn_length, insert_update_select_ratio,
                                            &block_store_, &buffer_pool_, &generator_, true, log_manager_);
-    log_manager_->Process();  // log all of the Inserts from table creation
+    // log all of the Inserts from table creation
+    StartLogging();
```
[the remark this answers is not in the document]
thepulkitagarwal: I think that log_test is also creating a different thread ([link](https://github.com/cmu-db/terrier/blob/master/test/storage/log_test.cpp#L17)).

src/storage/write_ahead_log/log_writer.cpp
utkarsh39: We h

src/include/storage/write_ahead_log/log_manager.h
utkarsh39: As of now, what do you suggest would be the right way to do this?

benchmark/storage/logging_benchmark.cpp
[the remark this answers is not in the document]
utkarsh39: To do this, we would need to add Pause mechanism to LogManager which would complicate the log manager and I don't think is necessary. log_test doesn't have to do this as it doesn't exclude some specific logging for benchmarks. In a general scenario, nobody Start and Shutdown would be called once so the current API would suffice for that.

src/storage/write_ahead_log/log_writer.cpp
tli2: I meant you should probably decouple the write and wakeup part of this function.

utkarsh39: There doesn't seem to be a straightforward way to do this. We designed this keeping in mind the writer thread upon waking up knows what action to take. If you could give a guideline, it would be really hellpful

GustavoAngulo:
I believe the original issue was fixed by updating the branch with master. The issue I was seeing on dev4 occurs when the logfile is not deleted after a test case. This can happen by calling CTRL-C or a segfault during the test. If this happened, then the next test iteration would read the logfile of the previous test and read garbage data. The fix is to just delete the log file if it exists during test setup.

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -1,23 +1,31 @@
...
+// TODO(Utkarsh): Get rid of magic constants
+#define MAX_BUF 2
+
 namespace terrier::storage {
 /**
  * A LogManager is responsible for serializing log records out and keeping track of whether changes from a transaction
  * are persistent.
  */
-class LogManager {
+class LogManager : public DedicatedThreadOwner {
```
tli2: The log manager probably does not own the `LogConsumerTask`. The ownership should be left for a more upper level component, maybe? Let's talk about this in a team meeting maybe.

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -68,25 +98,112 @@ class LogManager {
...
   // TODO(Tianyu): benchmark for if these should be concurrent data structures, and if we should apply the same
-  // optimization we applied to the GC queue.
+  //  optimization we applied to the GC queue.
   std::queue<RecordBufferSegment *> flush_queue_;
 
   // These do not need to be thread safe since the only thread adding or removing from it is the flushing thread
   std::vector<std::pair<transaction::callback_fn, void *>> commits_in_buffer_;
 
+  // System path for log file
+  const char *log_file_path_;
```
tli2: Probably needs to be lifted up to settings manager

src/network/connection_dispatcher_task.cpp
```diff
@@ -19,10 +19,10 @@ ConnectionDispatcherTask::ConnectionDispatcherTask(int num_handlers, int listen_
 
   // create worker threads.
   for (int task_id = 0; task_id < num_handlers; task_id++) {
-    auto handler = std::make_shared<ConnectionHandlerTask>(task_id, connection_handle_factory);
+    auto handler = new ConnectionHandlerTask(task_id, connection_handle_factory);
```
tli2: Who owns this pointer from now on then?

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -68,25 +98,112 @@ class LogManager {
...
+  std::vector<BufferedLogWriter> buffers_;
+  // This is the buffer the serializer thread will write to
+  BufferedLogWriter *filled_buffer_;
+  // The queue containing empty buffers which the serializer thread will use
+  common::ConcurrentBlockingQueue<BufferedLogWriter *> empty_buffer_queue_;
+  // The queue containing filled buffers pending flush to the disk
+  common::ConcurrentBlockingQueue<BufferedLogWriter *> filled_buffer_queue_;
+
+  // The log consumer task which flushes filled buffers to the disk
+  LogConsumerTask *log_consumer_task_ = nullptr;
```
tli2: Do you need a direct reference to the thread task, or just a queue (which can be processed on the back end by multiple threads?)

src/include/common/dedicated_thread_registry.h
```diff
@@ -52,29 +53,49 @@ class DedicatedThreadRegistry {
...
+    // Terminate task
+    task->Terminate();
+    threads_table_[task].join();
+
+    // Clear Metadata
+    threads_table_.erase(task);
+    thread_owners_table_[requester].erase(task);
+
+    // Notify requester of removal
+    requester->NotifyThreadRemoved(task);
```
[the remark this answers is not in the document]
GustavoAngulo: I envisioned `NotifyThreadRemoved` as letting the requester know it has lost the thread, to clean up any resources, and perhaps request a new thread. Perhaps we need something like `NotifyThreadOfPlannedRemoval` which tells the requester its about to lose the thread to ensure proper termination?

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -95,44 +95,54 @@ class BufferedLogWriter {
...
    * @param data memory location of the bytes to write
    * @param size number of bytes to write
+   * @return number of bytes written. This function only writes until the buffer gets full, so this can be used as the
+   * offset when calling this function again after flushing.
    */
-  void BufferWrite(const void *data, uint32_t size) {
-    if (!CanBuffer(size)) FlushBuffer();
-    // If we still do not have buffer space after flush, the write is too large to be buffered. We should bypass the
-    // buffer and write directly to disk
+  uint32_t BufferWrite(const void *data, uint32_t size) {
```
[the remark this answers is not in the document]
GustavoAngulo: This method is part of the BufferedLogWriter though, which owns only a single buffer, so it can't do much once it fulls up its buffer. It's up to the caller to find a new buffer to put the rest of the data in. What you're describing is [`LogManager::WriteValue`](https://github.com/cmu-db/terrier/pull/370/files#diff-70736a64a01e1c14098ca447fcbbb6d5R163), which is the only place `BufferWrite` is called.

src/include/storage/write_ahead_log/log_consumer.h
```diff
@@ -0,0 +1,50 @@
...
+ */
+class LogConsumerTask : public DedicatedThreadTask {
+ public:
+  /**
+   * Constructs a new LogConsumerTask
+   * @param log_manager pointer to the LogManager
+   */
+  explicit LogConsumerTask(LogManager *log_manager) : log_manager_(log_manager) {}
+
+  virtual ~LogConsumerTask() = default;
```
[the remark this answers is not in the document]
GustavoAngulo: I removed it, I was getting a compiler warning cause I was missing a virtual destructor on the base class (`DedicatedThreadTask`).

src/network/connection_dispatcher_task.cpp
GustavoAngulo: The `ConnectionDispatcherTask` owns the handler tasks, and is also responsible for [cleaning them up](https://github.com/cmu-db/terrier/pull/370/files/f5acda921b9dc1697e403291b819608cccc21acd#diff-cc4757df8e709f5364acdaf54c476669R40). I didn't write the code to handle when `ConnectionDispatcherTask` loses a thread from the thread registry, but the registry doesn't autonomously remove threads yet so thats okay for now.

src/include/storage/write_ahead_log/log_manager.h
GustavoAngulo: Currently we only support a single task, so having a pointer to the task works fine for now. Having multiple tasks would require more changes that are out of the scope of the PR.

src/include/storage/write_ahead_log/log_manager.h
GustavoAngulo: Done.

src/include/storage/write_ahead_log/log_consumer.h
```diff
@@ -0,0 +1,50 @@
...
+namespace terrier::storage {
+
+// Forward declaration for class LogManager
+class LogManager;
+
+/**
+ * A LogConsumerTask is responsible for writing serialized log records out to disk by processing buffers in the log
+ * manager's filled buffer queue
+ */
+class LogConsumerTask : public DedicatedThreadTask {
```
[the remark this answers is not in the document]
GustavoAngulo: Changed to `DiskLogWriterTask`, I wanted its purpose to be more explicit than `consume`. When I add a replication task, they can probably both inherit from a `LogConsumerTask`

src/include/common/dedicated_thread_registry.h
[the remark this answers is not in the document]
GustavoAngulo: Good catch, I added a latch to ThreadRegistry methods to protect tables.

src/include/common/dedicated_thread_owner.h
```diff
@@ -16,35 +18,65 @@ namespace terrier {
...
-   * TODO(tianyu) turn into async if need be
+   * that this function blocks until the thread can be dropped safely.
+   * @param task task that was removed from thread
+   * @return
    */
-  virtual void OnThreadRemoved(const std::shared_ptr<DedicatedThreadTask> &task) {}
+  virtual bool OnThreadRemoved(common::ManagedPointer<DedicatedThreadTask> task) { return false; }
 
  private:
+  std::mutex thread_count_latch_;
```
tli2: Maybe use a TBB one?

src/include/common/dedicated_thread_owner.h
```diff
@@ -16,35 +18,65 @@ namespace terrier {
...
+    return thread_count_;
+  }
+
+  /**
+   * Notifies the owner that a new thread can be given to it. The thread owner has the opportunity to decline the new
+   * thread if it does not need it
+   * @warning Should only be called by self driving infrastructure.
+   * @return true if owner accepts thread, false if it declines it
+   */
+  bool NotifyNewThread() { return OnThreadGranted(); }
```
tli2: It's unclear what the difference between this and `GrantNewThread` is for the caller. If all the caller should do is call them in sequence, we might as well merge the two methods.

src/include/common/dedicated_thread_owner.h
```diff
@@ -16,35 +18,65 @@ namespace terrier {
...
-    OnThreadRemoved(task);
   }
 
  protected:
+  /**
+   * Custom code to be run when offered a thread by each owner. The owner has the option to decline the new thread if it
+   * does not need it. If the owner accepts, its up to the owner to call RegisterDedicatedThread to register their task
+   * @return true if owner accepts thread, false if it declines it
+   */
+  virtual bool OnThreadGranted() { return false; }
```
tli2: Unless we merge the two public methods above, having this be separate from `NotifyNewThread` doesn't buy us much.

src/include/common/dedicated_thread_owner.h
```diff
@@ -16,35 +18,65 @@ namespace terrier {
...
+
   /**
    * Custom code to be run when removing a thread by each owner. It is expected
-   * that this function blocks until the thread can be dropped safely
-   *
-   * TODO(tianyu) turn into async if need be
+   * that this function blocks until the thread can be dropped safely.
+   * @param task task that was removed from thread
+   * @return
```
tli2: probably finish writing

src/include/common/dedicated_thread_owner.h
```diff
@@ -16,35 +18,65 @@ namespace terrier {
  */
 class DedicatedThreadOwner {
  public:
+  virtual ~DedicatedThreadOwner() = default;
```
tli2: I suspect we will need to write a lot more documentation detailing the sequence of how things work here. This module will probably need a design doc.

src/include/common/dedicated_thread_registry.h
```diff
@@ -52,29 +55,64 @@ class DedicatedThreadRegistry {
...
+      thread_owners_table_[requester].erase(task);
+    }
+    return true;
+  }
+
+  // TODO(tianyu, gus): Add code for thread removal from thread owner without specifying task
 
  private:
+  // Latch to protect internal tables
+  std::mutex table_latch_;
```
tli2: same as above

src/include/common/dedicated_thread_registry.h
```diff
@@ -52,29 +55,64 @@ class DedicatedThreadRegistry {
...
    * @param task The task to run in the dedicated thread
-   * @return the DedicatedThreadTask running on new thread
    */
-  template <typename Task>
-  void RegisterDedicatedThread(DedicatedThreadOwner *requester, std::shared_ptr<Task> task) {
-    thread_owners_table_[requester].push_back(task);
-    requester->NotifyNewThread();
-    TERRIER_ASSERT(threads_table_.find(task.get()) == threads_table_.end(), "Task is already registered");
-    threads_table_.emplace(task.get(), std::thread([=] { task->RunTask(); }));
+  void RegisterDedicatedThread(DedicatedThreadOwner *requester, common::ManagedPointer<DedicatedThreadTask> task) {
```
tli2: Who should own the task object?

src/include/common/notifiable_task.h
```diff
@@ -43,7 +43,7 @@ class NotifiableTask : public DedicatedThreadTask {
    * Destructs this NotifiableTask. All events currently registered to its base
    * are also deleted and freed.
    */
-  virtual ~NotifiableTask();
+  ~NotifiableTask() override;
```
tli2: unintentional change?

src/include/common/dedicated_thread_owner.h
```diff
@@ -16,35 +18,65 @@ namespace terrier {
...
+   * @warning Should only be called by self driving infrastructure.
+   * @return true if owner accepts thread, false if it declines it
+   */
+  bool NotifyNewThread() { return OnThreadGranted(); }
 
   /**
    * Notifies the owner that a new thread has been given to it
    */
-  void NotifyNewThread() { thread_count_++; }
+  void GrantNewThread() {
```
tli2: Should this pass the granted thread object? Or maybe the constructed task object?

src/include/common/dedicated_thread_owner.h
GustavoAngulo: Done, changed to use spin latch.

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -52,41 +105,267 @@ class LogManager {
...
   // TODO(Tianyu): benchmark for if these should be concurrent data structures, and if we should apply the same
-  // optimization we applied to the GC queue.
+  //  optimization we applied to the GC queue.
   std::queue<RecordBufferSegment *> flush_queue_;
 
   // These do not need to be thread safe since the only thread adding or removing from it is the flushing thread
   std::vector<std::pair<transaction::callback_fn, void *>> commits_in_buffer_;
 
+  // This stores all the buffers the serializer or the log consumer threads use
+  std::vector<BufferedLogWriter> buffers_;
```
tli2: nit:  can you also break the logic into the two tasks? It should be cleaner that way.

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -52,41 +105,267 @@ class LogManager {
...
+      std::this_thread::sleep_for(serialization_interval_);
+      log_manager_->Process();
+    }
   }
 };
+
+/**
+ * Task signals disk log consumer to flush logs into disk
+ */
+class LogFlusherTask : public DedicatedThreadTask {
```
tli2: nit: If you break up LogManager logic into these two classes, they should get their own files.

src/include/storage/write_ahead_log/log_manager.h
GustavoAngulo: Done.

src/include/storage/write_ahead_log/log_manager.h
GustavoAngulo: Done.

GustavoAngulo:
Benchmark numbers for `LoggingBenchmark` and `TPCCBenchmark` looked better when I switched the serializer task to moving the entire queue instead if popping elements from it.

src/storage/write_ahead_log/disk_log_consumer_task.cpp
```diff
@@ -0,0 +1,75 @@
...
+  // If the task hasn't run yet, sleep
+  // TODO(Gus): Get rid of magic number
+  while (!run_task_) std::this_thread::sleep_for(std::chrono::milliseconds(10));
+  TERRIER_ASSERT(run_task_, "Cant terminate a task that isnt running");
+  // Signal to terminate and force a flush so task persists before LogManager closes buffers
+  run_task_ = false;
+  log_manager_->disk_log_writer_thread_cv_.notify_one();
+}
+
+void DiskLogConsumerTask::FlushAllBuffers() {
```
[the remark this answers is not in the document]
GustavoAngulo: Done

src/storage/write_ahead_log/disk_log_consumer_task.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+void DiskLogConsumerTask::RunTask() {
+  run_task_ = true;
+  DiskLogConsumerTaskLoop();
+}
+
+void DiskLogConsumerTask::Terminate() {
+  // If the task hasn't run yet, sleep
+  // TODO(Gus): Get rid of magic number
+  while (!run_task_) std::this_thread::sleep_for(std::chrono::milliseconds(10));
```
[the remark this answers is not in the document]
GustavoAngulo: Done

src/include/storage/write_ahead_log/log_serializer_task.h
```diff
@@ -0,0 +1,130 @@
...
+   * @param buffer_segment the (perhaps partially) filled log buffer ready to be consumed
+   */
+  void AddBufferToFlushQueue(RecordBufferSegment *const buffer_segment) {
+    common::SpinLatch::ScopedSpinLatch guard(&flush_queue_latch_);
+    flush_queue_.push(buffer_segment);
+  }
+
+ private:
+  // LogManager that created task
+  LogManager *log_manager_;
```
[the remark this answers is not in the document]
GustavoAngulo: Done

PR #371, opened 2019-05-06
Adding tpl

amlatyrngom:
Adding folder.
Work is still in progress (rename namespace, add test files, pass CI checks, ...).

**TODO (FEATURES)**
- [x] Seq scan to stdout (tested to work for arbitrary columns picked out of a single table)
- [ ] Index scan to stdout
- [ ] @insert/update/delete(table, tuple) extension

**TODO (QUALITY)**
- [x] clang-tidy
- [x] cpplint
- [ ] DOXYGEN <-- this will be painful
- [ ] Final cleanup pass, after which we run format

**NICE TO HAVE**
- [ ] The interpreter should grab some hardcoded output_struct struct as its outputAlloc format
- [ ] Name conflict resolution probably will have bugs, e.g. if two tables have the same name. Aliasing issues in general.


lmwnshn:
(That was an intentional force push, keeping compiler_test compiling.)

lmwnshn:
Some questions for @GustavoAngulo and @wenxuanqiu , possibly for binder related group meeting in general @apavlo :

What will the binder do, for:

1. Insert-from-select in Peloton was modeled as a InsertPlanNode with a SelectPlanNode child, this is the only time an insert plan node had a child and so it could be special cased. Will we go this route again?
2. An InsertPlanNode should definitely be able to insert multiple values at once, but right now the mechanism for this is not clear.
3. UpdatePlanNode currently does not contain the values that we are updating it to be. It has the target of the update and that's it.

In general, the current status is we should be good for having all the (TPL to storage to some kind of callback) part; we think we understand this well and have dummy stuff in place. In particular, single inserts work; updates and deletes are basically copy-paste logic except with a different call to SQLTable wrapper. But the input to TPL doesn't seem stable yet.

wenxuanqiu:
> Some questions for @GustavoAngulo and @wenxuanqiu , possibly for binder/optimizer related group meeting in general @apavlo :
> 
> What will the binder/optimizer do, for:
> 
> 1. Insert-from-select in Peloton was modeled as a InsertPlanNode with a SelectPlanNode child, this is the only time an insert plan node had a child and so it could be special cased. Will we go this route again?
> 2. An InsertPlanNode should definitely be able to insert multiple values at once, but right now the mechanism for modeling the multiple values is not clear. Will we have a vec<vec>, should I treat vec as an array with (size / bulk_insert_count) elements?
> 3. UpdatePlanNode currently does not contain the values that we are updating it to be. It has the target of the update and that's it.
> 
> In general, the current status is we should be good for having all the (TPL to storage to some kind of callback) part; we think we understand this well and have dummy stuff in place. In particular, single inserts work; updates and deletes are basically copy-paste logic except with a different call to SQLTable wrapper. But the input to TPL doesn't seem stable yet.

I am encountering similar concerns when trying to create operators for INSERT, UPDATE, etc. I am not completely certain that whether the operators/plan_nodes should have ownership of values to be inserted or updated. I think these issues arise from building upstream and downstream objects from a basic version of plan nodes. I feel that we need to solve them at a higher level by figuring out where the values for INSERT and UPDATE need to be from the parser to the execution engine. I think this is somewhat related to the Context objects we briefly touched on during the meetings. Otherwise, we might have too many moving parts added over time that will bite back at us later.

sample_tpl/join.tpl
```diff
@@ -0,0 +1,55 @@
...
+fun main() -> int32 {
+  var state: State
+
+  // Initialize state
+  initState(&state)
+
+  // Run pipeline 1
+  pipeline_1(&state)
+
+  // Build table
```
apavlo: Is this function really building the hash table? If yes, then what is is `pipeline_1` doing?

mbutrovich:
There will be a new PR for this after #398 merges, since @amlatyrngom has a different branch that has been kept more up to date.

PR #373, opened 2019-05-11
clang-tidy heartbeat

lmwnshn:
clang-tidy can be so slow that Travis kills us while waiting for our progress bar to update.
This just prints stuff to stdout every 300 seconds (half Travis timeout) so that we don't get killed.

### DIFF
--- a/build-support/run-clang-tidy.py
+++ b/build-support/run-clang-tidy.py
@@ -311,9 +311,15 @@ def main():
 
         # Fill the queue with files.
         for i, name in enumerate(files):
-            if file_name_re.search(name):
-                task_queue.put(name)
-            update_progress(i, len(files))
+            put_file = False
+            while not put_file:
+                try:
+                    if file_name_re.search(name):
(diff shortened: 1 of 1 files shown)


jrolli:
This CI failure for this PR is fixed by #372.

codeworm96:
We are still experiencing this problem in our PR(#341).
The problems are:
1. This heartbeat message is not forced to be flushed immediately. So it may retain in the buffer.
2. `update_progress()` function does not print a update for each invocation. It is possible that we are continuously putting some files into the task_queue at such a speed that 1) it does not trigger the timeout and 2) makes `update_progress()` function not print a update for a period longer than Travis timeout.

Currently, we are mitigating this problem with https://github.com/cmu-db/terrier/pull/341/commits/94154c1d78b3bde76a03c865d0fff39dcd5dbadb.

mbutrovich:
@codeworm96 Assuming your changes resolve the issue, can you please cherry pick those changes to a separate PR?

codeworm96:
@mbutrovich Sure.

PR #377, opened 2019-05-14
Transformation to Arrow

tli2:
This PR adds a compaction component that transforms cold blocks into Arrow storage.

TODO: write tests for the AccessObserver. This is a little bit complicated because I have to mock things.

TODO: Use executors instead of calling data table primitives for moving tuples in the compaction phase. This will also take care of index updates for us.


tli2:
No dependent on #447

src/include/storage/block_access_controller.h
```diff
@@ -19,8 +45,92 @@ namespace terrier::storage {
...
+        case BlockState::COOLING:
+          if (!GetBlockState()->compare_exchange_strong(current_state, BlockState::HOT)) continue;
+          // wait until the compactor finishes before doing anything
+          // intentional fall through
+        case BlockState::FROZEN:
+          GetBlockState()->store(BlockState::HOT);
+          // intentional fall through
+        case BlockState::HOT:
+          // Although the block is already hot, we may need to wait for any straggling readers to finish
+          while (GetReaderCount()->load() != 0) __asm__ __volatile__("pause;");
```
mbutrovich: We've got _mm_pause() in the codebase elsewhere. Are they the same?

src/include/storage/block_compactor.h
```diff
@@ -0,0 +1,99 @@
...
+#include "storage/storage_defs.h"
+#include "transaction/transaction_manager.h"
+namespace terrier::storage {
+
+/**
+ * Typedef for a standard hash map with varlen entry as the key. The map uses deep equality checks (whether
+ * the stored underlying byte string is the same) for key comparison.
+ */
+template <class T>
+using VarlenEntryMap = std::unordered_map<VarlenEntry, T, VarlenContentHasher, VarlenContentDeepEqual>;
```
mbutrovich: Move to storage_defs?

src/include/storage/block_compactor.h
```diff
@@ -0,0 +1,99 @@
...
+/**
+ * The block compactor is responsible for taking hot data blocks that are considered to be cold, and make them
+ * arrow-compatible. In the process, any gaps resulting from deletes or aborted transactions are also eliminated.
+ * If the compaction is successful, the block is considered to be fully cold and will be accessed mostly as read-only
+ * data.
+ */
+class BlockCompactor {
+ private:
+  // A Compaction group is a series of blocks all belonging to the same data table. We compact them together
+  // so slots can be freed up. If we only eliminate gaps, deleted slots will never be reclaimed.
```
mbutrovich: Can you clarify the distinction between "compacting them" and "eliminat[ing] gaps"? Compacting means across blocks while eliminating gaps is just within blocks in this terminology?

src/include/storage/block_compactor.h
```diff
@@ -0,0 +1,99 @@
...
+    CompactionGroup(transaction::TransactionContext *txn, DataTable *table)
+        : txn_(txn),
+          table_(table),
+          all_cols_initializer_(ProjectedRowInitializer::Create(table_->accessor_.GetBlockLayout(),
+                                                                table_->accessor_.GetBlockLayout().AllColumns())),
+          read_buffer_(all_cols_initializer_.InitializeRow(
+              common::AllocationUtil::AllocateAligned(all_cols_initializer_.ProjectedRowSize()))) {}
+
+    ~CompactionGroup() {
+      // Deleting nullptr is just a noop
```
mbutrovich: I think we tested it and it will cause an ASAN warning though. IIRC we can't really win in this codebase. tidy says nullptr check is unnecessary before delete, while ASAN will fail if you delete nullptr.

src/include/storage/block_compactor.h
```diff
@@ -0,0 +1,99 @@
...
+
+  /**
+   * Processes the compaction queue and mark processed blocks as cold if successful. The compaction can fail due
+   * to live versions or contention. There will be a brief window where user transactions writing to the block
+   * can be aborted, but no readers would be blocked.
+   *
+   */
+  void ProcessCompactionQueue(transaction::TransactionManager *txn_manager);
+
+  // TODO(Tianyu): Should a block know about its own data table? We seem to need this back pointer awfully often.
```
mbutrovich: RawBlocks now have this pointer.

src/include/storage/storage_util.h
```diff
@@ -137,5 +137,7 @@ class StorageUtil {
    * @return vector of column ids
    */
   static std::vector<storage::col_id_t> ProjectionListAllColumns(const storage::BlockLayout &layout);
+
+  static void DeallocateVarlens(RawBlock *block, const TupleAccessStrategy &accessor);
```
mbutrovich: How is this not failing doxygen?

src/include/storage/tuple_access_strategy.h
```diff
@@ -60,31 +60,28 @@ class TupleAccessStrategy {
     ArrowBlockMetadata &GetArrowBlockMetadata() { return *reinterpret_cast<ArrowBlockMetadata *>(block_.content_); }
 
     // return reference to attr_offsets. Use as an array.
-    uint32_t *AttrOffets(const BlockLayout &layout) {
+    uint32_t *AttrOffsets(const BlockLayout &layout) {
```
mbutrovich: Nice.

src/storage/access_observer.cpp
```diff
@@ -0,0 +1,21 @@
+#include "storage/access_observer.h"
+#include "storage/block_compactor.h"
+
+namespace terrier::storage {
+void AccessObserver::ObserveGCInvocation() {
+  gc_epoch_++;
+  for (auto it = last_touched_.begin(), end = last_touched_.end(); it != end;) {
+    if (it->second + COLD_DATA_EPOCH_THRESHOLD < gc_epoch_) {
+      compactor_->PutInQueue(it->first);
+      it = last_touched_.erase(it);
```
mbutrovich: Should we be asserting that erase succeeded?

src/storage/access_observer.cpp
```diff
@@ -0,0 +1,21 @@
...
+    } else {
+      ++it;
+    }
+  }
+}
+
+void AccessObserver::ObserveWrite(RawBlock *block) {
+  if (block->insert_head_ == block->data_table_->GetBlockLayout().NumSlots()) last_touched_[block] = gc_epoch_;
```
mbutrovich: This conditional's purpose is non-obvious. You're checking that the block is full first to even be eligible to be tracked by the AccessObserver?

src/storage/block_compactor.cpp
```diff
@@ -0,0 +1,360 @@
...
+#include <unordered_map>
+#include <utility>
+#include <vector>
+#include "storage/index/bwtree_index.h"
+#include "storage/index/index_defs.h"
+#include "storage/sql_table.h"
+namespace terrier::storage {
+namespace {
+// for empty callback
+void NoOp(void * /* unused */) {}
```
mbutrovich: We've got an empty callback defined in transaction_util. Can we do something similar in storage_util?

src/storage/block_compactor.cpp
```diff
@@ -0,0 +1,360 @@
...
+}  // namespace
+
+void BlockCompactor::ProcessCompactionQueue(transaction::TransactionManager *txn_manager) {
+  std::forward_list<RawBlock *> to_process = std::move(compaction_queue_);
```
mbutrovich: Oddly enough, this is O(n). Not sure why that would be the case on a linked list.

https://en.cppreference.com/w/cpp/container/forward_list/operator%3D

src/storage/block_compactor.cpp
```diff
@@ -0,0 +1,360 @@
...
+    if (!column_bitmap->Test(i)) continue;
+
+    // Only do a gather operation if the column is varlen
+    VarlenEntry &entry = values[i];
+    std::memcpy(new_col.Values() + acc, entry.Content(), entry.Size());
+
+    // Need to GC
+    if (entry.NeedReclaim()) loose_ptrs->push_back(entry.Content());
+
+    // TODO(Tianyu): Describe why this is still safe
```
mbutrovich: Why is this still safe? :)

src/storage/block_compactor.cpp
```diff
@@ -0,0 +1,360 @@
...
+    auto ret = dictionary.emplace(values[i], 0);
+    // If the string has not been seen before, should add it to dictionary when counting total length.
+    if (ret.second) varlen_size += values[i].Size();
+  }
+  ArrowColumnInfo new_col_info;
+  new_col_info.Type() = col->Type();
+  new_col_info.Indices() = common::AllocationUtil::AllocateAligned<uint32_t>(metadata->NumRecords());
+  auto &new_col = new_col_info.VarlenColumn() = {varlen_size, static_cast<uint32_t>(dictionary.size() + 1)};
+
+  // TODO(Tianyu): This is retarded, but apparently you cannot retrieve the index of elements in your
```
mbutrovich: Is there a proposed long-term solution?

src/storage/garbage_collector.cpp
```diff
@@ -78,7 +79,10 @@ uint32_t GarbageCollector::ProcessUnlinkQueue() {
   while (!txns_to_unlink_.empty()) {
     txn = txns_to_unlink_.front();
     txns_to_unlink_.pop_front();
-    if (txn->undo_buffer_.Empty()) {
+
+    if (txn->IsReadOnly()) {
+      // TODO(Tianyu): Deal with this edge case
```
mbutrovich: TODO in this PR?

mbutrovich:
We should plan a meeting to discuss future work on augmenting storage access to use `TryAcquireInPlaceRead()`. This will have to be after this and the execution engine are in, but presumably would yield significant performance gains on the right workloads.

src/include/storage/block_compactor.h
tli2: This alias is internal to the block compactor. We can move when someone else finds a need for it.

src/storage/access_observer.cpp
tli2: No, because this is calling the erase method on a valid iterator handed out by the unordered_map, which will always succeed. In fact I don't think this version of erase even  bothers with error behavior.

src/storage/block_compactor.cpp
tli2: I can just use that one. I was under the impression we only had one in tests.

src/storage/block_compactor.cpp
tli2: I'll just use a queue then?

src/storage/garbage_collector.cpp
tli2: This is stale as I fixed it using the deferred event framework.

PR #381, opened 2019-05-15
Catalog accessor

jrolli:
This is the reduced interface for the catalog that should make it easier for other parts of the system to use the catalog without becoming tightly coupled to the internal implementation of the catalog (i.e. all of the Postgres-isms).

The intent is that reading and including `catalog/catalog_accessor.h` is sufficient for people learning the system to interface with the catalog enough to support their future projects.

**What is complete:**  The API should be mostly stable.  People who need to interface with the catalog (to include touching a `Schema` object in any way) should review `catalog_accessor.h` and make sure that the functions there support everything they need to do in the catalog.  If not, they should start a review and request a change detailing exactly what they need and why they need it.

**Work that still needs to be completed:**
- [x] Remove need for `namespace_oid_t` when we have an OID to something in `pg_class` table
- [ ] Improve fine-grained access into `pg_attribute` for supporting column addition, deletions, and modifications
- [x] Add support for storing indexes
- [ ] Fill in stubbed out implementations
- [ ] Use indexes rather than table scans


jrolli:
The current state of this PR should make the overall catalog API stable with the exception of how to pass and store expressions inside of default values and index key schemas.  Additionally, none of the operations that modify the catalog are thread safe, but concurrent lookups (without modifications) should be thread safe.

This should be able to support end-to-end testing of static schemas like TPC-C, while we write tests for the catalog's API and then refactor the backend.

jrolli:
While trying to switch the catalog test over to GC and the accessor, several memory leaks deep inside the implementation appeared.  These appear to be caused by a call to `BeginTransaction` inside the catalog logic that is not paired with a call to `Abort` or `Commit` and instead manually delete `TransactionContext`.  This effectively disables GC by causing the transaction manager to have a hung transaction.

src/catalog/catalog.cpp
```diff
@@ -24,19 +25,29 @@ Catalog::Catalog(transaction::TransactionManager *txn_manager, transaction::Tran
...
   // get database handle
   auto db_handle = GetDatabaseHandle();
   auto db_entry = db_handle.GetDatabaseEntry(txn, db_name);
+  if (db_entry == nullptr) return false;
+
   auto oid = db_entry->GetOid();
   // remove entry from pg_database
-  db_handle.DeleteEntry(txn, db_entry);
+  bool ret = db_handle.DeleteEntry(txn, db_entry);
+  if (!ret) return false;
```
apavlo: I feel like this the lack of an error message may cause confusion when this returns `false`. Can you at least add some debug messages?

src/catalog/catalog.cpp
```diff
@@ -45,6 +56,27 @@ void Catalog::DeleteDatabase(transaction::TransactionContext *txn, const std::st
...
+  if (db_entry == nullptr) return false;
+
+  // remove entry from pg_database
+  bool ret = db_handle.DeleteEntry(txn, db_entry);
+  if (!ret) return false;
+
+  // TODO(pakhtar): delete all user tables
+  // destroy all the non-global tables
+  // this should become just catalog tables
+  DeleteDatabaseTables(db);
```
apavlo: Once we have foreign keys, this will fail. You will need to delete the tables first, then delete the database.

src/catalog/catalog.cpp
```diff
@@ -45,6 +56,27 @@ void Catalog::DeleteDatabase(transaction::TransactionContext *txn, const std::st
...
+
+  // delete from the map
+  cat_map_.erase(db);
+  return true;
 }
 
 namespace_oid_t Catalog::CreateNameSpace(transaction::TransactionContext *txn, db_oid_t db_oid,
```
apavlo: `CreateNameSpace` -> `CreateNamespace`

src/catalog/catalog.cpp
```diff
@@ -78,26 +118,28 @@ table_oid_t Catalog::CreateUserTable(transaction::TransactionContext *txn, db_oi
   auto table_handle = ns_handle.GetTableHandle(txn, ns_oid);
 
   // creates the storage table and adds to pg_class
-  auto tbl_rw = table_handle.CreateTable(txn, schema, table_name);
+  auto tbl_rw = table_handle.CreateUserTable(txn, schema, table_name);
+
+  // Modification to support indexes
+  table_index_map_.emplace(tbl_rw->Oid(), std::unordered_set<index_oid_t>());
```
apavlo: I don't know how this map is being used (or why we need it), but I feel like it's going to cause problems later on because it's not transactional.

src/catalog/catalog.cpp
```diff
@@ -173,10 +219,9 @@ void Catalog::Bootstrap(transaction::TransactionContext *txn) {
 }
 
 // TODO(pakhtar): resolve second arg.
-void Catalog::AddColumnsToPGAttribute(transaction::TransactionContext *txn, db_oid_t db_oid,
-                                      const std::shared_ptr<storage::SqlTable> &table) {
-  Schema schema = table->GetSchema();
-  std::vector<Schema::Column> cols = schema.GetColumns();
+void Catalog::AddColumnsToPGAttribute(transaction::TransactionContext *txn, db_oid_t db_oid, SqlTableHelper *table) {
```
apavlo: I would not call this `PgAttribute`. We shouldn't leak the Postgres naming convention for tables into our API.

src/catalog/catalog_accessor.cpp
```diff
@@ -0,0 +1,316 @@
...
+  for (auto oid : columns) {
+    // TODO(John): Implement the metadata deletion by deleting the row in pg_attribute
+    col_successes.emplace_back(oid != INVALID_COLUMN_OID);
+  }
+
+  TERRIER_ASSERT(true, "This function is not implemented yet");
+  return col_successes;
+}
+
+bool CatalogAccessor::SetColumnNullable(table_oid_t table, col_oid_t column, bool nullable) {
```
apavlo: Instead of having separate methods for changing the column in different ways (e.g., Nullable, Default, ColumnType), would it better to have a single function where you give a new `ColumnDefinition` for an existing column and then the system figures out what changed?

src/catalog/namespace_handle.cpp
```diff
@@ -46,7 +46,8 @@ std::shared_ptr<NamespaceCatalogEntry> NamespaceCatalogTable::GetNamespaceEntry(
 namespace_oid_t NamespaceCatalogTable::NameToOid(transaction::TransactionContext *txn, const std::string &name) {
   auto nse = GetNamespaceEntry(txn, name);
   if (nse == nullptr) {
-    throw CATALOG_EXCEPTION("namespace does not exist");
+    // no such namespace
+    return namespace_oid_t(NULL_OID);
```
apavlo: Why use `NULL_OID`? Should it be `INVALID_NAMESPACE_OID`?

src/include/catalog/catalog_accessor.h
```diff
@@ -0,0 +1,476 @@
...
+ */
+class CatalogAccessor {
+ public:
+  /**
+   * This class decouples the definition of a column from its existence in a
+   * schema.  This allows us to delegate all OID assignments to the catalog
+   * (rather than requiring the caller to reason about them) and allows us to
+   * treat Schema and Column objects as immutable outside of the catalog.
+   */
+  class ColumnDefinition {
```
apavlo: Should we move `ColumnDefinition` and `IndexKeyDefinition` into standalone classes? We may want to reuse these in other parts of the system (e.g., CreatePlanNodes).

src/include/catalog/catalog_accessor.h
```diff
@@ -0,0 +1,476 @@
...
+   * @return OID of the index, INVALID_INDEX_OID if the index was not found
+   */
+  index_oid_t GetIndexOid(namespace_oid_t ns, const std::string &name);
+
+  /**
+   * Given a table, find all indexes for data in that table
+   * @param table OID being queried
+   * @return vector of index OIDs that reference the queried table
+   *
+   * TODO(John): Should this return the actual index pointers as well?
```
apavlo: No, just have it return `index_oids`

src/catalog/type_handle.cpp
```diff
@@ -48,7 +48,8 @@ TypeCatalogTable::TypeCatalogTable(Catalog *catalog, SqlTableHelper *pg_type)
 type_oid_t TypeCatalogTable::TypeToOid(transaction::TransactionContext *txn, const std::string &type) {
   auto te = GetTypeEntry(txn, type);
   if (te == nullptr) {
-    throw CATALOG_EXCEPTION("no such type");
+    // type does not exist
+    return type_oid_t(NULL_OID);
```
apavlo: `INVALID_TYPE_OID`

src/include/catalog/catalog_sql_table.h
```diff
@@ -27,7 +27,7 @@ class SqlTableHelper {
    * Constructor
    * @param table_oid the table oid of the underlying sql table
    */
-  explicit SqlTableHelper(catalog::table_oid_t table_oid) : table_oid_(table_oid) {}
```
apavlo: Students are going to misuse this class. We should rename it to `CatalogSqlTableHelper`

src/include/catalog/default_value.h
```diff
@@ -0,0 +1,38 @@
...
+#include <utility>
+
+#include "type/type_id.h"
+
+namespace terrier::catalog {
+/**
+ * Wraps the concept of a default value expression into an object that can be
+ * passed between the catalog and the consumers of these expressions.
+ */
+class DefaultValue {
```
apavlo: Is this not the same type of class as `CatalogAccessor::ColumnDefinition` (i.e., it's information that can be decoupled from the `Schema`)?

If yes, then we should rename this class to `DefaultValueDefinition` and maybe store it **src/include/catalog/definitions/** with `ColumnDefinition` and `IndexKeyDefinition`.

src/include/catalog/default_value.h
```diff
@@ -0,0 +1,38 @@
...
+#include <string>
+#include <utility>
+
+#include "type/type_id.h"
+
+namespace terrier::catalog {
+/**
+ * Wraps the concept of a default value expression into an object that can be
+ * passed between the catalog and the consumers of these expressions.
+ */
```
apavlo: I would be explicit in the comments that this stores a serialized `AbstractExpression`. 
To explain what a serialized expression is, I would maybe add

```
@see src/include/parser/expression/abstract_expression.h
```

src/include/catalog/index_key_schema.h
```diff
@@ -0,0 +1,95 @@
...
+#include <vector>
+#include "catalog/catalog_defs.h"
+#include "type/type_id.h"
+
+namespace terrier::catalog {
+
+/**
+ * A column of the index key has an identifier, type, and describes whether it can be null.
+ */
+class IndexKeyColumn {
```
apavlo: How is this different than `IndexKeyDefinition`?

jrolli: It doesn’t take an OID in its constructor which is the same separation as the schema columns.  This allows the catalog to be the single point for assigning OIDs.

src/catalog/catalog.cpp
jrolli: I agree, my plan for Postgres-isms is to cleanly isolate them into separate files and possibly a sub-directory to make it very clear where the boundary between internal and Postgres logic is.

src/include/catalog/catalog_accessor.h
jrolli: That probably makes sense.  Should they still live in the CatalogAccessor file or put them as standalone classes in their respective “schema” header files?

src/include/catalog/catalog_sql_table.h
jrolli: I plan on completely eliminating this class in the refactor.  I think it is obfuscating large portions of the catalog code rather than making it easier to read and reason about.

src/include/catalog/default_value.h
jrolli: I’ll admit that I need some mental help on what the right internal representation for default values are, but this is different from the other definition classes because the default values are independent of the catalog (whereas the other’s get defined by user code and assigned OIDs when submitted).

Maybe this should live under a different component?

mbutrovich:
Closed in lieu of #392.

PR #389, opened 2019-05-22
Fix bulk insert for InsertPlanNode

GustavoAngulo:
This PR changes how bulk inserts are represented in `InsertPlanNode`. Particularly, it makes the following changes:

* Change to `values_`
   * Values is now a vector of vectors. This represents a list of "tuples" to insert, where each tuple is a vector of the values for that tuple.
* Change to `parameter_info_`
  * Maps the index of a value to which column that value should be inserted into
  * This could have been a `std::vector`, but I wanted the mapping to be explicit
  * The `paramater_info_` allows for default values to be added during binding (in which case they would be in the `values_` vector and in `parameter_info`), or during execution (In which case there would be no entry for that column in `parameter_info`).
* Bulk insert count
   * This is now simply the size of `values_`

### DIFF
--- a/src/include/planner/plannodes/insert_plan_node.h
+++ b/src/include/planner/plannodes/insert_plan_node.h
@@ -3,6 +3,7 @@
 #include <memory>
 #include <string>
 #include <tuple>
+#include <unordered_map>
 #include <utility>
 #include <vector>
 #include "catalog/schema.h"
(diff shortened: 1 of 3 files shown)


src/include/planner/plannodes/insert_plan_node.h
```diff
@@ -260,19 +169,26 @@ class InsertPlanNode : public AbstractPlanNode {
...
+  const std::unordered_map<uint32_t, catalog::col_oid_t> &GetParameterInfo() const { return parameter_info_; }
+
+  /**
+   * @param value_idx index of value being inserted
+   * @return OID of column where value should be inserted
+   */
+  const catalog::col_oid_t GetColumnOidForValue(uint32_t value_idx) const { return parameter_info_.at(value_idx); }
 
   /**
    * @return number of times to insert
```
lmwnshn: number of values to insert, maybe?

src/include/planner/plannodes/insert_plan_node.h
```diff
@@ -301,22 +217,19 @@ class InsertPlanNode : public AbstractPlanNode {
...
-   * parameter information
+   * vector of values to insert. Multiple vector of values corresponds to a bulk insert. Values for each tuple are
+   * ordered the same across tuples. Parameter info provides column mapping of values
    */
-  std::vector<ParameterInfo> parameter_info_;
+  std::vector<std::vector<type::TransientValue>> values_;
 
   /**
-   * name of time to insert
+   * parameter information. Provides which column a value should be inserted into. For example, for a tuple t at
```
lmwnshn: What are best practices for documenting a member and also its accessor? It feels a bit redundant to have this here again.

src/include/planner/plannodes/insert_plan_node.h
GustavoAngulo: Updated comment.

GustavoAngulo:
> I do prefer the single vec design, but can work with this.

I added a TODO for now as a future optimization. I think when we bring in the optimizer and binder which will generate the plan node, we should circle back around and see if its still clean to flatten the vector.

PR #392, opened 2019-05-24
Catalog refactor

### LINKED ISSUE #311: catalog: thread safety
The one know issue, is that the maps in the catalog are not protected by locks. Discussed simply putting a lock around map access. The alternative of using a concurrent map was discussed.

To verify that we have no other concurrency bugs or design issues, a multi-threaded stress test for the catalog should be used. Deletion of catalog entries is only partially implemented. Until then, the test should only insert and verify.

### LINKED ISSUE #313: catalog: CreateTable
TableHandle::CreateTable doesn't do everything that we need.

The above, creates a storage table and inserts a row into pg_class. (It returns a raw pointer, which is probably not ok).

We need, at a higher level, for creating tables to:
- from input of schema and table name, to create a storage table
- insert the name and ptr into pg_class
- insert the schema (i.e. column information) into pg_attribute, with the correct references to the pg_type for the column types
 - insert any column default values into pg_attrdef

### LINKED ISSUE #323: catalog: storage pointers and shutdown
At the moment, the catalog uses shared ptrs to storage tables and stores are raw pointer to the storage table in pg_class.

We need to:
1. Switch to using raw pointers to storage tables.
2. Implement catalog shutdown, to safely shutdown.  The catalog does not know what transactions are active, nor does it know if the garbage collector is still accessing tables. Shutdown of the catalog should:
i) Wait for completion of all transactions
ii) Wait for the garbage collector to finish
Then, the catalog can delete both system and user tables safely.

### LINKED ISSUE #347: catalog: use of maps
See also issues #311 and #323. 

At time of writing, the catalog code uses two maps to facilitate lookup of information. All the information is in the catalog tables, but lookup in the catalog is more cumbersome and probably slower. The maps are a convenient cache for some operations.

Problems:

- The maps do not consider thread safety. They probably need locking added to ensure thread safety (or use some form of lock free structure).
- After recovery, the maps content would need to be reconstructed. The catalog doesn't support recovery yet.
- Lin raises a question whether the maps will work with multi-version schemas? I don't know enough about where the multi-version schema design stands. My understanding is that the schema versioning will be done under the sql table layer. The maps hold no schema information, but do hold information about table existence. Is this a problem? (Seems like it could be).

It is possible to eliminate the maps and use the catalog tables themselves.

### LINKED ISSUE #380: Catalog Limits Tables to 100 MiB
Currently the catalog handles allocation and initialization of the storage backend for SQL tables (i.e. `storage::SqlTable` instances).  It allocates these instances through its own wrapper that provides it a hard-coded block store with a max size of 100 (i.e. it can allocate up to 100x 1 MiB blocks).

The long-term solution for this is to separate "logical" creation of the table in the catalog from the "physical" creation of the storage object.

jrolli:
This refactors the catalog to give a simpler, non-PostgreSQL internal API to the rest of the system.  It will still store data in PostgreSQL's catalog layout for compatibility with admin tools that scan the catalog, but it makes it simpler for other parts of the system to read/write metadata in the catalog by not requiring them to learn the internal catalog schema.

The following items need to be finished before merging this PR:
- [x] Finish header files for `pg_attribute`, `pg_constraints`, and `pg_index`.
- [x] Finalize API for passing expressions (pointer to AbstractExpression vs serialized representation)
- [ ] (not required for TPC-C) Finalize API for altering tables (add/drop columns, column type/default/nullable changes, and constraints).
- [x] Implement the core API (everything except altering tables)
- [ ] (not required for TPC-C) Implement alter table API

Supersedes #381.
Fixes #380 
Fixes #347 
Fixes #323 
Fixes #313 
Fixes #311


jrolli:
Still need to implement initialization of `pg_namespace` and `pg_type` in the bootstrap routine (minimal functionality for user tables) and start implementing interface.  Entry objects probably need a rework due to SqlTable's API change.

jrolli:
Partially blocked on #433 because we will need 'ColumnValueExpression' objects to build the indexes.  However, we could dummy populate them until the fix is in.

@pavlo, should we wait to merge until after the expressions are added, or should I proceed with a throwaway key definition (probably null) and have the real definitions commented out with a TODO left to explain?

jrolli:
@apavlo, this should be ready for review.

jrolli:
Thoughts for review:
  @amlatyrngom should review the overall API as comparison and guidance for adapting to this API from his fake implementation.
  @mush-zhang should review the tests as an example of usage and possible identification of additional cases
  @mbutrovich should review the refactor of TPC-C and parts of storage layer that happened from moving IndexKeySchema.

mbutrovich:
This PR also introduces a flag in TransactionContext that indicates that a transaction _cannot_ Commit. It's currently tripped by a write-write conflict in the SqlTable or a unique-key conflict in the index wrapper. It cannot be set back, and the TransactionManager will not allow a transaction to commit with this flag set.

We didn't spin it off to a separate PR because there were other catalog-related changes in some of the files and undoing those so as to cherry-pick this feature seemed unnecessary with the catalog close to being ready for review.

apavlo:
@mbutrovich How is the "cannot commit" flag different than an abort? Who is responsible for checking that this flag is set before we go off and do more work (e.g., continue to execute subsequent queries in a batch)? What if somebody sets this flag and then never calls abort?

src/include/storage/index/index.h
```diff
@@ -58,12 +57,11 @@ class Index {
 
   /**
    * Creates a new index wrapper.
-   * @param oid identifier for the index
    * @param constraint_type type of index
    * @param metadata index description
    */
-  Index(const catalog::index_oid_t oid, const ConstraintType constraint_type, IndexMetadata metadata)
```
apavlo: Why did we remove the `index_oid_t`? The `SQLTable` has its own `table_oid_t`?

mbutrovich:
> How is the "cannot commit" flag different than an abort? 
Separation of responsibilities. The storage layer can trip a flag that this txn can no longer be allowed to commit, however it's the execution layer's job to actually abort in case there are operations it needs to handle (cleanup, etc.). The storage layer can't call abort directly.

> Who is responsible for checking that this flag is set before we go off and do more work (e.g., continue to execute subsequent queries in a batch)?

Currently no one. I proposed having the TransactionContext not allow to stage any more writes based on this flag (since in theory a txn could keep shipping out logs even though it has no possibility of committing now) but @jrolli suggested it wasn't needed at this time. Conversation is still open on this flag right now, it just prevents committing.

> What if somebody sets this flag and then never calls abort?
It would be no different any a transaction calling begin and never calling either abort or commit. It would just be hanging out in memory.

src/include/storage/index/index.h
jrolli: Objects should not need to reason about their own OID value as they should only be used to resolve logical "addresses" to physical ones.  `SqlTable` had the OID in its constructor removed as well.

src/catalog/database_catalog.cpp
```diff
@@ -0,0 +1,1643 @@
...
+  const auto oid_pm = columns_oid_index_->GetKeyOidToOffsetMap();
+  pr = oid_pri.InitializeRow(buffer);
+  // Write the attributes in the ProjectedRow. These hardcoded indexkeycol_oids come from
+  // Builder::GetColumnOidIndexSchema()
+  *(reinterpret_cast<uint32_t *>(pr->AccessForceNotNull(oid_pm.at(indexkeycol_oid_t(2))))) = col.GetOid();
+  *(reinterpret_cast<uint32_t *>(pr->AccessForceNotNull(oid_pm.at(indexkeycol_oid_t(1))))) = class_oid;
+
+  if (!columns_oid_index_->InsertUnique(txn, *pr, tupleslot)) {
+    // There was an oid conflict and we need to abort.  Free the buffer and return false to indicate failure
+    // TODO(Matt): Is this is possible at this point, or should we be asserting the insert result like the oid index on
```
mbutrovich: See TODO.

src/catalog/database_catalog.cpp
```diff
@@ -0,0 +1,1643 @@
...
+
+  // Step 4: Insert into class index
+  const auto class_pri = columns_class_index_->GetProjectedRowInitializer();
+  pr = class_pri.InitializeRow(buffer);
+  // Write the attributes in the ProjectedRow
+  *(reinterpret_cast<uint32_t *>(pr->AccessForceNotNull(0))) = class_oid;
+
+  if (!columns_class_index_->Insert(txn, *pr, tupleslot)) {
+    // There was an oid conflict and we need to abort.  Free the buffer and return false to indicate failure
+    // TODO(Matt): Is this is possible at this point, or should we be asserting the insert result like the oid index on
```
mbutrovich: See TODO.

src/catalog/database_catalog.cpp
```diff
@@ -0,0 +1,1643 @@
...
+  // Step 2: Scan the table to get the columns
+  std::vector<std::unique_ptr<Column>> cols;
+  pr = table_pri.InitializeRow(buffer);
+  for (const auto &slot : index_results) {
+    const auto UNUSED_ATTRIBUTE result = columns_->Select(txn, slot, pr);
+    TERRIER_ASSERT(result, "Index scan did a visibility check, so Select shouldn't fail at this point.");
+    cols.emplace_back(postgres::AttributeHelper::MakeColumn<Column>(pr, table_pm));
+  }
+
+  // TODO(Matt): do we have any way to assert that we got the number of attributes we expect? From another attribute in
```
mbutrovich: See TODO.

src/catalog/database_catalog.cpp
```diff
@@ -0,0 +1,1643 @@
...
+  // another catalog table maybe?
+
+  // Finish
+  delete[] buffer;
+  return cols;
+}
+
+// TODO(Matt): do we need a DeleteAttribute()?
```
mbutrovich: See TODO.

src/catalog/database_catalog.cpp
```diff
@@ -0,0 +1,1643 @@
...
+  const auto class_pri = columns_class_index_->GetProjectedRowInitializer();
+  // Buffer is large enough to hold all prs
+  byte *const buffer = common::AllocationUtil::AllocateAligned(table_pri.ProjectedRowSize());
+  // Scan the class index
+  auto *pr = class_pri.InitializeRow(buffer);
+  *(reinterpret_cast<uint32_t *>(pr->AccessForceNotNull(0))) = class_oid;
+  std::vector<storage::TupleSlot> index_results;
+  columns_class_index_->ScanKey(*txn, *pr, &index_results);
+  if (index_results.empty()) {
+    // TODO(Matt): do we need to indicate failure by returning false?
```
mbutrovich: See TODO.

src/catalog/database_catalog.cpp
```diff
@@ -0,0 +1,1643 @@
...
+    delete[] buffer;
+    return;
+  }
+
+  // TODO(Matt): do we have any way to assert that we got the number of attributes we expect? From another attribute in
```
mbutrovich: See TODO.

src/catalog/database_catalog.cpp
```diff
@@ -0,0 +1,1643 @@
...
+  // another catalog table maybe?
+
+  // Step 2: Scan the table to get the columns
+  pr = table_pri.InitializeRow(buffer);
+  for (const auto &slot : index_results) {
+    // TODO(Matt): this doesn't delete from any tables, just indexes
```
mbutrovich: See TODO.

src/include/catalog/schema.h
```diff
@@ -4,15 +4,29 @@
...
+
 namespace terrier::catalog {
 
+namespace postgres {
+class Builder;
+}
+
 /**
  * Internal object for representing SQL table schema. Currently minimal until we add more features to the system.
  * TODO(Matt): we should make sure to revisit the fields and their uses as we bring in a catalog to replace some of the
```
mbutrovich: See TODO. Obsolete now?

src/catalog/database_catalog.cpp
jrolli: This should be an assert because the comments here are inaccurate.  Failure here means the KV pair was already in the index which would imply that the insert into SqlTable returned an already in-use slot.

src/catalog/database_catalog.cpp
jrolli: I would lean towards this being an assert and force column OID reasoning to the class (i.e. index and table) logic.  That's where we should be doing OID assignments and management since we need to track it in order to support multi-versioned schemas in the future.

src/catalog/database_catalog.cpp
jrolli: At the moment, no.  The primary customer of this function will be @GustavoAngulo in the recovery logic when he is replaying a call to CreateTable (or index).

src/catalog/database_catalog.cpp
jrolli: Yes.

src/catalog/database_catalog.cpp
jrolli: Probably.  When we drop a table, we should drop the columns before dropping the objects in order to respect the foreign key relationships (not currently enforced).  That means we could get write-write conflicts that go undetected here and cause problems at commit time.

src/catalog/database_catalog.cpp
jrolli: At this point in the code, not really.  However, we could return a count of the number of columns deleted (reserving negatives for error?) and check it inside of the drop table logic (should be only customer for this function).

src/catalog/database_catalog.cpp
jrolli: We should be deleting from pg_attribute and deferring the deallocation of the `AbstractExpression` pointer stored in _ADBIN_.

src/include/catalog/schema.h
jrolli: I think this is obsolete.  The only pending change is properly storing the default values which we have a separate TODO for in the file (line 152).

src/catalog/database_catalog.cpp
mbutrovich: When would it be used? If you drop a column, wouldn't we replace all of the attributes with a new Schema?

jrolli: We would have to rebuild the new Schema object but why would we delete the remaining columns just to recreate them with the same OIDs?

mbutrovich: I wasn't sure they were keeping the same OIDs.

test/catalog/catalog_test.cpp
```diff
@@ -0,0 +1,350 @@
...
+  // Verify our columns exist
+  EXPECT_NE(schema.GetColumn("id").GetOid(), catalog::INVALID_COLUMN_OID);
+  EXPECT_NE(schema.GetColumn("user_col_1").GetOid(), catalog::INVALID_COLUMN_OID);
+
+  // Verify we can instantiate a storage object with the generated schema
+  auto table = new storage::SqlTable(&block_store_, schema);
+
+  EXPECT_TRUE(accessor->SetTablePointer(table_oid, table));
+  EXPECT_EQ(common::ManagedPointer(table), accessor->GetTable(table_oid));
+  txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
```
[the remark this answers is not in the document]
mbutrovich: We don't support any of those yet. :(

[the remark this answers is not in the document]
mush-zhang: Can we leave a TODO both in catalog accessor header file so that other people won't use it accidentally; and in the test cpp file so that we can add the related testcases when we have them implemented?

src/include/catalog/catalog_accessor.h
```diff
@@ -0,0 +1,280 @@
...
+
+  /**
+   * Given the index name and its specification, add it to the catalog
+   * @param ns is the namespace in which the index will exist
+   * @param table on which this index exists
+   * @param name of the index
+   * @param schema describing the new index
+   * @return OID for the index, INVALID_INDEX_OID if the operation failed
+   */
+  index_oid_t CreateIndex(namespace_oid_t ns, table_oid_t table, const std::string &name, const IndexSchema &schema);
```
amlatyrngom: Aren't the table's namespace and the index's namespace the same? If so, is the first parameter needed?

src/include/catalog/catalog_accessor.h
```diff
@@ -0,0 +1,280 @@
...
+   * @param name of the new table
+   * @param schema object describing the new table
+   * @return OID for the table, INVALID_TABLE_OID if the table already exists
+   * @warning The catalog accessor assumes it takes ownership of the schema object
+   * that is passed.  As such, there is no guarantee that the pointer is still
+   * valid when this function returns.  If the caller needs to reference the
+   * schema object after this call, they should use the GetSchema function to
+   * obtain the authoritative schema for this table.
+   */
+  table_oid_t CreateTable(namespace_oid_t ns, const std::string &name, const Schema &schema);
```
amlatyrngom: So to create a table, we need to:
* Call CreateTable with a schema that doesn't have the oids yet
* Then use GetSchema to get a schema with the oids.
* Then instanstiate a SqlTable and pass it into SetTablePointer.
Is that all?

src/include/catalog/index_schema.h
```diff
@@ -0,0 +1,198 @@
...
+    friend class tpcc::Schemas;
+    friend class terrier::StorageTestUtil;
+  };
+
+  /**
+   * Instantiates a new catalog description of an index
+   * @param columns describing the individual parts of the key
+   * @param is_unique indicating whether the same key can be (logically) visible repeats are allowed in the index
+   * @param is_primary indicating whether this will be the index for a primary key
+   * @param is_exclusion indicating whether this index is for exclusion constraints
```
amlatyrngom: This exclusion parameter is missing from the list of constraints. Is it always false for now?

src/include/catalog/catalog_accessor.h
jrolli: There is no requirement that an index and the table it indexes have to be in the same namespace, so we need to have it as a parameter here.

src/include/catalog/catalog_accessor.h
jrolli: At face value, yes.

However, if the schema has any constraints (primary key, uniqueness, etc.) then we also have to define, create, and instantiate the necessary indexes as well.  The catalog will not do this.

src/include/catalog/index_schema.h
jrolli: Can you clarify what you mean by this?

PR #395, opened 2019-05-30
add =operator overloading in managedPointer

mush-zhang:


mush-zhang:
use default = as only shallow copy is needed

PR #397, opened 2019-05-30
SqlTable API change: accept RedoRecords instead of ProjectedRows

mbutrovich:
This PR is in response to noticing that it's easy to interact with the storage layer without using buffers that get passed off to the LogManager. This forces SqlTables to accept RedoRecords which originate from StageWrite (allocated in the txn's RedoBuffer) instead of generic ProjectedRows. SqlTable unpacks the enclosed ProjectedRow and passes it to DataTable to avoid an API change at that level.

Other changes:
* StageWrite needed a TupleSlot, so you couldn't use the buffer returned by it for an Insert. You had to allocate a separate buffer, call Insert, call StageWrite, then memcpy the inserted tuple into the returned "staged" buffer. That seems backwards.
* temporarily removing catalog while we work on a refactor
* RedoRecords now store db_oid_t and table_oid_t instead of DataTable pointer
* tests that interact with SqlTable updated to use StageWrite for their buffers
* outdated TODOs removed
* removed DataTable::GetOid() since it was a placeholder
* TupleSlot is no longer an argument to SqlTable::Update(). It is read from the RedoRecord
* StrongTypeAlias' copy and move constructors are constexpr functions


src/include/storage/sql_table.h
```diff
@@ -71,8 +72,8 @@ class SqlTable {
    * @param redo the desired change to be applied. This should be the after-image of the attributes of interest.
    * @return true if successful, false otherwise
    */
-  bool Update(transaction::TransactionContext *const txn, const TupleSlot slot, const ProjectedRow &redo) const {
-    return table_.data_table->Update(txn, slot, redo);
+  bool Update(transaction::TransactionContext *const txn, const TupleSlot slot, RedoRecord *const redo) const {
```
jrolli: Should this take `TupleSlot` still or should it read it from the redo record as it passes it down to the `DataTable`?  I think that would semantically cleaner.

src/include/storage/write_ahead_log/log_record.h
```diff
@@ -150,15 +159,18 @@ class RedoRecord {
...
-   * @param table
+   * @param db_oid database oid of this redo record
+   * @param table_oid table oid of this redo record
    * @param tuple_slot
    * @return
    */
   static LogRecord *PartialInitialize(byte *const head, const uint32_t size, const transaction::timestamp_t txn_begin,
-                                      DataTable *const table, TupleSlot tuple_slot) {
+                                      const catalog::db_oid_t db_oid, const catalog::table_oid_t table_oid,
+                                      const TupleSlot tuple_slot) {
```
jrolli: Based off the TODO above, we should probably remove this function or at a minimum make it private with the logging test as a friend to prevent this from getting used in @GustavoAngulo replication code.

test/storage/bwtree_key_test.cpp
```diff
@@ -221,13 +221,13 @@ class BwTreeKeyTests : public TerrierTest {
     transaction::TransactionManager txn_manager(&buffer_pool, true, LOGGING_DISABLED);
     storage::GarbageCollector gc_manager(&txn_manager);
 
+    auto *const txn = txn_manager.BeginTransaction();
+
     // dummy tuple to insert for each key. We just need the visible TupleSlot
-    auto *const insert_buffer = common::AllocationUtil::AllocateAligned(tuple_initializer.ProjectedRowSize());
-    auto *const insert_tuple = tuple_initializer.InitializeRow(insert_buffer);
+    auto *const insert_redo = txn->StageWrite(catalog::db_oid_t(0), catalog::table_oid_t(0), tuple_initializer);
```
jrolli: Same comment as above about using a `#define` clause for tests and benchmarks so that it is easier to reason about how the OIDs compare to what the catalog will eventually assign.

src/include/storage/sql_table.h
mbutrovich: Confirmed with @tli2 and we're in agreement that we can lose it as an argument. I'll make some TERRIER_ASSERTs to make sure it's populated.

test/storage/bwtree_key_test.cpp
mbutrovich: Where do we want it defined? Which util?

jrolli: `catalog_[test|benchmark]_util.h`?  It would be a new file on the benchmark side, but I think that (or something like it) would be a good place for putting catalog mock-ups and helpers in the future.

src/include/storage/write_ahead_log/log_record.h
mbutrovich: #370 removes it so I'm not gonna worry about it in this PR.

src/include/storage/write_ahead_log/log_record.h
```diff
@@ -89,16 +89,24 @@ class RedoRecord {
   MEM_REINTERPRETATION_ONLY(RedoRecord)
 
   /**
-   * @return pointer to the DataTable that this Redo is concerned with
+   * @return the tuple slot changed by this redo record
    */
-  DataTable *GetDataTable() const { return table_; }
+  TupleSlot GetTupleSlot() const { return tuple_slot_; }
```
tli2: nit: I believe for some other fields we are returning a reference instead of having a separate setter

src/include/storage/write_ahead_log/log_record.h
```diff
@@ -201,32 +215,40 @@ class DeleteRecord {
...
+   * @return table oid for this delete record
    */
-  TupleSlot GetTupleSlot() const { return tuple_slot_; }
+  catalog::table_oid_t GetTableOid() const { return table_oid_; }
 
  private:
-  // TODO(Tianyu): Change to oid maybe?
-  DataTable *table_;
+  catalog::db_oid_t db_oid_;
+  catalog::table_oid_t table_oid_;
```
tli2: I would be cautious about this change. The LogWriter still needs to know about the schema in order to write out varlens. Alternatively, you can change redo records to always inline varlens instead of using a pointer.

test/include/util/catalog_test_util.h
```diff
@@ -33,5 +35,10 @@ struct CatalogTestUtil {
     }
     return catalog::Schema(columns);
   }
+
+  // Define contants here for when you need to fake having a catalog in tests. These values may change as the catalog
+  // comes in, and tests should be modified accordingly.
+  static constexpr catalog::db_oid_t generic_db_oid = catalog::db_oid_t(101);
```
tli2: nit: maybe put the word `test` in these names too, just to make sure.

src/include/storage/write_ahead_log/log_record.h
GustavoAngulo: The only thing the LogWriter needs is the `storage::BlockLayout` object from the `DataTable*`.

mbutrovich: RawBlocks store the DataTable* now, so in theory you can always get back to the DataTable from any TupleSlot. That should be enough to get BlockLayout, right?

src/include/storage/write_ahead_log/log_record.h
mbutrovich: I was gonna do that but TupleSlot is a type name. The signature can't be `TupleSlot &TupleSlot()`.

test/include/util/catalog_test_util.h
mbutrovich: In theory the `src` folder shouldn't be able to see it since it's in the test folder, but I can do that.

mbutrovich:
Regressions on the SingleStatement benchmarks are expected since we're now hitting the ObjectPool for RedoBuffers (as a real transaction would) as opposed to reusing an internally allocated buffer.

src/include/transaction/transaction_context.h
```diff
@@ -79,7 +79,7 @@ class TransactionContext {
    * @return a persistent pointer to the head of a memory chunk large enough to hold the undo record
    */
   storage::UndoRecord *UndoRecordForInsert(storage::DataTable *const table, const storage::TupleSlot slot) {
-    byte *result = undo_buffer_.NewEntry(sizeof(storage::UndoRecord));
+    byte *const result = undo_buffer_.NewEntry(sizeof(storage::UndoRecord));
```
tli2: lol

PR #399, opened 2019-06-02
BwTree memory leak fix

mbutrovich:
From @wangziqi2013 in https://github.com/mbutrovich/terrier/pull/2
> Problem Statement
> ---------------------
> We observed memory leaks while worker threads are inserting and scanning on the same leaf node. The leaked memory is always reported as being allocated inside the constructor of ForwardIterator, in which we perform a down traversal of the tree with the scan key. It seems that memory chunks allocated during this process is not properly added into the garbage chain or not properly released by the GC.
> 
> Solution
> ---------
> ~~The current solution is to avoid adding memory chunks into the garbage chain while performing scans. This is made possible by switching to TraverseReadOptimized(), which ignores SMOs and over-sized nodes on its down traversal path. No consolidation, node split, and completion of partial SMO will be considered during this process.~~
> 
> Currently I have identified a problem with AtomicStack which is the lock-free implementation of a stack structure used for recycling node IDs of deleted nodes. It seems that this structure has subtle bugs that only occur under heavy contention. I have disabled AtomicStack, and replaced it with a deque from the standard library protected by a mutex also from std. The memory leak disappears for now, but I am not sure whether it may emerge in the future.


mbutrovich:
@wangziqi2013 do we need to make any changes to the comments around the modified code? Otherwise I'll probably merge this today. Thanks in advance!

PR #405, opened 2019-06-05
Add BwTree garbage collection to centralized GC

mbutrovich:
This adds the ability for the `GarbageCollector` to call `PerformGarbageCollection()` on any registered indexes.

It also turns off every BwTree spinning up its own thread to call `PerformGarbageCollection()`.


src/include/storage/garbage_collector_thread.h
```diff
@@ -48,10 +48,14 @@ class GarbageCollectorThread {
     gc_paused_ = false;
   }
 
+  /**
+   * Access the underlying GC object, mostly to register indexes currently.
+   */
+  storage::GarbageCollector gc_;
```
tli2: nit: put into a getter

PR #409, opened 2019-06-06
HyPer bug fix

tli2:
Fixes a-b-a problem on version chain when aborting transactions unlink themselves. Aborting
transactions also need to check out a timestamp and "commit" after rolling back their changes to guard against this. The exact interleaving is this:
```
     transaction 1                 transaction 2
          begin
  read version_ptr
                                         begin
                                   write a -> a1
          read a1
                                   rollback a1 -> a
    check version_ptr
         return a1
```
For this to manifest, there has to be high contention on a given tuple slot, and insufficient CPU resouces (way more threads than there are cores, around 8x seems to work) such that threads are frequently swapped out.  compare-and-swap along with the pointer reduces the probability of this happening to be essentially infinitesimal, but it's still a probabilistic fix. To 100% prevent this race, we have to wait until no concurrent transaction with the abort that could have had a dirty read is alive to unlink this. The easiest way to achieve that is to take a timestamp as well when all changes have been rolled back for an aborted transaction, and let GC handle the unlinking.


mbutrovich:
More docs please since it's a really subtle race, difficult to reproduce reliably on all systems, and the code is non-obvious why it exists ("a-b-a" might not be enough for someone to understand). Otherwise LGTM and passes the tests locally that I had written to reproduce the issue on master.

PR #411, opened 2019-06-06
SqlTable::Insert API change

mbutrovich:
Change `SqlTable::Insert` to return `TupleSlot` again. You can't guarantee that it's safe to read after `Insert` or `Update` if the `RedoRecord` gets swapped out. Safest just to return the `TupleSlot` immediately. Also added some stronger warnings about `StageWrite`.


src/include/transaction/transaction_context.h
```diff
@@ -104,11 +104,16 @@ class TransactionContext {
 
   /**
    * Expose a record that can hold a change, described by the initializer given, that will be logged out to disk.
-   * The change can either be copied into this space, or written in the space and then used to change the DataTable.
+   * The change should be written in the space and then used to change the SqlTable.
```
jrolli: Replace "should" with "must" since it is not optional.

src/include/transaction/transaction_context.h
```diff
@@ -104,11 +104,16 @@ class TransactionContext {
...
    * @param db_oid the database oid that this record changes
    * @param table_oid the table oid that this record changes
    * @param initializer the initializer to use for the underlying record
    * @return pointer to the initialized redo record.
+   * @warning RedoRecords returned by StageWrite are not guaranteed to remain valid forever. If you call StageWrite
+   * again, the previous RedoRecord's buffer may be swapped out, written to disk, and handed back out to another
+   * transaction.
+   * @warning If you call StageWrite, its contents WILL be logged to disk. If you StageWrite anything that you didn't
+   * succeed in writing into the table or decide you don't want to use, the transaction MUST abort.
```
jrolli: These two warning make me think we need some more assert statements, more enforcement of the abort logic, or better documentation somewhere.  Specifically, this changes my understanding of the severity of write-write conflicts and makes opportunistic methods for migrating tuples between schema versions (or possibly compaction) more likely to abort (previously, I was thinking we could attempt an update and ignore the tuple if it failed).

Does the second warning also apply to `StageDelete`?  If so we should update that documentation as well.

Is it worth adding a "must abort" flag into transaction contexts as a sanity check?

mbutrovich: What are the semantics of the proposed "must abort" flag?

jrolli: Your asserts are cleaner than what I was suggesting with a "must abort" flag and I think are sufficient.

mbutrovich:
> added asserts to minimize self-harm in the execution layer.

There's still nothing that prevents someone from calling `Commit` on a transaction that needs to abort, but I'm not totally clear on if the flag is the right way to enforce that. We can discuss as the operators start getting written.

Issue #413, opened 2019-06-09
New connection dispatch can overwrite previous dispatch


tli2:
#328 rewrote the dispatcher (`network::ConnectionDispatchertask`) to communicate with handlers (`network::ConnectionHandlerTask`) with just manual events instead of through IPC with `pipe()`. 

The original IPC pipe functions as a queue. The new implementation instead has a singular field in the `network::ConnectionHandlerTask` that is set. This will not work when either:

1. The system is under load, and the dispatcher round-robins back to a handler before the said handler has had time to observe the first dispatch
2. Multiple dispatchers (probably running different protocols on different ports) dispatch to the same handler in short succession.

This can result in lost connections in either case. We still need a queue for dispatches.

Also, the libevent flags weren't updated, and still waits for available reads on fd -1, which makes no sense. Should be changed to `RegisterManualEvent`

Bonus points if one can cook up a test that will catch this scenario.

PR #414, opened 2019-06-09
Network Refactor

tli2:
We screwed up some dependencies in the network layer. I had to shuffle things around to make sure we are not hardcoding postgres dependencies in the general network layer.

This led to the discovery of #413. Some more work remains to be done to rip the dispatcher and handler creation out of the terrier server, so that in the future we can easily have multiple dispatchers and handlers with heterogenous protocols. Not done in this PR because it probably conflicts with @GustavoAngulo


src/include/network/protocol_interpreter.h
```diff
@@ -14,6 +15,10 @@ class ConnectionHandle;
  */
 class ProtocolInterpreter {
  public:
+  struct Provider {
```
apavlo: Add comment to explain what this is used for.

test/include/util/manual_packet_util.h
```diff
@@ -0,0 +1,97 @@
+/*
+ * This header includes some handy helper functions when testing with manual constructed packets.
```
apavlo: Move this to the be the description of the util class.

test/include/util/manual_packet_util.h
```diff
@@ -0,0 +1,97 @@
...
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include "loggers/test_logger.h"
+#include "network/connection_handle_factory.h"
+#include "util/manual_packet_util.h"
+
+namespace terrier::network {
+struct ManualPacketUtil {
```
apavlo: Why is this is a struct and not a class like the other util classes?

src/include/network/network_defs.h
```diff
@@ -23,7 +23,7 @@
 #include "traffic_cop/result_set.h"
 #include "type/type_id.h"
 
-namespace terrier::traffic_cop {
+namespace terrier::tcop {
```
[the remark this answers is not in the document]
tli2: Namespaces are not supposed to have underscores, and I thought `trafficcop` looked weird.

Can change if you don't like that

PR #415, opened 2019-06-09
Rewrite GC to use deferred action framework

tli2:
Refactors the storage layer to now rely exclusively on deferred action framework to do garbage collection.

Some design adjustments were made to the framework so it can accommodate earlier optimizations to the GC in #325


tli2:
Will break this up into separate PRs since @Gus might require parts of it for now.

PR #425, opened 2019-06-14
Add AbortRecord to logging

### LINKED ISSUE #423: Add an AbortRecord for recovery
Its possible for an aborted transaction to write its LogRecords on the log file, so we need some way to know when a transaction is aborted. We will only log an AbortRercord if the aborting transaction has already flushed a log buffer. Otherwise we don't need to log anything.

GustavoAngulo:
Resolves #423 

This PR adds abort records (`AbortRecord`) to logging. If an aborted transaction aborts before filling up its redo record buffer, these logs can be thrown out without being logged. However, if it fills up a record buffer and hands it off to the log manager before aborting, these records will appear in the log file. In recovery, we will need to know that the transaction aborted, otherwise it will appear as an unfinished transaction. Thus, we log an abort record **only** if an aborting transaction has previously handed off records to the log manager. 

In the average case, a transaction will not flush its redo buffer before aborting, so this should not be a big performance hit.

The test cases test that we see an abort record in the log if an aborted transaction previously flushed, and that we don't see an abort record if an aborted transaction did not previously flush.


src/include/storage/record_buffer.h
```diff
@@ -343,16 +343,24 @@ class RedoBuffer {
...
   byte *LastRecord() const { return last_record_; }
 
+  /**
+   * @return true if this buffer has previously flushed to the log manager
+   */
+  bool HasFlushed() const { return has_flushed_; }
+
  private:
+  // Flag to denote if this RedoBuffer has flushed records to the log manager already. Used to determine if we should
+  // write an abort record
```
tli2: nit: Can you also write down the reason behind this? (i.e. if no records from the abort transaction is written out, there is no need to write the record out. In the case where there is, this helps the recovery pass to quickly discard changes, etc.)

src/include/storage/write_ahead_log/log_record.h
```diff
@@ -333,4 +333,57 @@ class CommitRecord {
...
+   */
+  transaction::timestamp_t AbortTime() const { return txn_abort_; }
+
+  /**
+   * @return pointer to the aborting transaction.
+   */
+  transaction::TransactionContext *Txn() const { return txn_; }
+
+ private:
+  transaction::timestamp_t txn_abort_;
```
tli2: I don't believe you need any of these for recovery? A simple {ABORT, txn_start_time} should suffice.

You can probably remove this class in its entirety and just use the header.

GustavoAngulo: I removed the unnecessary meta data, but I left the class for code clarity.

src/include/storage/record_buffer.h
GustavoAngulo: Done.

Issue #407, opened 2019-06-05
Use string_view in HashUtil


apavlo:
Right now `HashUtil::Hash` has a specialized implementation for `std::string`. We need to switch it to the more lightweight `std::string_view`.

PR #426, opened 2019-06-14
Use string_view in HashUtil (#407)

MageMasher:


tli2:
#407  refers to this line https://github.com/cmu-db/terrier/blob/218d072a85f2f87453013cec3c743a6ead9a7aa9/src/include/common/hash_util.h#L106

Your changes fail the tests because it now hashes strings by their byte contents (the pointer to the actual string content).

PR #429, opened 2019-06-17
Dependency Injection

tli2:
I pulled in boost::di as third_party and wrote some extensions for it so that it works better with our codebase.

Converted the large tests to use this instead of manual dependency resolution. A bit of test refactor to make that possible as well.


tli2:
Dependent on #435 for bug fix

src/include/di/di_help.h
```diff
@@ -0,0 +1,269 @@
...
+     * @return see boost::di doc
+     */
+    template <class, class, class TProvider>
+    TerrierWrapper<TExpected, TGiven> create(const TProvider &provider) {
+      if (object_ == nullptr) object_ = std::unique_ptr<TGiven>(provider.get());
+      return object_.get();
+    }
+
+   private:
+    // TODO(Tianyu): In case we ever call this multi-threaded, it will not be same. Although I doubt
```
mbutrovich: "same" -> "safe" ?

src/include/di/di_help.h
```diff
@@ -0,0 +1,269 @@
+#pragma once
+#include <memory>
+
+#include "boost/di/di.h"
+#include "common/macros.h"
+#include "common/managed_pointer.h"
+#define DECLARE_ANNOTATION(name) static constexpr auto name = [] {};
```
mbutrovich: What is an annotation in this context?

src/include/di/injectors.h
```diff
@@ -0,0 +1,24 @@
...
+#include "transaction/transaction_manager.h"
+
+namespace terrier::di {
+/**
+ * When applied, returns an injector that contains object bindings with the correct scope for the storage
+ * engine (storage, logging, GC, transactions). Note that this does not contain bindings for constants
+ * such as object pool sizes. Unless those are supplied in the top level injector, the code will not compile.
+ */
+// TODO(Tianyu): Should we provide default values for constants here?
+auto storage_injector UNUSED_ATTRIBUTE = [] {
```
mbutrovich: What are the circumstances when we should define a new injector here? Any time we have a set of components that we want to bring up together?

src/include/di/di_help.h
```diff
@@ -0,0 +1,269 @@
...
+ public:
+  /**
+   * Implementation of the scope. see boost::di doc
+   * @tparam TExpected
+   * @tparam TGiven
+   */
+  template <class TExpected, class TGiven>
+  class scope {
+   public:
+    // TODO(Tianyu): Not sure about this. This is the referrable flag used for boost::di's singleton scope.
```
apavlo: What are you "not sure about"? Can you be more specific?

src/include/di/di_help.h
```diff
@@ -0,0 +1,269 @@
...
+    /**
+     * See boost::di doc
+     */
```
apavlo: Use `@see` with a link to the specific documentation page.

src/include/di/di_help.h
```diff
@@ -0,0 +1,269 @@
...
+   * Implementation of the scope. see boost::di doc
+   * @tparam TExpected
+   * @tparam TGiven
+   */
+  template <class TExpected, class TGiven>
+  class scope {
+   public:
+    // TODO(Tianyu): Not sure about this. This is the referrable flag used for boost::di's singleton scope.
+    /**
+     * See boost::di doc
```
apavlo: Can you explain what this actually is?

src/include/di/di_help.h
```diff
@@ -0,0 +1,269 @@
...
+     */
```
apavlo: Add `@see`

src/include/di/injectors.h
tli2: I think so. Whatever corresponds to a "layer" or "engine" in the system might qualify. I don't have a tried and true rule to provide.

src/include/di/di_help.h
tli2: @see only works for code concepts I think? I will just link to their webpage here.

tli2:
Instead of a README, I think I will do more in-line comments and a wiki page.

tli2:
https://github.com/cmu-db/terrier/wiki/Dependency-Injection

apavlo:
@tli2 Can you mark the comments as resolved if you think you've fixed them?

apavlo:
@tli2 Are we all done with this?

tli2:
Forgot about adding comments to the referable flag. Will fix that tonight after I get to my laptop. Should be all done after that

Issue #420, opened 2019-06-12
Restore AbstractExpression Meta-Data


apavlo:
@mush-zhang found that the Peloton `TupleValueExpression` has an internal triple of `oid_t` called the `bound_obj_id`:

https://github.com/cmu-db/peloton/blob/master/src/include/expression/tuple_value_expression.h#L193

We did not port this over to the new repo. It is not clear what it is but it is likely that the optimizer needs it.

We need to investigate (1) what it is, (2) how it is being set, and (3) how it is being used. If we decide that we still want it, then we should port it over.

-----------

## Update: 2019-06-14

There are four different internal data members that we need to add back to the expressions:
1. `expr_name`
2. `depth`
3. `alias`
4. `has_subquery`

We think that we can get back `alias` by adding back in this line:

https://github.com/cmu-db/terrier/blob/master/src/parser/postgresparser.cpp#L778-L783

The `depth` and `has_subquery` values can be derived in the binder, but we are not sure just yet how to do this in a single pass through the expression tree.

17zhangw:
This line may provide some insight: https://github.com/cmu-db/peloton/blob/master/src/include/expression/tuple_value_expression.h#L157

mush-zhang:
I am letting the postgres parser to write the alias to TupleValueExpression **only** in ColumnRefTransform() function. 
The `alias` field comes from the `ResTarget.name` in pg_parser. As its comment suggested, `name` is either:
1. column label in SELECT target list;
2. destination column in INSERT target-column-names list
3. destination column in UPDATE target list
Only column reference in select statement is stored as list of abstract expression, so I think that is the only place we need to call the constructor with `alias`. ( columns are stored as strings in INSERT and UPDATE)

lmwnshn:
This made it in.

Issue #404, opened 2019-06-05
Rename TupleValueExpression


apavlo:
We are going to break up the old `TupleValueExpression` into two separate classes:

1. `ColumnValueExpression`
   This is for indicating that a value comes from a tuple access from a SQLTable. There will be two constructors:
   1. `ColumnValueExpression(table_name, col_name)`
   2. `ColumnValueExpression(table_oid, col_oid)`
2. `DerivedValueExpression(type, tuple_idx, value_idx)`
   This is to indicate that the value comes from a derived/intermediate result.

apavlo:
@mush-zhang Be sure to switch the order of the constructor arguments so that `table_*` comes first before `column_*`.

PR #433, opened 2019-06-18
AbstractExpression & TupleValueExpression Fix

mush-zhang:
This pr will deal with issue #420 and  #404. 

- [x] Port over `alias_` as private member of AbstractExpression
  - [x] Constructor with `alias` in AbstractExpression and TupleValueExpression
  - [x] Call corresponding constructors in PostgresParser with alias as a parameter
  - [x] public method `GetAlias()`
- [x] Port over `expr_name_` as private member of AbstractExpression
  - [x] public method `GetExpressionName()`
  - [x] private/protected method `DeduceExpressionName()` in AbstractExpression
  - [x] override in ConstantValueExpression and TupleValueExpression
  - [x] Bring back `ExpressionTypeToString` function and `IsAggregateExpression` function for ExpressionType
  - [x] Bring in `TransientValue`'s `ToString()` method, for `GetExpressionName()` overriding in ConstantValueExpression
- [x] Port over `depth_` as private member of AbstractExpression
  - [x] public method `GetDepth()`
  - [x] private/protected method `DeriveDepth()` in AbstractExpression
  - [x] derive depth function override in subqueryExpression
- [x] Port over `has_subquery` as private member of AbstractExpression
  - [x] public method `HasSubquery()`
  - [x] private/protected method `DeriveSubqueryFlag()` in AbstractExpression
- [x] Public method `DeduceExpressionType()` in AbstractExpression as it will be used in binder and optimizer at least
  - [x] override in AggregateExpression and OperatorExpression
- [x] `Visit()`, `Accept()`, and `AcceptChildren()` for all expressions and related sql node visitor

- [x] Add more comprehensive testcases:
  - [x] AggregateExpressionTest
  - [x] ComparisonExpressionTest
  - [x] DefaultExpressionTest and DefaultExpressionJsonTest if needed
  - [x] FunctionExpressionTest
  - [x] OperatorExpressionTest
  - [x] ParameterValueExpressionTest
  - [x] StarExpressionTest and StarExpressionJsonTest (this is trivial)
  - [x] SubqueryExpressionTest
  - [x] ColumnValueExpressionTest and ColumnValueExpressionJsonTest
  - [x] DerivedValueExpressionTest and DerivedValueExpressionJsonTest
  - [x] TypeCastExpressionTest
  - [x] CaseExpressionJsonTest
  - [x] ConjunctionExpressionTest
  - [x] Testcases for functions added in TransientValue
  - [x] Testcases for functions added to Expression_defs
  - [x] Testcases for fucntions added in TypeUtils

 - [x] Split TupleValueExpression to ColumnValueExpression and DerivedValueExpression
  - [x] Use ColumnValueExpression in postgress parser
~~- [ ] Bring back `bound_obj_id_` and `is_bound_` for ColumnValueExpression~~


src/include/parser/expression/subquery_expression.h
```diff
@@ -37,6 +37,52 @@ class SubqueryExpression : public AbstractExpression {
...
+  int DeriveDepth() override {
+    int current_depth = this->GetDepth();
+    for (auto &select_elem : subselect_->GetSelectColumns()) {
+      int select_depth = select_elem->DeriveDepth();
+      if (select_depth >= 0 && (current_depth == -1 || select_depth < current_depth)) this->SetDepth(select_depth);
+    }
+    auto where = subselect_->GetSelectCondition();
+    if (where != nullptr) {
+      auto where_depth = where->DeriveDepth();
+      if (where_depth >= 0 && where_depth < current_depth) this->SetDepth(where_depth);
```
GustavoAngulo: What Depth are we supposed to find here? The smallest depth? the variable `current_depth` is never updated, so in the end, the depth will be set to just the last depth that is smaller than the `current_depth` we stored in `current_depth`. 

I think if this code is working as intended, there should be a comment explaining what depth we want in the end.

src/include/parser/expression/abstract_expression.h
```diff
@@ -125,9 +213,43 @@ class AbstractExpression {
   virtual void FromJson(const nlohmann::json &j);
 
  private:
-  ExpressionType expression_type_;                             // type of current expression
-  type::TypeId return_value_type_;                             // type of return value
-  std::vector<std::shared_ptr<AbstractExpression>> children_;  // list of children
+  /**
+   * Walks the expression trees and generate the correct expression name
+   */
+  virtual void DeduceExpressionName();
```
GustavoAngulo: Why is this one `Deduce` when the other ones are `Derive`?

src/parser/expression/abstract_expression.cpp
```diff
@@ -142,4 +161,53 @@ std::shared_ptr<AbstractExpression> DeserializeExpression(const nlohmann::json &
...
+  return depth_;
+}
+
+void AbstractExpression::DeduceExpressionName() {
+  // If alias exists, it will be used in TrafficCop
+  if (!alias_.empty()) return;
+
+  bool first = true;
+
+  if (expression_type_ == ExpressionType::FUNCTION) {
```
GustavoAngulo: Why is the logic for `ExpressionType::FUNCTION` here? Shouldn't it be in `FunctionExpression` in an overrided method similar to how you do it for `ConstantValueExpression`

src/include/parser/expression/abstract_expression.h
```diff
@@ -26,6 +28,18 @@ class AbstractExpression {
...
+   * Instantiates a new abstract expression with alias used for select statement column references.
+   * @param expression_type what type of expression we have
+   * @param return_value_type the type of the expression's value
+   * @param alias alias of the column (used in column value expression)
+   * @param children the list of children for this node
+   */
+  AbstractExpression(const ExpressionType expression_type, const type::TypeId return_value_type, const char *alias,
+                     std::vector<std::shared_ptr<AbstractExpression>> &&children)
+      : expression_type_(expression_type), return_value_type_(return_value_type), children_(std::move(children)) {
+    if (alias != nullptr) alias_ = std::string(alias);
```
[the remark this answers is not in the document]
mush-zhang: postgresparser parse the text with thrid party component and initialize an expression with whatever in the parse node. I can check if alias is nullptr in the postgres parser and call corresponding constructor depending on whether alias is there to avoid the above

mush-zhang:
I will need to write an operator== function for selection statement, for comparing subquery_expression.

src/include/parser/expression/abstract_expression.h
```diff
@@ -38,6 +52,21 @@ class AbstractExpression {
...
+
+  /**
+   * @param return_value_type Set the return value type of the current expression
+   */
+  void SetReturnValueType(type::TypeId return_value_type) { return_value_type_ = return_value_type; }
+
+  /**
+   * @param depth Set the depth of the current expression
+   */
+  void SetDepth(int depth) { depth_ = depth; }
```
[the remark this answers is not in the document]
mush-zhang: The binder will be using them. The reason they are protected instead of private is because subclasses of abstract_expression will be able to inherit and use them. For example, SetReturnValueType() is used in aggregate_expression and comparison_expression's DeduceReturnValueType()

src/include/parser/expression/abstract_expression.h
```diff
@@ -112,6 +149,57 @@ class AbstractExpression {
...
+
+  /**
+   * @param v Visitor pattern for the expression
+   */
+  virtual void Accept(SqlNodeVisitor *v) = 0;
+
+  /**
+   * @param v Visitor pattern for the expression
+   */
+  virtual void AcceptChildren(SqlNodeVisitor *v) {
```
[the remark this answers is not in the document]
mush-zhang: You are right. I can make Accept() and AcceptChildren() protected. But I will need to experiment if that would work for the binder, as friendship doesn't get inherited.

src/include/parser/expression/subquery_expression.h
mush-zhang: Depth will be set by the binder. So before expressions get into binder,  depth is never touched.

src/include/parser/expression/abstract_expression.h
mush-zhang: ah... I copied from Peloton.... I can standardize it. Which one makes more sense?

src/include/parser/expression/type_cast_expression.h
```diff
@@ -4,20 +4,19 @@
...
 #include "parser/expression/abstract_expression.h"
-#include "parser/expression_defs.h"
-#include "type/type_id.h"
 
 namespace terrier::parser {
 /**
  * Represents a type cast expression.
  */
 class TypeCastExpression : public AbstractExpression {
+  // TODO(Ling):  Do we need a separate class for operator_cast? We can put it in operatorExpression
```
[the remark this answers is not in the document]
mush-zhang: At least the parser is involved; I don't know if optimizer has implemented functions related to this class. Perfect question to bring up in today's optimizer meeting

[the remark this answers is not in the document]
17zhangw: The optimizer (as it stands) doesn't directly touch `TypeCastExpression`.

src/include/parser/expression/subquery_expression.h
GustavoAngulo: I see.

Does this code work as intended though? In AbstractExpression, it seems you calculate the [min depth](https://github.com/cmu-db/terrier/pull/433/files/a46ea0fb46997f6182be0a9ef60e6e49609c2ec5?file-filters%5B%5D=.cpp&file-filters%5B%5D=.h&hide-deleted-files=true#diff-a5c29509067f28d9df138208f3d05ff2R182), but the above code doesn't do that since you never update `current_depth` after calling `SetDepth`

src/include/parser/expression/abstract_expression.h
GustavoAngulo: I think `Derive` makes more sense here.

src/include/parser/expression/abstract_expression.h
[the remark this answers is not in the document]
GustavoAngulo: I see, then yea I would handle that logic in the parser and call the corresponding constructor to make the difference between the constructors explicit.

src/include/parser/expression/abstract_expression.h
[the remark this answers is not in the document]
GustavoAngulo: Ahh I missed the `protected` keyword. That sounds good.

src/include/parser/expression/abstract_expression.h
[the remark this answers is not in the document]
GustavoAngulo: I see, you're right about the inheritance of friendship. This is fine then.

src/include/parser/expression/type_cast_expression.h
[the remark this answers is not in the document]
GustavoAngulo: Since the catalog PR is currently waiting on this one, I would just make an issue for this and we can do it in a separate PR.

src/include/parser/expression/abstract_expression.h
```diff
@@ -26,6 +28,19 @@ class AbstractExpression {
...
                      std::vector<std::shared_ptr<AbstractExpression>> &&children)
       : expression_type_(expression_type), return_value_type_(return_value_type), children_(std::move(children)) {}
+  /**
+   * Instantiates a new abstract expression with alias used for select statement column references.
+   * @param expression_type what type of expression we have
+   * @param return_value_type the type of the expression's value
+   * @param alias alias of the column (used in column value expression)
+   * @param children the list of children for this node
+   */
+  AbstractExpression(const ExpressionType expression_type, const type::TypeId return_value_type, const char *alias,
```
apavlo: Why is `alias` a `const char*` and not `std::string`?

src/include/parser/expression/abstract_expression.h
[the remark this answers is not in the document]
apavlo: Create an issue for this so that we don't lose track.

src/include/parser/expression/derived_value_expression.h
```diff
@@ -0,0 +1,84 @@
...
+   * @param j json to deserialize
+   */
+  void FromJson(const nlohmann::json &j) override {
+    AbstractExpression::FromJson(j);
+    tuple_idx_ = j.at("tuple_idx").get<int>();
+    value_idx_ = j.at("value_idx").get<int>();
+  }
+
+ private:
+  int tuple_idx_;
```
apavlo: Missing documentation.

src/include/parser/expression/derived_value_expression.h
```diff
@@ -0,0 +1,84 @@
...
+  int value_idx_;
```
apavlo: Missing documentation.

src/include/parser/expression/derived_value_expression.h
```diff
@@ -0,0 +1,84 @@
...
+#include <memory>
+#include <string>
+#include <utility>
+#include <vector>
+#include "parser/expression/abstract_expression.h"
+
+namespace terrier::parser {
+
+/**
+ * Represents a logical tuple value.
```
apavlo: Provide a more descriptive explanation of what this object represents.

src/include/parser/postgresparser.h
```diff
@@ -88,13 +88,13 @@ class PostgresParser {
   static std::unique_ptr<parser::SQLStatement> NodeTransform(Node *node);
 
   // expressions
-  static std::unique_ptr<AbstractExpression> ExprTransform(Node *node);
+  static std::unique_ptr<AbstractExpression> ExprTransform(Node *node, char *alias = nullptr);
```
apavlo: No default argument values.

src/include/parser/expression_defs.h
```diff
@@ -72,4 +74,13 @@ enum class ExpressionType : uint8_t {
...
 };
 
+/**
+ * When short_str is true, return a short version of ExpressionType string
+ * For example, + instead of Operator_Plus. It's used to generate the expression name
+ * @param type Expression Type
+ * @param short_str Flag if a short version of the Expression Type should be returned; default false
+ * @return String representation of the Expression Type
+ */
+std::string ExpressionTypeToString(ExpressionType type, bool short_str = false);
```
apavlo: No default argument values.

src/include/parser/postgresparser.h
```diff
@@ -88,13 +88,13 @@ class PostgresParser {
...
   static ExpressionType StringToExpressionType(const std::string &parser_str);
 
   static std::unique_ptr<AbstractExpression> AExprTransform(A_Expr *root);
   static std::unique_ptr<AbstractExpression> BoolExprTransform(BoolExpr *root);
   static std::unique_ptr<AbstractExpression> CaseExprTransform(CaseExpr *root);
-  static std::unique_ptr<AbstractExpression> ColumnRefTransform(ColumnRef *root);
+  static std::unique_ptr<AbstractExpression> ColumnRefTransform(ColumnRef *root, char *alias = nullptr);
```
apavlo: No default argument values.

test/parser/expression_defs_test.cpp
```diff
@@ -0,0 +1,160 @@
+#include "parser/expression_defs.h"
+#include <string>
+
+#include "gtest/gtest.h"
+
+namespace terrier::parser {
+
+// NOLINTNEXTLINE
+TEST(ExpressionDefsTests, ExpressionTypeToStringTest) {
```
apavlo: This function is gross. Why not just make a map of input / outputs?

src/include/parser/expression_defs.h
```diff
@@ -38,6 +40,8 @@ enum class ExpressionType : uint8_t {
   CONJUNCTION_AND,
   CONJUNCTION_OR,
 
+  COLUMN_TUPLE,
```
apavlo: Shouldn't this be `COLUMN_VALUE`?

test/parser/expression_defs_test.cpp
mush-zhang: a map would also require us to hard code all its input-output pairs

src/include/parser/expression/abstract_expression.h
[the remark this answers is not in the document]
mush-zhang: I looked at the code and found out that the sqlNodeVisitor calls the AcceptChildren() methods of children expression classes. Moving these 2 methods to protected will lead to making SqlNodeVisitor class a friend to all expression classes, which sounds undesirable. These 2 functions, together with the DeriveExpression*() functions, modified the expression only based on the existing values of the expression. As the expressions cannot be forced modified by other classes besides the binder, public Accept*() and DeriveExpression*() functions are safe to be exposed to the public and illegal and unintentional calling of them would not cause fundamental harm on the system.

src/include/parser/expression/column_value_expression.h
```diff
@@ -0,0 +1,209 @@
...
+      : AbstractExpression(ExpressionType::COLUMN_VALUE, type::TypeId::INVALID, std::move(alias), {}),
+        namespace_name_(std::move(namespace_name)),
+        table_name_(std::move(table_name)),
+        column_name_(std::move(col_name)) {}
+
+  /**
+   * @param table_name table name
+   * @param col_name column name
+   */
+  ColumnValueExpression(std::string table_name, std::string col_name)
```
apavlo: I don't think we need this constructor. This class should always be given a namespace.

src/include/parser/expression/column_value_expression.h
```diff
@@ -0,0 +1,209 @@
...
+        namespace_name_(std::move(namespace_name)),
+        table_name_(std::move(table_name)),
+        column_name_(std::move(col_name)) {}
+
+  /**
+   * @param database_oid database OID
+   * @param table_oid table OID
+   * @param column_oid column OID
+   */
+  ColumnValueExpression(catalog::db_oid_t database_oid, catalog::table_oid_t table_oid, catalog::col_oid_t column_oid)
```
apavlo: Missing `namespace_oid_t`

src/include/parser/expression/abstract_expression.h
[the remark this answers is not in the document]
GustavoAngulo: Sounds good to me.

src/include/parser/expression/parameter_value_expression.h
```diff
@@ -57,6 +65,9 @@ class ParameterValueExpression : public AbstractExpression {
 
  private:
   // TODO(Tianyu): Can we get a better name for this?
+  /**
+   * Offset of the parameter
```
apavlo: "Offset of the value that this expression points to in the query's parameter list"

src/include/parser/expression/column_value_expression.h
jrolli: We need this constructor.  Unqualified object names are allowed because the binder is expected to use a search path of namespaces and bind to the first match.

src/include/parser/expression/column_value_expression.h
jrolli: We don’t need it here because we have the table OID.  Except for creating a table, we do not need the namespace if we already have the object OID.

src/include/parser/expression_defs.h
```diff
@@ -38,6 +40,8 @@ enum class ExpressionType : uint8_t {
   CONJUNCTION_AND,
   CONJUNCTION_OR,
 
+  COLUMN_VALUE,
```
17zhangw: How is this different from existing `ExpressionType::VALUE_TUPLE`?

src/include/parser/expression/constant_value_expression.h
```diff
@@ -35,13 +35,20 @@ class ConstantValueExpression : public AbstractExpression {
     return value_ == const_expr.GetValue();
   }
 
+  void DeriveExpressionName() override {
+    if (!this->GetAlias().empty()) return;
```
17zhangw: Just wondering, don't you have to call `this->SetExpressionName(this->GetAlias())`?

src/include/parser/expression_defs.h
mush-zhang: Value_tuple is the expression type of derivedValueExpression, and column_value for ColumnValueExpression

mbutrovich:
That benchmark failure on inserts is weird considering you didn't touch that code at all. Machine anomaly, I wonder?

mush-zhang:
@17zhangw The Copy() function is only a shallow copy, all it does is getting a shared pointer to the object. But we could still add some dummy tests for the Copy() function anyway

mush-zhang:
@mbutrovich I only modified one line of documentation... that is werid. Jenkins has been unstable recently. Btw I never succeeded in restarting build on jenkins. I will probably push some minor change tonight and see if it works by that time

src/include/parser/expression/column_value_expression.h
mush-zhang: Also I'm not clear when is this constructor called - certainly not in the parser, as the parser has no way to know oids.

jrolli: The binder should replace most (all?) of the ColumValueExpressions created by the parser with ones made through this constructor.

GustavoAngulo:
I've had the `DataTableBenchmark::SimpleInsert` fail randomly too on another PR. As far as I know, it fixed itself after some new commits. It's probably an anomaly.

crd477:
> @mbutrovich I only modified one line of documentation... that is werid. Jenkins has been unstable recently. Btw I never succeeded in restarting build on jenkins. I will probably push some minor change tonight and see if it works by that time

Are you able to login to Jenkins using the "login" button in the upper right corner?  If so, you should be given options to restart builds that failed anomalously on your own.

mush-zhang:
> > @mbutrovich I only modified one line of documentation... that is werid. Jenkins has been unstable recently. Btw I never succeeded in restarting build on jenkins. I will probably push some minor change tonight and see if it works by that time
> 
> Are you able to login to Jenkins using the "login" button in the upper right corner? If so, you should be given options to restart builds that failed anomalously on your own.

I can login, but the restart button never works for me

PR #439, opened 2019-06-19
Metrics Manager

mbutrovich:
Work in progress PR for Metrics Manager that's at least ready for CI to start yelling at me if I break something there.

Still TODO:
- [x] Metrics components can only be toggled by Settings Manager.
- [x] Logging metrics for @malin1993ml as the first customer.
- [ ] Design docs once we know that this is the architecture that we want.
- [ ] Meaningful benchmarks


mbutrovich:
Closing since we don't need it running CI and we'll defer merging. All of the fixes were cherry-picked into separate PRs already.

PR #441, opened 2019-06-20
Refactor deferred action framework

tli2:
Getting it out of the transaction manager and into its own component.

Some other minor tweaks to parameters supplied and such. Dependent on #429 because I was not going to manually change all the tests.

That said, somebody needs to do what was done in #429 in the benchmark folder still.

Also does not rewrite GC using DAF yet due to the number of conflicts and changes that would generate.Although the general idea of how that could be done is in #415


mbutrovich:
Find All > Replace > "Action" -> "Event"

src/include/storage/garbage_collector.h
```diff
@@ -93,7 +105,7 @@ class GarbageCollector {
   // queue of txns that need to be unlinked
   transaction::TransactionQueue txns_to_unlink_;
   // queue of unexecuted deferred actions
-  std::queue<std::pair<transaction::timestamp_t, transaction::Action>> deferred_actions_;
+  std::queue<std::pair<transaction::timestamp_t, transaction::TransactionAction>> deferred_actions_;
```
jrolli: See earlier note about changing label to “PlannedAction”. It seems like you’re using TransactionAction in a conflicting way.

src/include/storage/index/bwtree_index.h
```diff
@@ -40,13 +41,13 @@ class BwTreeIndex final : public Index {
...
         result,
         "non-unique index shouldn't fail to insert. If it did, something went wrong deep inside the BwTree itself.");
-
-    // Register an abort action with the txn context in case of rollback
-    txn->RegisterAbortAction([=]() {
-      const bool UNUSED_ATTRIBUTE result = bwtree_->Delete(index_key, location);
-      TERRIER_ASSERT(result, "Delete on the index failed.");
-    });
-
+    if (result) {
```
jrolli: Why are we adding this conditional when the assert immediately before it indicates it will always be true here?

src/include/storage/index/bwtree_index.h
```diff
@@ -93,11 +94,10 @@ class BwTreeIndex final : public Index {
...
 
     // Register a deferred action for the GC with txn manager. See base function comment.
-    auto *const txn_manager = txn->GetTransactionManager();
-    txn->RegisterCommitAction([=]() {
-      txn_manager->DeferAction([=]() {
+    txn->RegisterCommitAction([=](transaction::DeferredActionManager *deferred_action_manager) {
+      deferred_action_manager->RegisterDeferredAction([=]() {
         const bool UNUSED_ATTRIBUTE result = bwtree_->Delete(index_key, location);
-        TERRIER_ASSERT(result, "Deferred delete on the index failed.");
+        TERRIER_ASSERT(result, "Deferred delete on the index failed.");  // NOLINT
```
jrolli: What message is the NOLINT suppressing?

src/include/storage/storage_defs.h
```diff
@@ -19,9 +19,8 @@
 #include "transaction/transaction_defs.h"
 
 namespace terrier::storage {
+// TODO(Tianyu): Remove and replace with just the new DISABLED macro
```
jrolli: Is this TODO still needed?

src/include/transaction/deferred_action_manager.h
```diff
@@ -0,0 +1,108 @@
...
+    timestamp_t oldest_txn = timestamp_manager_->OldestTransactionStartTime();
+    processed += ClearBacklog(oldest_txn);
+    processed += ProcessNewActions(oldest_txn);
+    return processed;
+  }
+
+ private:
+  TimestampManager *timestamp_manager_;
+  // TODO(Tianyu): We might want to change this data structure to be more specialized than std::queue
+  // Both queue should stay sorted in timestamp
```
jrolli: The queues are not guaranteed to be monotonically increasing on time stamp.  See note on register method.

src/include/transaction/deferred_action_manager.h
```diff
@@ -0,0 +1,108 @@
...
+  explicit DeferredActionManager(TimestampManager *timestamp_manager) : timestamp_manager_(timestamp_manager) {}
+
+  /**
+   * Adds the action to a buffered list of deferred actions.  This action will
+   * be triggered no sooner than when the epoch (timestamp of oldest running
+   * transaction) is more recent than the time this function was called.
+   * @param a functional implementation of the action that is deferred. @see DeferredAction
+   */
+  timestamp_t RegisterDeferredAction(const DeferredAction &a) {
+    timestamp_t result = timestamp_manager_->CurrentTime();
```
jrolli: Grabbing the time stamp outside the latch below allows violations of your stated “sorted by time stamp” invariant for the queue (see below).  I don’t think this has any practical impacts at this time (we’re not reliant on the invariant), but we cannot guarantee a sorted queue with this ordering.

src/include/transaction/deferred_action_manager.h
```diff
@@ -0,0 +1,108 @@
...
+   * Clear the queue and apply as many actions as possible
+   * @return numbers of deferred actions processed
+   */
+  uint32_t Process() {
+    uint32_t processed = 0;
+    // Check out a timestamp from the transaction manager to determine the progress of
+    // running transactions in the system.
+    timestamp_t oldest_txn = timestamp_manager_->OldestTransactionStartTime();
+    processed += ClearBacklog(oldest_txn);
+    processed += ProcessNewActions(oldest_txn);
```
jrolli: Because the queue may not be in time stamp order (see above), this call can result in out-of-order processing of some parts of the queue.  Shouldn’t processing new actions be conditional on backlog being empty?

src/include/transaction/timestamp_manager.h
```diff
@@ -0,0 +1,73 @@
...
+#include "common/spin_latch.h"
+#include "common/strong_typedef.h"
+#include "transaction/transaction_defs.h"
+
+namespace terrier::transaction {
+class TransactionManager;
+/**
+ * Generates timestamps, and keeps track of the lifetime of transactions (whether they have entered or left the system)
+ */
+class TimestampManager {
```
jrolli: I’m confused as to why we needed to separate this class from the transaction manager if the majority of its code is specific to tracking transactions.

Specifically, why is it not solely the transaction manager’s job to track live transactions?

src/include/transaction/transaction_context.h
```diff
@@ -155,27 +149,37 @@ class TransactionContext {
...
+   */
+  void RegisterAbortAction(const TransactionAction &a) { abort_actions_.push_front(a); }
+
   /**
    * Defers an action to be called if and only if the transaction aborts.  Actions executed LIFO.
    * @param a the action to be executed
    */
-  void RegisterAbortAction(const Action &a) { abort_actions_.push_front(a); }
+  void RegisterAbortAction(const std::function<void()> &a) {
+    RegisterAbortAction([=](transaction::DeferredActionManager * /*unused*/) { a(); });
```
jrolli: Do we know how much of a hit we take by wrapping a lambda in a lambda?

src/include/storage/garbage_collector.h
tli2: This is unused. Seems like a merge problem. Removed.

src/include/transaction/deferred_action_manager.h
tli2: Yup. Leftover comment from previous version. Removed.

src/include/transaction/deferred_action_manager.h
tli2: Why? I don't think we should guarantee in-order processing for deferred actions. If there are dependencies, users should probably not queue them in parallel. (i.e. they should defer something that defers the second thing.)

The code should still work if we wrap the following line in a condition. Just want to discuss why you think it's necessary for correctness.

src/include/transaction/timestamp_manager.h
tli2: circular dependency between DeferredActionManager and TransactionManager for DI. (DAM cannot function without a pointer to TM, TM cannot function without owning a DAM pointer.) This is the easiest way to break that cycle.

src/include/transaction/transaction_context.h
tli2: This part of the code is not performance critical, so I didn't bother benchmarking it. There exists a TODO for this (https://github.com/cmu-db/terrier/blob/ded38bb7629b744eac555c28d7ef731cd06ebdf5/src/include/transaction/deferred_action_manager.h#L42)

Do we have an eager beginner who wants to look at this?

src/include/transaction/deferred_action_manager.h
jrolli: I’m not sure if it causes a correctness problem in practice, but our reasoning about double deferrals for drop index and drop table relies in part on the assumption that the deferral queue is processed in order.

We’re certainly making it significantly harder to reason about the deferrals, but I think in practice it only results in reordered execution of parallelizable tasks.  However, I cannot prove that at the moment and it would definitely require some strong analysis showing correctness for the paper.

src/include/transaction/transaction_context.h
jrolli: Create an issue and mark it beginner.  I’m with you that the penalty is not a major concern for now.

src/include/transaction/timestamp_manager.h
jrolli: Should we generalize the nomenclature to tracking “active timestamps“?  The dependency issue makes sense, but I think we need to make a cleaner separation between the two classes even if it’s just terminology.

src/include/transaction/deferred_action_manager.h
tli2: Leave it for now and open an issue?

src/include/transaction/timestamp_manager.h
tli2: I was thinking about that, but then, this concept exists to manage transactions. Inventing another concept of "active timestamps" seems counterproductive.

src/include/transaction/deferred_action_manager.h
jrolli: Sounds good.

tli2:
#519  and #520 

@mbutrovich John's review does not seem to qualify as a review here? Maybe some permissions need to be straightened out?

mbutrovich:
Just at a high level can you give an overview of the new class/behavior that was split off so folks who didn't review this have a mental model of how the framework now works?

GustavoAngulo:
Whats the plan for fixing #520 and #526, I don't think this PR should have been merged without the fix.

PR #465, opened 2019-07-08
Removed DedicatedThreadRegistry singleton, networking refactor

### LINKED ISSUE #418: Remove singleton DedicatedThreadRegistry
Currently the DedicatedThreadRegistry is a [singleton](https://github.com/cmu-db/terrier/blob/025d906a353430700bd4376fa066b3d7849b2ef8/src/include/common/dedicated_thread_registry.h#L48). We want to get rid of it because we hate singletons.

mbutrovich:
The infrastructure changes for #439 were starting to become overwhelming so I cherry-picked more of them out:
- Fix #418. This entailed modifying some classes to have the `DedicatedThreadRegistry` injected as a dependency. Most visibly, anything that derives from `DedicatedThreadOwner` now stashes the thread registry as a member so they always have access to it. Also moved all the threading stuff into the `common` namespace since it was in that folder.
- While looking at the threading and metrics stuff, it became clear that the network layer was making some incorrect assumptions since it was previously the only customer of that framework. The network layer is no longer allowed to call the thread registry's shut down method (impractical to assume the network layer can bring the whole system down) there were races with respect to the order tasks were shut down in. `ConnectionDispatcherTask` now has a proper `RunTask` and `Terminate` function, and allocates the `ConnectHandlerTask`s at the beginning of `RunTask` rather than in its constructor. The hierarchy of the tasks should now be more clear, and `ConnectionDispatcherTask` now properly registers itself with the framework rather than running as its own thread loop in `TerrierServer`. I simplified the `TerrierServer` logic for `Shutdown` and `Close` and got rid of `ServerLoop`.


src/main/db_main.cpp
```diff
@@ -21,63 +25,68 @@ void DBMain::Init() {
...
       settings_manager_->GetString(settings::Param::log_file_path),
       settings_manager_->GetInt(settings::Param::num_log_manager_buffers),
       std::chrono::milliseconds{settings_manager_->GetInt(settings::Param::log_serialization_interval)},
       std::chrono::milliseconds{settings_manager_->GetInt(settings::Param::log_persist_interval)},
-      settings_manager_->GetInt(settings::Param::log_persist_threshold), buffer_segment_pool_);
+      settings_manager_->GetInt(settings::Param::log_persist_threshold), buffer_segment_pool_,
+      common::ManagedPointer(thread_registry_));
   log_manager_->Start();
 
+  txn_manager_ = new transaction::TransactionManager(buffer_segment_pool_, true, log_manager_);
```
GustavoAngulo: Curious, what was the reasoning to changing the ordering of what order we bring up the components?

mbutrovich: `TransactionManager` was being brought up before `LogManager` so logging was never enabled.

Issue #432, opened 2019-06-18
Add Checksum to log file


GustavoAngulo:
We want to add a checksum to the log file to catch log file corruption.

PR #468, opened 2019-07-23
Checksum

jiulingz:
This PR adds the checksum to log files #432 .

Questions: 
- How to deal with partially filled buffers?
  In `BufferedLogWriter::BufferWrite`, I checked if the buffer is full (or almost full with size `LOG_BUFFER_PAYLOAD_SIZE = LOG_BUFFER_SIZE - LOG_BUFFER_SUM_SIZE`) after each write. If it is, I write the sum to the buffer as well. In my understanding of the workflow, this buffer is then handed over by the log serializer to the log consumer using `LogSerializerTask::HandFilledBufferToWriter`. 
However, this function is also called in `LogSerializerTask::Process` without checking if the `filled_buffer_` is full. I am not sure how I should handle this case.

### DIFF
--- a/src/include/common/constants.h
+++ b/src/include/common/constants.h
@@ -26,5 +26,13 @@ struct Constants {
    * The size of the buffers the log manager uses to buffer serialized logs and "group commit" them when writing to disk
    */
   static const uint32_t LOG_BUFFER_SIZE = (1 << 12);
+  /**
+   * The size of the checksum for the log buffer
+   */
+  static const uint32_t LOG_BUFFER_SUM_SIZE = (1 << 6);
+  /**
+   * The size of the payload in the log buffer
+   */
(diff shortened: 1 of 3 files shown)


GustavoAngulo:
Partially filled buffers is a tricky issue. You need to pass over partially filled buffers during `LogSerializerTask::Process` because transactions may be waiting for their logs to be persisted before committing. 

Because you need to have partially filled buffers, I don't think you'll be able to always write a checksum every 4K bytes. Off the top of my head, theres a couple possibilities for handling checksums regardless of buffer size:
* Write a checksum log record for each buffer. All the record holds is the checksum itself, and the number of bytes immediately following the record that the checksum covers. 
* Write a checksum per log record, from a paper I read I think H-Store actually does this. This is probably way more expensive than the first option.

PR #472, opened 2019-07-25
Normalize catalog names

jrolli:
This PR fixes type sensitivity for catalog items (naming objects and fetching objects by name) by forcing all user-defined input strings to lowercase before proceeding.  To avoid copying the string into an intermediary value, I opted to make the internal varlen field `contents_` mutable but this should not be too risky as the getter methods only expose immutable views.

### DIFF
--- a/src/catalog/catalog_accessor.cpp
+++ b/src/catalog/catalog_accessor.cpp
@@ -5,21 +5,34 @@
 #include "catalog/catalog.h"
 
 namespace terrier::catalog {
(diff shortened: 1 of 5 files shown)


jrolli:
Based off verbal feedback from @mbutrovich, I refactored the code to perform the normalization on "std::string" objects inside of `CatalogAccessor`.  This incurs an extra copy for string-based catalog calls, but better matches the semantics of our API design where the catalog accessor exists to translate user-facing catalog requests into the form expected by the catalog.

mush-zhang:
I had some newer experiments with binder test and begin to wonder if the postgres parser has already parsed all the name to lower case.

mbutrovich:
LGTM.

PR #476, opened 2019-07-29
Switch InsertPlanNode.parameter_info to use std::vector

### LINKED ISSUE #427: Switch InsertPlanNode.parameter_info to use std::vector
`InsertPlanNode.parameter_info` is an `unordered_map`:

https://github.com/cmu-db/terrier/blob/master/src/include/planner/plannodes/insert_plan_node.h#L120

The problem is that when you compute the hash of the node it may generate different results because the parameters are stored in a different order:

https://github.com/cmu-db/terrier/blob/master/src/planner/plannodes/insert_plan_node.cpp#L29-L32

#400 has a fix for this where we put the keys in a sorted map then read them back in order. But to the best of my knowledge, there is no reason that this has to be a map. We can just replace it with a `std::vector`.

LiuXiaoxuanPKU:
This commit tries to fix #427
1. Change parameter_info_ from `std::unordered_map<uint32_t /* value index */, catalog::col_oid_t>` to `std::vecto\<catalog::col_oid_t\>`
2. Change method signature AddParameterInfo from` Builder &AddParameterInfo(uint32_t value_idx, catalog::col_oid_t col_oid)` to `Builder &AddParameterInfo(catalog::col_oid_t col_oid)`. We rely on the caller to push col idx in order.
3. Fix hash function in src/planner/plannodes/drop_view_plan_node.cpp. From my understanding, we should calculate hash on the value (`! view_oid `) instead of on the address (`& view_oid`).

### DIFF
--- a/src/include/planner/plannodes/insert_plan_node.h
+++ b/src/include/planner/plannodes/insert_plan_node.h
@@ -69,7 +69,6 @@ class InsertPlanNode : public AbstractPlanNode {
     }
 
     /**
-     * @param value_idx index of value in values vector
      * @param col_oid oid of column where value at value_idx should be inserted
      * @return builder object
      */
--- a/src/include/planner/plannodes/insert_plan_node.h
+++ b/src/include/planner/plannodes/insert_plan_node.h
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <cstdint>
 #include <memory>
 #include <string>
 #include <tuple>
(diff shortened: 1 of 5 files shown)


src/include/planner/plannodes/insert_plan_node.h
```diff
@@ -69,12 +69,12 @@ class InsertPlanNode : public AbstractPlanNode {
...
 
     /**
-     * @param value_idx index of value in values vector
      * @param col_oid oid of column where value at value_idx should be inserted
      * @return builder object
      */
-    Builder &AddParameterInfo(uint32_t value_idx, catalog::col_oid_t col_oid) {
-      parameter_info_.emplace(value_idx, col_oid);
+    Builder &AddParameterInfo(catalog::col_oid_t col_oid) {
+      // We rely on the caller to push col idx in order
```
GustavoAngulo: nit: Can you add this assumption as a `@warning` in the doxygen comment

PR #481, opened 2019-07-29
Persist columns

### LINKED ISSUE #469: Update JSON (de)serialization for Catalog Objects
Schema and IndexSchema have been updated and need the JSON serialization/deserialization functions updated to reflect the recursion down into AbstractExpression.

### LINKED ISSUE #470: Persist column information into pg_attribute
Finish logic to persist column metadata into pg_attribute.

Blocked on #386.

jrolli:
This PR makes the changes necessary to persist the columns of both tables and indexes by inserting (and hence logging) them into `pg_attribute`.  The design has changed due to constraints on expression ownership so we are no longer using the `ADBIN` column and instead only store the materialized expressions in the corresponding schema.

In order to fully support this, I added the code to properly serialize/deserialize the (index) schema objects.

This PR also fixes the deallocation approach for the database catalog to rely on per table/schema abort deletions rather than trying to reuse the final teardown logic.

The PR also includes the bug fixes to transient value serialization (#480) but does not include the additional test in that PR.

Fixes #469 
Fixes #470

### DIFF
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -749,7 +749,8 @@ table_oid_t DatabaseCatalog::GetTableOid(transaction::TransactionContext *const 
 
 bool DatabaseCatalog::SetTablePointer(transaction::TransactionContext *const txn, const table_oid_t table,
                                       const storage::SqlTable *const table_ptr) {
-  txn->RegisterAbortAction([=]() { delete table_ptr; });
+  auto *tm = txn->GetTransactionManager();
+  txn->RegisterAbortAction([=]() { tm->DeferAction([=]() { delete table_ptr; }); });
   return SetClassPointer(txn, table, table_ptr);
 }
 
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -397,11 +397,6 @@ bool DatabaseCatalog::CreateColumn(transaction::TransactionContext *const txn, c
     return false;
   }
 
-  // Clean up the varlen's buffer in the case it wasn't inlined.
-  if (!name_varlen.IsInlined()) {
-    delete[] name_varlen.Content();
-  }
-
   // Step 3: Insert into oid index
   const auto oid_pri = columns_oid_index_->GetProjectedRowInitializer();
   const auto oid_pm = columns_oid_index_->GetKeyOidToOffsetMap();
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -1635,9 +1635,9 @@ Column DatabaseCatalog::MakeColumn(storage::ProjectedRow *const pr,
   auto col_len = *reinterpret_cast<uint16_t *>(pr->AccessForceNotNull(pr_map.at(ATTLEN_COL_OID)));
   auto col_null = !(*reinterpret_cast<bool *>(pr->AccessForceNotNull(pr_map.at(ATTNOTNULL_COL_OID))));
   auto *col_expr =
-      reinterpret_cast<VarlenEntry *>(pr->AccessForceNotNull(pr_map.at(ADBIN_COL_OID)));
+      reinterpret_cast<storage::VarlenEntry *>(pr->AccessForceNotNull(pr_map.at(ADBIN_COL_OID)));
 
-  auto expr = json::parse(col_expr->StringView()).get<parser::AbstractExpression>();
+  auto expr = nlohmann::json::parse(col_expr->StringView()).get<parser::AbstractExpression>();
 
   std::string name(reinterpret_cast<const char *>(col_name->Content()), col_name->Size());
   Column col = (col_type == type::TypeId::VARCHAR || col_type == type::TypeId::VARBINARY) ? Column(name, col_type, col_len, col_null, *col_expr) : Column(name, col_type, col_null, *col_expr);
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -355,7 +355,8 @@ bool DatabaseCatalog::CreateColumn(transaction::TransactionContext *const txn, c
   auto type_entry = reinterpret_cast<type::TypeId *>(redo->Delta()->AccessForceNotNull(table_pm[ATTTYPID_COL_OID]));
   auto len_entry = reinterpret_cast<uint16_t *>(redo->Delta()->AccessForceNotNull(table_pm[ATTLEN_COL_OID]));
   auto notnull_entry = reinterpret_cast<bool *>(redo->Delta()->AccessForceNotNull(table_pm[ATTNOTNULL_COL_OID]));
-  auto dbin_entry = reinterpret_cast<parser::AbstractExpression **>(redo->Delta()->AccessForceNotNull(table_pm[ADBIN_COL_OID]));
+  auto dbin_entry =
+      reinterpret_cast<parser::AbstractExpression **>(redo->Delta()->AccessForceNotNull(table_pm[ADBIN_COL_OID]));
   auto dsrc_entry =
       reinterpret_cast<storage::VarlenEntry *>(redo->Delta()->AccessForceNotNull(table_pm[ADSRC_COL_OID]));
   *oid_entry = col_oid;
@@ -403,7 +404,7 @@ bool DatabaseCatalog::CreateColumn(transaction::TransactionContext *const txn, c
   pr = oid_pri.InitializeRow(buffer);
   // Write the attributes in the ProjectedRow. These hardcoded indexkeycol_oids come from
   // Builder::GetColumnOidIndexSchema()
-  *(reinterpret_cast<ColOid *>(pr->AccessForceNotNull(oid_pm.at(indexkeycol_oid_t(2))))) =col_oid;
+  *(reinterpret_cast<ColOid *>(pr->AccessForceNotNull(oid_pm.at(indexkeycol_oid_t(2))))) = col_oid;
   *(reinterpret_cast<ClassOid *>(pr->AccessForceNotNull(oid_pm.at(indexkeycol_oid_t(1))))) = class_oid;
 
   bool UNUSED_ATTRIBUTE result = columns_oid_index_->InsertUnique(txn, *pr, tupleslot);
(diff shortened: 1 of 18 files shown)


src/catalog/database_catalog.cpp
```diff
@@ -368,8 +369,8 @@ bool DatabaseCatalog::CreateColumn(transaction::TransactionContext *const txn, c
   *len_entry = (col.Type() == type::TypeId::VARCHAR || col.Type() == type::TypeId::VARBINARY) ? col.MaxVarlenSize()
                                                                                               : col.AttrSize();
   *notnull_entry = !col.Nullable();
-  *dbin_entry = reinterpret_cast<uintptr_t>(
-      nullptr);  // TODO(Matt) implement deep copy of the abstract expression once #386 is in
+  // TODO(John) implement deep copy of the abstract expression and store the pointer once #386 is in
+  *dbin_entry = nullptr;
```
GustavoAngulo: I'm a bit confused, your PR description said we are no longer using the ADBIN column. Do you mean we should deprecate it, or just always set it to nullptr like done here.

src/catalog/database_catalog.cpp
```diff
@@ -550,6 +544,13 @@ bool DatabaseCatalog::DeleteColumns(transaction::TransactionContext *const txn,
         reinterpret_cast<const uint32_t *const>(pr->AccessWithNullCheck(table_pm[ATTNUM_COL_OID]));
     TERRIER_ASSERT(col_name != nullptr, "OID shouldn't be NULL.");
 
+    auto *stored_expression =
+        *reinterpret_cast<parser::AbstractExpression **>(pr->AccessForceNotNull(table_pm[ADBIN_COL_OID]));
```
GustavoAngulo: You're using `ADBIN_COL_OID` here, but it's not among the oids in `table_oids` above.

src/catalog/database_catalog.cpp
```diff
@@ -759,6 +749,8 @@ table_oid_t DatabaseCatalog::GetTableOid(transaction::TransactionContext *const
 
 bool DatabaseCatalog::SetTablePointer(transaction::TransactionContext *const txn, const table_oid_t table,
                                       const storage::SqlTable *const table_ptr) {
+  auto *tm = txn->GetTransactionManager();
+  txn->RegisterAbortAction([=]() { tm->DeferAction([=]() { delete table_ptr; }); });
```
GustavoAngulo: Can you add a comment explaining why we defer this action here, and why the catalog does it? I would expect the caller to check the return value of `SetTablePointer` and clean up the table it created if the call failed.

src/catalog/database_catalog.cpp
```diff
@@ -1027,6 +1019,8 @@ bool DatabaseCatalog::SetClassPointer(transaction::TransactionContext *const txn
 
 bool DatabaseCatalog::SetIndexPointer(transaction::TransactionContext *const txn, const index_oid_t index,
                                       const storage::index::Index *const index_ptr) {
+  auto *tm = txn->GetTransactionManager();
+  txn->RegisterAbortAction([=]() { tm->DeferAction([=]() { delete index_ptr; }); });
```
GustavoAngulo: Same comment as `SetTablePointer`.

src/include/catalog/index_schema.h
```diff
@@ -107,6 +106,38 @@ class IndexSchema {
...
+     * Deserializes a column
+     * @param j serialized column
+     */
+    void FromJson(const nlohmann::json &j) {
+      name_ = j.at("name").get<std::string>();
+      SetTypeId(j.at("type").get<type::TypeId>());
+      SetMaxVarlenSize(j.at("max_varlen_size").get<uint16_t>());
+      SetNullable(j.at("nullable").get<bool>());
+      SetOid(j.at("oid").get<indexkeycol_oid_t>());
+      definition_ = j.at("definition").get<std::shared_ptr<parser::AbstractExpression>>();
```
GustavoAngulo: This line won't work, use the `parser::expression::DeserializeExpression` method.

src/include/catalog/schema.h
```diff
@@ -145,6 +148,7 @@ class Schema {
       max_varlen_size_ = j.at("max_varlen_size").get<uint16_t>();
       nullable_ = j.at("nullable").get<bool>();
       oid_ = j.at("oid").get<col_oid_t>();
+      default_value_ = j.at("default_value").get<std::shared_ptr<parser::AbstractExpression>>();
```
GustavoAngulo: Use `parser::expression::DeserializeExpression`

src/catalog/database_catalog.cpp
```diff
@@ -550,6 +544,13 @@ bool DatabaseCatalog::DeleteColumns(transaction::TransactionContext *const txn,
         reinterpret_cast<const uint32_t *const>(pr->AccessWithNullCheck(table_pm[ATTNUM_COL_OID]));
     TERRIER_ASSERT(col_name != nullptr, "OID shouldn't be NULL.");
 
+    auto *stored_expression =
+        *reinterpret_cast<parser::AbstractExpression **>(pr->AccessForceNotNull(table_pm[ADBIN_COL_OID]));
+
+    if (stored_expression != nullptr) {
+      txn->RegisterCommitAction([=]() { delete stored_expression; });
```
GustavoAngulo: Who cleans `stored_expression` up in the case of an abort? We should probably register a Commit and Abort action.

src/include/catalog/database_catalog.h
```diff
@@ -265,7 +269,7 @@ class DatabaseCatalog {
    * @return the column from pg_attribute
    */
   template <typename Column, typename ClassOid, typename ColOid>
-  std::vector<std::unique_ptr<Column>> GetColumns(transaction::TransactionContext *txn, ClassOid class_oid);
+  std::vector<Column> GetColumns(transaction::TransactionContext *txn, ClassOid class_oid);
   // TODO(Matt): make this return stack object
```
GustavoAngulo: nit: You addressed this TODO

src/include/catalog/database_catalog.h
```diff
@@ -253,8 +257,8 @@ class DatabaseCatalog {
    * @param class_oid oid of table or index
    * @return the column from pg_attribute
    */
-  template <typename Column, typename ClassOid, typename ColOid>
-  std::unique_ptr<Column> GetColumn(transaction::TransactionContext *txn, ClassOid class_oid, ColOid col_oid);
+  // template <typename Column, typename ClassOid, typename ColOid>
+  // Column GetColumn(transaction::TransactionContext *txn, ClassOid class_oid, ColOid col_oid);
```
GustavoAngulo: Why is this commented out?

jrolli: As best as I could tell, there is no actual use-case for materializing a single column so I commented it out during the update.  I've now completely removed it.

src/catalog/database_catalog.cpp
```diff
@@ -759,6 +697,11 @@ table_oid_t DatabaseCatalog::GetTableOid(transaction::TransactionContext *const
 
 bool DatabaseCatalog::SetTablePointer(transaction::TransactionContext *const txn, const table_oid_t table,
                                       const storage::SqlTable *const table_ptr) {
+  auto *tm = txn->GetTransactionManager();
```
mbutrovich: Style nit: refer `txn_manager` over `tm`. I suspect this is actually all over the catalog.

src/include/catalog/index_schema.h
```diff
@@ -116,8 +161,9 @@ class IndexSchema {
     std::string name_;
     indexkeycol_oid_t oid_;
     uint32_t packed_type_;
-    common::ManagedPointer<const parser::AbstractExpression> definition_;
-    std::string serialized_expression_;
+
+    // TODO(John) this should go back to being a raw pointer once #386 is in
```
mbutrovich: Outdated TODO. Not sure if we know what PR to reference at this point.

src/catalog/database_catalog.cpp
```diff
@@ -1516,12 +1458,32 @@ bool DatabaseCatalog::CreateTableEntry(transaction::TransactionContext *const tx
...
+  *(reinterpret_cast<namespace_oid_t *>(index_pr->AccessForceNotNull(0))) = ns_oid;
   const auto result UNUSED_ATTRIBUTE = classes_namespace_index_->Insert(txn, *index_pr, tuple_slot);
   TERRIER_ASSERT(result, "Insertion into non-unique namespace index failed.");
 
   delete[] index_buffer;
 
+  // Write the col oids into a new Schema object
+  col_oid_t curr_col_oid(1);
+  for (auto &col : schema.GetColumns()) {
+    auto result = CreateColumn(txn, table_oid, curr_col_oid++, col);
```
mbutrovich: Shadows another `result` variable.

src/catalog/database_catalog.cpp
```diff
@@ -1310,6 +1231,27 @@ bool DatabaseCatalog::CreateIndexEntry(transaction::TransactionContext *const tx
   // Free the buffer, we are finally done
   delete[] index_buffer;
 
+  // Write the col oids into a new Schema object
+  indexkeycol_oid_t curr_col_oid(1);
+  for (auto &col : schema.GetColumns()) {
+    auto result = CreateColumn(txn, index_oid, curr_col_oid++, col);
```
mbutrovich: Shadows another `result` variable.

src/include/catalog/schema.h
```diff
@@ -154,7 +176,9 @@ class Schema {
     uint16_t max_varlen_size_;
     bool nullable_;
     col_oid_t oid_;
-    common::ManagedPointer<const parser::AbstractExpression> default_value_;
+
+    // TODO(John) this should go back to being a raw pointer once #386 is in
```
mbutrovich: Same issue with an outdated PR reference.

src/include/catalog/index_schema.h
jrolli: Changed to reference the new issue for smart pointers (#489).

src/include/catalog/schema.h
jrolli: Changed to reference the new issue for smart pointers (#489).

jrolli:
I addressed the lack of testing for `DeleteColumns` by making the deletions in `pg_class` cascade into `pg_attributes` which should address that oversight.

The current shortcomings for the catalog testing is testing behavior under unexpected, but non-fatal, API calls.

There are also additional behavior tests we will want to add as we implement features such as cascading deletes for namespaces and tables, but overall I think we are testing most of the known behavior.

PR #483, opened 2019-07-30
Expressions-only version of 386, breaks plan nodes

lmwnshn:
Separated out into a new PR, the old one fell quite far behind. Goal is to pull a couple of all nighters in the next couple of days and have this in by Friday.

Converts just expressions to use managed pointers. Temporarily disables plan nodes entirely, and plan nodes are not in scope for this PR, my understanding @jrolli is that just expressions are enough to unblock #470?

Questions remaining (Slack had some more context):
- [ ] For going from (Parent) to (More specific class), e.g. (SQLStatement -> SelectStatement), I don't think you can eliminate CastManagedPointerTo like @jrolli wanted?
- [ ] DeserializeJSON creates things which have managed pointers. If we eliminate .get() per @mush-zhang @GustavoAngulo ?, I don't see how you can clean this up without refactoring to pass in some deferred action magic thing to the expression classes.

Reviewer suggestion:
- Every time I create a raw pointer, make sure it gets added to the ParseResult... every couple of git diff passes I catch a new one leaking. Thinking about pairing up these with code comments like malloclab..

Work remaining:
- [x] Pull in my commits
- [x] Compiles terrier target
- [ ] Pull in Gus commits (I saw some smaller bug fixes, etc) -- goal Wednesday
- [ ] Compiles parser test targets, i.e. fix parser tests to use ParseResult and cleanup -- goal Thursday
- [ ] Check for memory leaks -- hopefully valgrind will help; see questions, some will definitely leak
- [ ] CI (format, tidy, lint, doxygen -- right now Windows isn't letting me do any of that) -- goal Friday

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -4,6 +4,13 @@
 
 file(GLOB_RECURSE TERRIER_SRCS ${PROJECT_SOURCE_DIR}/src/*.cpp ${PROJECT_SOURCE_DIR}/src/include/*.h)
 list(REMOVE_ITEM TERRIER_SRCS ${PROJECT_SOURCE_DIR}/src/main/terrier.cpp)
+set (EXCLUDE_DIR "/planner/")
+foreach (TMP_PATH ${TERRIER_SRCS})
+    string (FIND ${TMP_PATH} ${EXCLUDE_DIR} EXCLUDE_DIR_FOUND)
+    if (NOT ${EXCLUDE_DIR_FOUND} EQUAL -1)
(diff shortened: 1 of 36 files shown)


jrolli:
Talking through the API and design yesterday with @mbutrovich, we managed to unblock #470 in such a way that we hide the shared_ptr problem from the API.  Given the ongoing work on the binder and optimizer, it may be better to not do an intermediate step, but @mush-zhang and @17zhangw probably have a better perspective on that.

lmwnshn:
(Closed because I accidentally branched against this repo, wrong git origin.)

PR #485, opened 2019-07-31
Hash index based on libcuckoo

mbutrovich:
This implements a hash index based on [libcuckoo](https://github.com/efficient/libcuckoo) with a new key type. The tests and benchmarks for the bwtree are applied to the new index as well, with ordered scans removed of course. TPC-C microbenchmark is modified to use the hash index when it makes sense. This ends up 7 of the 10 indexes that don't need ordered scans.

There are docs within the code as well, but at a high level: the cuckoohash_map is not a multi map, so I rely on std::variant to pull it off.

I also simplified a bit of the index metadata now that we have a catalog. Stuff that is persisted in separate columns in the new catalog are no longer fields in IndexSchema. Also, outdated fields in the storage data structures themselves like oids are removed.

Other refactoring: CompactIntsKey's size argument is now in bytes instead of 8-byte slots so it's consistent with GenericKey and the new HashKey. This required some test refactoring. I also generalized a little bit more of the test code to exercise the new key type. I spotted some bugs in the tests in the process and fixed those up.

Third party stuff:
- latest version of libcuckoo, with one modification to use TBB's spinlock instead of their own rolled version
- latest version of xxHash3, aligns with what's used in #398. Not sure which PR will get in first but they should be the same files.


src/include/storage/index/hash_index.h
```diff
@@ -0,0 +1,272 @@
...
+
+    const bool UNUSED_ATTRIBUTE uprase_result = hash_map_->uprase_fn(index_key, key_found_fn, location);
+
+    TERRIER_ASSERT(insert_result != uprase_result,
+                   "Either a new key was inserted (uprase_result), or the value already existed and a new value was "
+                   "inserted (insert_result).");
+
+    // Register an abort action with the txn context in case of rollback
+    txn->RegisterAbortAction([=]() {
+      auto abort_key_found_fn = [location](ValueType &value) -> bool {
```
jrolli: A doxygen-like comment should precede this given the general complexity of this abort action (defering a capturing labmda that passes a capturing lambda to a third-party library).

src/include/storage/index/hash_index.h
```diff
@@ -0,0 +1,272 @@
...
+          TERRIER_ASSERT(value_map.count(location) == 1, "location must be the only value in the ValueMap.");
+          return true;
+        }
+        // ValueMap contains multiple elements, erase the element for location
+        const auto UNUSED_ATTRIBUTE erase_result = value_map.erase(location);
+        TERRIER_ASSERT(erase_result == 1, "Erasing from the ValueMap should not fail.");
+        return false;
+      };
+
+      const bool UNUSED_ATTRIBUTE uprase_result = hash_map_->uprase_fn(index_key, abort_key_found_fn);
```
jrolli: You should have a comment before this line that (both visually and conceptually) describes how the lambda above gets passed to the library.  Without the comment it is way too easy to miss the fact that there is a lambda defined above.

src/include/storage/index/hash_index.h
```diff
@@ -0,0 +1,272 @@
...
+    auto predicate = [txn](const TupleSlot slot) -> bool {
+      const auto *const data_table = slot.GetBlock()->data_table_;
+      const auto has_conflict = data_table->HasConflict(*txn, slot);
+      const auto is_visible = data_table->IsVisible(*txn, slot);
+      return has_conflict || is_visible;
+    };
+
+    bool UNUSED_ATTRIBUTE insert_result = false;
+
+    auto key_found_fn = [location, &insert_result, &predicate_satisfied, predicate](ValueType &value) -> bool {
```
jrolli: See above (doxygen-like comment)

src/include/storage/index/hash_index.h
```diff
@@ -0,0 +1,272 @@
...
+          insert_result = value_map.emplace(location).second;
+          TERRIER_ASSERT(insert_result,
+                         " index shouldn't fail to insert after predicate check. If it did, something went wrong deep "
+                         "inside the hash map itself.");
+        }
+      }
+      return false;
+    };
+
+    const bool UNUSED_ATTRIBUTE uprase_result = hash_map_->uprase_fn(index_key, key_found_fn, location);
```
jrolli: See above (separation from complex lambda)

src/include/storage/index/hash_index.h
```diff
@@ -0,0 +1,272 @@
...
+
+    TERRIER_ASSERT(predicate_satisfied || (insert_result != uprase_result),
+                   "Either a new key was inserted (uprase_result), or the value already existed and a new value was "
+                   "inserted (insert_result).");
+
+    const bool UNUSED_ATTRIBUTE overall_result = insert_result || uprase_result;
+
+    if (overall_result) {
+      txn->RegisterAbortAction([=]() {
+        auto abort_key_found_fn = [location](ValueType &value) -> bool {
```
jrolli: Maybe this lambda should be turned into a macro within this file.  It's the same logic that has to be inlined so I think only a macro would let you write it once.

src/include/storage/index/hash_index.h
```diff
@@ -0,0 +1,272 @@
...
+    const bool UNUSED_ATTRIBUTE find_result = hash_map_->find_fn(index_key, key_found_fn);
+
+    TERRIER_ASSERT(GetConstraintType() == ConstraintType::DEFAULT ||
+                       (GetConstraintType() == ConstraintType::UNIQUE && value_list->size() <= 1),
+                   "Invalid number of results for unique index.");
+  }
+
+  void ScanAscending(const transaction::TransactionContext &txn, const ProjectedRow &low_key,
+                     const ProjectedRow &high_key, std::vector<TupleSlot> *value_list) final {
+    TERRIER_ASSERT(false, "Range scans not supported on HashIndex.");
```
jrolli: These (and the following asserts) may need to be exceptions or something stronger.  Can we guarantee that a user cannot force us down this path?  I would like to get there but there is a lot of logic we haven't written to ensure this.

mbutrovich:
Oh this wasn't ready for review at all, but thanks for taking a pass! I'll go through your comments in time.

third_party/farmhash/farmhash.h
```diff
@@ -0,0 +1,321 @@
...
+// WARNING: This code has been only lightly tested on big-endian platforms!
+// It is known to work well on little-endian platforms that have a small penalty
+// for unaligned reads, such as current Intel and AMD moderate-to-high-end CPUs.
+// It should work on all 32-bit and 64-bit platforms that allow unaligned reads;
+// bug reports are welcome.
+//
+// By the way, for some hash functions, given strings a and b, the hash
+// of a+b is easily derived from the hashes of a and b.  This property
+// doesn't hold for any hash functions in this file.
+
```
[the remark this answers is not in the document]
mbutrovich: I mentioned the changes I made in versions.txt. If there's a better way to track that stuff we can start doing that.

third_party/versions.txt
```diff
@@ -29,3 +29,17 @@ GBENCHMARK_VERSION=1.4.1
 # branch: v1.x
 # commit hash: 8179b26388d118fa887030f2609560ec287531dd
 # commit hash date: 7 Aug 2018
+
+# farmhash
+# url: https://github.com/google/farmhash.git
```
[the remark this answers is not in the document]
mbutrovich: I have no idea how to set that up, but seems like a good idea. Can you commit to those sub-repositories or are they read-only?

src/include/storage/index/hash_index.h
mbutrovich: Yeah I'm considering how macro-heavy I want to go with this stuff. If I define them all in index.h I can use them in both derived index wrappers, but using macros feels so C99.

src/include/storage/index/compact_ints_key.h
```diff
@@ -4,7 +4,7 @@
 #include <functional>
 #include <vector>
 
-#include "common/hash_util.h"
+#include "farmhash/farmhash.h"
```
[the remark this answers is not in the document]
mbutrovich: Undecided.

src/include/storage/index/index_builder.h
```diff
@@ -23,6 +24,7 @@ class IndexBuilder {
   catalog::index_oid_t index_oid_{0};
   ConstraintType constraint_type_ = ConstraintType::INVALID;
   catalog::IndexSchema key_schema_;
+  bool ordered_ = true;
```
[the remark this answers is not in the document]
mbutrovich: Sure. Do we want it in IndexSchema (and eventually persisted in the catalog), or just an argument set on the IndexBuilder?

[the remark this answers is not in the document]
jrolli: It probably should be in IndexSchema with the goal of eventual persistence (`pg_index`?) since we would want to ensure replicas stay in sync with the master.

third_party/versions.txt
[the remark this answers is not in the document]
jrolli: Probably the best way to do what you describe is to fork the upstream projects in Github and then replace the appropriate folders here with git submodules.  That should allow us to periodically pull & merge upstream changes in a fashion that is isolated from terrier.  I'll try to give this an attempt and write it up if it seems to work.

test/storage/hash_index_test.cpp
```diff
@@ -0,0 +1,1645 @@
...
+  txn_manager_.Commit(scan_txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+}
+
+/**
+ * This test creates multiple worker threads that all try to insert [0,num_inserts) as tuples in the table and into the
+ * primary key index. At completion of the workload, all num_inserts_ txns * num_threads_ should have committed with
+ * visible versions in the index and table.
+ */
+// NOLINTNEXTLINE
+TEST_F(HashIndexTests, DefaultInsert) {
```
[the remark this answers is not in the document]
mbutrovich: Side effect of our index ConstraintType currently being DEFAULT and UNIQUE. We might need to get the interested parties in a room (not many people right now) to codify where our index metadata is living and the terminology being used, since all of this pre-dates the Postgres catalog.

.travis.yml
```diff
@@ -4,7 +4,7 @@ matrix:
   fast_finish: true
   include:
     - os: osx
-      osx_image: xcode10.1
+      osx_image: xcode10.2
```
mbutrovich: Needed for std::variant.

mbutrovich:
See updated PR description for overview. This is now ready for review. I was spurred to get this production-ready by the need for add/drop index to reason about index metadata that pre-dated our catalog, so I wanted to get that cleaned up.

This PR is good for about 15-20% improved performance on our TPC-C microbenchmark last time I peaked at a Jenkins run.

For review, if @jrolli has time that's great, but mostly want @yeshengm to take a pass and re-familiarize himself with the indexing layer.

src/include/storage/index/hash_key.h
```diff
@@ -0,0 +1,108 @@
...
+   * @param from ProjectedRow to generate HashKey representation of
+   * @param metadata index information, primarily attribute sizes and the precomputed offsets to translate PR layout to
+   * HashKey
+   */
+  void SetFromProjectedRow(const storage::ProjectedRow &from, const IndexMetadata &metadata) {
+    std::memset(key_data_, 0, key_size_byte);
+
+    const auto key_size = metadata.KeySize();
+
+    // NOLINTNEXTLINE (Matt): tidy thinks this has side-effects. I disagree.
```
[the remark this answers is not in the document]
mbutrovich: Not sure if a macro makes sense. Should I document it somewhere on the wiki maybe?

mbutrovich:
Also this PR now changes 17 of the catalog's 18 indexes to be hash maps instead of bwtrees. We'll compare the performance difference once #545 goes in.

mbutrovich:
Also now fixes #548.

mbutrovich:
Performance numbers from my laptop:
#545:
```
CatalogBenchmark/GetAccessor               3.12214M items/s
CatalogBenchmark/GetDatabaseOid            3.03497M items/s
CatalogBenchmark/GetDatabaseCatalog        3.5217M items/s
CatalogBenchmark/GetIndex                  3.05035M items/s
CatalogBenchmark/GetIndexOid               1.28681M items/s
CatalogBenchmark/GetIndexOids              3.20876M items/s
CatalogBenchmark/GetIndexSchema            2.55408M items/s
CatalogBenchmark/GetNamespaceOid           1.81154M items/s
CatalogBenchmark/GetSchema                 1.95754M items/s
CatalogBenchmark/GetTable                  2.94729M items/s
CatalogBenchmark/GetTableOid               1.25653M items/s
```
This PR:
```
CatalogBenchmark/GetAccessor               3.54384M items/s
CatalogBenchmark/GetDatabaseOid            3.34993M items/s
CatalogBenchmark/GetDatabaseCatalog        3.99479M items/s
CatalogBenchmark/GetIndex                  3.42306M items/s
CatalogBenchmark/GetIndexOid               1.73293M items/s
CatalogBenchmark/GetIndexOids              3.56294M items/s
CatalogBenchmark/GetIndexSchema            2.72928M items/s
CatalogBenchmark/GetNamespaceOid           2.21597M items/s
CatalogBenchmark/GetSchema                 2.13332M items/s
CatalogBenchmark/GetTable                  3.33497M items/s
CatalogBenchmark/GetTableOid               1.7926M items/s
```
Combined with #546 (which appears to be more impactful, at least on my machine) I'm hoping that @GustavoAngulo's performance issues hammering the catalog in recovery are mostly resolved, even without the CatalogAccessor cache.

src/catalog/database_catalog.cpp
```diff
@@ -1269,12 +1269,15 @@ bool DatabaseCatalog::CreateIndexEntry(transaction::TransactionContext *const tx
       pg_index_all_cols_prm_[postgres::INDISEXCLUSION_COL_OID]))) = schema.is_exclusion_;
   *(reinterpret_cast<bool *>(indexes_insert_pr->AccessForceNotNull(
       pg_index_all_cols_prm_[postgres::INDIMMEDIATE_COL_OID]))) = schema.is_immediate_;
+  // TODO(Matt): these should actually be set later based on runtime information about the index. @yeshengm
```
yeshengm: Sure. It's related to the pg_index & pg_indexes catalog.

src/include/storage/index/index.h
```diff
@@ -57,19 +45,23 @@ class Index {
...
+   * @return type of the index. Note that this is the physical type, not extracted from the underlying schema or other
+   * catalog metadata. This is mostly used for debugging purposes.
+   */
+  virtual IndexType Type() const = 0;
+
   /**
    * Invoke garbage collection on the index. For some underlying index types this may be a no-op.
    */
-  virtual void PerformGarbageCollection() = 0;
+  virtual void PerformGarbageCollection() {}
```
yeshengm: We can still keep it as a pure virtual function?

src/include/storage/index/bwtree_index.h
```diff
@@ -20,19 +22,18 @@ class BwTreeIndex final : public Index {
...
 
  private:
-  BwTreeIndex(const catalog::index_oid_t oid, const ConstraintType constraint_type, IndexMetadata metadata)
-      : Index(constraint_type, std::move(metadata)),
-        bwtree_{new third_party::bwtree::BwTree<KeyType, TupleSlot>{false}} {}
+  explicit BwTreeIndex(IndexMetadata metadata)
+      : Index(std::move(metadata)), bwtree_{new third_party::bwtree::BwTree<KeyType, TupleSlot>{false}} {}
 
-  third_party::bwtree::BwTree<KeyType, TupleSlot> *const bwtree_;
+  const std::unique_ptr<third_party::bwtree::BwTree<KeyType, TupleSlot>> bwtree_;
```
yeshengm: good raii style!

src/include/storage/index/index.h
mbutrovich: I guess it depends on which we believe is more common for future index types. I actually believe we won't have to invoke `PerformGarbageCollection` on future indexes, so I leave it a no-op in the base class and leave it up to the derived class to override if needed. If, in the future, we find that most index types do need it, we can change it back to pure virtual and have the hash_index override it as a no-op.

src/include/storage/index/index.h
```diff
@@ -116,7 +108,9 @@ class Index {
    * @param[out] value_list the values associated with the keys
    */
   virtual void ScanAscending(const transaction::TransactionContext &txn, const ProjectedRow &low_key,
-                             const ProjectedRow &high_key, std::vector<TupleSlot> *value_list) = 0;
+                             const ProjectedRow &high_key, std::vector<TupleSlot> *value_list) {
```
yeshengm: Same here.

mbutrovich: Similar to GC: not all indexes will support this operation. Pure virtual forces the derived types to implement it when they might not need to.

src/include/storage/index/hash_index.h
```diff
@@ -0,0 +1,269 @@
...
+          // replace the TupleSlot with a ValueMap containing both the old and new inserted value
+          value = ValueMap({{location}, {existing_location}}, 2);
+          insert_result = true;
+        }
+      } else {
+        auto &value_map = std::get<ValueMap>(value);
+
+        // Check all of the existing TupleSlots at this key for conflicts
+        for (const auto i : value_map) {
+          predicate_satisfied = predicate_satisfied || predicate(i);
```
yeshengm: Costly? Consider `std::any`/`std::all`?

mbutrovich: They're all O(n), but if it helps readability I can try to make it use the STL's stuff from <algorithm>

yeshengm: Yep, but algorithm utils should have short-circuiting stuff by default.

src/include/storage/index/compact_ints_key.h
```diff
@@ -4,17 +4,17 @@
 #include <functional>
 #include <vector>
 
-#include "common/hash_util.h"
 #include "portable_endian/portable_endian.h"
 #include "storage/index/index_metadata.h"
 #include "storage/projected_row.h"
 #include "storage/storage_defs.h"
+#include "xxHash/xxh3.h"
```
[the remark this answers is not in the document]
mbutrovich: I think that's a larger issue to consolidate the hashing stuff. There are wrappers in both `include/common` and `include/execution/util` that should be consolidated so as to present a consistent API throughout the codebase. I'm punting that from this PR. I could use the wrapper that came in with TPL, but that would just need to be refactored in the future anyway.

src/catalog/database_catalog.cpp
mbutrovich: The comment is referring to the fields below: live, ready, valid.

src/include/storage/index/index.h
```diff
@@ -168,6 +162,11 @@ class Index {
    * @return projected row initializer for the given key schema
    */
   const ProjectedRowInitializer &GetProjectedRowInitializer() const { return metadata_.GetProjectedRowInitializer(); }
+
+  /**
+   * @return IndexKeyKind selected by the IndexBuilder at index construction
```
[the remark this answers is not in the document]
mbutrovich: Right now, very deterministic as long as the schema is the same.

src/include/storage/index/index_builder.h
```diff
@@ -32,46 +31,26 @@ class IndexBuilder {
...
-   * @return the builder object
-   */
-  IndexBuilder &SetConstraintType(const ConstraintType constraint_type) {
-    constraint_type_ = constraint_type;
-    return *this;
+    if (key_schema_.Type() == IndexType::BWTREE) {
+      if (simple_key && metadata.KeySize() <= COMPACTINTSKEY_MAX_SIZE) return BuildBwTreeIntsKey(std::move(metadata));
+      return BuildBwTreeGenericKey(std::move(metadata));
+    }
+    // IndexType must be HASHMAP, since that's the only other option at the moment
```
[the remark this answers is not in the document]
mbutrovich: No preference. I can refactor it now.

src/include/storage/index/hash_key.h
```diff
@@ -0,0 +1,112 @@
...
+    TERRIER_ASSERT(std::invoke([&]() -> bool {
+                     for (const auto &i : metadata.GetSchema().GetColumns()) {
+                       if (i.Nullable()) return false;
+                     }
+                     return true;
+                   }),
+                   "There should not be any NULL attributes in this schema.");
+
+    // This assumes no padding between attributes, no good way to assert that right now. We'll rely on the IndexBuilder
+    // to properly vet the IndexSchema.
```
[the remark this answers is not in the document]
mbutrovich: Sum of attribute sizes would always be greater than the pointer distances due to the size of the last attribute, right? It's _an_ assertion, but a bit weak to me. I'm trying to think of what I can do with all of the IndexMetadata (including KeySchema, ProjectedRowInitializer, etc.) but it all sounds a bit ugly.

src/include/storage/index/compact_ints_key.h
```diff
@@ -67,16 +63,34 @@ class CompactIntsKey {
...
+    // NOLINTNEXTLINE (Matt): tidy thinks this has side-effects. I disagree.
+    TERRIER_ASSERT(std::invoke([&]() -> bool {
+                     for (const auto &i : metadata.GetSchema().GetColumns()) {
+                       if (i.Nullable()) return false;
+                     }
+                     return true;
+                   }),
+                   "There should not be any NULL attributes in this schema.");
+
+    std::memset(key_data_, 0, KeySize);
```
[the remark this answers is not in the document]
mbutrovich: Done.

mbutrovich:
My only hesitation with merging this right now is that it took a few attempts to get coverage to finish in time. The passing run finished with about a minute to spare.

jrolli:
It would lengthen the overall CI process, but we could split the code coverage into multiple disjoint runs (like C++ tests and TPL tests).  This is definitely not “efficient” but it might push the length of a single run low enough.

PR #486, opened 2019-07-31
Search path fix

### LINKED ISSUE #473: Fix search path resolution in CatalogAccessor
The current CatalogAccessor does not fully handle search paths properly.  Specifically, if `pg_catalog` is not in the explicit search path, then it is implicitly searched first when doing lookups (but is not the default namespace when creating objects).  However, if it is in the search path then it  is not searched first.

jrolli:
Updates the accessor to correctly handle the implicit placement of `pg_catalog` namespace in the search path when it is not explicitly set.  Specifically, PostgreSQL will always search `pg_catalog` first when resolving an unqualified object name unless `pg_catalog` was explicitly placed somewhere in the search path.  For unqualified object creation, the default namespace that is used is the first namespace in the explicit path.

Fixes #473

### DIFF
--- a/src/catalog/catalog_accessor.cpp
+++ b/src/catalog/catalog_accessor.cpp
@@ -32,8 +32,7 @@ void CatalogAccessor::SetSearchPath(std::vector<namespace_oid_t> namespaces) {
   search_path_.clear();
   search_path_.reserve(namespaces.size() + 1);
   search_path_.emplace_back(NAMESPACE_CATALOG_NAMESPACE_OID);
-  for (auto ns : namespaces)
-    search_path_.emplace_back(ns);
+  for (auto ns : namespaces) search_path_.emplace_back(ns);
 }
 
 namespace_oid_t CatalogAccessor::GetNamespaceOid(std::string name) {
--- a/src/catalog/catalog_accessor.cpp
+++ b/src/catalog/catalog_accessor.cpp
@@ -24,7 +24,7 @@ void CatalogAccessor::SetSearchPath(std::vector<namespace_oid_t> namespaces) {
   search_path_ = std::move(namespaces);
 
     // Check if 'pg_catalog is explicitly set'
-  for (auto ns : namespaces) {
+  for (auto &ns : namespaces) {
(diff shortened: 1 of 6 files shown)


src/catalog/catalog_accessor.cpp
```diff
@@ -17,7 +17,23 @@ db_oid_t CatalogAccessor::CreateDatabase(std::string name) {
...
+    // Check if 'pg_catalog is explicitly set'
+    if (ns == NAMESPACE_CATALOG_NAMESPACE_OID) {
+      search_path_ = std::move(namespaces);
+      return;
+    }
+  }
+
+  search_path_.clear();
+  search_path_.reserve(namespaces.size() + 1);
+  search_path_.emplace_back(NAMESPACE_CATALOG_NAMESPACE_OID);
```
mbutrovich: You'd get equivalent behavior from...
```
search_path_ = std::move(namespaces);
search_path_.emplace(search_path_ .begin(), NAMESPACE_CATALOG_NAMESPACE_OID);
```
... right? Complexity is the same, but the behavior is more clear to me. Might just be me though, not an explicit change request.

src/include/catalog/catalog_accessor.h
```diff
@@ -265,6 +265,7 @@ class CatalogAccessor {
   transaction::TransactionContext *txn_;
```
mbutrovich: Propose const-qualifying catalog_, dbc_, txn_, and db_oid_.

jrolli: Changing `catalog_` and `dbc_` required adding const qualifiers to more functions but was doable.  However, we cannot const qualify `txn_` because it is modified by transactions that write into the catalog (both through adding actions to the transaction's queue and creating redo records).

mbutrovich: Oh, I meant making the pointer const. transaction::TransactionContext *const txn_;

mbutrovich:
Jenkins failure is due to a huge drop in the DataTable benchmarks, which this PR doesn't touch. Something weird with those lately...

PR #491, opened 2019-08-01
CMake cleanup, jemalloc flag

mbutrovich:
It's been a year since this CMake stuff was written and was in need of a cleanup pass while I try to understand the jemalloc linking issue. Here's what I tried to do with my limited understanding of CMake:
- Simplify the library linking and dependency declarations. This should hopefully resolve the missing gflags race for people once and for all. We work around this in CI by explicitly making the gflags, google test, and gbenchmark targets first. I've removed those lines to see if the issue pops back up now. My testing on local machines seems to indicate that they always get build first now any time you build a target that depends on them. I also removed redundant linking and dependency declarations.
- There's a CMake variable to link jemalloc: DTERRIER_USE_JEMALLOC. Set it to ON to dynamically link the jemalloc library.
- Grabbed a newer, more exhaustive version of Findjemalloc.cmake from Apache Arrow in case the issues linking were related to a non-standard path. Modified it to be a little more verbose.
- removed the --no-as-needed jemalloc linker flag that was wrapped in an `if (LINUX)` for two reasons. First, that conditional is invalid in CMake. That's not how you check if you're on Linux, so it was never evaluating to true and appending the flag anyway. Second, it's not clear that flag is needed for jemalloc linking anyway. I tested on my 18.04 box and it links just fine. If people have trouble in the future we can look into it more.


GustavoAngulo:
Works fine on my machine. MacOS Mojave 10.14.4.

cmake output:
```
-- Terrier version: 0.1.0 (full: '0.1.0-prototype')
clang-tidy found at /usr/local/opt/llvm@6/bin/clang-tidy
clang-format found at /usr/local/opt/llvm@6/bin/clang-format
INFOCompiler command: /usr/local/Cellar/llvm@6/6.0.1/bin/clang++
INFOCompiler version: clang version 6.0.1 (tags/RELEASE_601/final)
Target: x86_64-apple-darwin18.5.0
Thread model: posix
InstalledDir: /usr/local/Cellar/llvm@6/6.0.1/bin
```

PR #496, opened 2019-08-03
Jenkins: Enable jemalloc for benchmark build on Jenkins

mbutrovich:
This uses the new CMake flag to dynamically link jemalloc only for the benchmark build on Jenkins. Assuming the numbers make sense and it appears to be working correctly, we'll update the nightly build pipeline and reset the benchmark lookback next week.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -75,7 +75,7 @@ pipeline {
                     steps {
                         sh 'echo y | sudo ./script/installation/packages.sh'
                         sh 'mkdir build'
-                        sh 'cd build && cmake -DCMAKE_BUILD_TYPE=Release -DTERRIER_USE_ASAN=OFF .. && make -j4'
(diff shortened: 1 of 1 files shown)


mbutrovich:
Select highlights from the benchmark suite:
```
pass 1.187e+06 9.007e+05 10     +31    history  10   GarbageCollectorBenchmark    UnlinkTime
pass 5.874e+06 3.398e+06 10     +72    history  10   GarbageCollectorBenchmark    ReclaimTime
pass 6.562e+04 5.401e+04 10     +21    history  10   TPCCBenchmark                ScaleFactor4WithLogging
pass 6.794e+04 5.716e+04 10     +18    history  10   TPCCBenchmark                ScaleFactor4WithoutLogging
```
The UnlinkTime benchmark doesn't make sense to me and makes me wonder if the data structures used to store the txns are internally using malloc and free frequently. They're forward_lists, which would make sense.

PR #501, opened 2019-08-09
LLVM 8 and accompanying CI checks

mbutrovich:
This bumps the LLVM toolchain to 8.0 in advance of merging #398. It also bumps the accompanying tools (clang-tidy, clang-format) to 8.0 as well, and includes fixes for some of the new tidy checks. I did disable some of the new checks, and will leave comments next to each in the PR on what to do with them going forward.


.clang-tidy
```diff
@@ -26,9 +26,18 @@ Checks:     '
             performance-*,
             portability-*,
             readability-*,
+            -bugprone-exception-escape,
```
mbutrovich: Suggest reenabling in the future as it seems to find a few legit issues with destructors that can throw. Will open an issue.

.clang-tidy
```diff
@@ -26,9 +26,18 @@ Checks:     '
...
+            -bugprone-macro-parentheses,
+            -bugprone-too-small-loop-variable,
```
mbutrovich: It complains about uint uin8_t or uint16_t when the limit on the loop is a container's .size() (size_t). We usually do this when we know the maximum size of the container though, so propose leaving disabled unless someone wants to refactor all of those for loops. Seems unnecessary.

.clang-tidy
```diff
@@ -26,9 +26,18 @@ Checks:     '
...
+            -clang-analyzer-cplusplus.NewDelete,
```
mbutrovich: Seems to generate false positives. Suggest relying on ASAN and valgrind for leak checks.

.clang-tidy
```diff
@@ -26,9 +26,18 @@ Checks:     '
...
+            -clang-analyzer-cplusplus.NewDeleteLeaks,
```
mbutrovich: Seems to generate false positives. Suggest relying on ASAN and valgrind for leak checks.

.clang-tidy
```diff
@@ -26,9 +26,18 @@ Checks:     '
...
             -google-readability-braces-around-statements,
-            -google-readability-namespace-comments,
+            -modernize-use-nodiscard,
```
mbutrovich: Don't fully understand this one, but seems to blow up all over the codebase. We can have a discussion on whether it's worth enforcing this and someone going through and copy-pasting [[nodiscard]] but seems like a waste of time.

.clang-tidy
```diff
@@ -26,9 +26,18 @@ Checks:     '
...
+            -modernize-avoid-c-arrays,
```
mbutrovich: This isn't really feasible for our codebase. Leave disabled.

.clang-tidy
```diff
@@ -26,9 +26,18 @@ Checks:     '
...
+            -performance-unnecessary-value-param,
```
mbutrovich: This actually found some good issues that I fixed, however it also blows up all over plan nodes with how shared_ptrs are used. Suggest reenabling once shared_ptrs are done being ripped out (@lmwnshn).

.clang-tidy
```diff
@@ -26,9 +26,18 @@ Checks:     '
...
             -readability-braces-around-statements,
+            -readability-isolate-declaration,
```
mbutrovich: Doesn't let you do:
`uint8_t x, y;`
I don't really like that, but am open to a conversation about it.

.clang-tidy
```diff
@@ -26,9 +26,18 @@ Checks:     '
...
+            -readability-magic-numbers,
```
mbutrovich: Yells at us about magic numbers mostly in tests. Not gonna worry about this check.

.clang-tidy
mbutrovich: Here's the relevant information about it:
https://en.cppreference.com/w/cpp/language/attributes/nodiscard

.clang-tidy
```diff
@@ -26,9 +26,18 @@ Checks:     '
             performance-*,
             portability-*,
             readability-*,
+            -bugprone-exception-escape,
+            -bugprone-macro-parentheses,
```
mbutrovich: This one seems reasonable to reenable in the future, I just didn't want to tackle it right now.

mbutrovich:
Travis:
- the 2 packages.sh systems work, so we're grabbing all of the new dependencies correctly
- the 2 Dockerfile systems fail, which is expected until @crd477 gets a change to update that

crd477:
>     * the 2 Dockerfile systems fail, which is expected until @crd477 gets a chance to update that

Sorry, I'm confused.  Where do you see the failure?

Jenkins does not use `Dockerfile`.  The Docker images that Jenkins uses are prepopulated with dependencies just to avoid spending time downloading gigabytes of packages every build but ultimately, they'll get whatever `packages.sh` specifies.

It looks to me like the correct thing happened in:
<http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-501/4/pipeline/13>

As far as I know, `Dockerfile` was just supplied by a previous developer as a convenience for the database that I will not [deadname](https://www.urbandictionary.com/define.php?term=deadname).

This is definitely an appropriate time to update the prepopulated Docker images but unless I'm missing something, it's not blocking #501, right?

mbutrovich:
You're right, my mistake: we only use Docker for Travis not for Jenkins, it looks like. I'll have to look into how to update that. Doesn't sound like there's anything to do on your part.

crd477:
OK, great!  I'll still update the Docker images with the current dependencies but that will not involve any source code changes and should be transparent to the CI process.

.clang-tidy
lmwnshn: Can confirm, ok

.clang-tidy
lmwnshn: Ditto

.clang-tidy
lmwnshn: My understanding is that it is "warn on unused result". I do think it is useful, but maybe not worth cluttering every function definition with...

.clang-tidy
lmwnshn: Leave a comment?

.clang-tidy
lmwnshn: Leave a comment that storage would be sad?

.clang-tidy
lmwnshn: I also don't like that, but would approve without further comment or discussion, so..

.clang-tidy
lmwnshn: We could disable it just for tests, I think, if that is an issue. I do remember someone going ??? at magic numbers before.

Dockerfile
```diff
@@ -7,9 +7,9 @@ ARG DEBIAN_FRONTEND=noninteractive
 RUN apt-get -y update && \
```
lmwnshn: I wish we 1. moved wget up there and 2. organized the package list alphabetically, but that can wait for a separate PR. I have a separate Dockerfile which is tidied up a little, which I'm using for my windows box.

src/planner/plannodes/csv_scan_plan_node.cpp
```diff
@@ -5,7 +5,7 @@
 namespace terrier::planner {
 
 common::hash_t CSVScanPlanNode::Hash() const {
-  common::hash_t hash = AbstractPlanNode::Hash();
+  common::hash_t hash = AbstractPlanNode::Hash();  // NOLINT: bugprone-parent-virtual-call: this appears intentional?
```
lmwnshn: Could @GustavoAngulo or @wenxuanqiu weigh in?

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -22,6 +22,7 @@ void LogSerializerTask::Process() {
   {
     common::SpinLatch::ScopedSpinLatch guard(&flush_queue_latch_);
     temp_flush_queue = std::move(flush_queue_);
+    flush_queue_ = std::queue<RecordBufferSegment *>();
```
lmwnshn: This is an actual fix riding on the rest of the PR, right?

mbutrovich: Yep. It complained about flush_queue_ being looked at again in the calling context, however a moved from container is left in a valid but otherwise non-deterministic state. Since std::queue doesn't have a clear or reset method, the solution is just a new object I guess.

.clang-tidy
mbutrovich: I'll open issues for each one and mark them beginner after this merges. I think that's more visible than comments in a clang-tidy file that no one looks at.

src/planner/plannodes/csv_scan_plan_node.cpp
mbutrovich: I think it actually is a legit bug, now that I look at the explanation a bit more.

https://clang.llvm.org/extra/clang-tidy/checks/bugprone-parent-virtual-call.html

Will fix.

PR #502, opened 2019-08-09
Canonical schemas

jrolli:
Update `GetColumns` and `DeleteColumns` functions to use `columns_oid_index_` which allows us to generate schemas in ascending OID order.  It also allows us to remove an entire index from the catalog.


src/catalog/database_catalog.cpp
```diff
@@ -456,26 +452,39 @@ std::vector<Column> DatabaseCatalog::GetColumns(transaction::TransactionContext
...
-  auto *pr = class_pri.InitializeRow(buffer);
+  auto *pr = oid_pri.InitializeRow(buffer);
+  auto *key_pr = oid_pri.InitializeRow(key_buffer);
+
+  // Write the attributes in the ProjectedRow
+  // Low key (class, INVALID_COLUMN_OID) [using uint32_t to avoid adding ColOid to template]
   *(reinterpret_cast<ClassOid *>(pr->AccessForceNotNull(0))) = class_oid;
+  *(reinterpret_cast<uint32_t *>(pr->AccessForceNotNull(1))) = 0;
+
+  // Low key (class + 1, INVALID_COLUMN_OID) [using uint32_t to avoid adding ColOid to template]
```
amlatyrngom: High Key.

src/catalog/database_catalog.cpp
```diff
@@ -406,39 +400,41 @@ bool DatabaseCatalog::CreateColumn(transaction::TransactionContext *const txn, c
...
-  auto *pr = class_pri.InitializeRow(buffer);
+  auto *pr = oid_pri.InitializeRow(buffer);
+  auto *pr_high = oid_pri.InitializeRow(key_buffer);
+
   // Write the attributes in the ProjectedRow
+  // Low key (class, INVALID_COLUMN_OID) [using uint32_t to avoid adding ColOid to template]
   *(reinterpret_cast<ClassOid *>(pr->AccessForceNotNull(0))) = class_oid;
+  *(reinterpret_cast<uint32_t *>(pr->AccessForceNotNull(1))) = 0;
+
+  // Low key (class + 1, INVALID_COLUMN_OID) [using uint32_t to avoid adding ColOid to template]
```
amlatyrngom: High Key.

src/catalog/database_catalog.cpp
```diff
@@ -406,39 +400,41 @@ bool DatabaseCatalog::CreateColumn(transaction::TransactionContext *const txn, c
...
+  byte *const key_buffer = common::AllocationUtil::AllocateAligned(oid_pri.ProjectedRowSize());
   // Scan the class index
-  auto *pr = class_pri.InitializeRow(buffer);
+  auto *pr = oid_pri.InitializeRow(buffer);
+  auto *pr_high = oid_pri.InitializeRow(key_buffer);
+
   // Write the attributes in the ProjectedRow
+  // Low key (class, INVALID_COLUMN_OID) [using uint32_t to avoid adding ColOid to template]
   *(reinterpret_cast<ClassOid *>(pr->AccessForceNotNull(0))) = class_oid;
+  *(reinterpret_cast<uint32_t *>(pr->AccessForceNotNull(1))) = 0;
```
amlatyrngom: `oid_pm` is used to insert into the index (in CreateColumn), but not to read from it (in GetColumns & DeleteColumns). These two uses should be the consistent. Probably don't need the map in CreateColumn?

PR #506, opened 2019-08-10
Metrics Manager

mbutrovich:
Opening a new PR since we want to start getting some eyes (and users) on this.


benchmark/integration/tpcc_benchmark.cpp
```diff
@@ -146,10 +147,90 @@ BENCHMARK_DEFINE_F(TPCCBenchmark, ScaleFactor4WithLogging)(benchmark::State &sta
...
+      }
+    }
+    state.SetItemsProcessed(state.iterations() * num_new_orders);
+  } else {
+    state.SetItemsProcessed(state.iterations() * num_precomputed_txns_per_worker_ * num_threads_);
+  }
+}
+
+// NOLINTNEXTLINE
+BENCHMARK_DEFINE_F(TPCCBenchmark, ScaleFactor4WithLoggingAndMetrics)(benchmark::State &state) {
```
malin1993ml: It looks like there're quite some duplicated code in the few functions in this file. Should we abstract them up?

benchmark/metrics/large_transaction_metrics_benchmark.cpp
```diff
@@ -0,0 +1,186 @@
+#include <string>
```
malin1993ml: It looks like the only difference between this benchmark and the original large transaction benchmark is whether the metrics are enabled. Should we put this inside the original benchmark file and pass an argument to indicate whether the metrics are enabled, to avoid duplicated code?

benchmark/metrics/logging_metrics_benchmark.cpp
```diff
@@ -0,0 +1,292 @@
+#include <string>
```
malin1993ml: Same comments as the above benchmark.

PR #508, opened 2019-08-13
Add clang-8 builds to Linux CI

mbutrovich:
Also rename the different configurations to be consistent. Remove ASAN related stuff from Travis since all we do is static analysis and building, not execute (except for Coverage where we run ASAN off).

I'll watch the build times tonight to see if this completely wrecks our CI time.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,14 +30,13 @@ matrix:
         - docker cp . build:/repo
     - os: linux
       dist: trusty
+      compiler: clang
       env:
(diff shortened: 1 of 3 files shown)


mbutrovich:
Tried 3 different ways of specifying clang for Travis. The second method was dumb in retrospect since it wouldn't affect the Docker container, but the first and third should have worked, I think. Regardless, Jenkins is working. We can look at Travis tomorrow.

PR #10, opened 2018-07-24
Enable Travis builds

lmwnshn:
Enables both Darwin and Ubuntu 18.04 Docker builds on Travis.

PR #509, opened 2019-08-13
Be smarter about Column copies

mbutrovich:
We noticed that `tpcc_test` was taking longer than it used to (2-3x), but throughput on the benchmark hasn't suffered so I suspected the loader. Quickly tracked it down to deep copies of `Schema::Column` and `IndexSchema::Column` objects in `tpcc/util.h` introduced with #481 when writing tuple and index key attributes. These aren't used in the workload because we cache all of the column offsets as if it were compiled code.

- Changed `Schema::GetColumn()` methods to return `const &` to avoid deep copies where possible.
- Changed `tpcc/util.h`'s call to `IndexSchema::GetColumns()` to be used as a `const auto &` to avoid deep copies. The perils of `auto`...

### DIFF
--- a/src/include/catalog/schema.h
+++ b/src/include/catalog/schema.h
@@ -216,7 +216,7 @@ class Schema {
    * @param col_offset offset into the schema specifying which Column to access
    * @return description of the schema for a specific column
    */
-  Column GetColumn(const uint32_t col_offset) const {
+  const Column &GetColumn(const uint32_t col_offset) const {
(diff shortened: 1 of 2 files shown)


mbutrovich:
Master:
`10/44 Test #10: tpcc_test ........................   Passed  685.19 sec`
#509:
`10/44 Test #10: tpcc_test ........................   Passed  185.47 sec`

Mission accomplished.

PR #511, opened 2019-08-14
Add 'timeout' for Ubuntu and 'gtimeout' for macOS

jrolli:
Adds a timeout mechanism for `make unittest` and `make runbenchmark` for Jenkins.  Motivated by a current bug in one PR that started causing resource starvation of our macOS build machines.

Uses a 1 hour timeout as a conservative baseline (double worst observed successful run).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -18,7 +18,7 @@ pipeline {
                         sh 'mkdir build'
                         sh 'cd build && cmake -DCMAKE_BUILD_TYPE=Debug -DTERRIER_USE_ASAN=ON .. && make -j4'
                         sh 'cd build && make check-clang-tidy'
-                        sh 'cd build && make unittest'
+                        sh 'cd build && gtimeout 1h make unittest'
                         sh 'cd build && python ../script/testing/junit/run_junit.py'
                     }
                 }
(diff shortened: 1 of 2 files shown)


mbutrovich:
Do we only want the timeout applied to the actual test (either unittest or runbenchmark)? Didn't we have a long-running build this morning due to packages.sh taking too long?

jrolli:
The install script can take a while (in fact one took 30+ minutes in Jenkins this time around), but I think they do generally exit eventually.  My focus here was to ensure we put a timeout on executing code that we generate and therefore may have errors that cause them not to exit.  Everything else should be (mostly) guaranteed to exit.  However, we can certainly extend this to encompass the other steps.

I'm not passionate one way or the other, so just let me know if you want me to put timeouts on the install script.

mbutrovich:
Fair enough. We may want to add it to the only configuration on Travis that runs `unittest` then.
https://github.com/cmu-db/terrier/blob/f1a0c78aff330fcccff885401d4c92aedd1ae201/.travis.yml#L52

mbutrovich:
Although Travis has a built-in timeout that we’ve run into so maybe not necessary.

PR #516, opened 2019-08-17
Cache contention

LiuXiaoxuanPKU:
This PR tries to solve the contention problems when multiple threads try to insert concurrently.

The steps to insert is first allocating a tupleslot in a block and then write to that tupleslot.
The previous implementation allows different threads to allocate tupleslots within the same block concurrently. Compare and swap is used to make sure only one thread can win. This implementation has two problems: (1) There is high contention on the tupleslot when multiple threads try to get the same tupleslot. Since only one thread can win, other threads need to try several times before succeed. (2) Since different threads are writing to the same block, the block header (offset field that indicates the offset of free tupleslot) is being written and invalidated, evicted from cached repeatedly, which degenerate the cache performance.

To solve the problem, each block uses the first bit of offset to indicate if other threads is trying to allocate tupleslot in the block. If yes, it will go to the next block with free slots or create a new block when no such block exists. This solves the two problems by preventing different threads from competing for the same tupleslot.

Experiments show that the fix makes insert operation scale better. The experient runs data_table_benchmark with ~70,000,000 insertions. It tries to test the throughput VS the number of threads.
insert_cache_fix

On the other hand, perf results show
(1) The fix reduces branch misses from xxx to xxx
(2) The fix reduces cache miss from xxx to xxx. (Which is not a great win)

I also count how many times the tread need to try before getting a tupleslot successfully. In the setting of 20 threads, the previous version needs to try 6.614556 times on average and the fix version only needs to try once.


apavlo:
This is amazing!

src/include/storage/tuple_access_strategy.h
```diff
@@ -253,9 +253,32 @@ class TupleAccessStrategy {
...
+    return block->insert_head_.compare_exchange_weak(old_val, clr_bit(old_val));
+  }
+
  private:
   const BlockLayout layout_;
   // Start of each mini block, in offset to the start of the block
   std::vector<uint32_t> column_offsets_;
+  // Not sure if I should put the two lambdas here
+  uint32_t (*clr_bit)(uint32_t) = [](uint32_t val) { return val & ~(1 << 31); };
+  uint32_t (*set_bit)(uint32_t) = [](uint32_t val) { return val | (1 << 31); };
```
jrolli: Make these two functions static class functions.  That should ensure we only have one copy of them (rather than one per tuple access strategy object) and will likely result in them being inlined into the compiled code.

src/storage/data_table.cpp
```diff
@@ -314,15 +359,13 @@ bool DataTable::CompareAndSwapVersionPtr(const TupleSlot slot, const TupleAccess
   return reinterpret_cast<std::atomic<UndoRecord *> *>(ptr_location)->compare_exchange_strong(expected, desired);
 }
 
-void DataTable::NewBlock(RawBlock *expected_val) {
+std::list<RawBlock *>::iterator DataTable::NewBlock() {
   common::SpinLatch::ScopedSpinLatch guard(&blocks_latch_);
-  // Want to stop early if another thread is already getting a new block
-  if (expected_val != insertion_head_) return;
   RawBlock *new_block = block_store_->Get();
   accessor_.InitializeRawBlock(this, new_block, layout_version_);
```
jrolli: Move these two lines to before the instantiation of the spin latch.  It should greatly reduce the time spent in the critical section without affecting correctness.

src/include/storage/storage_defs.h
```diff
@@ -82,7 +86,14 @@ struct alignas(common::Constants::BLOCK_SIZE) RawBlock {
...
   // A Block needs to always be aligned to 1 MB, so we can get free bytes to
-  // store offsets within a block in ine 8-byte word.
+  // store offsets within a block in ine 8-byte word
+
+  /**
+   * Get the offset of this block. Because the first bit insert_head_ is used to indicate the status
+   * of the block, we need to clear the status bit to get the real offset
+   * @return the offset which tells us where the next insertion should take place
+   */
+  std::atomic<uint32_t> GetInsertHead() { return (~(1 << 31)) & insert_head_.load(); }
```
tli2: Why is the return type also an atomic? This is not the same underlying value since you've copied it

src/include/storage/data_table.h
```diff
@@ -238,10 +238,14 @@ class DataTable {
...
+  // latch used to protect block list
   mutable common::SpinLatch blocks_latch_;
-  // to avoid having to grab a latch every time we insert. Failures are very, very infrequent since these
-  // only happen when blocks are full, thus we can afford to be optimistic
-  std::atomic<RawBlock *> insertion_head_ = nullptr;
+  // latch used to protect insertion_head_
+  mutable common::SpinLatch header_latch_;
+  std::list<RawBlock *>::iterator insertion_head_;
+  // Check if the block that insertion_head_ points to is full, if it's full, move the header
+  // This function uses header_latch_ to ensure correctness
```
tli2: Capital first letter

src/include/storage/storage_defs.h
```diff
@@ -82,7 +86,14 @@ struct alignas(common::Constants::BLOCK_SIZE) RawBlock {
   byte content_[common::Constants::BLOCK_SIZE - sizeof(uintptr_t) - sizeof(uint16_t) - sizeof(layout_version_t) -
                 sizeof(uint32_t) - sizeof(BlockAccessController)];
   // A Block needs to always be aligned to 1 MB, so we can get free bytes to
-  // store offsets within a block in ine 8-byte word.
+  // store offsets within a block in ine 8-byte word
```
tli2: one?

src/include/storage/tuple_access_strategy.h
```diff
@@ -253,6 +253,40 @@ class TupleAccessStrategy {
...
+  }
+
+  /**
+   * Compare and swap block status to idle. Expect that block to be busy, if yes, set the block status to be idle
+   * @param block the block to compare and set
+   * @return true if the set operation succeeded, false if the block is already idle
+   */
+  bool ClearBlockBusyStatus(RawBlock *block) const {
+    uint32_t old_val = set_bit(block->insert_head_.load());
+    return block->insert_head_.compare_exchange_weak(old_val, clr_bit(old_val));
```
tli2: Why case weak here. Will we lose a block forever in the face of a spurious failure?

LiuXiaoxuanPKU: You are right, it should be strong

src/include/storage/storage_defs.h
LiuXiaoxuanPKU: Will it cause any problem if I change the return type to uint32_t?
In other places, we use the offset to check if the block is full. They may use stale values to do the check?

tli2: You are making a copy anyways so it will be stale even if it is atomic.

LiuXiaoxuanPKU: I got that point. Any suggestions on how to fix it? I guess right now I will just leave some comments there because it passed all tests and seems not to affect too much.

LiuXiaoxuanPKU:
1. We do not support concurrent insertion into the same block anymore because the data table insert logic makes sure different txns write to different blocks concurrently. Thus I get rid of the concurrent insert test and benchmark for tuple access strategy. **If you think we should keep the concurrent insertion for tuple access strategy for some reason, please comment here**

2. The experiment graph above is performed on dev4 machine. When running on jenkins, the performance gain is less obvious. The fixed version only has better performance when there are **more than** 4 threads. If we want to pass the benchmark and merge the change, we may increase the thread number or overwrite the jenkins number manually.

apavlo:
@jrolli I don't follow what you are saying. Let's discuss tomorrow.

benchmark/storage/tuple_access_strategy_benchmark.cpp
```diff
@@ -78,40 +78,6 @@ BENCHMARK_DEFINE_F(TupleAccessStrategyBenchmark, SimpleInsert)(benchmark::State
...
-      block_store_.Release(raw_blocks_[i]);
-    }
-  }
-
-  state.SetItemsProcessed(state.iterations() * layout_.NumSlots() * num_blocks_);
-}
-
 BENCHMARK_REGISTER_F(TupleAccessStrategyBenchmark, SimpleInsert)->Unit(benchmark::kMillisecond);
 
-BENCHMARK_REGISTER_F(TupleAccessStrategyBenchmark, ConcurrentInsert)->Unit(benchmark::kMillisecond)->UseRealTime();
```
mbutrovich: Why did we remove this benchmark? I think it's still meaningful to know how fast multiple threads can insert into a block if needed. I realize that won't happen when going through the DataTable API anymore, but you never know what we'll do in the future.

jrolli: We would have to keep extra branching logic at the block layer to allow this to be safe and not error out.

tli2: If we are to assume that only one thread can insert into a block at a time, we should replace CAS instructions with normal write instructions for better performance on the bitmap level.

src/include/storage/tuple_access_strategy.h
```diff
@@ -253,6 +253,40 @@ class TupleAccessStrategy {
...
+    uint32_t old_val = SetBit(block->insert_head_.load());
+    return block->insert_head_.compare_exchange_strong(old_val, ClrBit(old_val));
+  }
+
+  /**
+   * Set the first bit of given val to 0, helper function used by ClearBlockBusyStatus
+   * @param val the value to be set
+   * @return the changed value with first bit set to 0
+   */
+  static uint32_t ClrBit(uint32_t val) { return val & ~(1 << 31); }
```
tli2: nit: don't use abbreviations here for clr as other parts of your code are not using them

src/storage/data_table.cpp
```diff
@@ -122,23 +125,67 @@ bool DataTable::Update(transaction::TransactionContext *const txn, const TupleSl
   return true;
 }
 
+void DataTable::checkMoveHead(std::list<RawBlock *>::iterator block) {
```
tli2: nit: Capitalize

tli2:
On a side note, we don't seem to have a check for naming conventions. (e.g. CamlCase for functions and classes, snake_case_ for members, snake_case for local vars, etc.)

benchmark/storage/tuple_access_strategy_benchmark.cpp
jrolli: That makes sense.  Do we need the atomic operations (of some kind) to ensure proper sequencing of writes from other threads' perspective?

src/storage/tuple_access_strategy.cpp
```diff
@@ -43,21 +43,18 @@ void TupleAccessStrategy::InitializeRawBlock(storage::DataTable *const data_tabl
...
-      block->insert_head_++;
-      return true;
-    }
-  }
-
-  return false;
+  // We do not support concurrent insertion to the same block anymore
+  // Assumption: Different threads cannot insert into the same block at the same time
+  // If the block is not full, the function should always succeed (Flip should always return true)
+  bitmap->Flip(pos, false);
```
tli2: Can you assert this instead of just comment?

src/include/storage/tuple_access_strategy.h
```diff
@@ -253,6 +253,40 @@ class TupleAccessStrategy {
...
+   * @return the changed value with first bit set to 0
+   */
+  static uint32_t ClearBit(uint32_t val) { return val & ~(1 << 31); }
+
+  /**
+   * Set the first bit of given val to 1, helper function used by SetBlockBusyStatus
+   * @param val val the value to be set
+   * @return the changed value with first bit set to 1
+   */
+  static uint32_t SetBit(uint32_t val) { return val | (1 << 31); }
```
mbutrovich: See previous comment. I think in this case you want `INT32_MIN` as your mask.

mbutrovich:
As I review the logic a bit more, I'm not thrilled at the lack of a guarantee of progress. It seems odd to me that a thread can allocate a block on data_table.cpp:164 and then loops through again and looks for an available block. It seems unlikely, but a block could be extremely unlikely and keep allocating blocks for other threads to insert into and not make efficient progress. It seems simpler to me that if a thread allocates a block, it has exclusive access to it, allocates a slot, and then pops it into the list of blocks when it's done. I also think this aligns more cleanly with a possible bulk allocation logic for bulk insert. Thoughts? We can discuss tomorrow in the lab as well.

mbutrovich:
Also as best as I can tell, the Block's busy status is bit is only used as a lock for Inserts right now? If so, we might just want to leave a full Block locked rather than clearing it as on data_table.cpp:172. This isn't exactly directions to change right now, just word vomiting thoughts as I review the logic.

src/storage/data_table.cpp
```diff
@@ -122,23 +131,79 @@ bool DataTable::Update(transaction::TransactionContext *const txn, const TupleSl
...
 }
 
+void DataTable::CheckMoveHead(std::list<RawBlock *>::iterator block) {
+  common::SpinLatch::ScopedSpinLatch guard_head(&header_latch_);
+  if (block == insertion_head_) {
+    // If the header block is full, move the header to point to the next block
+    insertion_head_++;
+  }
+
+  // If there are no more free blocks, create a new empty block and  point the insertion_head to it
```
jrolli: Drop this code.  `DataTable::Insert` already handles this case for you (which is necessary for correctness) and I don't think there is a compelling reason to do eager allocation of a 1 MiB chunk of memory.

src/storage/data_table.cpp
```diff
@@ -122,23 +131,79 @@ bool DataTable::Update(transaction::TransactionContext *const txn, const TupleSl
...
+        // The block is not full, succeed
+        break;
+      }
+      // Fail to insert into the block, flip back the status bit
+      accessor_.ClearBlockBusyStatus(*block);
+      // if the full block is the insertion_header, move the insertion_header
+      // Next insert txn will search from the new insertion_header
+      CheckMoveHead(block);
+      // The block is full, try next block
+      ++block;
```
jrolli: nit: Consolidate this line with line 198 by removing the else and unconditionally executing it after the `if` statement.

src/storage/data_table.cpp
```diff
@@ -122,23 +131,79 @@ bool DataTable::Update(transaction::TransactionContext *const txn, const TupleSl
...
+    } else {  // The block is inserting by other txn
+      // Try next block
+      ++block;
+    }
   }
+
+  // Do not need to wait unit finish inserting,
+  // can flip back the status bit once the thread gets the allocated tuple slot
+  accessor_.ClearBlockBusyStatus(*block);
```
jrolli: I may have initially asked this on Slack, but is there any significant performance difference between releasing the lock before or after `InsertInto`?  There's potential cache-thrashing throughout the insert process (not just in the header, but with each attribute's bitmap and value as well), but it may not be significant in practice if the lock is held long enough to space the threads out.

src/storage/data_table.cpp
```diff
@@ -122,23 +131,79 @@ bool DataTable::Update(transaction::TransactionContext *const txn, const TupleSl
...
   TupleSlot result;
+  auto block = insertion_head_;
   while (true) {
-    RawBlock *block = insertion_head_.load();
-    if (block != nullptr && accessor_.Allocate(block, &result)) break;
-    NewBlock(block);
+    // No free block left
+    if (block == blocks_.end()) {
+      RawBlock *new_block = NewBlock();
+      TERRIER_ASSERT(accessor_.SetBlockBusyStatus(new_block), "Status of new block should not be busy");
```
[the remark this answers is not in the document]
LiuXiaoxuanPKU: The logic here is 
If there is no free block left, create a new block, allocate a tupleslot, then push that new block into the block list. 
So we do not necessarily need to set/clear the block status, because the thread allocates first then adds it to the list.
The assert here just says a new block is always idle. Maybe we can get rid of it to avoid confusion?

src/storage/data_table.cpp
LiuXiaoxuanPKU: This function is called by the Insert function. I may misunderstand your comments. 'DataTable::Insert already handles this case for you' Insert handles the case by calling the checkMoveHead function.

PR #525, opened 2019-08-26
Make Jenkins print hostname, increase build parallelism on Jenkins Linux machines.

jrolli:
Adds `hostname` command to start of all Jenkins scripts in order to better enable performance issue tracking.


crd477:
You should use `sh 'echo $NODE_NAME'` instead of printing the hostname.  The output of `hostname` will be meaningless when the stage is running in a Docker container, whereas `NODE_NAME` is setup by Jenkins and will reflect the name of the actual worker node that the job is running on regardless of whether the stage is running in a container or not.

Issue #526, opened 2019-08-28
Race with deferred DDL changes and GC


GustavoAngulo:
The error involves the following schedule (Assume there is a table foo):

```
T0				                T1

BEGIN
write into foo
				                BEGIN
				                drop foo (defers a delete action)
				                COMMIT
COMMIT
T0 removes itself from 
curr_active_txns_
				                GC processes the drop foo deferred action which deallocates foo,
				                because current active txn > T1
ERROR*
```
Error: GC tries to unlink T0, which causes reads to foo’s block layout, which GC already deallocated

Stack frame of error:
```
2019-08-28 17:32:06.343057-0400 recovery_test[47728:5704403] =================================================================
2019-08-28 17:32:06.343510-0400 recovery_test[47728:5704403] ==47728==ERROR: AddressSanitizer: heap-use-after-free on address 0x611000025ec8 at pc 0x000103a972cc bp 0x7ffeec8cac60 sp 0x7ffeec8cac58
2019-08-28 17:32:06.343539-0400 recovery_test[47728:5704403] READ of size 4 at 0x611000025ec8 thread T0
2019-08-28 17:32:06.343567-0400 recovery_test[47728:5704403] pc_0x103a972cb###func_terrier::storage::BlockLayout::NumSlots() const###file_block_layout.h###line_81###obj_(recovery_test:x86_64+0x1007642cb)
2019-08-28 17:32:06.343584-0400 recovery_test[47728:5704403] pc_0x1039adf31###func_terrier::storage::TupleAccessStrategy::AccessWithoutNullCheck(terrier::storage::TupleSlot, terrier::common::StrongTypeAlias<terrier::storage::tags::col_id_t_typedef_tag, unsigned short>) const###file_tuple_access_strategy.h###line_163###obj_(recovery_test:x86_64+0x10067af31)
2019-08-28 17:32:06.343600-0400 recovery_test[47728:5704403] pc_0x1039a4559###func_terrier::storage::DataTable::AtomicallyReadVersionPtr(terrier::storage::TupleSlot, terrier::storage::TupleAccessStrategy const&) const###file_data_table.cpp###line_281###obj_(recovery_test:x86_64+0x100671559)
[...]

PR #527, opened 2019-09-04
GC queue and Deferred Action queue Ordering

### LINKED ISSUE #520: Actions in DeferredActionManager can appear out of order
With #441 decoupling actions into a separate component, it is no longer protected by the commit latch in the transaction manager. There is now possible out-of-order execution of the action queue.

We (me and jrolli) think that it does not matter, but we are not sure. It will be nice to take some time to investigate and verify that this is fine.

tli2:
fix #520 

This PR shuffles around latches to make sure that:

- Deferred Actions appear in order
- Transactions are not removed from the global table until it has been added to GC queue

I believe the combination of the two fixes above will allow #526 to be fixed with a double deferral


mbutrovich:
@yash620 @jrolli @GustavoAngulo How does this fit with what we discussed this morning?

yash620:
I think @jrolli mentioned that in DeferredActionsManager::Process() you would need to ensure the backlog is empty before processing new actions to fix 520. ClearBacklog() doesn't actually clear it, it just tries to, so we would just need to add an if check here. This would also mean that the new actions aren't be stored in a buffer of the deferred actions manager but rather who ever is registering the actions. This way whoever is registering the actions can backoff or stall for a while if the action queue is filling up.

Issue #437, opened 2019-06-19
Helper Function to Extract Referenced Column Oids from AbstractExpression Tree


apavlo:
The optimizer cannot assume that an index's schema contains a one-to-one mapping from key attribute to column oid. This is because you can have indexes like this:

```sql
CREATE INDEX ON xxx (col1 + 1, col2 + col3);
```

@17zhangw needs a helper method that recursively traverses an `AbstractExpression` tree and returns a set of the referenced `catalog::col_oid_t` values in the tree.

This should be done **after** we rename `TupleValueExpression` in #404.

## Instructions

1. Create a new `ExpressionUtil` class in 'src/include/parser/expression/expression_util.h'.
   See [`HashUtil`](https://github.com/cmu-db/terrier/blob/master/src/include/common/hash_util.h) for what these classes look like.

2. Create a static method called `GetColumnOids`:
    ```c++
    static void ExpressionUtil::GetColumnOids(std::set<catalog::col_oid_t> &col_oids, managed_ptr<AbstractExpression> exp) {
    ...
    }
    ```
    This method should recursively call itself for the left/right children (if not null) and populate the `col_oids` set from `TupleValueExpression` objects (this will change after #404).

3. Write a set of test cases in 'test/parser/expression_util_test.cpp'. See other test cases on how to create test `AbstractExpression` objects and run the tests.

jbx7106:
Hi apavio I'm a beginner. Would you assign this task to me?

apavlo:
@jbx7106 Go for it! I've updated the issue with instructions. Send us as pull request when you're ready. See these instructions on how to do that:

https://github.com/cmu-db/terrier/wiki/Pull-Request-Process

apavlo:
@jbx7106 See this incomplete method from the old code:

https://github.com/cmu-db/peloton/blob/master/src/include/expression/expression_util.h#L349-L385

This is not exactly what we want but it should give you the basic idea.

jbx7106:
Great. Thanks, @apavlo I look forward to contributing. :)

PR #539, opened 2019-09-11
Add helper function to extract referenced column oids from AbstractExpression

tpan496:
Addresses #437.
Added a new file parser/expression/expression_util.h that is based on https://github.com/apavlo/terrier/commit/69c8e6f89c45c380caf313cff0f75376614a86d8. Since ```TupleValueExpression``` has been updated, I resorted to using ```ColumnValueExpression``` instead.


apavlo:
This PR is for @17zhangw.

@17zhangw - Does this provide you with what you need?

src/include/parser/expression/expression_util.h
```diff
@@ -0,0 +1,45 @@
...
+  // Static utility class
+  ExpressionUtil() = delete;
+
+  /**
+   * Populate the given set with all of the column oids referenced
+   * in this expression tree.
+   * @param col_oids place to store column oids in the expression
+   * @param expr to extract column oids from
+   */
+  static void GetColumnOids(std::set<catalog::col_oid_t> *col_oids, common::ManagedPointer<AbstractExpression> expr) {
```
GustavoAngulo: What was the reason to use `std::set`? I actually think this should be a `std::vector`. The main reason is because a lot of our functions that use `col_oid_t` store them in an `std::vector` ([`SqlTable::InitializerForProjectedRow`](https://github.com/cmu-db/terrier/blob/94e999bfb9355dafb03f15b6c86803650599fe37/src/include/storage/sql_table.h#L180) most likely being the most popular). Converting from an `std::set` to `std::vector` is O(n) in c++ so we should avoid that if possible. 

If for some reason we do need a set, I can't think of a reason we need the ordering `std::set` provides, so we could just use `std::unordered_set`

src/include/parser/expression/expression_util.h
```diff
@@ -0,0 +1,45 @@
...
+    // Recurse into our children
+    for (const auto &child : expr->GetChildren()) {
+      ExpressionUtil::GetColumnOids(col_oids, common::ManagedPointer<AbstractExpression>(child.get()));
+    }
+
+    // If our mofo is a ColumnValueExpression, then pull out our
```
GustavoAngulo: Not familiar with the word/acronym "mofo", I'm guessing its not the slang term. Can you make this comment clearer

src/include/parser/expression/expression_util.h
tpan496: Is it possible for ```col_oid_t``` to be duplicated? I forked this from older code that was using a ```std::set```, but I am not entirely sure about the behavior. If duplicates are not a problem I can switch to ```std::vector```?

apavlo: I don't think this is supposed to be ordered.

@17zhangw Can you point us to the code that would use this information?

src/include/parser/expression/expression_util.h
apavlo: This is fine: https://www.urbandictionary.com/define.php?term=mofo

test/parser/expression_util_test.cpp
```diff
@@ -0,0 +1,55 @@
...
+namespace terrier::parser::expression {
+
+// NOLINTNEXTLINE
+TEST(ExpressionUtilTest, GetColumnOidsTest) {
+  // Test a single ColumnValueExpression
+  auto tve =
+      std::make_shared<ColumnValueExpression>(catalog::db_oid_t(1), catalog::table_oid_t(2), catalog::col_oid_t(3));
+  std::set<catalog::col_oid_t> oids;
+  ExpressionUtil::GetColumnOids(&oids, common::ManagedPointer<AbstractExpression>(tve.get()));
+  EXPECT_THAT(oids, ElementsAre(catalog::col_oid_t(3)));
```
apavlo: Nice!

src/include/parser/expression/expression_util.h
tpan496: I actually thought it was meant to be a legit database term so I kept it as it is

src/include/parser/expression/expression_util.h
apavlo: Yes, a `col_id_t` can be duplicated in an expression tree:

```
SELECT * FROM xxx WHERE id = 1 OR id = 2
```

17zhangw: Technically, it would be used within this file:
```
https://github.com/cmu-db/terrier/blob/optimizer/src/include/optimizer/index_util.h
```

The current optimizer (in the branch) only supports picking indexes built only on base columns. But I do anticipate we'll be needing this function in the near future.

PR #542, opened 2019-09-13
Update lookback after merging #516

mbutrovich:
#516 changed some performance numbers so we want to reset the lookback to reflect that.


mbutrovich:
As always, any PRs over the next few days should have their performance numbers inspected very carefully before approval because we won't get the auto rejection until enough nightly builds run.

PR #543, opened 2019-09-14
Added Index OID vector to InsertPlanNode

tanujnay112:
Added index oid vector to InserPlanNode in order to enable index insertion codegen.


src/include/planner/plannodes/insert_plan_node.h
```diff
@@ -79,6 +79,16 @@ class InsertPlanNode : public AbstractPlanNode {
       return *this;
     }
 
+
+    /**
+     * @param index_oid oid of insertion index that is to be added to this plan node
+     * @return builder object
+     */
+    Builder &AddIndexOid(catalog::index_oid_t index_oid) {
```
mbutrovich: Do we expect to add OIDs one at a time? It's going to come from the catalog, right? CatalogAccessor::GetIndexOids directly returns the vector for a given table.

mbutrovich:
Pinging @jrolli and @amlatyrngom since after a discussion in the lab they seem convinced that this is not necessary.

apavlo:
After a deep discussion, we decided to proceed with this PR.

apavlo:
We also need to update the physical operators to include this information.

src/include/planner/plannodes/insert_plan_node.h
```diff
@@ -79,6 +79,17 @@ class InsertPlanNode : public AbstractPlanNode {
       return *this;
     }
 
+    /**
+     * @param index_oids vector of index oids to insert into
+     * @return builder object
+     */
+    Builder &AddIndexOids(std::vector<catalog::index_oid_t> &&index_oids) {
```
mbutrovich: The entire vector's coming in as an r-value reference. You should be able to just do `index_oids_ = index_oids;` since vector should have a move-assignment operator.

src/optimizer/physical_operators.cpp
```diff
@@ -513,11 +515,12 @@ bool Insert::operator==(const BaseOperatorNode &r) {
 // InsertSelect
 //===--------------------------------------------------------------------===//
 Operator InsertSelect::Make(catalog::db_oid_t database_oid, catalog::namespace_oid_t namespace_oid,
-                            catalog::table_oid_t table_oid) {
+                            catalog::table_oid_t table_oid, std::vector<catalog::index_oid_t> &&index_oids) {
   auto *insert_op = new InsertSelect;
   insert_op->database_oid_ = database_oid;
   insert_op->namespace_oid_ = namespace_oid;
   insert_op->table_oid_ = table_oid;
+  insert_op->index_oids_ = std::move(index_oids);
```
mbutrovich: Does this work without the std::move? It's already an r-value reference.

src/include/planner/plannodes/insert_plan_node.h
tanujnay112: Does this not replace all oids that were already present in the index_oids_ vector?

mbutrovich: Yeah it would replace all of them. Do we expect to have to append to the list of index oids multiple times? I feel like this is a one-shot call by the binder and then the list is sort of immutable.

tanujnay112: OK that makes sense, I should probably change it to "SetIndexOids" then.

PR #545, opened 2019-09-16
Basic catalog microbenchmarks

mbutrovich:
We have several tasks ahead of us that should improve performance of the catalog, but we don't have simple microbenchmarks to evaluate their impact. We don't want to just rely on benchmarks of other components in the system that rely on the catalog to expose catalog regressions. This should make it easier to figure out if catalog performance issues are rolling downhill.


mbutrovich:
Note that this only profiles the read operations on the catalog. We want to optimize the writes as well, but in general are less concerned with profiling those right now.

benchmark/catalog/catalog_benchmark.cpp
```diff
@@ -0,0 +1,314 @@
...
+  }
+
+  txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+
+  state.SetItemsProcessed(state.iterations());
+}
+
+// NOLINTNEXTLINE
+BENCHMARK_DEFINE_F(CatalogBenchmark, GetNamespaceOid)(benchmark::State &state) {
+  // Create a database and check that it's immediately visible
```
GustavoAngulo: nit: Should be "Create a namespace"

benchmark/catalog/catalog_benchmark.cpp
```diff
@@ -0,0 +1,314 @@
...
+    return {table_oid, idx_oid};
+  }
+};
+
+// NOLINTNEXTLINE
+BENCHMARK_DEFINE_F(CatalogBenchmark, GetAccessor)(benchmark::State &state) {
+  auto *txn = txn_manager_->BeginTransaction();
+
+  // NOLINTNEXTLINE
+  for (auto _ : state) {
```
GustavoAngulo: I guess a general question I have is how do these benchmarks cover benchmarking future catalog caching. In particular, how many times does `for (auto _ : state) {...}` execute? Is that something set by us or gbenchmark? Is it deterministic?

mbutrovich: It is not deterministic, really. It's as many times as Google Benchmark needs to get a stable number. In our case, it will likely report the cached performance since it'll likely run enough iterations for it to report the stable, cached number.

benchmark/catalog/catalog_benchmark.cpp
mbutrovich: Owned by copy and paste.

PR #547, opened 2019-09-17
Parser/Plan Nodes/Catalog ManagedPointer and JSON refactor

lmwnshn:
**9/30 status**

Prior to merging upstream,

- [x] expressions own their children expressions, statements do not own any expressions, plan nodes do not own any expressions
- [x] JSON serialization and deserialization of parser expressions, plan nodes, some catalog things
- [x] clang-tidy
- [x] cpplint
- [x] Doxygen
- [x] All unit tests pass

After merging upstream,

- I had to fix some things (there were some new shared_ptrs added to the system as part of execution engine and expression util PRs, etc)
- I am no longer able to compile terrier as I run out of disk space and swap, so my system OOMs. I will repartition my drive tomorrow.
- I _think_ it should still pass everything, I'm heading home. See what Travis and Jenkins says.

**Concerns / notes**

- Why is DatabaseCatalog::MakeColumn deserializing from JSON? We don't have a particularly speedy JSON library.

**Next steps**

- Anyone else want to review this? About 2-3k lines added and removed since @yash620 last looked at it.


src/include/common/json.h
```diff
@@ -4,22 +4,23 @@
 #include "nlohmann/json.hpp"
 
 namespace terrier::common {
+// TODO(WAN): I think we should be using the adt_serializable, not this hack..
```
yash620: Yeah i agree, we should offload as much the json responsibility to some library that is meant to maintain it. Can you open an issue for this. Would help us avoid the difficulties you're facing right now in the future.

src/include/parser/create_function_statement.h
```diff
@@ -100,8 +85,8 @@ class CreateFunctionStatement : public SQLStatement {
    * @param as_type executable or query string
    */
   CreateFunctionStatement(bool replace, std::string func_name, std::vector<std::string> func_body,
-                          std::shared_ptr<ReturnType> return_type,
-                          std::vector<std::shared_ptr<FuncParameter>> func_parameters, PLType pl_type, AsType as_type)
+                          std::unique_ptr<ReturnType> return_type,
+                          std::vector<std::unique_ptr<FuncParameter>> func_parameters, PLType pl_type, AsType as_type)
```
yash620: Why is this required to be a unique ptr? This structure isn't storing the pointer anyway and it's just moving the data pointed by it. Is it a way of enforcing that func_parameters is deallocated by who ever is passing it in?

src/include/parser/create_statement.h
```diff
@@ -329,35 +296,27 @@ struct ColumnDefinition {
...
 
-  /**
-   * @return the expression that we're indexed on
-   */
-  std::shared_ptr<AbstractExpression> GetExpression() const {
+  /** @return the expression that we're indexed on */
+  common::ManagedPointer<AbstractExpression> GetExpression() const {
     TERRIER_ASSERT(expr_ != nullptr, "Names don't come with expressions.");
-    return expr_;
+    return common::ManagedPointer(expr_);
```
yash620: why the wrapper, isn't expr_ already a managed pointer?

src/include/parser/update_statement.h
```diff
@@ -54,38 +54,29 @@ class UpdateStatement : public SQLStatement {
...
-  std::shared_ptr<TableRef> GetUpdateTable() { return table_; }
+  /** @return update table target */
+  common::ManagedPointer<TableRef> GetUpdateTable() { return common::ManagedPointer(table_); }
 
-  /**
-   * @return update clauses
-   */
-  std::vector<std::shared_ptr<UpdateClause>> GetUpdateClauses() { return updates_; }
+  /** @return update clauses */
+  const std::vector<std::unique_ptr<UpdateClause>> &GetUpdateClauses() { return updates_; }
```
mush-zhang: Should we remove the const guard here? I may need to do std::move on the `updates_` when creating a logical update operator

src/include/parser/create_function_statement.h
yash620: Unique pointers seem like a good idea to me, it's enforcing the deallocation in code. I was just clarifying if that was the intended goal.

src/include/parser/insert_statement.h
```diff
@@ -37,55 +36,44 @@ class InsertStatement : public SQLStatement {
    * @param table_ref table
    * @param insert_values values to be inserted
    */
-  InsertStatement(std::shared_ptr<std::vector<std::string>> columns, std::shared_ptr<TableRef> table_ref,
-                  std::shared_ptr<std::vector<std::vector<std::shared_ptr<AbstractExpression>>>> insert_values)
+  InsertStatement(std::unique_ptr<std::vector<std::string>> columns, std::unique_ptr<TableRef> table_ref,
+                  std::unique_ptr<std::vector<std::vector<common::ManagedPointer<AbstractExpression>>>> insert_values)
```
yash620: Who is this vector of ManagedPointers? And how are they being deallocated?

src/include/parser/insert_statement.h
```diff
@@ -37,55 +36,44 @@ class InsertStatement : public SQLStatement {
...
   const InsertType type_;
-  const std::shared_ptr<std::vector<std::string>> columns_;
-  const std::shared_ptr<TableRef> table_ref_;
-  const std::shared_ptr<SelectStatement> select_;
-  // TODO(WAN): unsure about this one.
-  const std::shared_ptr<std::vector<std::vector<std::shared_ptr<AbstractExpression>>>> insert_values_;
+  const std::unique_ptr<std::vector<std::string>> columns_;
+  const std::unique_ptr<TableRef> table_ref_;
+  const std::unique_ptr<SelectStatement> select_;
+  const std::unique_ptr<std::vector<std::vector<common::ManagedPointer<AbstractExpression>>>> insert_values_;
```
yash620: I am confused with what's going on here. What is the need for a vector of vectors of managed pointers? Where are these managed pointers being created and how are they being deallocated? I'm having trouble following the execution flow here.

src/include/parser/postgresparser.h
```diff
@@ -11,9 +11,26 @@
...
+ public:
+  void AddStatement(std::unique_ptr<SQLStatement> statement) { statements_.emplace_back(std::move(statement)); }
+
+  void AddExpression(std::unique_ptr<AbstractExpression> expression) {
+    expressions_.emplace_back(std::move(expression));
+  }
+
+  const std::vector<std::unique_ptr<SQLStatement>> &GetStatements() { return statements_; }
+
+  const std::vector<std::unique_ptr<AbstractExpression>> &GetExpressions() { return expressions_; }
```
yash620: Should this be returned as a std::vector<std::managed_ptr<AbstractExpression>>? From the looks of it when ParseResult get destructed expressions_ get's cleaned up so all the AbstractExpressions would also be cleaned up. So doesn't that mean ParseResult is managing this data?

src/include/parser/postgresparser.h
```diff
@@ -36,7 +53,7 @@ class PostgresParser {
    * @param query_string query string to be parsed
    * @return unique pointer to parse tree
    */
-  std::vector<std::unique_ptr<parser::SQLStatement>> BuildParseTree(const std::string &query_string);
+  ParseResult BuildParseTree(const std::string &query_string);
```
yash620: Good call in abstracting this away into its own object, makes it much easier to understand

src/include/parser/postgresparser.h
```diff
@@ -78,106 +95,105 @@ class PostgresParser {
    * @param root list of parsed nodes
    * @return SQLStatementList corresponding to the parsed node list
    */
-  static std::vector<std::unique_ptr<parser::SQLStatement>> ListTransform(List *root);
+  static void ListTransform(ParseResult *parse_result, List *root);
```
yash620: Add documentation above for @param ParseResult

src/include/parser/postgresparser.h
```diff
@@ -78,106 +95,105 @@ class PostgresParser {
...
 
   /**
    * Transforms a single node in the parse list into a terrier SQLStatement object.
    * @param node parsed node
    * @return SQLStatement corresponding to the parsed node
    */
-  static std::unique_ptr<parser::SQLStatement> NodeTransform(Node *node);
+  static std::unique_ptr<SQLStatement> NodeTransform(ParseResult *parse_result, Node *node);
```
yash620: Update documentation to include `parse_result`

src/include/parser/postgresparser.h
```diff
@@ -78,106 +95,105 @@ class PostgresParser {
...
+  static std::unique_ptr<SQLStatement> CreateIndexTransform(ParseResult *parse_result, IndexStmt *root);
+  static std::unique_ptr<SQLStatement> CreateSchemaTransform(ParseResult *parse_result, CreateSchemaStmt *root);
+  static std::unique_ptr<SQLStatement> CreateTriggerTransform(ParseResult *parse_result, CreateTrigStmt *root);
+  static std::unique_ptr<SQLStatement> CreateViewTransform(ParseResult *parse_result, ViewStmt *root);
 
   // CREATE helpers
   using ColumnDefTransResult = struct {
     std::unique_ptr<ColumnDefinition> col_;
-    std::vector<std::shared_ptr<ColumnDefinition>> fks_;
+    std::vector<std::unique_ptr<ColumnDefinition>> fks_;
```
yash620: nit: fks_ doesn't seem like a very description name. I'm not sure what this is

src/include/parser/prepare_statement.h
```diff
@@ -29,8 +29,8 @@ class PrepareStatement : public SQLStatement {
    * @param query - the parsed form of statement
    * @param placeholders - placeholder values? (explain)
    */
-  PrepareStatement(std::string name, std::shared_ptr<SQLStatement> query,
-                   std::vector<std::shared_ptr<ParameterValueExpression>> placeholders)
+  PrepareStatement(std::string name, std::unique_ptr<SQLStatement> query,
+                   std::vector<common::ManagedPointer<ParameterValueExpression>> placeholders)
```
yash620: i don't see placeholders being assigned. I think you're missing `placeholders_(placeholders)`

src/include/parser/select_statement.h
```diff
@@ -297,69 +307,53 @@ class SelectStatement : public SQLStatement {
...
-  std::shared_ptr<TableRef> GetSelectTable() { return from_; }
+  /** @return table being selected from */
+  common::ManagedPointer<TableRef> GetSelectTable() { return common::ManagedPointer(from_); }
 
-  /**
-   * @return select condition
-   */
-  std::shared_ptr<AbstractExpression> GetSelectCondition() { return where_; }
+  /** @return select condition */
+  common::ManagedPointer<AbstractExpression> GetSelectCondition() { return common::ManagedPointer(where_); }
```
yash620: This is already stored as a ManagedPointer, I don't think the wrapper is necessary.

src/include/parser/table_ref.h
```diff
@@ -49,17 +49,17 @@ class JoinDefinition {
...
    * @return right table
    */
-  std::shared_ptr<TableRef> GetRightTable() { return right_; }
+  common::ManagedPointer<TableRef> GetRightTable() { return common::ManagedPointer(right_); }
 
   /**
    * @return join condition
    */
-  std::shared_ptr<AbstractExpression> GetJoinCondition() { return condition_; }
+  common::ManagedPointer<AbstractExpression> GetJoinCondition() { return common::ManagedPointer(condition_); }
```
yash620: condition_ is already stored as a managed pointer is the wrapper needed?

src/include/parser/update_statement.h
```diff
@@ -54,38 +54,29 @@ class UpdateStatement : public SQLStatement {
...
 
-  /**
-   * @return update condition
-   */
-  std::shared_ptr<AbstractExpression> GetUpdateCondition() { return where_; }
+  /** @return update condition */
+  common::ManagedPointer<AbstractExpression> GetUpdateCondition() { return common::ManagedPointer(where_); }
```
yash620: wrapper seems unnecessary

src/include/planner/plannodes/abstract_plan_node.h
```diff
@@ -42,29 +42,29 @@ class AbstractPlanNode {
...
   };
 
   /**
    * Constructor for the base AbstractPlanNode. Derived plan nodes should call this constructor to set output_schema
    * @param children child plan nodes
    * @param output_schema Schema representing the structure of the output of this plan node
    */
-  AbstractPlanNode(std::vector<std::shared_ptr<AbstractPlanNode>> &&children,
-                   std::shared_ptr<OutputSchema> output_schema)
+  AbstractPlanNode(std::vector<std::unique_ptr<AbstractPlanNode>> &&children,
```
yash620: If you're using the move constructor for children__ shouldn't the pattern be `std::unique_ptr<std::vector<.... `?

src/include/planner/plannodes/abstract_plan_node.h
```diff
@@ -84,7 +84,7 @@ class AbstractPlanNode {
   /**
    * @return child plan nodes
    */
-  const std::vector<std::shared_ptr<AbstractPlanNode>> &GetChildren() const { return children_; }
+  const std::vector<std::unique_ptr<AbstractPlanNode>> &GetChildren() const { return children_; }
```
yash620: I feel like this should be `std::vector<std::manged_ptr<AbstractPlanNode>>`. If this function is giving out a reference to this vector and let's say someone makes a deep copy of the vector, how would unique_ptrs work in that case? Would there be two unique_ptrs pointing to the same location in memory? And when that  copy unique_ptr goes out of scope is memory it's pointing to freed? Wouldn't that be bad since data would be freed while abstract_plan_node is still using it?

src/include/planner/plannodes/abstract_scan_plan_node.h
```diff
@@ -123,7 +123,7 @@ class AbstractScanPlanNode : public AbstractPlanNode {
   /**
    * @return predicate used for performing scan
    */
-  const std::shared_ptr<parser::AbstractExpression> &GetScanPredicate() const { return scan_predicate_; }
+  const std::unique_ptr<parser::AbstractExpression> &GetScanPredicate() const { return scan_predicate_; }
```
yash620: This should be returning a `std::managed_ptr<parser::AbstractExpression>`

src/include/planner/plannodes/aggregate_plan_node.h
```diff
@@ -119,7 +119,7 @@ class AggregatePlanNode : public AbstractPlanNode {
   /**
    * @return pointer to predicate for having clause
    */
-  const std::shared_ptr<parser::AbstractExpression> &GetHavingClausePredicate() const {
+  const std::unique_ptr<parser::AbstractExpression> &GetHavingClausePredicate() const {
```
yash620: This should be returning `std::ManagedPointer<parser::AbstractExpression>`

src/include/planner/plannodes/create_table_plan_node.h
```diff
@@ -595,15 +597,15 @@ class CreateTablePlanNode : public AbstractPlanNode {
...
+    Builder &ProcessCheckConstraint(const std::unique_ptr<parser::ColumnDefinition> &col) {
       auto check_cols = std::vector<std::string>();
 
       // TODO(Gus,Wen) more expression types need to be supported
       if (col->GetCheckExpression()->GetReturnValueType() == type::TypeId::BOOLEAN) {
         check_cols.push_back(col->GetColumnName());
 
-        parser::ConstantValueExpression *const_expr_elem =
-            dynamic_cast<parser::ConstantValueExpression *>(col->GetCheckExpression()->GetChild(1).get());
+        auto const_expr_elem =
```
yash620: why change this to auto? Using auto makes code harder to read.

src/include/planner/plannodes/create_trigger_plan_node.h
```diff
@@ -138,8 +138,8 @@ class CreateTriggerPlanNode : public AbstractPlanNode {
      * Build the create trigger plan node
      * @return plan node
      */
-    std::shared_ptr<CreateTriggerPlanNode> Build() {
-      return std::shared_ptr<CreateTriggerPlanNode>(new CreateTriggerPlanNode(
+    std::unique_ptr<CreateTriggerPlanNode> Build() {
```
yash620: I am kind of confuse. What is the purpose of Build and how is it different from `CreateTriggerPlanNode` . Why not expose the constructor?

src/include/planner/plannodes/hash_join_plan_node.h
```diff
@@ -54,25 +54,26 @@ class HashJoinPlanNode : public AbstractJoinPlanNode {
       return *this;
     }
 
+    // TODO(WAN) do we want to invalidate the builder after build?
```
yash620: Wouldn't there be issues from reusing the builder? A lot of the variables are moved from the Builder to the Node, so I don't think you can call build again

src/include/planner/plannodes/hash_plan_node.h
```diff
@@ -81,7 +81,7 @@ class HashPlanNode : public AbstractPlanNode {
   /**
    * @return keys to be hashed on
    */
-  const std::vector<std::shared_ptr<parser::AbstractExpression>> &GetHashKeys() const { return hash_keys_; }
+  const std::vector<std::unique_ptr<parser::AbstractExpression>> &GetHashKeys() const { return hash_keys_; }
```
yash620: I think this should be `std::vector<std::ManagedPointer<....>>`

src/include/planner/plannodes/output_schema.h
```diff
@@ -178,7 +182,7 @@ class OutputSchema {
     /**
      * @return the expression used to derive the intermediate column
      */
-    const std::shared_ptr<parser::AbstractExpression> &GetExpression() const { return expr_; }
+    const std::unique_ptr<parser::AbstractExpression> &GetExpression() const { return expr_; }
```
yash620: This should be a managed pointer

src/include/parser/postgresparser.h
```diff
@@ -11,9 +11,26 @@
...
 
+class ParseResult {
+ public:
+  void AddStatement(std::unique_ptr<SQLStatement> statement) { statements_.emplace_back(std::move(statement)); }
+
+  void AddExpression(std::unique_ptr<AbstractExpression> expression) {
+    expressions_.emplace_back(std::move(expression));
+  }
+
+  const std::vector<std::unique_ptr<SQLStatement>> &GetStatements() { return statements_; }
```
yash620: This should be returned as a std::vector<std::managed_ptr>? From the looks of it when ParseResult get destructed expressions_ get's cleaned up so all the AbstractExpressions would also be cleaned up. So doesn't that mean ParseResult is managing this data?

src/include/parser/create_statement.h
lmwnshn: Good catch.

src/include/catalog/schema.h
```diff
@@ -178,7 +191,7 @@ class Schema {
     col_oid_t oid_;
 
     // TODO(John) this should become a unique_ptr as part of addressing #489
```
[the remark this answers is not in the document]
lmwnshn: :+1:

src/include/parser/postgresparser.h
lmwnshn: Modified to iterate over the vector instead.

src/include/planner/plannodes/output_schema.h
lmwnshn: Good catch

src/include/planner/plannodes/hash_plan_node.h
lmwnshn: Yup

src/include/planner/plannodes/hash_join_plan_node.h
lmwnshn: Yeah. I'm not sure how we'd enforce it right now though.

src/include/planner/plannodes/create_trigger_plan_node.h
lmwnshn: Not particularly familiar with this usage.
@wenxuanqiu Opinions?

src/include/planner/plannodes/create_table_plan_node.h
lmwnshn: We have the type on the RHS with the .CastManagedPointerTo<Type>, but I have no objection to writing it out. Changed.

src/include/planner/plannodes/aggregate_plan_node.h
lmwnshn: Search and replace strikes again. Thanks!

src/include/planner/plannodes/abstract_scan_plan_node.h
lmwnshn: Nice

src/include/planner/plannodes/abstract_plan_node.h
lmwnshn: Yep

src/include/parser/table_ref.h
lmwnshn: Nope, a remnant of worse days.

src/include/parser/select_statement.h
lmwnshn: I think you think right.

src/include/parser/prepare_statement.h
lmwnshn: https://github.com/cmu-db/terrier/blob/5d983d30e42dcd52c11e07c36a6141ded2acbae5/src/include/parser/prepare_statement.h#L37 ?

src/include/parser/postgresparser.h
lmwnshn: Foreign keys, I will add a comment.

src/include/parser/postgresparser.h
lmwnshn: Done, also removed the @return

src/include/parser/postgresparser.h
lmwnshn: Done

src/include/parser/postgresparser.h
lmwnshn: I'm a little unhappy about how it gets passed around through every single function call..

src/include/parser/postgresparser.h
lmwnshn: Yes

src/include/parser/create_function_statement.h
lmwnshn: If I understand correctly, your question @yash620 is "why is the class storing a pointer to the thing instead of storing the thing directly".

This is also a question that I had back when I originally ported the parser last year. I believe a meta reason for keeping it this way is that the rest of the Peloton system (optimizer, binder, etc) would have expected it to remain this way, and this minimized friction for future porting.

If the user needs another reference to a unique pointer, they should be able to get a managed pointer from it. Let me know if you see somewhere like that. We do want to avoid accidentally doing costly deep copies everywhere, supposedly that's how Peloton became a lot slower.

But to return to the original question; I think with the introduction of managed pointers there is actually an advantage to keeping it as a unique pointer. Now, when someone asks you for your ReturnType or whatever, you hand off a managed pointer and it remains a managed pointer and you know that you don't own it.

It is possible that your question is instead "why isn't it a vector<FuncParameter>". I _think_ the optimizer and binder would prefer it to be vector<unique_ptr<FuncParameter>>, but I could be mistaken.

lmwnshn: Long story short; the parser was initially ported without any customers, so we kept a lot of internal representation the same to make it easier to port the customers over. That said, if a refactor makes sense, we should probably do it.

src/include/parser/create_function_statement.h
```diff
@@ -126,7 +111,7 @@ class CreateFunctionStatement : public SQLStatement {
   /**
    * @return return type
    */
-  std::shared_ptr<ReturnType> GetFuncReturnType() { return return_type_; }
+  common::ManagedPointer<ReturnType> GetFuncReturnType() { return common::ManagedPointer(return_type_); }
```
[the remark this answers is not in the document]
lmwnshn: The entire CREATE FUNCTION stuff is pretty weird in general.

src/include/parser/expression/abstract_expression.h
```diff
@@ -215,48 +213,45 @@ class AbstractExpression {
...
+   * tree modifications.
    */
   int depth_ = -1;
   /**
-   * The flag indicating if there's sub-query in the current expression
-   */
+   * MUTABLE Flag indicating if there's a sub-query in the current expression or in any of its children.
+   *
+   * TODO(WAN): check with LING on why we need this and whether we can roll DeriveSubqueryFlag into DeriveDepth.
+   * */
```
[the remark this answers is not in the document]
mush-zhang: has_subquery: 
this is essentially use to detect if the query predicate IsSupportedConjunctivePredicate ; according to Peloton comment:
```
 // Currently support : 1. expr without subquery
  // 2. subquery without disjunction. Since the expr is already one of the
  // conjunctive exprs, we'll only need to check if the root level is an
  // operator with subquery
```
I wonder if that is true; we might be able to remove the check if we get everything supported

Depth:
according to Peloton comment:
```
// Depth is used to detect correlated subquery, it is set in the binder,
    // if a TVE has depth less than the depth of the current operator, then it is correlated predicate
 ```

File: src/parser/postgresparser.cpp [hunk 29]
```diff
@@ -70,112 +72,108 @@ std::vector<std::unique_ptr<SQLStatement>> PostgresParser::BuildParseTree(const
...
 
-std::vector<std::unique_ptr<SQLStatement>> PostgresParser::ListTransform(List *root) {
-  std::vector<std::unique_ptr<SQLStatement>> result;
-
+void PostgresParser::ListTransform(ParseResult *parse_result, List *root) {
   if (root != nullptr) {
     for (auto cell = root->head; cell != nullptr; cell = cell->next) {
       auto node = static_cast<Node *>(cell->data.ptr_value);
-      result.emplace_back(NodeTransform(node));
+      parse_result->AddStatement(NodeTransform(parse_result, node));
```
yash620: This seems kind of iffy to me that you have to call AddStatement on parse result and you're also passing parse_result on NodeTransform. In some cases NodeTransform seems to directly modify parse_result, such as in the cases where WhereTransform is called, the WhereTransform calls `parse_result->AddExpression(std::move(expr));` So why not do something similar and let the Transforms add the statement to parse_result when necessary?

src/parser/postgresparser.cpp
```diff
@@ -454,12 +452,12 @@ std::unique_ptr<AbstractExpression> PostgresParser::AExprTransform(A_Expr *root)
 }
 
 // Postgres.BoolExpr -> terrier.ConjunctionExpression
-std::unique_ptr<AbstractExpression> PostgresParser::BoolExprTransform(BoolExpr *root) {
+std::unique_ptr<AbstractExpression> PostgresParser::BoolExprTransform(ParseResult *parse_result, BoolExpr *root) {
   std::unique_ptr<AbstractExpression> result;
-  std::vector<std::shared_ptr<AbstractExpression>> children;
+  std::vector<std::unique_ptr<AbstractExpression>> children;
```
yash620: This should be a vector of managed pointers. When you move children in and children is a vector of unique_ptrs are the unique_ptrs maintained? When children is cleaned up at the end of this function would the unique_ptr destructor be called or is children empty at this point because all of its contents would have been moved?

yash620: Nvm I think it should be fine based on move semantics

src/parser/postgresparser.cpp
```diff
@@ -556,42 +555,42 @@ std::unique_ptr<AbstractExpression> PostgresParser::ColumnRefTransform(ColumnRef
...
-    std::vector<std::shared_ptr<AbstractExpression>> children;
+    std::vector<std::unique_ptr<AbstractExpression>> children;
     if (root->agg_star_) {
       auto child = std::make_unique<StarExpression>();
       children.emplace_back(std::move(child));
       result = std::make_unique<AggregateExpression>(agg_fun_type, std::move(children), root->agg_distinct_);
     } else if (root->args_->length < 2) {
       auto expr_node = reinterpret_cast<Node *>(root->args_->head->data.ptr_value);
-      auto child = ExprTransform(expr_node);
+      auto child = ExprTransform(parse_result, expr_node, nullptr);
```
yash620: this isn't being explicitly being casted so try to avoid using auto so the type is more evident

src/parser/postgresparser.cpp
```diff
@@ -454,12 +452,12 @@ std::unique_ptr<AbstractExpression> PostgresParser::AExprTransform(A_Expr *root)
...
 // Postgres.BoolExpr -> terrier.ConjunctionExpression
-std::unique_ptr<AbstractExpression> PostgresParser::BoolExprTransform(BoolExpr *root) {
+std::unique_ptr<AbstractExpression> PostgresParser::BoolExprTransform(ParseResult *parse_result, BoolExpr *root) {
   std::unique_ptr<AbstractExpression> result;
-  std::vector<std::shared_ptr<AbstractExpression>> children;
+  std::vector<std::unique_ptr<AbstractExpression>> children;
   for (auto cell = root->args_->head; cell != nullptr; cell = cell->next) {
     auto node = reinterpret_cast<Node *>(cell->data.ptr_value);
-    children.emplace_back(ExprTransform(node));
+    children.emplace_back(ExprTransform(parse_result, node, nullptr));
```
yash620: should this be move(ExprTransform) ? Doesn't emplace_back make a constructor call and unique_ptrs aren't copiable so this has to be an explicit move constructor.

src/parser/postgresparser.cpp
```diff
@@ -556,42 +555,42 @@ std::unique_ptr<AbstractExpression> PostgresParser::ColumnRefTransform(ColumnRef
...
     // normal functions (built-in functions or UDFs)
     func_name = (reinterpret_cast<value *>(root->funcname_->tail->data.ptr_value))->val_.str_;
-    std::vector<std::shared_ptr<AbstractExpression>> children;
+    std::vector<std::unique_ptr<AbstractExpression>> children;
 
     if (root->args_ != nullptr) {
       for (auto cell = root->args_->head; cell != nullptr; cell = cell->next) {
         auto expr_node = reinterpret_cast<Node *>(cell->data.ptr_value);
-        children.emplace_back(ExprTransform(expr_node));
+        children.emplace_back(ExprTransform(parse_result, expr_node, nullptr));
```
yash620: Same as above, does it need an explicit move call?

src/parser/postgresparser.cpp
```diff
@@ -688,17 +686,17 @@ std::unique_ptr<AbstractExpression> PostgresParser::SubqueryExprTransform(SubLin
...
 
 // Postgres.TypeCast -> terrier.TypeCastExpression
-std::unique_ptr<AbstractExpression> PostgresParser::TypeCastTransform(TypeCast *root) {
+std::unique_ptr<AbstractExpression> PostgresParser::TypeCastTransform(ParseResult *parse_result, TypeCast *root) {
   auto type_name = reinterpret_cast<value *>(root->type_name_->names_->tail->data.ptr_value)->val_.str_;
   auto type = ColumnDefinition::StrToValueType(type_name);
-  std::vector<std::shared_ptr<AbstractExpression>> children;
-  children.emplace_back(ExprTransform(root->arg_));
+  std::vector<std::unique_ptr<AbstractExpression>> children;
+  children.emplace_back(ExprTransform(parse_result, root->arg_, nullptr));
```
yash620: I think this would need an explicit move call

src/parser/postgresparser.cpp
```diff
@@ -751,42 +749,45 @@ std::unique_ptr<SelectStatement> PostgresParser::SelectTransform(SelectStmt *roo
...
+      PARSER_LOG_AND_THROW("SelectTransform", "Select operation", root->type_);
     }
   }
 
   return result;
 }
 
 // Postgres.SelectStmt.targetList -> terrier.SelectStatement.select_
-std::vector<std::shared_ptr<AbstractExpression>> PostgresParser::TargetTransform(List *root) {
+std::vector<common::ManagedPointer<AbstractExpression>> PostgresParser::TargetTransform(ParseResult *parse_result,
```
yash620: Why is this different from the other transforms? All the other transforms return unique pointers

src/parser/postgresparser.cpp
```diff
@@ -751,42 +749,45 @@ std::unique_ptr<SelectStatement> PostgresParser::SelectTransform(SelectStmt *roo
...
   }
 
-  std::vector<std::shared_ptr<AbstractExpression>> result;
+  std::vector<common::ManagedPointer<AbstractExpression>> result;
   for (auto cell = root->head; cell != nullptr; cell = cell->next) {
     auto target = reinterpret_cast<ResTarget *>(cell->data.ptr_value);
-    result.emplace_back(ExprTransform(target->val_, target->name_));
+    auto expr = ExprTransform(parse_result, target->val_, target->name_);
+    auto expr_managed = common::ManagedPointer(expr);
+    parse_result->AddExpression(std::move(expr));
```
yash620: This seems kind of awkward to me. Can there be situations where parse_result is deallocated before the result from this function is used. In those cases you would have segfaults

src/parser/postgresparser.cpp
```diff
@@ -846,35 +846,41 @@ std::unique_ptr<TableRef> PostgresParser::FromTransform(SelectStmt *select_root)
...
+  auto having = common::ManagedPointer<AbstractExpression>(nullptr);
   if (having_node != nullptr) {
-    having = ExprTransform(having_node);
+    auto expr = ExprTransform(parse_result, having_node, nullptr);
+    having = common::ManagedPointer(expr);
+    parse_result->AddExpression(std::move(expr));
   }
 
-  auto result = std::make_unique<GroupByDescription>(std::move(columns), std::move(having));
+  auto result = std::make_unique<GroupByDescription>(std::move(columns), having);
```
yash620: Same issue as above, what is preventing parse_result from being deallocated before the result is used?

src/parser/postgresparser.cpp
lmwnshn: Currently, it is the responsibility of the caller of BuildParseTree to keep the parsed results alive while people need to use it. How else would you manage this?

src/parser/postgresparser.cpp
lmwnshn: The caller receives both the "tree that relies on ParseResult being alive" and "ParseResult".

src/parser/postgresparser.cpp
lmwnshn: This is one of the helpers for creating a select statement. If you see the header file, the top level SQLStatements get unique ptrs, the helper and stuff-in-them-to-abstract-expression are managed ptrs.

src/parser/postgresparser.cpp
lmwnshn: See above.

src/parser/postgresparser.cpp
lmwnshn: Doesn't emplace_back use std::forward? https://en.cppreference.com/w/cpp/container/vector/emplace_back

src/parser/postgresparser.cpp
lmwnshn: I'm not particularly convinced, given that everyone is developing in CLion and the type is a control-mouseover away.. the code is messy enough already. There's also a lot of places that this comment could apply, and I'd rather keep it consistently terse (a reader is expected to remember what ExprTransform returns) than to have it in some places / all places.

Though I'm not hugely opposed to it either.

src/parser/postgresparser.cpp
lmwnshn: I vaguely remember wanting a convention of, "automatically updates the parse result with non-owned stuff, but always return the top-level thing".

I believe there are some instances in the code that need the result of a NodeTransform, to _not_ actually be in the parse result but for some further "stuff it into some other class". For example, PrepareTransform.

src/include/planner/plannodes/hash_join_plan_node.h
lmwnshn: In general it seems ridiculous to have a builder that can only be used once.

apavlo:
@mush-zhang has given this her seal of approval. We're merging this now!!!

PR #550, opened 2019-09-18
ASAN flag wasn't being passed to the IR compiler

mbutrovich:
This fixes it, in theory.


apavlo:
What does this do?

mbutrovich:
I actually don't think the coverage flag should go to the IR compiler. The flag just tells the compiler files you want to generate coverage _for_, which I don't think anything that the IR compiler handles we want coverage calculated on anyway. It's all TPL files, right?

This PR now just makes sure that ASAN is enabled on the IR compilation pass if it's enabled for the overall project.

jrolli:
This is probably best practice, but I'm not sure it actually does anything since the IR and TPL are basically wrappers around our C++ libraries.  Are there any (de)allocations or potentially unsafe memory accesses by the functions in the IR that we can test to see if (and how) this changes ASAN behavior?

mbutrovich:
After speaking with @amlatyrngom about it, this would probably only catch a leak in the bytecode_handlers.h file. It's unlikely that will ever happen, but it probably doesn't hurt. He said our existing setup already catches leaks in TPL files.

PR #553, opened 2019-09-19
Better GetIndexes method

### LINKED ISSUE #513: Create new Catalog GetIndexes function
Currently, `DatabaseCatalog::GetIndexes` returns a vector of index oids for indexes on a table. However, to get all the index pointers and schemas, you need to make individual calls with `DatabaseCatalog::GetIndex` and `DatabaseCatalog::GetIndexSchema`, **one oid at a time.** Each of these calls is allocating memory to build table and index key ProjectedRows, which scales with the number of indexes on a table. 

Recovery, and later the insert and delete executors, have use cases for getting all the indexes and index schemas on a table. We can have a single call that takes a vector of oids and returns a vector of pairs of index pointers and schemas, allowing us to reduce the number of memory allocations from # of oids (by using the current individual call style) to a constant number, as well as only having to construct a single `ProjectedRowInitializer`.

GustavoAngulo:
Resolves #513

As #513 describes, to get all index objects and schemas on a table, we have to make a call to get all the index oids, then individual calls for each oid to get the index object and index schema. This function condenses all that logic into one call, achieving much better performance by reducing number of allocations and `ProjectedRow` initializations.


src/catalog/catalog_accessor.cpp
```diff
@@ -88,6 +88,11 @@ std::vector<constraint_oid_t> CatalogAccessor::GetConstraints(table_oid_t table)
 
 std::vector<index_oid_t> CatalogAccessor::GetIndexes(table_oid_t table) const { return dbc_->GetIndexes(txn_, table); }
 
+std::vector<std::pair<common::ManagedPointer<storage::index::Index>, const IndexSchema &>>
+CatalogAccessor::GetIndexObjects(terrier::catalog::table_oid_t table) {
```
mbutrovich: Style nit: shouldn't need the fully qualified namespace for table_oid_t here.

src/catalog/database_catalog.cpp
```diff
@@ -1083,6 +1088,88 @@ const IndexSchema &DatabaseCatalog::GetIndexSchema(transaction::TransactionConte
   return *reinterpret_cast<IndexSchema *>(ptr_pair.first);
 }
 
+std::vector<std::pair<common::ManagedPointer<storage::index::Index>, const IndexSchema &>>
+DatabaseCatalog::GetIndexObjects(terrier::transaction::TransactionContext *txn, terrier::catalog::table_oid_t table) {
```
mbutrovich: For namespace nits.

src/catalog/database_catalog.cpp
```diff
@@ -1083,6 +1088,88 @@ const IndexSchema &DatabaseCatalog::GetIndexSchema(transaction::TransactionConte
...
+  std::vector<storage::TupleSlot> index_scan_results;
+  indexes_table_index_->ScanKey(*txn, *indexes_key_pr, &index_scan_results);
+
+  // If we found no indexes, return an empty list
+  if (index_scan_results.empty()) {
+    delete[] buffer;
+    return {};
+  }
+
+  std::vector<index_oid_t> index_oids;
```
mbutrovich: Stupid performance optimization: can we reserve the size of index_oids since we know it from the size of index_scan_results?

PR #557, opened 2019-09-23
TrafficCop

YuzeLiao:
Roadmap:
1. ~~Run TPC-C with sqlite TrafficCop.~~(Done). 
- (Solved)~~Error happens when loading TPC-C database,  i.e., when --load=true with oltp-bench. It typically happens when doing batch insert -- Error Message: "ERROR - BatchUpdateException Cause => Expected command status BEGIN, got ." （Not sure if this is related to abort.)~~
- (Solved)~~Infinite loop between BIND, DESCRIBE, and EXECUTE for the same query. Suspect that something is wrong with the returned message.~~
- (Solved)~~Can successfully load DB. But Update fails when executing the benchmark. Causes have been found, expected to fix it soon.~~

2. Replace sqlite's parser,  binder, and engine with terrier's.


apavlo:
This has been replaced with #601

PR #558, opened 2019-09-24
Add recovery benchmark to nightly runs and new index recovery benchmark

GustavoAngulo:
Adds the recovery benchmark to the nightly benchmark runs, as well as adds a new index recovery benchmark. The existing high abort rate benchmark was removed because it was non-deterministic in the amount of aborts, and thus the performance of the benchmark.


GustavoAngulo:
Waiting on #559

mbutrovich:
@GustavoAngulo Do these numbers sound about right to you?
```
pass 3.337e+04 0         0       +0    config   0    RecoveryBenchmark            OLTPWorkload
pass 8062      0         0       +0    config   0    RecoveryBenchmark            HighAbortRate
pass 3.003e+04 0         0       +0    config   0    RecoveryBenchmark            HighStress

```

GustavoAngulo:
@mbutrovich Yes that looks about right. However what I'm realizing now is that the throughput I'm exporting is how many **committed** recovered txns did we process, not how many txns we processed or executed before recovery. That's why the Abort benchmark has such low throughput. Which do you throughput do you think I should use? How many txns we executed before recovery, and how fast we recovered from the log file?

mbutrovich:
In looking at the benchmarks, I think the input might be a bit malformed for what we're trying to measure. You run random workloads to generate the log files, then recover from them. The problem as I see it is that the input can change based on the characteristics of the system (or the random number generator), leading to changes in your recovery benchmark. I would have more deterministic inputs and then measure your recovery rate based on that.

GustavoAngulo:
Re-requesting a review since I added the new benchmark.

PR #559, opened 2019-09-24
Update LLVM_DIR path for llvm@8 Homebrew package

mbutrovich:
The `llvm` Homebrew package now defaults to 9. We don't want that yet (all the clang-tools and LLVM are untested with the repo right now), so we need to specify 8.


mbutrovich:
Mac builds passed which is all this is relevant to. Coverage is gonna time out. Gonna merge this since it's blocking @GustavoAngulo.

PR #563, opened 2019-09-26
Multiplatform thread CPU timer

mbutrovich:
@malin1993ml is interested in collecting CPU time metrics in the future. Motivated by that, this is a way to get thread-scoped CPU time measurements. macOS has the `get_rusage` syscall, but doesn't allow it on a per-thread basis like modern Linux kernels do. I had to resort to the Mach `thread_info` stuff instead. See the comments for sources, but I compared the numbers to what Google Benchmark gives out for its timers and they seem comparable.


test/common/thread_cpu_timer_test.cpp
```diff
@@ -0,0 +1,50 @@
...
+TEST(ThreadCPUTimerTests, BasicTest) {
+  common::ThreadCPUTimer timer;
+  volatile uint64_t j;
+  const uint64_t num_iters = 100000;
+
+  timer.Start();
+  j = 0;
+  for (uint64_t i = 0; i < num_iters; i++) {
+    j = i * 2;
+    EXPECT_EQ(j, i * 2);
```
malin1993ml: Is this for keeping the compiler away from optimizing out the computation?

mbutrovich: That and GCC would yell at me if I didn't read the value back out even with UNUSED_ATTRIBUTE on the declaration.

Issue #564, opened 2019-09-26
Modify run_micro_bench.py to only have a single list of benchmarks


apavlo:
Any time we add a new microbenchmark, you have add it to `run_micro_bench.py` in two places:

1. https://github.com/cmu-db/terrier/blob/master/script/micro_bench/run_micro_bench.py#L31-L40
2. https://github.com/cmu-db/terrier/blob/master/script/micro_bench/run_micro_bench.py#L716-L725

From the what I can tell, there is no reason to repeat this list and thus it is unnecessarily complex. We should refactor the code to have a single list of benchmarks at the top.

I will fix this.

apavlo:
Fixed in #565

PR #565, opened 2019-09-27
Cleaning up nightly microbench script

apavlo:
This is a quick and dirty clean up of the microbench python script to make the configuration more obvious. I removed unused variables/methods and pass around the `Config` object as needed. This is for #564 

https://youtu.be/dIdynsc0yBc


mbutrovich:
It looks like it's not grabbing the lookback data correctly for the benchmarks anymore.

apavlo:
I am not sure that this script is reliable? Sometimes it reports the previous results and sometimes it does not. There is nothing in my changes that would affect this. Here is my latest run:

http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-565/3/pipeline/20#step-128-log-788

And here is one from the nightly run that shows that it has zero previous results:

http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier-nightly/detail/terrier-nightly/378/pipeline#log-1784

PR #574, opened 2019-10-01
Refactoring for network protocols

tpan496:
This refactoring separates ITP classes from Postgres and establishes abstract classes that will define future APIs. 

- `ITPNetworkCommand` and `PostgresNetworkCommand` now both inherit from `AbstractNetworkCommand`. In this way, I will be able to separate out the `PostgresPacketWriter`.
- Updated network command classes to return 'AbstractNetworkCommand' pointers. 
- Added `AbstractPacketWriter` as an API for packet writers. `AbstractPacketWriter` would also allow the `transition::Exec` function to look the same across ITP/Postgres commands.


src/include/network/abstract_packet_writer.h
```diff
@@ -0,0 +1,399 @@
...
+        return AppendRawValue(val);
+      case 2:
+        return AppendRawValue(_CAST(T, htobe16(_CAST(uint16_t, val))));
+      case 4:
+        return AppendRawValue(_CAST(T, htobe32(_CAST(uint32_t, val))));
+      case 8:
+        return AppendRawValue(_CAST(T, htobe64(_CAST(uint64_t, val))));
+        // Will never be here due to compiler optimization
+      default:
+        throw NETWORK_PROCESS_EXCEPTION("");
```
apavlo: Empty exception message

src/include/network/abstract_packet_writer.h
```diff
@@ -0,0 +1,399 @@
...
+   * @tparam T type of value to read off. Has to be size 1, 2, 4, or 8.
+   * @param val value to write
+   * @return self-reference for chaining
+   */
+  template <typename T>
+  AbstractPacketWriter &AppendValue(T val) {
+    // We only want to allow for certain type sizes to be used
+    // After the static assert, the compiler should be smart enough to throw
+    // away the other cases and only leave the relevant return statement.
+    static_assert(sizeof(T) == 1 || sizeof(T) == 2 || sizeof(T) == 4 || sizeof(T) == 8, "Invalid size for integer");
```
apavlo: How is `static_assert` different than `TERRIER_ASSERT`?

src/include/network/connection_dispatcher_task.h
```diff
@@ -37,7 +38,8 @@ class ConnectionDispatcherTask : public common::NotifiableTask {
    * @param thread_registry DedicatedThreadRegistry dependency needed because it eventually spawns more threads in
    * RunTask
    */
-  ConnectionDispatcherTask(uint32_t num_handlers, int listen_fd, common::DedicatedThreadOwner *dedicated_thread_owner,
+  ConnectionDispatcherTask(bool hot_standby, uint32_t num_handlers, int listen_fd,
```
apavlo: Why is this flag needed for the connection dispatcher? What other protocols need this?

src/include/network/connection_dispatcher_task.h
```diff
@@ -54,7 +56,7 @@ class ConnectionDispatcherTask : public common::NotifiableTask {
    * @param flags Unused. This is here to conform to libevent callback function
    * signature.
    */
-  void DispatchPostgresConnection(int fd, int16_t flags);
+  void DispatchConnection(int fd, int16_t);
```
apavlo: Missing variable name

GustavoAngulo:
Andy's right, I had written some stuff in the log manager to support replication, but this pr should be only for ITP. I would just revert the files under `storage/write_ahead_log` to what terrier/master has. I can work on those changes in a separate branch.

src/include/network/connection_dispatcher_task.h
GustavoAngulo: The flag will tell the dispatcher whether to listen for replication messages (start event loop on replication port)

src/include/network/postgres/postgres_packet_writer.h
```diff
@@ -0,0 +1,20 @@
+#pragma once
+
+#include <memory>
+
+#include "network/abstract_packet_writer.h"
+
+namespace terrier::network {
+/**
+ * Wrapper around an I/O layer WriteQueue to provide Postgres-sprcific
```
GustavoAngulo: nit: specific*

src/include/network/abstract_packet_writer.h
tpan496: Seems like `static_assert` throws a compile-time error, `TERRIER_ASSERT` uses `assert` so it throws a run-time error

src/main/db_main.cpp
```diff
@@ -50,10 +50,14 @@ DBMain::DBMain(std::unordered_map<settings::Param, settings::ParamInfo> &&param_
   thread_pool_->Startup();
 
   t_cop_ = new trafficcop::TrafficCop;
-  command_factory_ = new network::PostgresCommandFactory;
-
   connection_handle_factory_ = new network::ConnectionHandleFactory(common::ManagedPointer(t_cop_));
-  provider_ = new network::PostgresProtocolInterpreter::Provider(common::ManagedPointer(command_factory_));
+
```
GustavoAngulo: This is probably not the right solution. For instance I don't know why `TerrierServer` needs to take in the providers and command factory from the main. I think the server should bring up those things itself, and DBMain should pass the hot_standby telling it to also bring up the replication event loop. @tli2 do you know why DBMain is the owner of the command factory and provider?

src/network/connection_dispatcher_task.cpp
```diff
@@ -19,12 +19,12 @@ ConnectionDispatcherTask::ConnectionDispatcherTask(
...
       next_handler_(0) {
-  RegisterEvent(listen_fd, EV_READ | EV_PERSIST,
-                METHOD_AS_CALLBACK(ConnectionDispatcherTask, DispatchPostgresConnection), this);
+  RegisterEvent(listen_fd, EV_READ | EV_PERSIST, METHOD_AS_CALLBACK(ConnectionDispatcherTask, DispatchConnection),
+                this);
   RegisterSignalEvent(SIGHUP, METHOD_AS_CALLBACK(NotifiableTask, ExitLoop), this);
 }
 
-void ConnectionDispatcherTask::DispatchPostgresConnection(int fd, int16_t) {  // NOLINT
+void ConnectionDispatcherTask::DispatchConnection(int fd, int16_t) {  // NOLINT
```
GustavoAngulo: Add a TODO that `DispatchConnection` will also need to take an argument telling it which interpreter to use for the connection.

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -47,6 +47,13 @@ struct PosixIoWrappers {
    */
   static void Close(int fd);
 
+  /**
```
GustavoAngulo: Like I mentioned in a comment on the pr, revert back all the logging code. I'll work on that in a separate pr parallel to you.

src/include/network/abstract_command_factory.h
```diff
@@ -0,0 +1,24 @@
...
+ public:
+  /**
+   * Convert a packet to a command.
+   * @param packet the input packet
+   * @return a shared_ptr to the converted command
+   */
+  virtual std::shared_ptr<AbstractNetworkCommand> PacketToCommand(InputPacket *packet) = 0;
+
+  virtual ~AbstractCommandFactory() = default;
+};
```
tli2: I am not sure why this is needed as the abstraction layer drawn to the rest of the network layer is `ProtocolInterpreter`. Do you use the type `AbstractCommandFactory` anywhere in the codebase where it actually exhibits polymorphism?

src/include/network/abstract_network_command.h
```diff
@@ -0,0 +1,45 @@
...
+#include "network/network_defs.h"
+#include "network/network_types.h"
+#include "network/protocol_interpreter.h"
+
+namespace terrier::network {
+
+/**
+ * Interface for the execution of the AbstractNetworkCommands
+ */
+class AbstractNetworkCommand {
```
tli2: ditto

src/include/network/abstract_packet_writer.h
```diff
@@ -0,0 +1,213 @@
...
+#include "network/postgres/postgres_protocol_utils.h"
+#include "type/transient_value_peeker.h"
+
+namespace terrier::network {
+
+/**
+ * Wrapper around an I/O layer WriteQueue to provide network protocol
+ * helper methods.
+ */
+class AbstractPacketWriter {
```
tli2: ditto

src/main/db_main.cpp
tli2: dependency injection. Otherwise I need to change server code if I need to change the protocol started.

What you should probably do is to change the injected type from a single provider to a list of providers if you plan to run multiple protocols off of the same network I/O layer

src/main/db_main.cpp
```diff
@@ -50,10 +50,14 @@ DBMain::DBMain(std::unordered_map<settings::Param, settings::ParamInfo> &&param_
...
+  // TODO(Gus): Change later to take in both providers
+  bool hot_standby = false;
+  if (!hot_standby) {
```
tli2: I am guessing this will go away once you fix the dependency thing above?

src/include/network/abstract_command_factory.h
tpan496: Both `PostgresCommandFactory` and `ITPCommandFactory` inherit from this class. I am not sure what future APIs we will need in this abstract though.

tli2: In that this case I would not implement this abstract class now as it serves no purpose.

tpan496: @GustavoAngulo

src/include/network/network_defs.h
```diff
@@ -71,6 +72,10 @@ enum class NetworkMessageType : unsigned char {
   // Messages that don't have headers (like Startup message)
   NO_HEADER = 255,
 
+  ////////////////////////////
+  // Postgres message types //
```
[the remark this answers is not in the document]
tpan496: Updated

src/network/connection_dispatcher_task.cpp
tpan496: Updated

src/include/storage/write_ahead_log/log_io.h
tpan496: Updated

src/include/network/terrier_server.h
```diff
@@ -83,6 +83,7 @@ class TerrierServer : public common::DedicatedThreadOwner {
   // For logging purposes
   // static void LogCallback(int severity, const char *msg);
 
+  const bool hot_standby_ = false;  // this server is currently in hot standby
```
tanujnay112: Why is this const

tpan496: This was added here to fix a compilation issue because the `ConnectionDispatcherTask` has been updated.
I am not sure but I think before the ITP protocol has been implemented, replication will be disabled so hot_standby_ would always be false. However, if needed I could also change the code to pass it as a parameter for starting a `TerrierServer`

src/network/postgres/postgres_protocol_interpreter.cpp
```diff
@@ -27,11 +28,12 @@ Transition PostgresProtocolInterpreter::Process(std::shared_ptr<ReadBuffer> in,
...
   }
-  std::shared_ptr<PostgresNetworkCommand> command = command_factory_->PostgresPacketToCommand(&curr_input_packet_);
+  std::shared_ptr<PostgresNetworkCommand> command =
+      std::static_pointer_cast<PostgresNetworkCommand>(command_factory_->PacketToCommand(&curr_input_packet_));
   PostgresPacketWriter writer(out);
   if (command->FlushOnComplete()) out->ForceFlush();
-  Transition ret = command->Exec(common::ManagedPointer(this), common::ManagedPointer(&writer), t_cop,
-                                 common::ManagedPointer(context), callback);
+  Transition ret = command->Exec(common::ManagedPointer<ProtocolInterpreter>(this),
+                                 common::ManagedPointer<PostgresPacketWriter>(&writer), t_cop, context, callback);
```
GustavoAngulo: @tanujnay112 @tli2 Why does the `ForceFlush` occur before the command is executed? Shouldn't it happen after?

src/include/network/connection_dispatcher_task.h
GustavoAngulo: @tpan496 Lets remove the flag for now.

src/include/network/abstract_network_command.h
GustavoAngulo: I would say this one should have this base class to prevent having duplicate code of the flush on complete and the `ReadBufferView`

src/include/network/postgres/postgres_command_factory.h
```diff
@@ -1,24 +1,26 @@
...
-class PostgresCommandFactory : public AbstractCommandFactory {
+class PostgresCommandFactory {
  public:
   /**
    * Convert a Postgres packet to command.
    * @param packet the Postgres input packet
    * @return a shared_ptr to the converted command
    */
-  std::shared_ptr<AbstractNetworkCommand> PacketToCommand(InputPacket *packet) override;
+  virtual std::shared_ptr<PostgresNetworkCommand> PacketToCommand(InputPacket *packet);
```
GustavoAngulo: nit: Why did you make these virtual? `PostgresCommandFactory` should probably be a final class

tpan496: I can probably remove the `virtual` keyword. This was just how the old code looks before we added `AbstractCommandFactory`.

src/include/network/packet_writer.h
```diff
@@ -0,0 +1,214 @@
...
+  /**
+   * Write out a packet with a single type. Some messages will be
+   * special cases since no size field is provided. (PG_SSL_YES, PG_SSL_NO)
+   * @param type Type of message to write out
+   */
+  void WriteSingleTypePacket(NetworkMessageType type) {
+    // Make sure no active packet being constructed
+    TERRIER_ASSERT(curr_packet_len_ == nullptr, "packet length is null");
+    switch (type) {
+      case NetworkMessageType::PG_SSL_YES:
```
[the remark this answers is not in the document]
tpan496: I think we can rename it and move it to the PostgresPacketWriter class? @GustavoAngulo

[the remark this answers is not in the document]
GustavoAngulo: I would add a `WriteSSLPacket` to the `PostgresPacketWriter` that does the logic in this function, and then keep the `WriteSingleTypePacket` method in `PacketWriter`, but add a `static_assert` that you don't call `WriteSingleTypePacket` with the SSL types, and to use `WriteSSLPacket` instead.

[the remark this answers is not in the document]
tpan496: Updated

src/include/network/postgres/postgres_command_factory.h
tli2: It might have been virtual for test stub purposes. Weichen would probably know more.

src/include/network/packet_writer.h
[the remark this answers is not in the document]
tli2: I think `PacketWriter` was Postgres-specific code from Peloton packaged together in a cleaner API. We never intended for it to be general purpose. 

You might want to have a PacketWriter class shared between the Postgres protocol and the internal protocol, but that class would probably look quite different from this one.

I think it's better to leave this as-is for now. and draw an abstraction in the next PR with your internal protocol.

src/include/network/postgres/postgres_command_factory.h
tpan496: @apavlo @GustavoAngulo Now that Tianyu has mentioned about the tests, I notice `network_test.cpp` has a `FakeCommandFactory' that inherits `PostgresCommandFactory` that overrides this function. The `virtual` keyword is therefore needed for the current tests.

GustavoAngulo: Ok that makes sense.

src/include/network/packet_writer.h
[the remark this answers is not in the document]
tpan496: The current version of this PR should be in accordance with Tianyu’s comment.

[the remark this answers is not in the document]
tpan496: The current version of this PR should be in accordance with Tianyu’s comment.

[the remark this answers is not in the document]
tpan496: The current version of this PR should be in accordance with Tianyu’s comment.

apavlo:
Microbenchmark failure on jenkins is an anomaly. Merging now.

PR #578, opened 2019-10-01
Modify TPC-C microbenchmark framework to use Catalog

mbutrovich:
When the microbenchmark was originally written we didn't have a Catalog. The Builder class allocated OIDs, and the relevant metadata and pointers were stashed in a Database object. This PR modifies the setup sequence of TPC-C to use the Catalog for OID allocation,  table and index creation, and finally transfers ownership of the tables and indexes to the Catalog. As a sanity check on the Catalog, we use the schemas and pointers returned from the Catalog and stash them in the Database object for the rest of the execution.

The workload itself is unchanged. This change was done so that we can leverage the Builder and Loader code for the future TPL implementation of TPC-C, which will need to talk to the Catalog for information about the tables and indexes.

While I was poking around the code and running tests over and over I also decided to parallelize the Loader. This should cut down on CI time a bit.


test/util/tpcc/delivery.cpp
```diff
@@ -42,7 +42,7 @@ bool Delivery::Execute(transaction::TransactionManager *const txn_manager, Datab
     const auto no_o_id = *reinterpret_cast<int32_t *>(new_order_select_tuple->AccessWithNullCheck(0));
 
     // Delete the corresponding New Order table row
-    txn->StageDelete(db->db_oid_, db->order_table_oid_, new_order_slot);
+    txn->StageDelete(db->db_oid_, db->new_order_table_oid_, new_order_slot);
```
mbutrovich: Noticed this bug last week while being confused about Delivery's performance.

mbutrovich:
I'll fix the merge conflicts in the AM.

test/include/util/tpcc/loader.h
```diff
@@ -129,229 +126,254 @@ struct Loader {
...
-        for (int32_t s_i_id = 0; s_i_id < 100000; s_i_id++) {
-          // For each row in the WAREHOUSE table:
-          // 100,000 rows in the STOCK table
-
-          // insert in table
-          auto *const stock_redo = txn->StageWrite(db->db_oid_, db->stock_table_oid_, stock_tuple_pr_initializer);
-          BuildStockTuple(s_i_id + 1, static_cast<int8_t>(w_id + 1), stock_original[s_i_id], stock_redo->Delta(),
-                          stock_tuple_pr_map, db->stock_schema_, generator);
-          const auto stock_slot = db->stock_table_->Insert(txn, stock_redo);
+      thread_pool->SubmitTask([w_id, txn_manager, db, &workers, &warehouse_tuple_pr_initializer,
```
apavlo: Gross...

PR #579, opened 2019-10-02
Add `make clean` as final step of Jenkins builds

mbutrovich:
This should reduce the archived size of builds.


crd477:
LGTM.  This ought to help bigly with the disk space shortages.

PR #580, opened 2019-10-03
Restructure Jenkins pipeline

### LINKED ISSUE #534: Trigger Jenkins build on successful Travis run
Right now I believe Jenkins polls for any PRs to run. Ideally we'd like Jenkins to only be triggered when Travis passes. Hopefully there are Travis triggers and Jenkins APIs to make this possible.

crd477:
Fix #534.

unflatten the pipeline by performing some quick checks, followed by all of the other time-consuming testing builds, and ending with the benchmark stage if everything else works


mbutrovich:
My only concern is that we wouldn't have a macOS build in the first `Check` stage. In fact, we wouldn't have any macOS coverage for clang-format or the other simple checks. In theory the only thing in Terrier that we've seen platform divergence on are clang-tidy, compilation, and execution, which we run across all platforms in the `Test` stage. However, on Peloton when we used older versions of clang-format (i.e.  early 3.x releases) we would see different results depending on the Linux or macOS version of the tool.

If we're not concerned with platform differences on clang-format (there shouldn't be anymore), check-censored, and check-list (it's just a Python 2 script) then I think this looks good. I am not actually concerned with macOS coverage in the `Check` stage, but just wanted to raise it for discussion.

crd477:
I overlooked "on all systems it runs" from https://github.com/cmu-db/terrier/issues/534#issuecomment-535508985.
Would it make more sense then to just keep the current structure, do check-{format,lint,censored} in the Release build stages on both platforms, and use the `parallelsAlwaysFailFast` option to abort if one of those fails?

mbutrovich:
I'm not too concerned with the wasted build effort in the `Check` stage as that will hopefully catch dumb mistakes. I still like that stage on its own.

mbutrovich:
This Jenkins pipeline now looks pretty good to me. The Check state replaces Travis, and Benchmark only gets run last if everything else is good.

@apavlo If we merge this then we can remove Travis as a required check. We're probably due for that soon since we occasionally hit the 50-minute Travis timeout just compiling the codebase. It'll only get worse as we merge more in.

pmenon:
I noticed that building the project happens both in `Check` and `Test`, but why bother? Why not just run CMake, format, lint, clang-tidy in `Check` and to the build+test in `Test`?

mbutrovich:
That's what @crd477 suggested as well, so as not to waste the build in the Check stage. We can do that but then we're basically just back to a flat Jenkins pipeline and firing off a ton of builds on potentially bad commits. `parallelsAlwaysFailFast` would at least kill those jobs, but the end result is probably a lot more short-running busted jobs in the Jenkins queue.

mbutrovich:
It's definitely redundant, but the hope was to check compilation at a smaller scale first. If we push all compilation to the `Test` stage then I think we need to enable `parallelsAlwaysFailFast` on `Test` so as not to waste a ton of capacity compiling bad builds for too long.

crd477:
OK, right  That refreshes my memory on the topic now.  Would it be alright with you if I enable `parallelsAlwaysFailFast` as part of #672 or would you prefer I do it as a separate PR?

mbutrovich:
We can enable it with #672. I expect `parallelsAlwaysFailFast` may save our Jenkins cluster with the class starting up shortly.

crd477:
Good.  I'll open a PR for the other one soon.
One of the blockers for enabling `parallelsAlwaysFailFast` was the instability of the macOS workers.  I'm pretty confident that's been addressed with the cleanup post steps and a few improvements I made to the way the VMs run.

Issue #484, opened 2019-07-31
Logical + Physical Operators for Create and Drop


17zhangw:
Currently we don't have logical/physical operators for `CREATE` and `DROP` but we do have the plans (i.e `src/include/planner/plannodes/create_*.h` and `src/include/planner/plannodes/drop_*.h`).

Ideally for create/drop operations, the optimizer should take in a logical operator, convert it to a physical operator, and then emit the correct plan (as opposed to special casing in Peloton). To do this, we need logical + physical operators for `CREATE` and `DROP` that contain all the information required by the plan nodes.

apavlo:
@zhou0ricky Start with `CreateDatabasePlanNode`

1. Look at the create plan nodes to see what meta-data they store:

https://github.com/cmu-db/terrier/blob/master/src/include/planner/plannodes/create_database_plan_node.h#L111

2. Add new logical operators to this file: https://github.com/cmu-db/terrier/blob/master/src/include/optimizer/logical_operators.h

3. Add new physical operators to this file: https://github.com/cmu-db/terrier/blob/master/src/include/optimizer/physical_operators.h

PR #582, opened 2019-10-07
Logical + Physical Operators for Create and Drop #484

zhou0ricky:
Created Logical Operators for Create operations mentioned in plannodes directory in logical_operators.cpp and logical_operators.h.

Also updated OperatorVisator class in operator_visator.h as well as add new Logical Operators in OpType class in optimizer_def.h. 

Compilation failures still arise due to issues in optimizer_def.h and operator_visator.h. 
Clarification on physical operator implementation is needed.


src/include/optimizer/logical_operators.h
```diff
@@ -958,5 +958,230 @@ class LogicalExportExternalFile : public OperatorNode<LogicalExportExternalFile>
...
 
+/**
+ * Logical operator for CreateDatabase
+ */
+class LogicalCreateDatabase : public OperatorNode<LogicalCreateDatabase> {
+ public:
+  /**
+   * @return
+   */
+  static Operator Make();
```
apavlo: I think you need a `std::string name_` member.

src/include/optimizer/logical_operators.h
```diff
@@ -958,5 +958,230 @@ class LogicalExportExternalFile : public OperatorNode<LogicalExportExternalFile>
...
+ * Logical operator for CreateTable
+ */
+class LogicalCreateTable : public OperatorNode<LogicalCreateTable> {
+ public:
+  /**
+   * @param database_oid OID of the database
+   * @param namespace_oid OID of the namespace
+   * @return
+   */
+  static Operator Make(catalog::db_oid_t database_oid, catalog::namespace_oid_t namespace_oid);
```
apavlo: Again, a table needs to have a name.

PR #591, opened 2019-10-11
Increase recovery benchmark iterations

GustavoAngulo:
The `ReadWriteWorkload` and `HighStress` recovery benchmarks seem to only be running for 2 iterations each on jenkins. We should bump these up as anomalies (such as disk slowdowns) will sometimes produce low numbers.


GustavoAngulo:
New parameters result in the following new iteration counts:

* HighStress: 2 -> 5          
* ReadWriteWorkload: 2 -> 6

PR #593, opened 2019-10-15
Refactoring garbage collector to use deferred action framework

yash620:
WIP, most of implementation complete. Still have to refactor all the tests


src/transaction/deferred_action_thread.cpp
```diff
@@ -0,0 +1,11 @@
+#include "transaction/deferred_action_thread.h"
+#include "transaction/deferred_action_manager.h"
+
+namespace terrier::transaction {
+    void DeferredActionThread::DeferredActionThreadLoop() {
+    while (run_deferred_events_) {
+      std::this_thread::sleep_for(deferred_actions_period_);
```
jrolli: I don't think we want to unconditionally sleep.  @GustavoAngulo can probably provide insight on how to efficiently check whether sleeping is appropriate since he had to address that issue in the log manager.

util/execution/tpl.cpp
```diff
@@ -252,8 +252,8 @@ static void CompileAndRun(const std::string &source, const std::string &name = "
       parse_ms, typecheck_ms, codegen_ms, interp_exec_ms, adaptive_exec_ms, jit_exec_ms);
   txn_manager.Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
   catalog.TearDown();
-  gc.PerformGarbageCollection();
-  gc.PerformGarbageCollection();
+  da_manager.Process();
+  da_manager.Process();
```
jrolli: Doesn't your destructor handle this now?

yash620:
Yeah I agree with @jrolli  that there is no guarantee that the unlink will happen before the double deferral on that transaction. This is a result of the bulk gc. We should discuss on how to maintain the bulk gc along with maintaining the double deferral semantics. 

For double deferrals that is only for DLL changes right, mainly drop table? Anything else that would need it?

yash620:
After talking it over with Andy and Matt at today's meeting we think we could try forcing the DDL changes (drop table, etc...) to clear the unlink queue before executing their changes. 

The biggest issue I see with this is from an engineering perspective I can't see how to enforce this in code. Whoever is writing the deferred event for the DDL change also has to include a call to gc's perform gc, so it will be an adhoc thing. This will also be exposing gc to outside the deferred event framework which seems kind of iffy to me. Thoughts on this @jrolli @tli2 ?

jrolli:
Some additional observations, DDL is the majority of double deferral issues but there may be (are?) others.  Specifically, `GarbageCollector::ReclaimSlotIfDeleted` might need to be a double deferral because it should only occur *after* the deferred index deletion occurs.  In the current indexes (both hash and Bwtree), the insert could fail, even on a non-unique index, because the KV pair is already in the index.

With regard to how to engineer this, I think the key to making it work is two specific characteristics:  (1) The transaction manager should have full control of the buffer, and (2) double deferrals become the transaction manager's responsibility to coordinate.

With regard to (1), you currently have flipped the way we should be batching by deferring the call to `CleanupTransaction` and batching inside the deferred action manager.  The transaction manager should be buffering and deferring a "batch" as a single deferred event.  This reduces the number of lambdas and invocations and returns control to the transaction manager.

Once (1) is in place, (2) becomes adding two new deferral queues to the transaction context that flush and defer the lambdas after a deferral on commit and abort (yes, its complicated).  The TM would be responsible for wrapping the lambdas in these queues in a new lambda that flushes and defers the unlink queue and then defers the original lambda.

Given the complexity of the above, how sure are we that not batching is a performance bottleneck?

src/include/storage/garbage_collector.h
```diff
@@ -73,45 +66,36 @@ class GarbageCollector {
    */
   void UnregisterIndexForGC(common::ManagedPointer<index::Index> index);
 
- private:
   /**
-   * Process the deallocate queue
-   * @return number of txns (not UndoRecords) processed for debugging/testing
+   * Unlink the transaction, if unlink successful defers an event to deallocate
```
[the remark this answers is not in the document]
yash620: Yeah I don't think unlink will fail I'll modify that comment

src/include/transaction/deferred_action_thread.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include <chrono>  //NOLINT
+#include <thread>  //NOLINT
+#include "di/di_help.h"
+#include "transaction/deferred_action_manager.h"
+
+namespace terrier::transaction {
+
+/**
+ * Class for spinning off a thread that runs garbage collection at a fixed interval. This should be used in most cases
```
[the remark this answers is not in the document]
yash620: Yeah it's for deferred actions at large, this is left over from code I was copying I'll modify that.

src/include/transaction/deferred_action_thread.h
```diff
@@ -0,0 +1,70 @@
...
+    // Make sure all garbage is collected. This takes 3 runs for unlink and deallocate, as well as catalog deallocations
+    deferred_actions_manager_->Process();
+    deferred_actions_manager_->Process();
+    deferred_actions_manager_->Process();
+  }
+
+  /**
+   * Pause the GC from running, typically for use in tests when the state of tables need to be fixed.
+   */
+  void DeferredActionsGC() {
```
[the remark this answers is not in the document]
yash620: Yeah

util/execution/tpl.cpp
yash620: Yup, i'll modify that

src/include/storage/garbage_collector.h
```diff
@@ -5,10 +5,11 @@
 #include <utility>
 #include "common/shared_latch.h"
 #include "storage/access_observer.h"
-#include "storage/index/index.h"
 #include "transaction/transaction_context.h"
 #include "transaction/transaction_defs.h"
-#include "transaction/transaction_manager.h"
+#include "transaction/timestamp_manager.h"
+
+#define MAX_OUTSTANDING_UNLINK_TRANSACTIONS 20
```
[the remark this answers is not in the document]
mbutrovich: Now no longer used?

src/include/storage/garbage_collector.h
```diff
@@ -33,85 +34,36 @@ class GarbageCollector {
...
   void TruncateVersionChain(DataTable *table, TupleSlot slot, transaction::timestamp_t oldest) const;
 
-  void ProcessIndexes();
-
   transaction::TimestampManager *timestamp_manager_;
   transaction::DeferredActionManager *deferred_action_manager_;
-  transaction::TransactionManager *const txn_manager_;
   AccessObserver *observer_;
   // timestamp of the last time GC unlinked anything. We need this to know when unlinked versions are safe to deallocate
   transaction::timestamp_t last_unlinked_;
```
mbutrovich: No longer needed?

src/storage/garbage_collector.cpp
```diff
@@ -7,109 +7,37 @@
...
-  if (txns_unlinked > 0) {
-    // Only update this field if we actually unlinked anything, otherwise we're being too conservative about when it's
-    // safe to deallocate the transactions in our queue.
-    last_unlinked_ = timestamp_manager_->CheckOutTimestamp();
+  if (txn->IsReadOnly()) {
+    // This is a read-only transaction so this is safe to immediately delete
+    delete txn;
+  } else {
+    timestamp_manager_->CheckOutTimestamp();
+    const transaction::timestamp_t oldest_txn = timestamp_manager_->OldestTransactionStartTime();
```
mbutrovich: I don't think we can call this on a per-txn basis (see: @GustavoAngulo's warning on the function). It takes the latch every time. It was designed to be done once per GC wakeup. Deferred action manager already calls it in Process(), so it updates the cached value. The actions should be reading the cached value.

I'm wondering if we need to rename that method to `AdvanceEpoch` or something, and then have `CachedEpoch`.

src/storage/garbage_collector.cpp
```diff
@@ -7,109 +7,37 @@
...
+    UnlinkTransaction(oldest_txn, txn);
+    deferred_action_manager_->RegisterDeferredAction([=]() { delete txn; });
```
mbutrovich: Is there any way to batch all of the txns to be deleted together to a single deferral?

src/storage/garbage_collector.cpp
```diff
@@ -7,109 +7,37 @@
...
-  txns_to_unlink_ = transaction::TransactionQueue(std::move(requeue));
-
-  return txns_processed;
-}
-
-void GarbageCollector::ProcessDeferredActions(transaction::timestamp_t oldest_txn) {
-  if (deferred_action_manager_ != DISABLED) {
-    // TODO(Tianyu): Eventually we will remove the GC and implement version chain pruning with deferred actions
-    deferred_action_manager_->Process(oldest_txn);
+    if (observer_ != nullptr) observer_->ObserveWrite(undo_record.Slot().GetBlock());
```
mbutrovich: nit: `observer_ != DISABLED`

src/storage/garbage_collector.cpp
```diff
@@ -7,109 +7,37 @@
...
 #include "transaction/transaction_context.h"
 #include "transaction/transaction_defs.h"
-#include "transaction/transaction_manager.h"
 #include "transaction/transaction_util.h"
 
 namespace terrier::storage {
 
-std::pair<uint32_t, uint32_t> GarbageCollector::PerformGarbageCollection() {
+void GarbageCollector::CleanupTransaction(transaction::TransactionContext *txn) {
   if (observer_ != nullptr) observer_->ObserveGCInvocation();
```
mbutrovich: nit: `observer_ != DISABLED`

src/transaction/deferred_action_manager.cpp
```diff
@@ -0,0 +1,107 @@
...
+}
+
+uint32_t DeferredActionManager::ProcessNewActions(timestamp_t oldest_txn) {
+  uint32_t processed = 0;
+  // swap the new actions queue with a local queue, so the rest of the system can continue
+  // while we process actions
+  std::queue<std::pair<timestamp_t, DeferredAction>> new_actions_local;
+  {
+    common::SpinLatch::ScopedSpinLatch guard(&deferred_actions_latch_);
+    new_actions_local = std::move(new_deferred_actions_);
```
mbutrovich: Performance concern: this is O(n). We should keep an eye on this.

src/transaction/deferred_action_manager.cpp
```diff
@@ -0,0 +1,107 @@
...
+  // Iterate through the new actions queue and execute as many as possible
+  while (!new_actions_local.empty() && oldest_txn >= new_actions_local.front().first) {
+    new_actions_local.front().second(oldest_txn);
+    processed++;
+    new_actions_local.pop();
+  }
+
+  // Add the rest to back log otherwise
+  while (!new_actions_local.empty()) {
+    back_log_.push(new_actions_local.front());
```
mbutrovich: Performance concern: yikes. We should keep an eye on this.

src/transaction/transaction_manager.cpp
```diff
@@ -177,11 +178,10 @@ timestamp_t TransactionManager::Abort(TransactionContext *const txn) {
 
   // We hand off txn to GC, however, it won't be GC'd until the LogManager marks it as serialized
   if (gc_enabled_) {
-    common::SpinLatch::ScopedSpinLatch guard(&timestamp_manager_->curr_running_txns_latch_);
     // It is not necessary to have to GC process read-only transactions, but it's probably faster to call free off
```
mbutrovich: I'm not convinced this is true now that we're taking a latch that other txns will be trying to get in the middle of query execution, rather than just at commit.. I would change this to a TODO: that says "evaluate performance of deleting read-only transactions on the critical path".

jrolli:
@yash620, are you still maintaining this?  I have some spare time over the next couple weeks, but I don't want to duplicate work you've done.

PR #598, opened 2019-10-19
Allow Exception constructor to take std::string instead of char*

CarltonBranch:
Implements exception constructor to take std::string as an input parameter instead of char*


apavlo:
@CarltonBranch Please format your code per these instructions:

https://github.com/cmu-db/terrier/wiki/Pull-Request-Process

PR #603, opened 2019-10-22
Expand DataTable Microbenchmarks

### LINKED ISSUE #568: Expand DataTable Microbenchmarks
@malin1993ml found in his experiments that there are scalability issues with the `DataTable`'s `SlotIterator` class:

https://github.com/cmu-db/terrier/blob/master/src/include/storage/data_table.h#L47

We need to add a new microbenchmark to evaluate the performance of `DataTable::Scan` method. We can also use this opportunity to clean up the `DataTableBenchmark` class a little bit.

1. Rename `DataTableBenchmark` functions to make it more clear what is being tested:
    * [RandomRead](https://github.com/cmu-db/terrier/blob/master/benchmark/storage/data_table_benchmark.cpp#L156) -> `SelectRandom`
    * [SequentialRead](https://github.com/cmu-db/terrier/blob/master/benchmark/storage/data_table_benchmark.cpp#L134) -> `SelectSequential`
   * [ConcurrentRandomRead](https://github.com/cmu-db/terrier/blob/master/benchmark/storage/data_table_benchmark.cpp#L180) -> `ConcurrentSelectRandom`

2. Create a new multi-threaded test called `ConcurrentSelectSequential`. Each thread will scan a different segment of the table in sequential order. It doesn't matter if the segments overlap and they are reading the same tuples. Your new test code should work the same as `ConcurrentSelectRandom`. But instead of duplicating all of the multi-threading setup code in `ConcurrentSelectRandom`, you should refactor it out into a single function that supports the ability to do both. Then both `ConcurrentSelectSequential` and `ConcurrentSelectRandom` will call this utility function to perform the test.

3. Create a new test `DataTableBenchmark::ConcurrentScan`. We are not going to bother with writing the single-threaded version since we will eventually throw it away. It should be similar to the new `ConcurrentSelectSequential` test but you will want to use the [DataTable::Scan]() function instead of `DataTable::Select`. See our existing [DataTableTests::SimpleSequentialScan](https://github.com/cmu-db/terrier/blob/master/test/storage/data_table_test.cpp#L195) unit test on how to use this API.

harsh141994:
Fix #568 

I'll update the comments in the final clean up.

This is my local run : 
Screenshot 2019-10-28 at 6 41 31 PM


mbutrovich:
Closed in favor of #613.

PR #607, opened 2019-10-25
Fix occasional thread_start_container_test failures.

### LINKED ISSUE #561: thread_state_container_test fails occasionally
We're seeing failures rarely on CI for `thread_state_container_test`. I believe @amlatyrngom investigated it briefly and said that it required a large number of local iterations before it manifested as a memory leak.

mbutrovich:
According to @amlatyrngom this fixes #561.


pmenon:
Wait, how does this fix the problem if the leak is originating in TBB?

Also, the diff changes the behavior of the test: it no longer calls the constructor as it should. If you add a check for `EXPECT_EQ(44u, arr_2_[0])` the test will fail.

mbutrovich:
Right, I recall the conversation now that we need a whitelist instead. I'll close this.

PR #611, opened 2019-11-01
Export throughput metrics for Recovery Manager

tpan496:
- [x] New metric format for recovery
- [x] Throughput metric export for `recovery_manager.cpp`
- [x] Tests similar to those in `metric_test.cpp` for this new metric


src/storage/recovery/recovery_manager.cpp
```diff
@@ -77,19 +79,44 @@ void RecoveryManager::ProcessCommittedTransaction(terrier::transaction::timestam
...
 
-    if (IsSpecialCaseCatalogRecord(buffered_record)) {
-      idx += ProcessSpecialCaseCatalogRecord(txn, &buffered_changes_map_[txn_id], idx);
-    } else if (buffered_record->RecordType() == LogRecordType::REDO) {
-      ReplayRedoRecord(txn, buffered_record);
-    } else {
-      ReplayDeleteRecord(txn, buffered_record);
+    // Update metrics
+    total_elapsed_us += elapsed_us;
+    num_txns += 1;
```
GustavoAngulo: This is incorrect. A transaction can have multiple records, you are incrementing the txn counter once **per record** here.

src/storage/recovery/recovery_manager.cpp
```diff
@@ -77,19 +79,44 @@ void RecoveryManager::ProcessCommittedTransaction(terrier::transaction::timestam
...
+    num_bytes += buffered_record->Size();
+
+    // Check if we need to export throughput metric
+    if (total_elapsed_us >= recovery_metric_interval_) {
+      // Export throughput metric
+      if (num_txns > 0 && common::thread_context.metrics_store_ != nullptr &&
+          common::thread_context.metrics_store_->ComponentEnabled(metrics::MetricsComponent::RECOVERY)) {
+        common::thread_context.metrics_store_->RecordRecoveryData(num_txns, num_bytes);
```
GustavoAngulo: I would also export `total_elapsed_us`, that way we can compute txns/sec and bytes/s

src/storage/recovery/recovery_manager.cpp
```diff
@@ -77,19 +79,44 @@ void RecoveryManager::ProcessCommittedTransaction(terrier::transaction::timestam
...
+        num_txns = 0;
+        num_bytes = 0;
+      }
+      total_elapsed_us = 0;
```
GustavoAngulo: Why are you reseting the time period here? If the conditions to export a metric are not met, we should extend the period (by not reseting it) until they are.

src/storage/recovery/recovery_manager.cpp
```diff
@@ -77,19 +79,44 @@ void RecoveryManager::ProcessCommittedTransaction(terrier::transaction::timestam
...
+  uint64_t num_bytes = 0;
+
   // Apply all buffered changes. They should all succeed. After applying we can safely delete the record
   for (uint32_t idx = 0; idx < buffered_changes_map_[txn_id].size(); idx++) {
     auto *buffered_record = buffered_changes_map_[txn_id][idx].first;
     TERRIER_ASSERT(
         buffered_record->RecordType() == LogRecordType::REDO || buffered_record->RecordType() == LogRecordType::DELETE,
         "Buffered record must be a redo or delete.");
+    {
+      common::ScopedTimer<std::chrono::microseconds> scoped_timer(&elapsed_us);
```
GustavoAngulo: I'm not sure on the overhead of a scoped timer, but pay attention to the benchmarking results from Jenkins. If the overhead of per record timing is too much, we can probably get away with doing a per transaction timing. The average transaction should not last more than 1 second.

src/storage/recovery/recovery_manager.cpp
```diff
@@ -77,19 +79,44 @@ void RecoveryManager::ProcessCommittedTransaction(terrier::transaction::timestam
...
-      ReplayDeleteRecord(txn, buffered_record);
+    // Update metrics
+    total_elapsed_us += elapsed_us;
+    num_txns += 1;
+    num_bytes += buffered_record->Size();
+
+    // Check if we need to export throughput metric
+    if (total_elapsed_us >= recovery_metric_interval_) {
+      // Export throughput metric
+      if (num_txns > 0 && common::thread_context.metrics_store_ != nullptr &&
```
GustavoAngulo: nit: `common::thread_context.metrics_store_ != DISABLED`

src/include/storage/recovery/recovery_manager.h
```diff
@@ -154,6 +154,9 @@ class RecoveryManager : public common::DedicatedThreadOwner {
   // Number of recovered committed txns. Used for benchmarking
   uint32_t recovered_txns_;
 
+  // How often the recovery throughput metric is collected
+  const uint32_t recovery_metric_interval_ = 1;
```
GustavoAngulo: This should be `std::chrono::milliseconds`, and pass it in as a `RecoveryManager` parameter.

src/include/metrics/recovery_metric.h
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Columns to use for writing to CSV.
+   */
+  static constexpr std::array<std::string_view, 1> COLUMNS = {"now,num_txns,num_bytes"};
+
+ private:
+  friend class RecoveryMetric;
+  FRIEND_TEST(MetricsTests, RecoveryCSVTest);
+
+  void RecordRecoveryData(const uint64_t num_txns, const uint64_t num_bytes) {
```
GustavoAngulo: Like I said in a below comment, I would also export the time period length over which these values occurred.

tpan496:
I added a simple test in `recovery_test.cpp` that checks against total number of transactions and bytes. I planned to add this inside `metric_test.cpp`, but I feel like the Setup function inside `recovery_test` is better suited.
I created a `metrics_manager` inside `recovery_manager` itself that will be used inside the background `recovery_task`; I am not sure if this is the correct way to do this.

src/include/storage/recovery/recovery_manager.h
```diff
@@ -85,6 +103,10 @@ class RecoveryManager : public common::DedicatedThreadOwner {
         catalog::postgres::Builder::GetConstraintTableSchema();
     catalog_table_schemas_[catalog::postgres::INDEX_TABLE_OID] = catalog::postgres::Builder::GetIndexTableSchema();
     catalog_table_schemas_[catalog::postgres::TYPE_TABLE_OID] = catalog::postgres::Builder::GetTypeTableSchema();
+
+    if (enable_metric_) {
+      metrics_manager_ = common::ManagedPointer(new metrics::MetricsManager);
```
GustavoAngulo: The `RecoveryManager` should not instantiate a new `MetricsManager`, it should be passed in as a managed pointer to the `RecoveryManager`. Your test should instantiate a `MetricsThread`, and you should pass in the `MetricsThread`'s `MetricsManager`. 

If metrics is disabled, pass in `DISABLED` instead. This will also remove the need for the `enable_metric` flag, just instead check if `metrics_manager_ != DISABLED`

src/storage/recovery/recovery_manager.cpp
```diff
@@ -144,15 +122,40 @@ void RecoveryManager::DeferRecordDeletes(terrier::transaction::timestamp_t txn_i
...
+      num_bytes += ProcessCommittedTransaction(*it);
+    }
     txns_processed++;
+
+    // Update metrics
+    total_elapsed_us += elapsed_us;
+    num_txns += 1;
+
+    // Check if we need to export throughput metric
+    if (std::chrono::milliseconds(total_elapsed_us) >= recovery_metric_interval_ || std::next(it) == upper_bound_it) {
```
GustavoAngulo: @mbutrovich I want your input on this. Should components export metrics as soon as they have data collected, or should they input them on an interview basis (i.e. output a metrics every second for the processing done for the last second)

src/include/storage/recovery/recovery_manager.h
```diff
@@ -43,7 +43,20 @@ class RecoveryManager : public common::DedicatedThreadOwner {
     /**
      * Runs the recovery task. Our task only calls Recover on the log manager.
      */
-    void RunTask() override { recovery_manager_->Recover(); }
+    void RunTask() override {
+      if (recovery_manager_->enable_metric_) {
+        // Enable metrics
+        recovery_manager_->metrics_manager_->RegisterThread();
+        recovery_manager_->metrics_manager_->EnableMetric(metrics::MetricsComponent::RECOVERY);
```
GustavoAngulo: When you change the metrics manager to be passed in to the recovery manager, the `RegisterThread` is correctly placed here, but the `EnableMetric` should happen in the test code  (like how its done in `TPCCBenchmark`).

src/include/storage/recovery/recovery_manager.h
```diff
@@ -43,7 +43,20 @@ class RecoveryManager : public common::DedicatedThreadOwner {
...
+      }
+
+      recovery_manager_->Recover();
+
+      // Save results to metrics manager, otherwise the results are lost
+      if (recovery_manager_->enable_metric_) {
+        recovery_manager_->metrics_manager_->Aggregate();
```
GustavoAngulo: The `RecoveryManager` should not call `Aggregate`, thats the role of the `MetricsThread`. If you want to test, you can use a generous `sleep` to ensure the `MetricsThread` has enough time to wake up and Aggregate the data.

test/storage/recovery_test.cpp
```diff
@@ -244,6 +251,30 @@ class RecoveryTests : public TerrierTest {
...
+      // Calculate total number of txns
+      uint64_t total_num_txns = 0;
+      uint64_t total_num_bytes = 0;
+      for (auto &data : aggregated_data->recovery_data_) {
+        total_num_txns += data.num_txns_;
+        total_num_bytes += data.num_bytes_;
+      }
+
+      EXPECT_EQ(total_num_txns, 101);
+      EXPECT_EQ(total_num_bytes, 91176);
```
GustavoAngulo: How did you get these values? My guess is trial and error, in which case this is probably not a good idea. `tested->SimulateOltp(100, 4)` is a **random** execution, and while we use a seed, I'm not convinced it will always succeed. Even when random testing, your tests should not fail even if you change the seed. Changing the seed would probably cause this to fail as varlen entries would be a different size, thus changing the number of byes.

I would do what `MetricsTest` does and just execute 1-3 predetermined transactions and test with those.

test/storage/recovery_test.cpp
```diff
@@ -315,7 +363,7 @@ TEST_F(RecoveryTests, DropDatabaseTest) {
   DiskLogProvider log_provider(LOG_FILE_NAME);
   RecoveryManager recovery_manager(&log_provider, common::ManagedPointer(recovery_catalog_), recovery_txn_manager_,
                                    recovery_deferred_action_manager_, common::ManagedPointer(thread_registry_),
-                                   &block_store_);
+                                   &block_store_, recovery_metric_interval_, false);
```
GustavoAngulo: Like I mentioned in another comment, create a `MetricsThread`, and pass in that `MetricsThread`'s `MetricsManager` here

test/storage/recovery_test.cpp
tpan496: After discussion we would just be testing whether the throughput is being logged or not, so this part can be left out.

src/include/storage/recovery/recovery_manager.h
tpan496: No longer needed as I am using the metrics thread, which does the registering for us

tpan496:
The recovery task is being reaped right after finishing the recovery task, which makes it hard for the metrics thread to perform deterministic tests. So right now the test just involves checking if any metrics are being output or not.

src/storage/recovery/recovery_manager.cpp
```diff
@@ -136,7 +136,7 @@ uint32_t RecoveryManager::ProcessDeferredTransactions(terrier::transaction::time
   uint64_t num_bytes = 0;
   for (auto it = deferred_txns_.begin(); it != upper_bound_it; it++) {
     {
-      common::ScopedTimer<std::chrono::microseconds> scoped_timer(&elapsed_us);
+      common::ScopedTimer<std::chrono::milliseconds> scoped_timer(&elapsed_us);
```
GustavoAngulo: If you're changing this to milliseconds, you need to update to code to be `elapsed_ms` instead of `elapsed_us`. Especially in the csv file where the column name is still `elapsed_us`

src/storage/recovery/recovery_manager.cpp
GustavoAngulo: Lets take out the interval stuff and just output a metric with the same logic below

PR #612, opened 2019-11-01
Rename 'util' dirs to 'benchmark_util' and 'test_util'

apavlo:
To avoid the confusion between the benchmark/test util dirs and src util dir, I renamed the following two directories:

* `benchmark/util` -> `benchmark/benchmark_util`
* `test/util` -> `test/test_util`

https://www.youtube.com/watch?v=ZYkFgpEX3F0


test/CMakeLists.txt
```diff
@@ -25,6 +25,7 @@ if (TERRIER_BUILD_TESTS)
     add_subdirectory(traffic_cop)
     add_subdirectory(transaction)
     add_subdirectory(type)
+    add_subdirectory(util)
```
mbutrovich: This is the source of the CMake issue in CI. Just gotta delete this line.

PR #613, opened 2019-11-04
data table benchmarking

harsh141994:


mbutrovich:
Needs a more descriptive title and description (I think it fixes #568). I'll review it today.

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -220,19 +223,110 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentRandomRead)(benchmark::State &s
...
+  for (auto _ : state) {
+    auto workload = [&](uint32_t id) {
+      // We can use dummy timestamps here since we're not invoking concurrency control
+      transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
+                                          DISABLED);
+
+      std::vector<storage::col_id_t> all_cols = StorageTestUtil::ProjectionListAllColumns(layout_);
+      EXPECT_NE((!all_cols[all_cols.size() - 1]), -1);
+      storage::ProjectedColumnsInitializer initializer(layout_, all_cols, num_reads_);
+      auto *buffer = common::AllocationUtil::AllocateAligned(initializer.ProjectedColumnsSize());
```
mbutrovich: This memory is never reclaimed.

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -220,19 +223,110 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentRandomRead)(benchmark::State &s
...
+
+  // NOLINTNEXTLINE
+  for (auto _ : state) {
+    auto workload = [&](uint32_t id) {
+      // We can use dummy timestamps here since we're not invoking concurrency control
+      transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
+                                          DISABLED);
+
+      std::vector<storage::col_id_t> all_cols = StorageTestUtil::ProjectionListAllColumns(layout_);
+      EXPECT_NE((!all_cols[all_cols.size() - 1]), -1);
```
mbutrovich: What is this testing?

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -220,19 +223,110 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentRandomRead)(benchmark::State &s
...
+      storage::ProjectedColumnsInitializer initializer(layout_, all_cols, num_reads_);
```
mbutrovich: If the PCI is requested as `num_reads_` then in theory the entire table should be materialized in one shot. This probably isn't realistic, and we'd want to use smaller chunks. As currently defined, the while loop around Scan shouldn't be looping.

I would change it to `common::Constants::K_DEFAULT_VECTOR_SIZE` since that's what the execution engine seems to use.

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -220,19 +223,110 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentRandomRead)(benchmark::State &s
...
+      auto *buffer = common::AllocationUtil::AllocateAligned(initializer.ProjectedColumnsSize());
+      storage::ProjectedColumns *columns = initializer.Initialize(buffer);
+      auto it = read_table.begin();
+
+      while (it != read_table.end()) {
+        read_table.Scan(&txn, &it, columns);
```
mbutrovich: This is really the only operation we're interested in profiling. There's a lot of other setup being done in the workload lambda that we're not terribly interested in. Our benchmarks should be targeted and only time the operations we care about.

benchmark/storage/data_table_benchmark.cpp
harsh141994: this is the all_cols I just created to see, its not empty. Removing it.

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -220,19 +223,118 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentRandomRead)(benchmark::State &s
...
+  // We can use dummy timestamps here since we're not invoking concurrency control
+  transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0), &buffer_pool_,
+                                      DISABLED);
+  std::vector<storage::TupleSlot> read_order;
+  for (uint32_t i = 0; i < num_reads_; ++i) {
+    read_order.emplace_back(read_table.Insert(&txn, *redo_));
+  }
+
+  std::vector<storage::col_id_t> all_cols = StorageTestUtil::ProjectionListAllColumns(layout_);
+  storage::ProjectedColumnsInitializer initializer(layout_, all_cols, num_reads_);
```
mbutrovich: See earlier comment about num_reads_ and why that's not what we want here.. Change it to common::Constants::K_DEFAULT_VECTOR_SIZE since that's what the execution engine seems to use.

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -220,19 +223,118 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentRandomRead)(benchmark::State &s
...
+    }
+    state.SetIterationTime(static_cast<double>(elapsed_ms) / 1000.0);
+  }
+  for (auto p : buf) {
+    delete p;
+  }
+  buf.clear();
+  // projected columns are deleted automatically
+  all_columns.clear();
+  state.SetItemsProcessed(state.iterations() * num_reads_);
```
mbutrovich: `state.iterations() * num_reads_ * num_threads_`

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -220,19 +223,118 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentRandomRead)(benchmark::State &s
...
+        thread_pool.SubmitTask([j, &workload] { workload(j); });
+      }
+      thread_pool.WaitUntilAllFinished();
+    }
+    state.SetIterationTime(static_cast<double>(elapsed_ms) / 1000.0);
+  }
+  for (auto p : buf) {
+    delete p;
+  }
+  buf.clear();
```
mbutrovich: No reason to call `clear` on these vectors since they immediately fall out of scope anyway.

benchmark/storage/data_table_benchmark.cpp
```diff
@@ -220,19 +223,118 @@ BENCHMARK_DEFINE_F(DataTableBenchmark, ConcurrentRandomRead)(benchmark::State &s
...
+  // projected columns are deleted automatically
```
mbutrovich: Not really, they're the same memory as the buffers that you just deleted. I would remove the comment.

benchmark/storage/data_table_benchmark.cpp
harsh141994: done

benchmark/storage/data_table_benchmark.cpp
harsh141994: done

benchmark/storage/data_table_benchmark.cpp
harsh141994: yeah, true

benchmark/storage/data_table_benchmark.cpp
harsh141994: sorry, I missed it last time. Corrected it.

PR #616, opened 2019-11-04
Dump Table

YuzeLiao:
This PR implements the function of dumping tables to disk. We build this function on top of the data table layer because we want to make full use of the relaxed Arrow format and the data table is where we have access to blocks.

*This PR also fixed two bugs in block_compactor (CopyToArrowVarlen() and BuildDictionary())*

RoadMap:
- [x] Implement exporting fixed-length columns
- [x] Implement exporting variable-length columns
- [x] Implement exporting dictionary-compressed columns
- [x] (Manually) Debugged and verified that PyArrow can correctly load fixed-length columns
- [x] (Manually) Debugged and verified that PyArrow can correctly load variable-length columns
- [x] (Manually) Debugged and verified that PyArrow can correctly load dictionary-compressed columns
- [x] Automate tests
- [x] Improve code style, modularity, etc.


apavlo:
@YuzeLiao If the data is already in the Arrow format, why not write out raw Arrow blocks to disk? What does Flatbuffers provide for us?

YuzeLiao:
> @YuzeLiao If the data is already in the Arrow format, why not write out raw Arrow blocks to disk? What does Flatbuffers provide for us?

I think Flatbuffers mainly does metadata serializations. In Arrow's doc, it says "_The Arrow columnar memory layout only applies to data and not metadata. Implementations are free to represent metadata in-memory in whichever form is convenient for them. We handle metadata serialization in an implementation-independent way using Flatbuffers, detailed below._", so if we want to use them in different scenarios (Pandas/Numpy for now), we may need to be consistent with Apache Arrow Spec. Actually, in your recently submitted paper,  what the latter half of the third graph of section 2.2 describes is exactly something implemented with Flatbuffers (DDL .........)

src/storage/data_table.cpp
```diff
@@ -246,6 +274,189 @@ bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSl
...
+    if (!layout.IsVarlen(col_id) || col_info.Type() == ArrowColumnType::FIXED_LENGTH) {
+      uint8_t byte_width = accessor_.GetBlockLayout().AttrSize(col_id);
+      type = flatbuf::Type_FixedSizeBinary;
+      type_offset = flatbuf::CreateFixedSizeBinary(*flatbuf_builder, byte_width).Union();
+    } else {
+      switch (col_info.Type()) {
+        case ArrowColumnType::DICTIONARY_COMPRESSED:
+          dictionary = flatbuf::CreateDictionaryEncoding(
+              *flatbuf_builder, dictionary_id, flatbuf::CreateInt(*flatbuf_builder, 8 * sizeof(uint64_t), true), false);
+          (*dictionary_ids)[col_id] = dictionary_id++;
```
GustavoAngulo: Use `TERRIER_FALLTHROUGH`

src/storage/data_table.cpp
```diff
@@ -246,6 +274,189 @@ bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSl
...
+        case ArrowColumnType::GATHERED_VARLEN:  // fall through
+          type = flatbuf::Type_LargeBinary;
+          type_offset = flatbuf::CreateLargeBinary(*flatbuf_builder).Union();
+          break;
+        default:
+          throw std::runtime_error("unexpected control flow");
+      }
+    }
+    std::vector<flatbuffers::Offset<org::apache::arrow::flatbuf::Field>> fake_children;
```
GustavoAngulo: nit: Can you add a comment explaining why this is called `fake_children`. Something like "We don't care about the children because...."

src/storage/data_table.cpp
```diff
@@ -246,6 +274,189 @@ bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSl
...
+  flatbuffers::FlatBufferBuilder flatbuf_builder;
+  std::ofstream outfile(file_name, std::ios_base::out | std::ios_base::trunc | std::ios_base::binary);
+  std::unordered_map<col_id_t, int64_t> dictionary_ids;
+  WriteSchemaMessage(outfile, &dictionary_ids, &flatbuf_builder);
+
+  const BlockLayout &layout = accessor_.GetBlockLayout();
+  auto column_ids = layout.AllColumns();
+  common::SpinLatch::ScopedSpinLatch guard(&blocks_latch_);
+  // We may choose to scan twice instead of using a buffer?
+  auto dump_buffer = new char[MAX_DUMP_BUFFER_BLOCK * common::Constants::BLOCK_SIZE];
```
GustavoAngulo: The purpose of this `dump_buffer` use in these methods is not clear. Why are we copying into this, and then writing it out to the file? Copying seem's like an expensive step that we can just avoid by writing to the file directly. 

Your comment seems to indicate that you can scan once for all the metadata you need, write the metadata, and then write out the data directly to file without having to use this buffer? I would run perf and confirm it, but my intuition tells me that using this `dump_buffer` is expensive. It's also like 10MB alone which seems like a large amount of stack space to allocate.

src/storage/data_table.cpp
```diff
@@ -246,6 +274,189 @@ bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSl
...
+  for (RawBlock *block : blocks_) {
+    std::vector<flatbuf::FieldNode> field_nodes;
+    std::vector<flatbuf::Buffer> buffers;
+    while (!block->controller_.TryAcquireInPlaceRead()) {
```
GustavoAngulo: nit: Can you add a comment here? It's not clear why this exists

src/storage/data_table.cpp
```diff
@@ -246,6 +274,189 @@ bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSl
...
+          }
+          default:
+            throw std::runtime_error("unexpected control flow");
+        }
+      } else {
+        int32_t cur_buffer_len;
+        if (i == column_id_size - 1) {
+          auto casted_column_start = reinterpret_cast<uintptr_t>(column_start);
+          uintptr_t mask = common::Constants::BLOCK_SIZE - 1;
+          cur_buffer_len = ((casted_column_start + mask) & (~mask)) - casted_column_start;
```
GustavoAngulo: Add a comment explaining why we special case this

test/storage/block_compactor_test.cpp
```diff
@@ -371,4 +371,96 @@ TEST_F(BlockCompactorTest, DictionaryCompressionTest) {
   }
 }
 
+// A fake test that exports a random data table to increase test coverage. Will later be replaced by an formal
+// automated export table test.
+// NOLINTNEXTLINE
+TEST_F(BlockCompactorTest, ExportDictionaryCompressedTableTest) {
+  storage::BlockLayout layout = StorageTestUtil::RandomLayoutWithVarlens(100, &generator_);
```
GustavoAngulo: Like I said in slack, try doing like 1000 columns and seeing if codecov passes

src/include/storage/data_table.h
```diff
@@ -281,6 +293,25 @@ class DataTable {
...
+   * @param outfile the output file
+   * @param dictionary_ids when encountered columns that are dictionary-encoded, we pre-assign the
+   *                       dictionary id
+   * @param flatbuf_builder flatbuffer builder
+   */
+  void WriteSchemaMessage(std::ofstream &outfile, std::unordered_map<col_id_t, int64_t> *dictionary_ids,
+                          flatbuffers::FlatBufferBuilder *flatbuf_builder) const;
+
+  /**
+   * Write a dictionary message
```
GustavoAngulo: Assuming its not a ton of metadata, can you add a comment showing the serialized format, similar to how [`ProjectedRows`](https://github.com/cmu-db/terrier/blob/18c62295b42cc6acd06926989318bd2d94a1e416/src/include/storage/projected_row.h#L19-L23) is shown here

src/include/storage/data_table.h
```diff
@@ -281,6 +293,25 @@ class DataTable {
   bool CompareAndSwapVersionPtr(TupleSlot slot, const TupleAccessStrategy &accessor, UndoRecord *expected,
                                 UndoRecord *desired);
 
+  /**
+   * Write a schema message
```
GustavoAngulo: Can you add a comment showing the serialized format, similar to how [`ProjectedRows`](https://github.com/cmu-db/terrier/blob/18c62295b42cc6acd06926989318bd2d94a1e416/src/include/storage/projected_row.h#L19-L23) is shown here

src/include/storage/data_table.h
```diff
@@ -213,6 +219,12 @@ class DataTable {
    */
   const BlockLayout &GetBlockLayout() const { return accessor_.GetBlockLayout(); }
 
+  /**
+   * Dump a table to disk in arrow IPC format.
```
GustavoAngulo: Can you write a high level diagram of how tables are serialized. Similar to how [`ProjectedRows`](https://github.com/cmu-db/terrier/blob/18c62295b42cc6acd06926989318bd2d94a1e416/src/include/storage/projected_row.h#L19-L23) is shown here

src/include/storage/arrow_block_metadata.h
```diff
@@ -97,7 +97,7 @@ class ArrowVarlenColumn {
   /**
    * @return the offsets array
    */
-  uint32_t *Offsets() const { return offsets_; }
+  uint64_t *Offsets() const { return offsets_; }
```
GustavoAngulo: Why were all these changed to `uint64_t`

YuzeLiao: Because the offsets array defined in arrow is 64-byte. If we use 32-byte in the system, we cannot directly use the offsets block.

src/storage/data_table.cpp
YuzeLiao: Changed to two-pass writing.

src/include/storage/data_table.h
YuzeLiao: I added comments describes the high-level format, for example, the first part is a Schema Message, then some RecordBatch messages + dictionary messages. But I may not be able to state their internal structures because they are all defined by flatbuffer schema files. I provided pointers to those files.

src/storage/data_table.cpp
```diff
@@ -246,6 +258,251 @@ bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSl
...
+
+void DataTable::ExportTable(const std::string &file_name) {
+  flatbuffers::FlatBufferBuilder flatbuf_builder;
+  std::ofstream outfile(file_name, std::ios_base::out | std::ios_base::trunc | std::ios_base::binary);
+  std::unordered_map<col_id_t, int64_t> dictionary_ids;
+  WriteSchemaMessage(outfile, &dictionary_ids, &flatbuf_builder);
+
+  const BlockLayout &layout = accessor_.GetBlockLayout();
+  auto column_ids = layout.AllColumns();
+  common::SpinLatch::ScopedSpinLatch guard(&blocks_latch_);
```
malin1993ml: This will block all the other concurrent scans and inserts into the table while exporting. A fix for now is to copy the list of blocks (`blocks_`) while holding this lock, and release the lock before doing the actual exporting. This is not optimal but good enough for us here (and better than the current version.)

src/storage/data_table.cpp
```diff
@@ -246,6 +258,251 @@ bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSl
...
+            break;
+          }
+          default:
+            throw std::runtime_error("unexpected control flow");
+        }
+      } else {
+        int32_t cur_buffer_len;
+        // Calculate the length of the data region of current column. For the columns except the last one, we calculate
+        // their length by using the start of next column's bit map - the start of current column's data. For the
+        // last column, we calculate the length by using the begining address of the next block - the start of current
```
malin1993ml: typo -> beginning

src/include/storage/data_table.h
```diff
@@ -281,6 +312,62 @@ class DataTable {
...
+   *    https://arrow.apache.org/docs/format/Columnar.html
+   *
+   * For the flatbuffer schema of Schema message, please refer to:
+   *    https://github.com/apache/arrow/blob/master/format/Schema.fbs
+   *
+   * @param dictionary_ids The dictionary entries and the indices for a batch of rows are written seperately.
+   *                       Therefore, when a column is dictionary-compressed, we need to assign an id to it,
+   *                       so that the dictionary and the indices can be paired.
+   * @param flatbuf_builder flatbuffer builder
+   */
```
malin1993ml: Missing the `outfile` param

src/include/storage/data_table.h
```diff
@@ -281,6 +312,62 @@ class DataTable {
...
+   *    https://arrow.apache.org/docs/format/Columnar.html
+   *
+   * For the flatbuffer schema of Dictionary message, please refer to:
+   *    https://github.com/apache/arrow/blob/master/format/Message.fbs
+   *
+   * @param outfile the output file
+   * @param dictionary_id id of this dictionary, should have been assigned previously when writing the schema message.
+   * @param flatbuf_builder flatbuffer builder
+   */
+  void WriteDictionaryMessage(std::ofstream &outfile, int64_t dictionary_id, const ArrowVarlenColumn &varlen_col,
```
malin1993ml: missing the `varlen_col` param.

src/include/storage/data_table.h
```diff
@@ -213,6 +219,31 @@ class DataTable {
...
+   *
+   * Where Dictionary messages and RecordBatch messages can interleave. Table data are primarily
+   * stored in RecordBatch messages, each of which has the following format:
+   *
+   *    metadata_size: int32
+   *    metadata_flatbuffer: bytes
+   *    padding to 8-byte
+   *    message body
+   * Message body is the data, which are organized as several buffers. Each buffer is described by
+   * a metadata_flatbuffer
```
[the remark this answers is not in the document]
YuzeLiao: Thank you! Yeah, I think it was actually not clear here...I've rewritten this part of comments.

src/storage/data_table.cpp
```diff
@@ -246,6 +258,251 @@ bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSl
...
+}
+
+void DataTable::WriteSchemaMessage(std::ofstream &outfile, std::unordered_map<col_id_t, int64_t> *dictionary_ids,
+                                   flatbuffers::FlatBufferBuilder *flatbuf_builder) {
+  RawBlock *block = blocks_.front();
+  const BlockLayout &layout = accessor_.GetBlockLayout();
+  ArrowBlockMetadata &metadata = accessor_.GetArrowBlockMetadata(block);
+  std::vector<flatbuffers::Offset<flatbuf::Field>> fields;
+  int64_t dictionary_id = 0;
+  for (col_id_t col_id : layout.AllColumns()) {
```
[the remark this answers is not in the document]
YuzeLiao: Done.

src/storage/data_table.cpp
```diff
@@ -246,6 +258,251 @@ bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSl
...
+  flatbuf_builder->Finish(message);
+  int32_t flatbuf_size = flatbuf_builder->GetSize();
+  auto padded_flatbuf_size = StorageUtil::PadUpToSize(ARROW_ALIGNMENT, flatbuf_size);
+  outfile.write(reinterpret_cast<const char *>(&FLATBUF_CONTINUZATION), sizeof(int32_t));
+  outfile.write(reinterpret_cast<const char *>(&padded_flatbuf_size), sizeof(int32_t));
+  outfile.write(reinterpret_cast<const char *>(flatbuf_builder->GetBufferPointer()), flatbuf_size);
+  if (padded_flatbuf_size != static_cast<uint32_t>(flatbuf_size)) {
+    outfile.write(ALIGNMENT, padded_flatbuf_size - flatbuf_size);
+  }
+  outfile.flush();
```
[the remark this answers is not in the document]
YuzeLiao: Yes, the suggestion makes a lot of sense. I've encapsulated this part of code into function AssembleMetadataBuffer()

src/storage/data_table.cpp
YuzeLiao: Thank you for pointing it out!

src/include/storage/data_table.h
YuzeLiao: Solved.

src/include/storage/data_table.h
YuzeLiao: Solved!

src/storage/data_table.cpp
```diff
@@ -246,6 +258,251 @@ bool DataTable::Delete(transaction::TransactionContext *const txn, const TupleSl
...
+    // and its children type is Int64. Another example, for type List<List<Int64>>, List<Int64> is the children
+    // of the outer List, and Int64 is the children of the inner List. Since we don't have nested types, we use
+    // an empty array as fake children.
+    std::vector<flatbuffers::Offset<org::apache::arrow::flatbuf::Field>> fake_children;
+    fields.emplace_back(flatbuf::CreateField(*flatbuf_builder, name, true, type, type_offset, dictionary,
+                                             flatbuf_builder->CreateVector(fake_children)));
+  }
+  auto schema =
+      flatbuf::CreateSchema(*flatbuf_builder, flatbuf::Endianness_Little, flatbuf_builder->CreateVector(fields));
+  auto message = flatbuf::CreateMessage(*flatbuf_builder, flatbuf::MetadataVersion_V4, flatbuf::MessageHeader_Schema,
```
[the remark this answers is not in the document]
YuzeLiao: Apache Arrow uses MetadataVersion_V4 when they write an arrow-format message. When they read a message, they support all the possible versions (V1-V4):

[https://github.com/apache/arrow/blob/4c63bef8de079d4e6e559adab822d0f44dc8645d/cpp/src/arrow/ipc/metadata_internal.h#L65](https://github.com/apache/arrow/blob/4c63bef8de079d4e6e559adab822d0f44dc8645d/cpp/src/arrow/ipc/metadata_internal.h#L65)

[https://github.com/apache/arrow/blob/4c63bef8de079d4e6e559adab822d0f44dc8645d/cpp/src/arrow/ipc/metadata_internal.cc#L851](https://github.com/apache/arrow/blob/4c63bef8de079d4e6e559adab822d0f44dc8645d/cpp/src/arrow/ipc/metadata_internal.cc#L851)

Since we only write in our system, we may use some other versions. But use V4 seems to be consistent with Apache Arrow. As long as they guarantee "read" is always compatible with all the versions, we don't need to change this in the future because we only do "write" things.

src/storage/data_table.cpp
YuzeLiao: Solved by just copying the list of blocks (as suggested). It seems to be safe because the block will never be deallocated until the data table is destroyed.

malin1993ml:
Hi @crd477, we're facing some strange issue with this PR on the Jenkins build. Basically we can build it locally and on Travis, but the build always failed during installing gflags on Jenkins under many retries, like [here](http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-616/91/pipeline/40). And this PR does not touch gflags at all. We're wondering would you be able to help us diagnose this issue, or guide us to get the gflags installation error log (`gflags_ep-install-*.log`) printed out under build failure?

crd477:
> Hi @crd477, we're facing some strange issue with this PR on the Jenkins build. Basically we can build it locally and on Travis, but the build always failed during installing gflags on Jenkins under many retries, like [here](http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-616/91/pipeline/40). And this PR does not touch gflags at all. We're wondering would you be able to help us diagnose this issue, or guide us to get the gflags installation error log (`gflags_ep-install-*.log`) printed out under build failure?

I'm looking at this now and it's odd.  I can't reproduce it manually even in the same exact environment that Jenkins uses.

Unfortunately, the gflags logs mentioned in that result got automatically cleaned up but I'll see if I can workaround that temporarily for this branch.  I'll let you know what I find.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -276,4 +276,7 @@ if (${LLVM_PACKAGE_VERSION} VERSION_LESS "8.0")
 endif ()
 llvm_map_components_to_libnames(LLVM_LIBRARIES core mcjit nativecodegen native ipo)
 include_directories(SYSTEM ${LLVM_INCLUDE_DIRS})
-list(APPEND TERRIER_LINK_LIBS ${LLVM_LIBRARIES})
\ No newline at end of file
+list(APPEND TERRIER_LINK_LIBS ${LLVM_LIBRARIES})
+
+#flatbuffers
+include_directories(SYSTEM "${THIRDPARTY_DIR}/flatbuffers/include")
```
crd477: You might want to add a final newline here.

crd477:
OK, I believe this is fixed now.  @YuzeLiao, I recommend backing out your control experiment change(s) and trying again.

The error that was generated inside the container was the result of a nonexistent directory inside the Docker builder image that cmake was trying to write to.  I corrected it by adding that directory and re-rolling the image.  While I was doing that, I took the opportunity to refresh the pre-installed package dependencies since I noticed the significant amount of time that was being spent on that step.  The previous version of the image was about 8 months old already.

I still cannot understand how this was never triggered before...

cmake_modules/ThirdpartyToolchain.cmake
YuzeLiao: Thanks!

YuzeLiao:
> OK, I believe this is fixed now. @YuzeLiao, I recommend backing out your control experiment change(s) and trying again (sorry, was looking at a reversed diff - leave the pandas stuff in...) .
> 
> The error that was generated inside the container was the result of a nonexistent directory inside the Docker builder image that cmake was trying to write to. I corrected it by adding that directory and re-rolling the image. While I was doing that, I took the opportunity to refresh the pre-installed package dependencies since I noticed the significant amount of time that was being spent on that step. The previous version of the image was about 8 months old already.
> 
> I still cannot understand how this was never triggered before...

Thank you Chad @crd477!

malin1993ml:
Thanks a lot, @crd477!

test/storage/export_table_test.cpp
```diff
@@ -0,0 +1,312 @@
...
+  "pa_table = pa_table.to_pandas()\n"                 \
+  "pa_table.to_csv('" CSV_TABLE_NAME "', index=False, header=False)\n"
+
+namespace terrier {
+
+struct ExportTableTest : public ::terrier::TerrierTest {
+  bool ParseNext(std::ifstream &csv_file, char stop_char, char *tmp_char) {
+    csv_file.get(*tmp_char);
+    bool end = false;
+    switch (*tmp_char) {
```
malin1993ml: Could you comment the differences between these cases and the logic to handle them?

test/storage/export_table_test.cpp
```diff
@@ -0,0 +1,312 @@
...
+      if (i < num_tuples) {
+        table.Select(common::ManagedPointer(txn), slot, read_row);
+        for (int j = 0; j < read_row->NumColumns(); j++) {
+          auto col_id = read_row->ColumnIds()[j];
+
+          std::string integer;
+          std::vector<byte> bytes;
+          char tmp_char;
+          csv_file.get(tmp_char);
+          switch (tmp_char) {
```
malin1993ml: Similar here. Could you comment the logic for these cases?

apavlo:
@YuzeLiao You can ignore the Travis error. The build timed out. Your code had nothing to do with that.

The Jenkins errors are real though.

malin1993ml:
@mbutrovich @apavlo This PR is only failing `DataTableBenchmark::SelectSequential` on Jenkins. But I don't think it modified any layout or existing operations in the `DataTable`. It only adds the ability to dump the table in the Arrow format. Should we just merge this?

YuzeLiao:
@malin1993ml It seems `DataTableBenchmark::SelectSequential` also failed on other PRs such as 768 and 790. It may be good to go but feel free to let me know your decision and I can make any further modifications to this PR. Thanks!

malin1993ml:
@YuzeLiao Sounds good, thanks! We'll discuss this during tomorrow's meeting.

mbutrovich:
I have some changes to request that I'll type up later. Don't merge yet please.

malin1993ml:
@mbutrovich I think the review comment is addressed and all the Jenkins tests are passing after merging with the latest master. Shall we merge this one?

PR #619, opened 2019-11-07
Remove DI framework

mbutrovich:
The folks in the development meeting last week came to the conclusion that we should remove the DI framework because it adds too much complexity to the codebase. While we'd like to streamline how we bring up components in tests, this doesn't really save lines of code and creates a high barrier to entry for new developers. While the framework is good at relieving the burden of instantiating components in the correct dependency order, I don't believe the current number of components in the system makes that too burdensome. If that changes in the future, we have the commit history to bring this back if that's what we want.

This PR is marked question/discussion for anyone not at the meeting last week to raise their concerns.


src/include/metrics/metrics_thread.h
```diff
@@ -13,16 +12,16 @@ namespace terrier::metrics {
...
    */
-  BOOST_DI_INJECT(MetricsThread, (named = METRICS_PERIOD) std::chrono::milliseconds metrics_period)  // NOLINT
-  : run_metrics_(true),
-    /// @cond DOXYGEN_IGNORE // TODO(Matt): no idea why this is currently necessary. Doxygen thinks these are functions
-    metrics_paused_(false),
-    metrics_period_(metrics_period),
-    metrics_thread_(std::thread([this] { MetricsThreadLoop(); })) {}
+  MetricsThread(std::chrono::milliseconds metrics_period)  // NOLINT
+      : run_metrics_(true),
+        /// @cond DOXYGEN_IGNORE // TODO(Matt): no idea why this is currently necessary. Doxygen thinks these are
```
lmwnshn: Is this still the case once you get rid of DI?

PR #401, opened 2019-06-03
Banning shared_ptr and std::cerr

lmwnshn:
Remove or NOLINT usages of shared_ptr and std::cerr from the codebase.

PR #621, opened 2019-11-09
Remove shared_ptr from the network layer

mbutrovich:
Needed for #401. This clears up the ownership in the network layer.

I still want to make a pass adding some comments now that I understand the behavior a bit better,  but I can at least see how it does in CI.

### DIFF
--- a/src/network/postgres/postgres_command_factory.cpp
+++ b/src/network/postgres/postgres_command_factory.cpp
@@ -2,7 +2,8 @@
 #include <memory>
 namespace terrier::network {
 
-PostgresNetworkCommand *PostgresCommandFactory::PacketToCommand(InputPacket *packet) {
+std::unique_ptr<PostgresNetworkCommand> PostgresCommandFactory::PacketToCommand(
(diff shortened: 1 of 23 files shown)


src/network/postgres/postgres_command_factory.cpp
```diff
@@ -2,7 +2,7 @@
 #include <memory>
 namespace terrier::network {
 
-std::shared_ptr<PostgresNetworkCommand> PostgresCommandFactory::PacketToCommand(InputPacket *packet) {
+PostgresNetworkCommand *PostgresCommandFactory::PacketToCommand(InputPacket *packet) {
```
tanujnay112: Any reason this shouldn't be a ManagedPointer that's being returned?

mbutrovich: In theory it would be a unique_ptr, if anything, since ownership belongs to the caller. I'll see if I can make that happen with the casting stuff since unique_ptr doesn't have the *_pointer_cast methods that shared_ptr does.

PR #625, opened 2019-11-12
Refactor pointers in Operator and OperatorExpression

mbutrovich:
#401 sent me down a rabbit hole cleaning up the memory leaks from removing shared_ptr. This cleans up the Operator, OperatorExpression, and QueryToOperatorTransformer classes to use unique_ptr and ManagedPointer, removing the raw pointers and making object ownership explicit.


src/optimizer/logical_operators.cpp
```diff
@@ -699,26 +666,21 @@ bool LogicalSemiJoin::operator==(const BaseOperatorNode &r) {
...
-  group_by->having_ = {};
-  return Operator(group_by);
-}
+Operator LogicalAggregateAndGroupBy::Make() { return Operator(std::make_unique<LogicalAggregateAndGroupBy>()); }
 
 Operator LogicalAggregateAndGroupBy::Make(std::vector<common::ManagedPointer<parser::AbstractExpression>> &&columns) {
-  auto *group_by = new LogicalAggregateAndGroupBy;
+  auto group_by = std::make_unique<LogicalAggregateAndGroupBy>();
   group_by->columns_ = std::move(columns);
   group_by->having_ = {};
```
wenxuanqiu: This line can be omitted.

src/include/optimizer/operator_expression.h
```diff
@@ -21,31 +21,34 @@ class OperatorExpression {
...
-  const std::vector<OperatorExpression *> &GetChildren() const { return children_; }
+  OperatorExpression(OperatorExpression &&op) noexcept : op_(std::move(op.op_)), children_(std::move(op.children_)) {}
 
   /**
-   * @return underlying operator
+   * @return vector of children
    */
-  const Operator &GetOp() const { return op_; }
+  std::vector<common::ManagedPointer<OperatorExpression>> GetChildren() const {
+    std::vector<common::ManagedPointer<OperatorExpression>> result;
```
lmwnshn: reserve()

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -28,11 +28,11 @@ QueryToOperatorTransformer::QueryToOperatorTransformer(std::unique_ptr<catalog::
...
 }
 
-OperatorExpression *QueryToOperatorTransformer::ConvertToOpExpression(common::ManagedPointer<parser::SQLStatement> op,
-                                                                      parser::ParseResult *parse_result) {
+std::unique_ptr<OperatorExpression> QueryToOperatorTransformer::ConvertToOpExpression(
+    common::ManagedPointer<parser::SQLStatement> op, parser::ParseResult *parse_result) {
   output_expr_ = nullptr;
   op->Accept(this, parse_result);
-  return output_expr_;
+  return std::move(output_expr_);
```
lmwnshn: Maybe I'm still recovering from hitting my head, is this std::move(nullptr)??

mbutrovich: I don't think so. output_expr_ can change after calling Accept.

lmwnshn: I'm confused where this changes, doesn't come up in a [search](https://github.com/cmu-db/terrier/search?q=output_expr_&unscoped_q=output_expr_). Unless maybe in the optimizer branch?

mbutrovich: I don't really want to be responsible for logic changes in here. I'm just cleaning up pointers. If logic needs to change I'd open an issue.

PR #627, opened 2019-11-12
Extern templates for indexes

mbutrovich:
Shaved about a minute off of the dev4 build time, and on the dinky MacBook Air it goes from 13m down to 11m. It's not huge but it's somewhere to start.

Need to see benchmark results to see if restrictions on inlining from this resulted in any noticeable runtime performance loss.

Sources:
https://en.cppreference.com/w/cpp/language/class_template
> An explicit instantiation declaration (an extern template) skips implicit instantiation step: the code that would otherwise cause an implicit instantiation instead uses the explicit instantiation definition provided elsewhere (resulting in link errors if no such instantiation exists). This can be used to reduce compilation times by explicitly declaring a template instantiation in all but one of the source files using it, and explicitly defining it in the remaining file.

https://arne-mertz.de/2019/02/extern-template-reduce-compile-times/


mbutrovich:
Also removed an unnecessary include from object_pool.h.

mbutrovich:
Compilation times from the Jenkins configs that are dedicated instances, so they're fairly consistent between requests:
```
macos-10.14/AppleClang-1001.0.46.4 (Debug/ASAN/unittest)
master: 7m52s
#627: 6m21s
```

```
macOS-10.14/AppleClang-1001.0.46.4 (Release/unittest)
master: 10m30s
#627: 8m21s
```

```
ubuntu-18.04/gcc-7.3.0 (Release/runbenchmark)
master: 4m56s
#627: 3m41s
```
It's a start. I suspect this will be an interactive process across several PRs.

src/include/common/object_pool.h
```diff
@@ -4,7 +4,6 @@
 #include <string>
 #include <utility>
 #include "common/allocator.h"
-#include "common/container/concurrent_queue.h"
```
mbutrovich: This was an unnecessary include that propagated to quite a few .cpp files in the system.

mbutrovich:
Benchmark regressions are one-off outliers on unrelated changes: parser and DataTable.

PR #632, opened 2019-11-15
Per-connection temporary namespaces, many related fixes

wenxuanqiu:
This PR creates per connection temporary namespaces when a new connection to the server is made. The temporary namespace is dropped when the client disconnects. In addition, I made the following changes,

- Make TransactionManager and Catalog dependencies of TrafficCop and initialize them in DBMain
- Add Int64 as new settings data type
- Update relevant tests


src/traffic_cop/traffic_cop.cpp
```diff
@@ -0,0 +1,37 @@
...
+  auto ns_oid =
+      catalog_->GetAccessor(txn, db_oid)->CreateNamespace(std::string(TEMP_NAMESPACE_PREFIX) + std::to_string(sockfd));
+  txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+  return std::pair<catalog::db_oid_t, catalog::namespace_oid_t>{db_oid, ns_oid};
+}
+
+bool TrafficCop::DropTempNamespace(catalog::namespace_oid_t ns_oid, catalog::db_oid_t db_oid) {
+  auto txn = txn_manager_->BeginTransaction();
+  auto result = catalog_->GetAccessor(txn, db_oid)->DropNamespace(ns_oid);
+  txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
```
mbutrovich: I'm not sure we can commit this unconditionally. In theory through programmer discipline we could make sure this never gets called with invalid args, but we should either add conditional logic on the commit stuff, or asserts to make sure this never gets called with exceptional behavior.

src/traffic_cop/traffic_cop.cpp
```diff
@@ -0,0 +1,51 @@
...
+  auto txn = txn_manager_->BeginTransaction();
+  auto db_oid = catalog_->GetDatabaseOid(txn, database_name);
+  if (db_oid == catalog::INVALID_DATABASE_OID) {
+    return {catalog::INVALID_DATABASE_OID, catalog::INVALID_NAMESPACE_OID};
+  }
+
+  auto ns_oid =
+      catalog_->GetAccessor(txn, db_oid)->CreateNamespace(std::string(TEMP_NAMESPACE_PREFIX) + std::to_string(sockfd));
+
+  if (db_oid == catalog::INVALID_DATABASE_OID) {
```
mbutrovich: Copy and paste error? I think you meant to check ns_oid.

src/traffic_cop/traffic_cop.cpp
```diff
@@ -0,0 +1,51 @@
...
+  TERRIER_ASSERT(replication_log_provider_ != DISABLED, "Should not be handing off logs if no log provider was given");
+  replication_log_provider_->HandBufferToReplication(std::move(buffer));
+}
+
+std::pair<catalog::db_oid_t, catalog::namespace_oid_t> TrafficCop::CreateTempNamespace(
+    int sockfd, const std::string &database_name) {
+  auto txn = txn_manager_->BeginTransaction();
+  auto db_oid = catalog_->GetDatabaseOid(txn, database_name);
+  if (db_oid == catalog::INVALID_DATABASE_OID) {
+    return {catalog::INVALID_DATABASE_OID, catalog::INVALID_NAMESPACE_OID};
```
mbutrovich: Need to abort the txn before returning.

src/traffic_cop/traffic_cop.cpp
wenxuanqiu: Oh shoot. Let me fix that.

src/include/network/postgres/postgres_protocol_utils.h
```diff
@@ -28,7 +28,7 @@ namespace terrier::network {
       {"is_superuser", "on"},
       {"server_encoding", "UTF8"},
       {"server_version", "9.5devel"},
-      {"session_authorization", "postgres"},
```
mbutrovich: Was this done to match the default database name in the Catalog?

src/include/network/connection_context.h
```diff
@@ -13,6 +13,21 @@ namespace terrier::network {
...
 struct ConnectionContext {
+  /**
+   * Commandline arguments parsed from protocol interpreter
+   */
+  std::unordered_map<std::string, std::string> cmdline_args_;
+
+  /**
+   * The OID of the database accessed by this connection
+   */
+  catalog::db_oid_t db_oid_ = catalog::INVALID_DATABASE_OID;
```
mbutrovich: Since it's just a struct we don't have much control over how people use this. If we don't want to make it a proper class, we should document some invariants about db_oid_. Presumably it shouldn't change after the connection is established, and then can't be changed again until the connection is closed and then the context can possibly be reused?

src/include/transaction/deferred_action_manager.h
```diff
@@ -69,6 +73,20 @@ class DeferredActionManager {
...
   }
 
+  /**
+   * Invokes GC and log manager enough times to fully GC any outstanding transactions and process deferred events.
+   * Currently, this must be done 3 times. The log manager must be called because transactions can only be GC'd once
+   * their logs are persisted.
+   * @param gc gc to use for garbage collection
+   * @param log_manager log manager to use for flushing logs
+   */
+  void FullyPerformGC(storage::GarbageCollector *const gc, storage::LogManager *const log_manager) {
```
mbutrovich: Is this the right place for this @GustavoAngulo? What about StorageUtil?

src/main/db_main.cpp
```diff
@@ -32,26 +31,46 @@ DBMain::DBMain(std::unordered_map<settings::Param, settings::ParamInfo> &&param_
...
+  // Initialize the catalog, depends on transaction manager, block store.
+  catalog_ = std::make_unique<catalog::Catalog>(txn_manager_, block_store_);
+
+  // Bootstrap the default database in the catalog.
+  auto *bootstrap_txn = txn_manager_->BeginTransaction();
+  catalog_->CreateDatabase(bootstrap_txn, catalog::DEFAULT_DATABASE, true);
+  txn_manager_->Commit(bootstrap_txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+
+  // Run the GC to get a clean system
+  garbage_collector_->PerformGarbageCollection();
```
mbutrovich: Should we call FullyPerformGC here?

src/include/network/postgres/postgres_protocol_utils.h
wenxuanqiu: This field actually needs to match the user who is requesting access to the database. In all of our tests, the client is currently sending {User : DefaultDatabaseName, Database : DefaultDatabaseName}, with DefaultDatabaseName currently being set to "terrier" by the catalog.

src/main/db_main.cpp
wenxuanqiu: I agree.

src/include/transaction/deferred_action_manager.h
wenxuanqiu: I talked to @yash620 about this. His take is that FullyPerformGC should be something that is bound to an instance of the Garbage Collector. So it should be a member variable of either the GC or the DeferredActionManager.

PR #634, opened 2019-11-19
Traffic Cop, terrier engine instead of Sqlite

lmwnshn:
Basically just glue code that can send incoming SQL queries all the way to the binder.

```
Network -> Parser -> Binder ->  ??? (optimizer) -> TPL
                       ^ we are here!
```

---

Status:

- [x] Create works!
- [ ] I'm apparently not awake enough to fix a memory leak. Will ask Matt for help later.
- [ ] Once the leak is resolved, I'll uncomment four lines in the postgres network commands. If things go well, then it binds and this should be ready to merge by today. If not, then ???


mbutrovich:
So blocked by #484?

lmwnshn:
Yes. But in addition to having those nodes, we need to be able to execute those nodes. Otherwise we will continue to try to bind non-existent things.

mbutrovich:
Should be unblocked now. Will you update the description to reflect the current state and an overview of changes so we can get it reviewed pretty please?

yvxiang:
Does this branch works now?

mbutrovich:
No it's blocked at the moment.

mbutrovich:
Closing since we have a separate branch that this functionality will come from.

Issue #635, opened 2019-11-20
Figure out IF EXISTS for DROP statements


mbutrovich:
`DROP` statements allow for an `IF EXISTS <name>` with the following semantics:

> Do not throw an error if the [object] does not exist. A notice is issued in this case.

For example, case 1:
```
matt=# begin;
BEGIN
matt=# drop table foo;
DROP TABLE
matt=# drop table foo;
ERROR:  table "foo" does not exist
matt=# commit;
ROLLBACK
```
The txn must be rolled back because the second `DROP TABLE` command didn't find the table anymore.

And then we have case 2:
```
matt=# begin;
BEGIN
matt=# drop table foo;
DROP TABLE
matt=# drop table if exists foo;
NOTICE:  table "foo" does not exist, skipping
DROP TABLE
matt=# commit;
COMMIT
```
The txn is allowed to commit, despite the second `DROP TABLE` command not finding the table anymore.

This has ramifications at multiple layers of our architecture:
* Catalog and Execution: Return values from drop functions are not verbose enough to explain why a request may fail. They just return `false` to the caller. This is caused by either:
    * The object was already dropped by the same txn (i.e. the Catalog's index scan will come up empty when looking for the name on second attempt to `DROP`). This would only happen if we reused the plan node for the `DROP TABLE` statement. If we force the statement to go through binding again, then binding should fail. I wrote TODOs in the Catalog code several months ago about this behavior (https://github.com/cmu-db/terrier/blob/643fdf49212d8ab0f9f3eaefb9a53006ef5a9b13/src/catalog/database_catalog.cpp#L656) wondering if it was even reasonable for this code flow to happen.
    * Write-write conflict on the object
* Binder: failing to bind a `DROP` name without an `IF EXISTS` seems like it could short-circuit a failure and trip the `MustAbort` flag in the `TransactionContext`.
* Caching: any binding checks that we expect to happen every time cannot be bypassed. Avoid opportunistically caching DDL queries entirely?

It seems like the binder could handle this stuff, rather than pushing `IF EXISTS` logic down to the Execution and Catalog layers, as is implied by the presence of the flag in the physical plan nodes. This allows Execution and Catalog to be more deterministic and strict in their code flow if they can rely on the binder to be handling these cases, but I'd like to open the issue to discussion in case there's something I'm overlooking.

lmwnshn:
We currently support IF EXISTS (#687), the code may just need to be pushed around

PR #636, opened 2019-11-20
CREATE/DROP plan node cleanup and executors

mbutrovich:
Overview of changes:

- There's a static util class in `execution/sql` that can take in `CREATE` or `DROP` plan nodes along with an `ExecutionContext` and perform the operation. Is this the API we want? Names are undecided and easy to change. I'm not happy with them right now.
- I gutted any fields that weren't required to execute these plan nodes. Note: I did not remove fields that we would eventually want to support (i.e. constraints for CREATE TABLE).  I only removed fields that legitimately seemed redundant.
- Removed `SetFromCreateStatement` and `SetFromDropStatement` methods on plan node Builders. These were used for a few contrived tests, but represent a special case code path that will not be used in the real system. All plan nodes will be generated by going through the parser, binder, and optimizer.
- Removed references to skiplist and ART indexes from `PostgresParser`. I'm not a fan of dead code coming over from Peloton.
- Typing up #635 pretty much convinced me that `IF EXISTS` should not be in physical plan nodes so I removed that field, but that issue exists to discuss it.

### DIFF
--- a/src/include/parser/parser_defs.h
+++ b/src/include/parser/parser_defs.h
@@ -60,8 +60,9 @@ enum class IndexType {
   INVALID = INVALID_TYPE_ID,  // invalid index type
   BWTREE = 1,                 // bwtree
   HASH = 2,                   // hash
-  SKIPLIST = 3,               // skiplist
-  ART = 4,                    // ART
+  // TODO(Matt) are we ever gonna support these again? if not we should remove them
+  //  SKIPLIST = 3,               // skiplist
+  //  ART = 4,                    // ART
 };
 
 enum class InsertType {
--- a/src/include/parser/parser_defs.h
+++ b/src/include/parser/parser_defs.h
@@ -60,9 +60,6 @@ enum class IndexType {
   INVALID = INVALID_TYPE_ID,  // invalid index type
   BWTREE = 1,                 // bwtree
   HASH = 2,                   // hash
-  // TODO(Matt) are we ever gonna support these again? if not we should remove them
-  //  SKIPLIST = 3,               // skiplist
-  //  ART = 4,                    // ART
 };
 
 enum class InsertType {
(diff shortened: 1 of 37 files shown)


mbutrovich:
While working on this, I realized that the only drop operation that currently cascades in the Catalog is drop database. In an effort to limit the scope of this PR, I'm pushing cascading to the next PR since we'll still need that behavior, particularly for drop namespace for #632.

src/include/parser/parser_defs.h
```diff
@@ -60,8 +60,9 @@ enum class IndexType {
   INVALID = INVALID_TYPE_ID,  // invalid index type
   BWTREE = 1,                 // bwtree
   HASH = 2,                   // hash
-  SKIPLIST = 3,               // skiplist
-  ART = 4,                    // ART
+  // TODO(Matt) are we ever gonna support these again? if not we should remove them
```
lmwnshn: We could just go ahead and remove them. If/when someone adds them, they can easily add it here.

src/parser/postgresparser.cpp
```diff
@@ -1314,10 +1314,11 @@ std::unique_ptr<SQLStatement> PostgresParser::CreateIndexTransform(ParseResult *
     index_type = IndexType::BWTREE;
   } else if (strcmp(access_method, "hash") == 0) {
     index_type = IndexType::HASH;
-  } else if (strcmp(access_method, "skiplist") == 0) {
-    index_type = IndexType::SKIPLIST;
-  } else if (strcmp(access_method, "art") == 0) {
-    index_type = IndexType::ART;
+    // TODO(Matt) are we ever gonna support these again? if not we should remove them
```
lmwnshn: Can probably be removed.

mbutrovich:
Want to also document that this PR strengthens the asserts in the Catalog and removes some TODOs related to `IF EXISTS` and future expected behavior from the binder. I will open a new issue for that.

PR #638, opened 2019-11-21
Larger attribute size

### LINKED ISSUE #630: Change block layout column attribute size to be 16 bits
Changing attribute size for block layout to be 16 bits. The most significant bit will be a flag for whether the attribute is a varlen or not. This will give room for having 2^15 byte attributes.

Mostly implemented in: https://github.com/yash620/terrier/tree/yash-enforce-projection-delimiters

Need to fix 4 failing test cases:
block_compactor_test: seg_fault
large_garbage_collector_test: trying to free memory that wasn't allocated
log_test: not empty, not all txns left are read only
recovery_test: Unknown log record type during deserialization: 0

tli2:
fix #630 

We were writing ad-hoc bit masking tricks to get the varlen sign bit out in multiple places of the code base. I added a macro for that so modifying it in the future will be easier.

Otherwise, attribute sizes are not present in any code where we manually lay out bytes, so going to 32 bit, if needed in the future, would not be hard.

### DIFF
--- a/src/include/storage/block_layout.h
+++ b/src/include/storage/block_layout.h
@@ -9,7 +9,7 @@ namespace terrier::storage {
 // Internally we use the sign bit to represent if a column is varlen or not. Down to the implementation detail though,
 // we always allocate 16 bytes for a varlen entry, with the first 8 bytes being the pointer to the value and following
 // 4 bytes be the size of the varlen. There are 4 bytes of padding for alignment purposes.
-#define VARLEN_COLUMN static_cast<uint8_t>(0x90)  // 16 with the first (most significant) bit set to 1
+#define VARLEN_COLUMN static_cast<uint16_t>(0x90)  // 16 with the first (most significant) bit set to 1
 /**
  * Stores metadata about the layout of a block.
  */
@@ -25,7 +25,7 @@ class BlockLayout {
    *
    * @param attr_sizes vector of attribute sizes.
    */
-  explicit BlockLayout(std::vector<uint8_t> attr_sizes);
+  explicit BlockLayout(std::vector<uint16_t> attr_sizes);
 
   /**
    * @return number of columns.
@@ -37,9 +37,9 @@ class BlockLayout {
    * @param col_id the column id to check for
    * @return attribute size at given col_id.
    */
-  uint8_t AttrSize(col_id_t col_id) const {
+  uint16_t AttrSize(col_id_t col_id) const {
     // mask off the first bit as we use that to check for varlen
-    return static_cast<uint8_t>(INT8_MAX & attr_sizes_.at(!col_id));
+    return static_cast<uint16_t>(INT16_MAX & attr_sizes_.at(!col_id));
   }
 
   /**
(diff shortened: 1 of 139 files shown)


src/include/storage/block_layout.h
```diff
@@ -9,7 +9,8 @@ namespace terrier::storage {
 // Internally we use the sign bit to represent if a column is varlen or not. Down to the implementation detail though,
 // we always allocate 16 bytes for a varlen entry, with the first 8 bytes being the pointer to the value and following
 // 4 bytes be the size of the varlen. There are 4 bytes of padding for alignment purposes.
-#define VARLEN_COLUMN static_cast<uint8_t>(0x90)  // 16 with the first (most significant) bit set to 1
+#define VARLEN_COLUMN static_cast<uint16_t>(0x8010)  // 16 with the first (most significant) bit set to 1
```
yash620: When I experimented with this 0x9000 and 0x8000 wouldn't work and only 0x8010 would. I was wondering what the second 1 is used for. From the rest of the code it seems like only the most significant bit being set to 1 is the indicator for that attribute size being a varlen.

src/storage/recovery/recovery_manager.cpp
```diff
@@ -320,7 +320,7 @@ void RecoveryManager::UpdateIndexesOnTable(transaction::TransactionContext *txn,
       if (table_pr->IsNull(pr_map[table_col_oid])) {
         index_pr->SetNull(index->GetKeyOidToOffsetMap().at(index_col_oid));
       } else {
-        auto size = col.AttrSize() & INT8_MAX;
+        auto size = col.AttrSize() & INT_MAX;
```
yash620: INT_MAX might be misinterpreted. The initial way I would interpret this is that this is a 32 bit int max since that's the default size of int. But in this case I think we are looking for a 16 bit int max.

src/include/storage/block_layout.h
tli2: the remaining bits have to be 16 as explained by the comment

src/storage/recovery/recovery_manager.cpp
tli2: This looks like I slipped up. Nice catch!

src/include/storage/block_layout.h
```diff
@@ -9,7 +9,8 @@ namespace terrier::storage {
...
+#define ATTR_SIZE_BYTES(size) static_cast<uint16_t>((size)&INT16_MAX)
```
apavlo: Add a comment to say what this function does and when would somebody need to use it.

src/include/storage/block_layout.h
```diff
@@ -25,7 +26,7 @@ class BlockLayout {
    *
    * @param attr_sizes vector of attribute sizes.
    */
-  explicit BlockLayout(std::vector<uint8_t> attr_sizes);
+  explicit BlockLayout(std::vector<uint16_t> attr_sizes);
```
apavlo: Are we likely to ever need to change this? Should we make a `#define` for it so that we know all the locations where it is referenced in the code?

tli2: @mbutrovich probably has a better idea than me?

mbutrovich: Do we anticipate inline attributes larger than 32KB? Moving to uint32_t for attr_sizes would necessitate a change to block size.

tli2: We require that each block can hold at least one tuple. Up until now we have been statically defining a "max number of columns" such that even if all columns are the largest possible, a block can still hold at least one tuple.

This one probably already breaks the old value, but we have asserts in place to catch it when it goes wrong. I actually believe that checking on the fly is preferable to statically fixing a max number of columns. The code to do so is already in there as asserts and tests. 

I can file an issue for this if we believe that is the way forward.

PR #641, opened 2019-11-22
DDL lock in DatabaseCatalog

mbutrovich:
This is a quick way to avoid write skew anomalies in the catalog when we have concurrent DDL changes that won't trigger write-write conflicts in the storage layer. Any transactions attempting to perform DDL changes must first acquire the database-level lock.

Long term this could be handled with more robust solutions like precision locking (see HyPer paper) but this will save us for now. In the meantime, any DDL change implementations added to the catalog should make sure to honor these semantics and acquire the lock first.

This will allow me to implement DROP SCHEMA (namespace) without worrying about another transaction adding objects to that namespace at the same time.

### DIFF
--- a/script/installation/packages.sh
+++ b/script/installation/packages.sh
@@ -64,8 +64,6 @@ install_mac() {
   fi
   # Update Homebrew.
   brew update
-  brew upgrade
-  brew cleanup
   # Install packages.
   brew ls --versions cmake || brew install cmake
   brew ls --versions coreutils || brew install coreutils
--- a/script/installation/packages.sh
+++ b/script/installation/packages.sh
@@ -64,6 +64,8 @@ install_mac() {
   fi
   # Update Homebrew.
   brew update
+  brew upgrade
+  brew cleanup
   # Install packages.
   brew ls --versions cmake || brew install cmake
   brew ls --versions coreutils || brew install coreutils
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -1787,8 +1787,8 @@ Column DatabaseCatalog::MakeColumn(storage::ProjectedRow *const pr, const storag
 bool DatabaseCatalog::TryLock(transaction::TransactionContext *const txn) {
   auto current_val = write_lock_.load();
 
-  const transaction::timestamp_t txn_id = txn->FinishTime();
-  const transaction::timestamp_t start_time = txn->StartTime();
+  const transaction::timestamp_t txn_id = txn->FinishTime();     // this is the uncommitted txn id
+  const transaction::timestamp_t start_time = txn->StartTime();  // this is the unchanging start time of the txn
 
   const bool already_hold_lock = current_val == txn_id;
   if (already_hold_lock) return true;
@@ -1800,13 +1800,14 @@ bool DatabaseCatalog::TryLock(transaction::TransactionContext *const txn) {
   if (owned_by_other_txn || newer_committed_version) return false;
 
   if (write_lock_.compare_exchange_strong(current_val, txn_id)) {
+    // acquired the lock
     auto *const write_lock = &write_lock_;
-    auto release = [=]() -> void { write_lock->store(txn->FinishTime()); };
-    txn->RegisterCommitAction(release);
-    txn->RegisterAbortAction(release);
+    txn->RegisterCommitAction([=]() -> void { write_lock->store(txn->FinishTime()); });
+    txn->RegisterAbortAction([=]() -> void { write_lock->store(current_val); });
     return true;
   }
-  txn->MustAbort();
+  txn->MustAbort();  // though no changes were written to the storage layer, we'll treat this as a DDL change failure
+                     // and force the txn to rollback
   return false;
 }
 
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -1804,7 +1804,11 @@ bool DatabaseCatalog::TryLock(transaction::TransactionContext *const txn) {
   const bool newer_committed_version = transaction::TransactionUtil::Committed(current_val) &&
                                        transaction::TransactionUtil::NewerThan(current_val, start_time);
 
-  if (owned_by_other_txn || newer_committed_version) return false;
+  if (owned_by_other_txn || newer_committed_version) {
+    txn->MustAbort();  // though no changes were written to the storage layer, we'll treat this as a DDL change failure
+    // and force the txn to rollback
+    return false;
+  }
 
   if (write_lock_.compare_exchange_strong(current_val, txn_id)) {
     // acquired the lock
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -1786,16 +1786,22 @@ Column DatabaseCatalog::MakeColumn(storage::ProjectedRow *const pr, const storag
 
 bool DatabaseCatalog::TryLock(transaction::TransactionContext *const txn) {
   auto current_val = write_lock_.load();
-  if (current_val == txn->FinishTime()) return true;  // already hold the lock
 
-  if (transaction::TransactionUtil::NewerThan(txn->StartTime(), current_val) &&
-      write_lock_.compare_exchange_strong(current_val, txn->FinishTime())) {
+  const transaction::timestamp_t txn_id = txn->FinishTime();
+  const transaction::timestamp_t start_time = txn->StartTime();
+
+  const bool already_hold_lock = current_val == txn_id;
+  if (already_hold_lock) return true;
+
+  const bool owned_by_other_txn = (!transaction::TransactionUtil::Committed(current_val) && !already_hold_lock);
+  const bool newer_committed_version = transaction::TransactionUtil::Committed(current_val) &&
+                                       transaction::TransactionUtil::NewerThan(current_val, start_time);
+
+  if (owned_by_other_txn || newer_committed_version) return false;
+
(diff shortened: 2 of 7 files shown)


script/installation/packages.sh
```diff
@@ -64,6 +64,8 @@ install_mac() {
   fi
   # Update Homebrew.
   brew update
+  brew upgrade
+  brew cleanup
```
jrolli: This seems like its fixing an unrelated problem.  Is it OK for these changes to wait on this PR?

src/catalog/database_catalog.cpp
```diff
@@ -785,6 +789,7 @@ table_oid_t DatabaseCatalog::GetTableOid(transaction::TransactionContext *const
 
 bool DatabaseCatalog::SetTablePointer(transaction::TransactionContext *const txn, const table_oid_t table,
                                       const storage::SqlTable *const table_ptr) {
+  if (!TryLock(txn)) return false;
```
jrolli: This should probably be replaced by an assertion that the current transaction already owns the lock since this should only occur after a call to `CreateTable`.

src/catalog/database_catalog.cpp
```diff
@@ -1073,6 +1082,7 @@ bool DatabaseCatalog::SetClassPointer(transaction::TransactionContext *const txn
 
 bool DatabaseCatalog::SetIndexPointer(transaction::TransactionContext *const txn, const index_oid_t index,
                                       const storage::index::Index *const index_ptr) {
+  if (!TryLock(txn)) return false;
```
jrolli: This should probably be replaced by an assertion that the current transaction already owns the lock since this should only occur after a call to `CreateIndex`.

src/catalog/database_catalog.cpp
```diff
@@ -1774,6 +1784,33 @@ Column DatabaseCatalog::MakeColumn(storage::ProjectedRow *const pr, const storag
...
+bool DatabaseCatalog::TryLock(transaction::TransactionContext *const txn) {
+  auto current_val = write_lock_.load();
+
+  const transaction::timestamp_t txn_id = txn->FinishTime();     // this is the uncommitted txn id
+  const transaction::timestamp_t start_time = txn->StartTime();  // this is the unchanging start time of the txn
+
+  const bool already_hold_lock = current_val == txn_id;
+  if (already_hold_lock) return true;
+
+  const bool owned_by_other_txn = (!transaction::TransactionUtil::Committed(current_val) && !already_hold_lock);
```
jrolli: nit: `!already_hold_lock` is always true because of the if-statement and return above.

src/catalog/database_catalog.cpp
```diff
@@ -1774,6 +1784,33 @@ Column DatabaseCatalog::MakeColumn(storage::ProjectedRow *const pr, const storag
...
+  const bool newer_committed_version = transaction::TransactionUtil::Committed(current_val) &&
+                                       transaction::TransactionUtil::NewerThan(current_val, start_time);
+
+  if (owned_by_other_txn || newer_committed_version) return false;
```
jrolli: Do we not have a function/macro that encapsulates this logic?  It seems like we should since this is identical to asking "do I have a write-write conflict".

src/include/catalog/database_catalog.h
```diff
@@ -348,6 +349,18 @@ class DatabaseCatalog {
   friend class postgres::Builder;
   friend class storage::RecoveryManager;
 
+  /**
+   * Internal function to DatabaseCatalog to disallow concurrent DDL changes. This also disallows older txns to enact
+   * DDL chances after a newer transaction has committed one. This effectively follows the same timestamp ordering logic
```
jrolli:
```suggestion
   * DDL changes after a newer transaction has committed one. This effectively follows the same timestamp ordering logic
```

src/include/catalog/database_catalog.h
```diff
@@ -348,6 +349,18 @@ class DatabaseCatalog {
...
+   * as the version pointer MVCC stuff in the storage layer.
```
jrolli:
```suggestion
   * as the version pointer MVCC stuff in the storage layer.  It also serializes all DDL within a database.
```

src/include/catalog/database_catalog.h
```diff
@@ -348,6 +349,18 @@ class DatabaseCatalog {
...
+   * @param txn Requesting txn. This is used to inspect the timestamp and register commit/abort events to release the
+   * lock if it is acquired.
+   * @return true if lock was acquired, false otherwise
+   * @warning this requires that commit and abort actions be performed after the commit time is stored in the
+   * TransactionContext's FinishTime.
+   */
+  bool TryLock(transaction::TransactionContext *txn);
```
jrolli: Is there a use-case for returning the lock's "timestamp" instead of a boolean?  I'm wondering if there is a situation where you may want to spin on it when it is held but abort if it is newer than you.

test/catalog/catalog_test.cpp
```diff
@@ -502,4 +502,64 @@ TEST_F(CatalogTests, GetIndexObjectsTest) {
   txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
 }
 
+/*
+ * Exercise a bunch of scenarios of the DDL lock semantics
+ */
+// NOLINTNEXTLINE
+TEST_F(CatalogTests, DDLLockTest) {
+  // txn2 is used to verify that older txns can't acquire the lock
+  auto *txn2 = txn_manager_->BeginTransaction();
```
jrolli: nit: It's easier to understand the tested ordering if the the number in `txn#` corresponds to the BEGIN ordering since it will at least be consistent with the begin timestamp ordering.

test/catalog/catalog_test.cpp
```diff
@@ -502,4 +502,64 @@ TEST_F(CatalogTests, GetIndexObjectsTest) {
...
+  EXPECT_EQ(ns_oid, catalog::INVALID_NAMESPACE_OID);  // Should fail to get the lock because txn0 holds it (txn1 > txn0)
+  txn_manager_->Abort(txn1);
+
+  auto accessor2 = catalog_->GetAccessor(txn2, db_);
+  ns_oid = accessor2->CreateNamespace("test_namespace2");
+  EXPECT_EQ(ns_oid, catalog::INVALID_NAMESPACE_OID);  // Should fail to get the lock because txn0 holds it (txn2 < txn0)
+
+  txn_manager_->Commit(txn0, transaction::TransactionUtil::EmptyCallback, nullptr);  // txn0 releases the lock
+
+  ns_oid = accessor2->CreateNamespace("test_namespace2");
```
jrolli: Semantic violation of failure makes this undefined behavior.  The first failure should force `txn2` to abort so this should fail regardless of the lock state.

test/catalog/catalog_test.cpp
```diff
@@ -502,4 +502,64 @@ TEST_F(CatalogTests, GetIndexObjectsTest) {
...
+  auto accessor4 = catalog_->GetAccessor(txn4, db_);
+  EXPECT_NE(accessor4, nullptr);
+  VerifyCatalogTables(*accessor4);  // Check visibility to me
+  ns_oid = accessor4->GetNamespaceOid(
+      "test_namespace0");  // Should succeed as txn4 acquires the lock (txn4 < txn3) but txn3 aborted
+  EXPECT_TRUE(accessor4->DropNamespace(ns_oid));
+  ns_oid = accessor4->GetNamespaceOid("test_namespace0");
+  EXPECT_EQ(ns_oid, catalog::INVALID_NAMESPACE_OID);
+  txn_manager_->Commit(txn4, transaction::TransactionUtil::EmptyCallback, nullptr);  // txn4 releases the lock
+}
```
jrolli: Missing sequence:
```
T1              T2
--------------------------------
BEGIN
                BEGIN
TryLock
COMMIT
                TryLock (fails)
```

Still need to test because currently tested via an undefined code path.
```
T1              T2
--------------------------------
                BEGIN
BEGIN
TryLock
COMMIT
                TryLock (fails)
```

script/installation/packages.sh
mbutrovich: Nope, this wouldn't have gotten through CI without it. Needs to be triggered from time to time and this was faster than bugging Chad. Plan was to revert on the review commit.

src/catalog/database_catalog.cpp
mbutrovich: It's HasConflict in the storage layer but the function signature is different. I can see what I can come up with.

src/catalog/database_catalog.cpp
mbutrovich: Breaks Bootstrap because it calls directly to the internal CreateTableEntry API. Bah.

src/include/catalog/database_catalog.h
mbutrovich: Punt.

test/catalog/catalog_test.cpp
mbutrovich: I think I added this schedule.

mbutrovich:
Same reason that was failing recover_test. TPCC's setup calls SetTablePointer without acquiring the lock first. It's because I broke each step out with its own txn, but semantically that doesn't really work with how the catalog is supposed to be used. I'll fix that.

mbutrovich:
Benchmark machine showed a regression on catalog_benchmark, which doesn't make sense since we only profile the getters for the catalog which are oblivious of the DDL lock. I can't reproduce the regression locally on any machines, so will chalk it up to a deviation in CI. I will keep an eye on it though.

PR #642, opened 2019-11-25
 More GCC 8+ fixes

gonzalezjo:
This changeset allows for compilation under GCC9 with LLVM 9 libraries. 

This is not-for-merge for a variety of reasons, including but not limited to:

1. ~~The Wclass-memaccess workaround is a bit cludgy (though, interestingly, RocksDB did the same thing I'm doing here when they switched to GCC9)~~ Still somewhat cludgy, but I think that adding comments has resolved most of the issue here.

2. I'm not convinced there's any interest in a PR regarding toolchain updates at the moment

3. Code review is of course still needed.

The specific changes that have been made are: 

1. Fixing a misleading part of the build configuration that mandates that the LLVM version be exactly 8.0 (and not greater), while stating that it wants an LLVM version greater than *or equal to* 8. This can be made into its own PR.

2.  Fix seemingly erroneous use of std::move, which is something that newer versions of GCC throw warnings for. This can also be brought into its own PR.

3. A workaround for -Waddress-of-packed-member. There's a potentially unaligned read that GCC9 warns (and with -Werror, errors) about. 

4. Patches surrounding various type-related checks that are new in GCC8 and 9. I'm not a fan of the solution used to address the new -Wclass-memaccess checks, but I went with the approach seen here for a few reasons: (1) other large projects like RocksDB did it, so surely it can't be that bad; and, (2) addressing it "properly" would require tweaking ProjectedRow, RawBlock, and Random, which might introduce more complexity than would be optimal. ~~A middle-ground between clearer code and more code might be to add comments clarifying why I'm re-casting to `void*`~~ Edit: Comments have been added.

### DIFF
--- a/benchmark/storage/tuple_access_strategy_benchmark.cpp
+++ b/benchmark/storage/tuple_access_strategy_benchmark.cpp
@@ -61,7 +61,8 @@ BENCHMARK_DEFINE_F(TupleAccessStrategyBenchmark, SimpleInsert)(benchmark::State 
       // Get a Block, zero it, and initialize
       storage::RawBlock *raw_block = block_store_.Get();
       raw_blocks_.emplace_back(raw_block);
-      std::memset(raw_block, 0, sizeof(storage::RawBlock));
+      // Recast raw_block as a -Wclass-memaccess workaround
+      std::memset(static_cast<void *>(raw_block), 0, sizeof(storage::RawBlock));
       tested.InitializeRawBlock(nullptr, raw_block, storage::layout_version_t(0));
       for (uint32_t j = 0; j < layout_.NumSlots(); j++) {
         storage::TupleSlot slot;
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -546,7 +546,7 @@ std::vector<Column> DatabaseCatalog::GetColumns(transaction::TransactionContext 
   // Finish
   delete[] buffer;
   delete[] key_buffer;
-  return std::move(cols);
+  return cols;
 }
 
 // TODO(Matt): we need a DeleteColumn()
--- a/src/execution/vm/bytecode_emitter.cpp
+++ b/src/execution/vm/bytecode_emitter.cpp
@@ -75,7 +75,7 @@ void BytecodeEmitter::Bind(BytecodeLabel *label) {
     auto &jump_locations = label->ReferrerOffsets();
 
     for (const auto &jump_location : jump_locations) {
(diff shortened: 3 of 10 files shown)


mbutrovich:
FYI this branch seems to compile without issues on Ubuntu 20.04 (which will be our next targeted platform), so it's definitely on the right track! Nice job!

gonzalezjo:
Thank you. 


I think this is ready for review. I'm curious about the benchmarks failing on CI, though. This shouldn't be a performance regression (it's practically all syntactic changes) in any way, and the failed benchmarks especially shouldn't be affected. The error bars on Jenkins benchmarks seem somewhat big as well.


Based on [this](https://github.com/cmu-db/terrier/pull/641#issuecomment-558313906) I feel like it's almost certainly just statistical noise, but I've re-merged so we should see new CI results soon.

gonzalezjo:
Huh. No clue what happened there. Is there a way to get the CI bot to re-execute?

mbutrovich:
Similar to the disclaimer I gave in #317, happy to review and merge this with the caveat that we may break compatibility with GCC8/9 before it's part of our CI (likely next summer when we add Ubuntu 20.04 LTS). This should at least reduce the amount of work when that time comes though.

gonzalezjo:
Sounds good to me.

cmake_modules/ThirdpartyToolchain.cmake
```diff
@@ -269,7 +269,7 @@ include_directories(SYSTEM ${PQXX_INCLUDE_DIRECTORIES})
 list(APPEND TERRIER_LINK_LIBS ${PQXX_LIBRARIES})
 
 # LLVM 8.0
-find_package(LLVM 8.0 REQUIRED CONFIG)
```
mbutrovich: This is an artifact of when we started by supporting LLVM 6+, but once we moved to requiring 8 only we left the logic. I'm concerned that removing that will cause issues for people who have multiple versions of LLVM installed. Let's restrict the scope of this PR just to GCC fixes, and revisit LLVM later.

gonzalezjo: Makes sense. This should be fine for now, then.

mbutrovich:
Something weird happened when you force-pushed that last commit. The git diff includes the 2 most recent PRs from master. Maybe merge in upstream/master then push again? It'll probably complain about merge conflicts now since it says you modified those lines.

gonzalezjo:
Huh, strange. I think we're good now? I think that came from hitting the "Update branch" button on GitHub that shows up on this PR page for me. I'm just going to _not_ press that.

PR #644, opened 2019-11-26
Cascading DROP logic for table and namespace

mbutrovich:
- `DatabaseCatalog::DeleteTable()` now drops any indexes on the table being dropped. Documentation already said it did this, so no change there.
- `DatabaseCatalog::DeleteNamespace()` now drops any tables and indexes on the namespace being dropped. Documentation already said it did this, so no change there.
- Added a couple of internal functions to make it easier, though these could be exposed via `CatalogAccessor` if desired in the future.
- Test cases.

### DIFF
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -371,10 +371,11 @@ bool DatabaseCatalog::DeleteNamespace(transaction::TransactionContext *const txn
     return false;
   }
 
+  // Step 4: Cascading deletes
   // Get the objects in this namespace
   auto ns_objects = GetNamespaceObjectOids(txn, ns_oid);
   for (const auto object : ns_objects) {
-    // delete all of the tables. This should get most of the indexes
+    // Delete all of the tables. This should get most of the indexes
     if (object.second == postgres::ClassKind::REGULAR_TABLE) {
       result = DeleteTable(txn, static_cast<table_oid_t>(object.first));
       if (!result) {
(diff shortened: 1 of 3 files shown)


src/catalog/database_catalog.cpp
```diff
@@ -371,13 +371,46 @@ bool DatabaseCatalog::DeleteNamespace(transaction::TransactionContext *const txn
...
+    if (object.second == postgres::ClassKind::INDEX) {
+      result = DeleteIndex(txn, static_cast<index_oid_t>(object.first));
+      if (!result) {
+        // Someone else has a write-lock. Free the buffer and return false to indicate failure
+        delete[] buffer;
+        return false;
+      }
+    }
+  }
+
```
jrolli: It is probably worth adding a third fetch inside an assert here that ensures the results are empty.  This will catch new object types (like views) as we add them in the future.

src/catalog/database_catalog.cpp
```diff
@@ -637,6 +670,21 @@ table_oid_t DatabaseCatalog::CreateTable(transaction::TransactionContext *const
...
 
+bool DatabaseCatalog::DeleteIndexes(transaction::TransactionContext *const txn, const table_oid_t table) {
+  if (!TryLock(txn)) return false;
+  // Get the indexes
+  const auto index_oids = GetIndexOids(txn, table);
+  // Delete all indexes
+  for (const auto index_oid : index_oids) {
+    auto result = DeleteIndex(txn, index_oid);
+    if (!result) {
+      // write-write conflict. Someone beat us to this operation.
```
jrolli: Is this possible with the database-level DDL lock?

src/catalog/database_catalog.cpp
```diff
@@ -1661,6 +1712,41 @@ bool DatabaseCatalog::CreateTableEntry(transaction::TransactionContext *const tx
   return true;
 }
 
+std::vector<std::pair<uint32_t, postgres::ClassKind>> DatabaseCatalog::GetNamespaceObjectOids(
```
jrolli: nit: This is labeled "object" OIDs but it is only "class" OIDs and does not include querying `pg_type`.

I know we haven't implemented that really, but where are we thinking that would get hooked in?  Here or in the main namespace function?

test/catalog/catalog_test.cpp
```diff
@@ -274,6 +274,203 @@ TEST_F(CatalogTests, UserIndexTest) {
...
+
+  EXPECT_TRUE(accessor->SetTablePointer(table_oid, table));
+  txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+
+  // Create the index
+  txn = txn_manager_->BeginTransaction();
+  accessor = catalog_->GetAccessor(txn, db_);
+  EXPECT_NE(accessor, nullptr);
+  std::vector<catalog::IndexSchema::Column> key_cols{catalog::IndexSchema::Column{
+      "id", type::TypeId::INTEGER, false, parser::ColumnValueExpression(db_, table_oid, schema.GetColumn("id").Oid())}};
```
jrolli: You should have this index be on a table *not* in the dropped namespace in order to test the second pass logic in your code.  As is, this index is really a hidden second test of `DeleteTable`'s cascading logic.

src/catalog/database_catalog.cpp
mbutrovich: It is not. But since the DDL lock is a possible temporary hack, I figured I'd be as consistent with existing logic as possible.

test/catalog/catalog_test.cpp
mbutrovich: Isn't that what the test after this does?

PR #645, opened 2019-11-30
Exec engine microbenchmark

thepinetree:
This benchmark simulates a sequential scan over a filled table to determine the total amount of time required for a complete search through the table. We simulate a table with random initialization to mirror realistic transaction behavior.


pmenon:
@thepinetree You've accidentally added intermediate CMake build files. Please remove them from the PR.

benchmark/transaction/seqscan_benchmark.cpp
```diff
@@ -0,0 +1,152 @@
+#include <common/scoped_timer.h>
```
mbutrovich: Should probably make a new folder called `execution` rather than have it in `transaction`.

PR #648, opened 2019-12-07
DBMain and various other refactors

### LINKED ISSUE #604: Remove global LOG_* and TEST_LOG_* targets
Per our discussion today, we want to remove the global and generic test log targets:

https://github.com/cmu-db/terrier/blob/master/src/include/loggers/main_logger.h
https://github.com/cmu-db/terrier/blob/master/src/include/loggers/test_logger.h

This will break any `LOG_*` and `TEST_LOG_*` usages. We should convert any test code that uses this target to use their component specific target.

mbutrovich:
Overview of changes:
- `DBMain` now has some nested classes defining "layers" within the system for sub-components that need to be brought up together. This should make dependencies more explicit.
- Debug loggers are now instantiated in the outermost scope of all executables since they need to be brought up before any parts of the system, and down after any parts of the system. That means it can't be `DBMain`'s job. This could be `DBMain`'s job if you have a dummy handle object to the loggers, but then you need special case logic to avoid redundant calls depending on if you're in a Google Benchmark, Google Test, or Terrier itself.
- `LargeDataTableBenchmarkObject` takes the `MetricsManager` instead of `MetricsThread` to simplify the dependency. Also removed some random PR population code that just slows it down. We don't do correctness testing in the benchmarks so true randomness isn't needed.
- `DBMain` now has a `Builder` that should yell at you if you request an invalid configuration. It also has a set of reasonable (for tests) defaults. Some scenarios will need to override these, while the real system uses the SettingsManager.
- Fix #604. The test logger is gone. The global one is now `common_logger`. I wanted to get rid of it entirely, but there are some debug logging statements from stuff in the common folder and wasn't sure what to do with that.
- Refactored a lot of tests and some benchmarks to use the new `DBMain` builder. Benchmarks could still be refactored further, but because the life cycle of the `BlockStore` is now tied to a `DBMain` instance rather than the life of the benchmark, I noticed a benchmark regression because the `BlockStore` was no longer "warmed up" after the first iteration. I reverted most of the benchmarks for that reason. Someone can tackle reasoning about that in a later PR.
- Refactored a lot of components of the system to take `ManagedPointer` or `std::unique_ptr` for their dependencies to make ownership (or lack of) explicit.
- Remove some hardcoded settings from `settings.h`.
- Added methods to start and stop GC for `GarbageCollectorThread` since some tests expect this behavior. This class will likely go away in the future with #593 anyway.
- Moved the parameter map ownership to the `SettingsManager` rather than leaving it in `DBMain`. Previously there was a dependency back to `DBMain` from the `SettingsManager` which created circular dependencies if you wanted `DBMain` to own the `SettingsManager`.
- Removed dependency between `SettingsManager` and `Optimizer`.

For the review I'm mostly interested in feedback on the `DBMain` class itself: is it documented enough, is it usable, etc. The rest of the massive diff is mostly related to refactoring tests for `ManagedPointer` or the new `DBMain` class. Functionally all the tests are identical, and since they pass CI (actually more reliably now that the races on shutting down sub-components should be resolved) I'm not that concerned about going over the tests too closely.


benchmark/benchmark_util/data_table_benchmark_util.cpp
```diff
@@ -156,7 +159,7 @@ void LargeDataTableBenchmarkObject::PopulateInitialTable(uint32_t num_tuples, Ra
   for (uint32_t i = 0; i < num_tuples; i++) {
     auto *const redo =
         initial_txn_->StageWrite(CatalogTestUtil::TEST_DB_OID, CatalogTestUtil::TEST_TABLE_OID, row_initializer_);
-    StorageTestUtil::PopulateRandomRow(redo->Delta(), layout_, 0.0, generator);
+    //    StorageTestUtil::PopulateRandomRow(redo->Delta(), layout_, 0.0, generator);
```
lmwnshn: ?

mbutrovich: It is. I was testing if it affected the benchmark numbers at all in CI. I'm content to remove it.

src/include/main/db_main.h
```diff
@@ -102,37 +58,648 @@ class DBMain {
...
+   */
+  class Builder {
+   public:
+    /**
+     * Validate configuration and construct requested DBMain.
+     * @return DBMain that you get to own. YES, YOU! We're just giving away DBMains over here!
+     */
+    std::unique_ptr<DBMain> Build() {
+      auto db_main = std::make_unique<DBMain>();
+
```
lmwnshn: Do you think it is worth having a high level comment on dependencies here? Or maybe a link to a wiki page that has the graph which you drew on the board.

I _think_ we add new components infrequently enough that this shouldn't fall too far out of date / will remain quite useful overall? Defer to your judgment.

src/loggers/parser_logger.cpp
```diff
@@ -1,14 +1,16 @@
 #include "loggers/parser_logger.h"
+
 #include <memory>
-#include "loggers/main_logger.h"
 
 namespace terrier::parser {
 
-std::shared_ptr<spdlog::logger> parser_logger;  // NOLINT
+std::shared_ptr<spdlog::logger> parser_logger = nullptr;  // NOLINT
```
mbutrovich: I added this initialization and then the nullptr check below to all of the loggers when trying to debug why loggers were getting started and shutdown multiple times in the current system, along with races associated to that. I've cleaned that all up, but should I change the `if` below to just be a `TERRIER_ASSERT` so we don't reintroduce duplicate initialization calls in the future?

src/include/settings/settings_defs.h
```diff
@@ -66,17 +66,6 @@ SETTING_int(
     terrier::settings::Callbacks::NoOp
 )
 
-// Number of worker pool threads
```
lmwnshn: Did this setting go somewhere?

src/main/db_main.cpp
```diff
@@ -1,119 +1,28 @@
...
   running_ = true;
-  server_->SetPort(static_cast<uint16_t>(
-      type::TransientValuePeeker::PeekInteger(param_map_.find(settings::Param::port)->second.value_)));
-  server_->RunServer();
+  network_layer_->GetServer()->RunServer();
 
   {
-    std::unique_lock<std::mutex> lock(server_->RunningMutex());
-    server_->RunningCV().wait(lock, [=] { return !(server_->Running()); });
+    std::unique_lock<std::mutex> lock(network_layer_->GetServer()->RunningMutex());
```
mbutrovich: This could be refactored to not call `GetServer()` 3 times.

src/include/main/db_main.h
mbutrovich: I could document that here, yeah. I considered writing a lot more about how to add a dependency to `DBMain` but I'm hoping in the next 2 weeks everything is in anyway.

src/include/settings/settings_defs.h
mbutrovich: It doesn't really have any meaning until we architect the task/thread infrastructure so I removed it.

test/storage/log_test.cpp
```diff
@@ -198,7 +186,8 @@ TEST_F(WriteAheadLoggingTests, LargeLogTest) {
   while (in.HasMore()) {
     storage::LogRecord *log_record = ReadNextRecord(&in);
     if (log_record->TxnBegin() == transaction::INITIAL_TXN_TIMESTAMP) {
-      // TODO(Tianyu): This is hacky, but it will be a pain to extract the initial transaction. The LargeTransactionTest
+      // TODO(Tianyu): This is hacky, but it will be a pain to extract the initial transaction. The
```
lmwnshn: huh, how did this reformat happen? Nothing important, just looks a bit goofy, but I'm surprised.

src/include/main/db_main.h
```diff
@@ -99,40 +59,642 @@ class DBMain {
...
+            static_cast<uint64_t>(settings_manager->GetInt64(settings::Param::num_log_manager_buffers));
+        log_serialization_interval_ = settings_manager->GetInt(settings::Param::log_serialization_interval);
+        log_persist_interval_ = settings_manager->GetInt(settings::Param::log_persist_interval);
+        log_persist_threshold_ =
+            static_cast<uint64_t>(settings_manager->GetInt64(settings::Param::log_persist_threshold));
+
+        gc_interval_ = settings_manager->GetInt(settings::Param::gc_interval);
+
+        network_port_ = static_cast<uint16_t>(settings_manager->GetInt(settings::Param::port));
+      }
```
malin1993ml: Shall we extract this thing into a separate private function? This function is really long at the moment.

mbutrovich:
Will merge in the AM.

PR #608, opened 2019-10-29
Operations needed for TPCC.

amlatyrngom:
Brings in Insert, Update & Delete from TPL.

PR #600, opened 2019-10-20
Optimizer PR

17zhangw:
Brings over Peloton optimizer

Remaining tasks before merging:
- [x] Remerge binder with InnerJoin changes
- [x] Fix the rule_impls.cpp for PushFilterThroughJoin
- [x] Enable TPCC Join Test
- [x] Enable TPCC Aggregate Test
- [x] GroupBy Offsets change in plan node
- [x] OutputSchema changes
- [x] Rip out HashPlanNode
- [x] Enable TPCC Delete/Update Tests
- [x] Test for OutputSchemas
- [x] Fix DISTINCT operator

PR #650, opened 2019-12-10
Tpl with codegen

amlatyrngom:
This is the codegen side of the exec engine. Here are the things to check for each reviewer:

@17zhangw should review the plannodes, and see if they are used correctly.
@mbutrovich, @lmwnshn and @jrolli check if the catalog and the managed pointers are well used. Also check the API to the rest of the system. 
@pmenon and @apavlo check the overall architecture.

This should merge after #600 and #608.


src/include/execution/compiler/storage/pr_filler.h
```diff
@@ -0,0 +1,106 @@
...
+#include <utility>
+#include "execution/compiler/function_builder.h"
+#include "execution/compiler/operator/operator_translator.h"
+#include "execution/compiler/translator_factory.h"
+
+namespace terrier::execution::compiler {
+/**
+ * PRFiller is used to generate functions that fill index PRs from table PRs
+ */
+class PRFiller : public ExpressionEvaluator {
```
jrolli: If this is an `ExpressionEvaluator`, why is it not in the `expression` sub-directory?

src/include/execution/compiler/translator_factory.h
```diff
@@ -0,0 +1,118 @@
...
+   */
+  static std::unique_ptr<ExpressionTranslator> CreateExpressionTranslator(
+      const terrier::parser::AbstractExpression *expression, CodeGen *codegen);
+
+  /**
+   * Whether this is a comparison operation
+   */
+  static bool IsComparisonOp(terrier::parser::ExpressionType type) {
+    return ((type) <= terrier::parser::ExpressionType::COMPARE_IS_DISTINCT_FROM &&
+            (type) >= terrier::parser::ExpressionType::COMPARE_EQUAL);
```
jrolli: Using range comparisons seems pretty brittle given we may add operators in the future (applies below as well).  It may be more verbose, but I think it is more maintainable to do a long series of equality comparisons (switch statement?) since it does not rely on enum layout.  A good compiler should consolidate the ranges anyway.

src/execution/compiler/operator/seq_scan_translator.cpp
```diff
@@ -0,0 +1,219 @@
...
+      case terrier::type::TypeId::BIGINT:
+        filter_val =
+            codegen_->IntLiteral(static_cast<int32_t>(terrier::type::TransientValuePeeker::PeekBigInt(trans_val)));
+        break;
+      default:
+        UNREACHABLE("Impossible vectorized predicate!");
+    }
+    ast::Expr *filter_call = codegen_->PCIFilter(pci_, predicate->GetExpressionType(), col_idx, col_type, filter_val);
+    builder->Append(codegen_->MakeStmt(filter_call));
+  }
```
jrolli: What's supposed to happen if neither of the if statements match?  Right now we implicitly return, but that seems like a mistake rather than design.

src/execution/compiler/pipeline.cpp
```diff
@@ -0,0 +1,3 @@
+#include "execution/compiler/pipeline.h"
+
+namespace terrier::execution::compiler {}  // namespace terrier::execution::compiler
```
jrolli: Why does this file exist?

src/include/execution/compiler/operator/operator_translator.h
```diff
@@ -0,0 +1,191 @@
...
+      ast::Expr *type = codegen_->TplType(col.GetExpr()->GetReturnValueType());
+      fields->emplace_back(codegen_->MakeField(field_name, type));
+      attr_idx++;
+    }
+  }
+
+  /**
+   * @return the plan node as a generic nodeq
+   */
+  virtual const planner::AbstractPlanNode *Op() = 0;
```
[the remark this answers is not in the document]
amlatyrngom: Some alternatives that came up when I was doing this:
1. Have one abstract `op_` field in the `OperatorTranslator`: this requires static or dynamic casts everywhere.
2. Make this class templatized: unnecessarily complicated.
3. Have two plannode fields: an abstract one that lives in `OperatorTranslator`, and a concrete one that lives in each translator. I find this less elegant (subjectly speaking).

src/include/execution/compiler/operator/output_translator.h
```diff
@@ -0,0 +1,72 @@
...
+  void InitializeHelperFunctions(util::RegionVector<ast::Decl *> *decls) override {}
+
+  // Does nothing
+  void InitializeSetup(util::RegionVector<ast::Stmt *> *setup_stmts) override {}
+
+  // Does nothing
+  void InitializeTeardown(util::RegionVector<ast::Stmt *> *teardown_stmts) override {}
+
+  // Should never called since this the last layer.
+  ast::Expr *GetOutput(uint32_t attr_idx) override { UNREACHABLE("Should not be called on this translator"); }
```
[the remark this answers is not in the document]
amlatyrngom: The query's output will be placed in the `OutputBuffer` of the execution context.
This `GetOutput` is used by parent translators to get their children's outputs. Because this is the final layer, it does not have a parent. So this function should not be called.

src/include/execution/compiler/operator/operator_translator.h
```diff
@@ -0,0 +1,191 @@
...
+   * Add statements to the teardown function
+   * @param teardown_stmts list of statements in the teardown function
+   */
+  virtual void InitializeTeardown(util::RegionVector<ast::Stmt *> *teardown_stmts) = 0;
+
+  /**
+   * Add fields to the state struct
+   * @param state_fields list of fields of the state struct
+   */
+  virtual void InitializeStateFields(util::RegionVector<ast::FieldDecl *> *state_fields) = 0;
```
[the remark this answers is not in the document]
amlatyrngom: Will be addressed in meeting.

src/execution/compiler/operator/aggregate_translator.cpp
```diff
@@ -0,0 +1,397 @@
...
+  }
+
+  // Make the struct
+  decls->emplace_back(codegen_->MakeStruct(payload_struct_, std::move(fields)));
+}
+
+/*
+ * Generate the aggregation's input values
+ */
+void AggregateBottomTranslator::GenValuesStruct(util::RegionVector<ast::Decl *> *decls) {
```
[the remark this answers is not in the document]
jrolli: I'll leave it up to your assessment.  If we're unlikely to ever need to modify this code or add another function that reuses the pattern, then I'm ok with it as is.

src/include/execution/compiler/operator/operator_translator.h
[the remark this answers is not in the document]
jrolli: Just add a TODO in the doxygen that we should keep an eye on the performance of this.

src/execution/compiler/codegen.cpp
```diff
@@ -0,0 +1,568 @@
...
+    case type::TypeId::DECIMAL:
+      builtin = nullable ? ast::Builtin::PRGetDoubleNull : ast::Builtin::PRGetDouble;
+      break;
+    case type::TypeId::DATE:
+      builtin = nullable ? ast::Builtin::PRGetDateNull : ast::Builtin::PRGetDate;
+      break;
+    case type::TypeId::VARCHAR:
+      builtin = nullable ? ast::Builtin::PRGetVarlenNull : ast::Builtin::PRGetVarlen;
+      break;
+    default:
```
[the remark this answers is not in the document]
weihang7: What about using a template engine like jinja?

src/execution/compiler/pipeline.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+  // Function parameter
+  util::RegionVector<ast::FieldDecl *> params = codegen_->ExecParams();
+
+  // Function return type (nil)
+  ast::Expr *ret_type = codegen_->BuiltinType(ast::BuiltinType::Kind::Nil);
+
+  FunctionBuilder builder{codegen_, fn_name, std::move(params), ret_type};
+
+  // for (const auto & translator: pipeline_) {
```
lmwnshn: What is this?

src/include/execution/compiler/operator/nested_loop_translator.h
```diff
@@ -0,0 +1,137 @@
...
+namespace terrier::execution::compiler {
+
+class NestedLoopRightTranslator;
+
+/**
+ * Nested Loop Inner translator.
+ * Because a nested loop join can be expressed without any special operation, this translator
+ * does not do much. It just binds the two sides of the join together through GetOutput.
+ */
+class NestedLoopLeftTransaltor : public OperatorTranslator {
```
lmwnshn: If not too onerous, fix typo? Translator

src/include/execution/compiler/operator/nested_loop_translator.h
```diff
@@ -0,0 +1,137 @@
...
+ private:
+  const planner::NestedLoopJoinPlanNode *op_;
+};
+
+/**
+ * Nested Loop Right Translator.
+ * Just like the left translator, this one does not do much. However, it needs to produces an if statement
+ * for the join predicate.
+ */
+class NestedLoopRightTransaltor : public OperatorTranslator {
```
lmwnshn: Similarly Translator

src/include/planner/plannodes/index_scan_plan_node.h
```diff
@@ -46,42 +67,96 @@ class IndexScanPlanNode : public AbstractScanPlanNode {
...
     /**
-     * Build the Index scan plan node
-     * @return plan node
+     * Sets the index cols.
      */
-    std::unique_ptr<IndexScanPlanNode> Build() {
-      return std::unique_ptr<IndexScanPlanNode>(new IndexScanPlanNode(std::move(children_), std::move(output_schema_),
-                                                                      scan_predicate_, is_for_update_, is_parallel_,
-                                                                      database_oid_, namespace_oid_, index_oid_));
+    Builder &AddIndexColum(catalog::indexkeycol_oid_t col_oid, const IndexExpression &expr) {
```
lmwnshn: AddIndexColumn

src/include/planner/plannodes/index_scan_plan_node.h
```diff
@@ -46,42 +67,96 @@ class IndexScanPlanNode : public AbstractScanPlanNode {
...
+      lo_index_cols_.emplace(col_oid, expr);
+      return *this;
+    }
+
+    /**
+     * Sets the lower bound index cols.
+     */
+    Builder &AddLoIndexColum(catalog::indexkeycol_oid_t col_oid, const IndexExpression &expr) {
```
lmwnshn: AddLoIndexColumn

src/include/planner/plannodes/index_scan_plan_node.h
```diff
@@ -46,42 +67,96 @@ class IndexScanPlanNode : public AbstractScanPlanNode {
...
+      lo_index_cols_.emplace(col_oid, expr);
+      return *this;
+    }
+
+    /**
+     * Sets the index upper bound cols.
+     */
+    Builder &AddHiIndexColum(catalog::indexkeycol_oid_t col_oid, const IndexExpression &expr) {
```
lmwnshn: AddHiIndexColumn

src/execution/compiler/compiler.cpp
```diff
@@ -0,0 +1,179 @@
...
+      curr_pipeline->Add(std::move(left_translator));
+      // Inner loop
+      MakePipelines(*op.GetChild(0), curr_pipeline);
+      curr_pipeline->Add(std::move(right_translator));
+      return;
+    }
+    default: {
+      // Every other operation just adds itself to the current pipeline.
+      auto translator = TranslatorFactory::CreateRegularTranslator(&op, codegen_);
+      if (op.GetChildrenSize() != 0) MakePipelines(*op.GetChild(0), curr_pipeline);
```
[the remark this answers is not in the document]
lmwnshn: Added assertion.

src/execution/compiler/compiler.cpp
```diff
@@ -0,0 +1,179 @@
...
+      curr_pipeline->Add(std::move(right_translator));
+      return;
+    }
+    case terrier::planner::PlanNodeType::NESTLOOP: {
+      // The two sides of the nested loop join belong to the same pipeline. They are just concatenated together.
+      // These two translator glue the two sides together and ensure that expression evaluation is correctly done.
+      auto left_translator = TranslatorFactory::CreateLeftTranslator(&op, codegen_);
+      auto right_translator = TranslatorFactory::CreateRightTranslator(&op, left_translator.get(), codegen_);
+      // Outer loop
+      MakePipelines(*op.GetChild(1), curr_pipeline);
```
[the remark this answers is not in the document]
lmwnshn: Added documentation to the plan node itself.

lmwnshn:
I've been poking at it over the past couple of days. Some remarks:

1. addressed the remaining review comments, mostly just documentation
2. merged in master
3. the optimizer and exec engine differed in IndexScanPlanNode, right now it just has both versions in there; this should be unified. Had brief chat with @17zhangw about it. Main difficulties are (1) optimizer currently doesn't produce the required information of e.g. lower bounds and upper bounds (2) limit information is produced in a separate node and it isn't clear how to get this to the Visit of the IndexScan operator (3) some minor indexkeycol_oid_t vs col_oid_t differences, I have a fix unpushed which involves building and caching the reverse map of indexed_oids_ and using that during the optimizer's plan generation.

I don't know if fixing the last point's (1) and (2) should fall under the scope of 650. Without it, we lose range scans and limits on scans. This can maybe be punted.

PR #656, opened 2019-12-18
parallelize data table benchmarks

harsh141994:
Need to check on Jenkins(when benchmark runs), if it is able to get the environment variable correctly.


apavlo:
I will submit a new PR with this.

PR #658, opened 2019-12-22
pg_language and pg_proc catalog tables

tanujnay112:
Adding pg_language and pg_proc en route adding UDF support in the execution engine.


src/catalog/postgres/builder.cpp
```diff
@@ -294,6 +302,33 @@ Schema Builder::GetTypeTableSchema() {
...
+  columns.back().SetOid(LANPLTRUSTED_COL_OID);
+
+  columns.emplace_back("lanplcallfoid", type::TypeId::INTEGER, false, MakeNull(type::TypeId::INTEGER));
+  columns.back().SetOid(LANPLCALLFOID_COL_OID);
+
+  columns.emplace_back("laninline", type::TypeId::INTEGER, false, MakeNull(type::TypeId::INTEGER));
+  columns.back().SetOid(LANINLINE_COL_OID);
+
+  columns.emplace_back("lanvalidator", type::TypeId::INTEGER, false, MakeNull(type::TypeId::INTEGER));
+  columns.back().SetOid(LANVALIDATOR_COL_OID);
```
jrolli: These columns are not set in `InsertLanguage`.  Additionally, you have marked them as not nullable but are not setting them (@mbutrovich, is there a graceful way to catch this kind of behavior with asserts?).  Looking at Postgres' documentation, I think `lanispl = FALSE` and `lanpltrusted = TRUE` are reasonable initialization values (still leave the default as `NULL`) and the rest should probably be `NULL` and nullable until we go to implement them.

src/include/catalog/database_catalog.h
```diff
@@ -238,6 +238,10 @@ class DatabaseCatalog {
   std::vector<std::pair<common::ManagedPointer<storage::index::Index>, const IndexSchema &>> GetIndexes(
       transaction::TransactionContext *txn, table_oid_t table);
 
+  bool InsertLanguage(transaction::TransactionContext *txn, const std::string &lanname);
+
+  bool DeleteLanguage(transaction::TransactionContext *txn, const std::string &lanname);
```
jrolli: To be consistent with the rest of the catalog, you should only support deletions by OID.

src/include/catalog/database_catalog.h
```diff
@@ -238,6 +238,10 @@ class DatabaseCatalog {
   std::vector<std::pair<common::ManagedPointer<storage::index::Index>, const IndexSchema &>> GetIndexes(
       transaction::TransactionContext *txn, table_oid_t table);
 
+  bool InsertLanguage(transaction::TransactionContext *txn, const std::string &lanname);
```
jrolli: To be consistent with the rest of the catalog, this should return the OID assigned to the language or `INVALID_LANGUAGE_OID` on failure (will need a definition in `catalog_defs.h`).

src/catalog/database_catalog.cpp
```diff
@@ -1610,6 +1642,11 @@ void DatabaseCatalog::BootstrapTypes(transaction::TransactionContext *const txn)
              postgres::Type::BASE);
 }
 
+void DatabaseCatalog::BootstrapLanguages(transaction::TransactionContext *const txn)
+{
+  InsertLanguage(txn, "plpgsql");
```
jrolli: This breaks the current catalog convention by assigning a "userspace" OID value to an internal catalog object.  The OID number should be a constant defined in `pg_language.h` and forced during bootstrap.  It may be useful to pull most of the table logic in `InsertLanguage` into a private helper that takes the OID as a parameter so that you can use it during bootstrap and still provide the OID-free interface for the main API.

src/include/catalog/database_catalog.h
```diff
@@ -344,6 +348,12 @@ class DatabaseCatalog {
   storage::index::Index *constraints_index_index_;
   storage::index::Index *constraints_foreigntable_index_;
 
+  storage::SqlTable *languages_;
+  storage::index::Index *languages_oid_index_;
+  storage::index::Index *languages_name_index_;  // indexed on language OID and name
```
jrolli: Comment is not accurate.  It looks like the name index is just on the name string.

src/include/storage/storage_util.h
```diff
@@ -189,13 +189,52 @@ class StorageUtil {
...
   }
 
+  /**
+   * Helper method to serialize a vector of strings into a VarlenEntry
+   * @param vec input whose elements are to be serialized into a VarlenEntry
+   * @return varlen entry representing this data vector
+   * @warning checking IsInlined() to see if you need to possibly clean up a buffer
+   * @warning the length of the data vector is not serialized
+   */
+  static storage::VarlenEntry CreateVarlen(const std::vector<const std::string> &vec) {
```
apavlo: We probably want to rename this to `CreateArray` since we don't want everyone to use this.
Another issue is that it only supports `std::string`. What if it took in a vector of `TransientValue`?

jrolli: I think we should avoid transient values since that was a major source of complexity in the previous catalog and we should always know a column type in the catalog at compile time.  Maybe we have a group of helpers that are Create[Type]Array?

tanujnay112: @apavlo For now I made a template function for other types below that actually for other fixed size types.

src/catalog/catalog_accessor.cpp
```diff
@@ -132,6 +132,39 @@ common::ManagedPointer<storage::index::Index> CatalogAccessor::GetIndex(index_oi
...
+  return dbc_->GetLanguageOid(txn_, lanname);
+}
+
+bool CatalogAccessor::DropLanguage(language_oid_t language_oid) { return dbc_->DropLanguage(txn_, language_oid); }
+
+proc_oid_t CatalogAccessor::CreateProcedure(const std::string &procname, language_oid_t lanoid, namespace_oid_t procns,
+                                            const std::vector<const std::string> &args,
+                                            const std::vector<type::TypeId> &arg_types,
+                                            const std::vector<type::TypeId> &all_arg_types,
+                                            const std::vector<const char> &arg_modes, type_oid_t rettype,
```
apavlo: What are the acceptable values for `arg_modes`? Should we use an `enum class`?

src/catalog/catalog_accessor.cpp
```diff
@@ -132,6 +132,39 @@ common::ManagedPointer<storage::index::Index> CatalogAccessor::GetIndex(index_oi
...
+  return dbc_->CreateLanguage(txn_, language);
+}
+
+language_oid_t CatalogAccessor::GetLanguageOid(const std::string &lanname) {
+  return dbc_->GetLanguageOid(txn_, lanname);
+}
+
+bool CatalogAccessor::DropLanguage(language_oid_t language_oid) { return dbc_->DropLanguage(txn_, language_oid); }
+
+proc_oid_t CatalogAccessor::CreateProcedure(const std::string &procname, language_oid_t lanoid, namespace_oid_t procns,
```
apavlo: Rename `lanoid` to `language_oid`

src/catalog/catalog_accessor.cpp
```diff
@@ -132,6 +132,39 @@ common::ManagedPointer<storage::index::Index> CatalogAccessor::GetIndex(index_oi
...
+                                            const std::vector<const std::string> &args,
+                                            const std::vector<type::TypeId> &arg_types,
+                                            const std::vector<type::TypeId> &all_arg_types,
+                                            const std::vector<const char> &arg_modes, type_oid_t rettype,
+                                            const std::string &src, bool is_aggregate) {
+  return dbc_->CreateProcedure(txn_, procname, lanoid, procns, args, arg_types, all_arg_types, arg_modes, rettype, src,
+                               is_aggregate);
+}
+
+bool CatalogAccessor::DropProcedure(proc_oid_t proc) { return dbc_->DropProcedure(txn_, proc); }
```
apavlo: Rename `proc` to `proc_oid`

src/catalog/database_catalog.cpp
```diff
@@ -1,18 +1,19 @@
-#include "catalog/database_catalog.h"
-
 #include <memory>
 #include <string>
 #include <utility>
 #include <vector>
 
 #include "catalog/catalog_defs.h"
+#include "catalog/database_catalog.h"
```
apavlo: Shouldn't this be at the top?

mbutrovich: I think this is a `clang-format` version discrepancy. I don't think the version of `clang-format` that we're using in our toolchain enforces that ordering, while the newer one bundled with CLion does which is why you see that ordering in quite a few files. I suspect we'll eventually standardize on it when we bump the toolchain next, but for now it's probably too difficult to enforce outside of manual code review.

src/catalog/postgres/builder.cpp
```diff
@@ -294,6 +308,33 @@ Schema Builder::GetTypeTableSchema() {
...
+  columns.back().SetOid(LANOID_COL_OID);
+
+  columns.emplace_back("lanname", type::TypeId::VARCHAR, MAX_NAME_LENGTH, false, MakeNull(type::TypeId::VARCHAR));
+  columns.back().SetOid(LANNAME_COL_OID);
+
+  columns.emplace_back("lanispl", type::TypeId::BOOLEAN, true, MakeNull(type::TypeId::BOOLEAN));
+  columns.back().SetOid(LANISPL_COL_OID);
+
+  columns.emplace_back("lanpltrusted", type::TypeId::BOOLEAN, true, MakeNull(type::TypeId::BOOLEAN));
+  columns.back().SetOid(LANPLTRUSTED_COL_OID);
```
jrolli: These should probably be not nullable.  The values for the bootstrapped languages should be known at compile time (i.e. false for `ispl` and true for `trusted`).

src/include/catalog/catalog_accessor.h
```diff
@@ -269,6 +269,66 @@ class CatalogAccessor {
    */
   common::ManagedPointer<storage::BlockStore> GetBlockStore() const;
 
+  /**
+   * Adds a language to the catalog (with default parameters for now) if
+   * it doesn't exist in pg_language already
+   * @param lanname name of language
+   * @return oid of added language if it didn't exist before or INVALID_LANGUAGE_OID if else
+   */
+  language_oid_t CreateLanguage(const std::string &lanname);
```
jrolli: This seems like an insufficient prototype.  I would lean toward asking for everything we *will* need and dropping them unimplemented things inside the catalog as this is apart of the API and changing the signature will require refactoring more and more places as this gets leveraged.

src/include/catalog/database_catalog.h
```diff
@@ -246,7 +247,85 @@ class DatabaseCatalog {
...
       common::ManagedPointer<transaction::TransactionContext> txn, table_oid_t table);
 
+  /**
+   * Creates a language entry into the pg_language table
+   * @param txn transaction to use
+   * @param lanname name of language to insert
+   * @return oid of created entry if successful else INVALID_LANGUAGE_OID
+   */
+  language_oid_t CreateLanguage(const common::ManagedPointer<transaction::TransactionContext> txn,
+                                const std::string &lanname);
```
jrolli: Probably an insufficient API prototype (see comment in catalog accessor header).

src/include/catalog/database_catalog.h
```diff
@@ -246,7 +247,85 @@ class DatabaseCatalog {
...
  private:
+  /**
+   * Creates a language entry into the pg_language table
+   * @param txn transaction to use
+   * @param lanname name of language to insert
+   * @param oid oid of entry
+   * @return true if insertion is successful
+   */
+  bool CreateLanguage(common::ManagedPointer<transaction::TransactionContext> txn, const std::string &lanname,
+                      language_oid_t oid);
```
jrolli: Probably insufficient parameters (see above).  However, this might be the right place to hide unimplemented parameters for now.  I'm open to simplicity here for the sake of making bootstrap simpler.  If we do that, add a TODO stating something to that effect.

src/catalog/postgres/builder.cpp
```diff
@@ -548,4 +589,136 @@ IndexSchema Builder::GetConstraintForeignTableIndexSchema(db_oid_t db) {
...
+  columns.emplace_back("proargdefaults", type::TypeId::VARBINARY, 4096, false, MakeNull(type::TypeId::VARCHAR));
+  columns.back().SetOid(PROARGDEFAULTS_COL_OID);
+
+  columns.emplace_back("proargnames", type::TypeId::VARBINARY, 4096, false, MakeNull(type::TypeId::VARBINARY));
+  columns.back().SetOid(PROARGNAMES_COL_OID);
+
+  columns.emplace_back("prosrc", type::TypeId::VARCHAR, 4096, false, MakeNull(type::TypeId::VARCHAR));
+  columns.back().SetOid(PROSRC_COL_OID);
+
+  columns.emplace_back("proconfig", type::TypeId::VARBINARY, 4096, false, MakeNull(type::TypeId::VARBINARY));
```
jrolli: Magic number (4096).  This should probably be a constant defined in `pg_proc.h`.

src/catalog/postgres/builder.cpp
```diff
@@ -548,4 +589,136 @@ IndexSchema Builder::GetConstraintForeignTableIndexSchema(db_oid_t db) {
...
+  columns.emplace_back("proname", type::TypeId::VARCHAR, MAX_NAME_LENGTH, false,
+                       parser::ColumnValueExpression(db, PRO_TABLE_OID, PRONAME_COL_OID));
+  columns.back().SetOid(indexkeycol_oid_t(2));
+
+  columns.emplace_back("proallargs", type::TypeId::VARBINARY, MAX_NAME_LENGTH, false,
+                       parser::ColumnValueExpression(db, PRO_TABLE_OID, PROALLARGTYPES_COL_OID));
+  columns.back().SetOid(indexkeycol_oid_t(3));
+
+  // Unique, not primary
+  IndexSchema schema(columns, storage::index::IndexType::HASHMAP, true, false, false, true);
```
jrolli: This should be the default bwtree index because we may need to do range scans on the index (e.g. all procs in a given namespace).

src/catalog/postgres/builder.cpp
```diff
@@ -548,4 +589,136 @@ IndexSchema Builder::GetConstraintForeignTableIndexSchema(db_oid_t db) {
...
+
+IndexSchema Builder::GetLanguageNameIndexSchema(db_oid_t db) {
+  std::vector<IndexSchema::Column> columns;
+
+  columns.emplace_back("lanname", type::TypeId::VARCHAR, MAX_NAME_LENGTH, false,
+                       parser::ColumnValueExpression(db, LANGUAGE_TABLE_OID, LANNAME_COL_OID));
+  columns.back().SetOid(indexkeycol_oid_t(1));
+
+  // Unique, not primary
+  IndexSchema schema(columns, storage::index::IndexType::HASHMAP, true, false, false, true);
```
jrolli: This should be a bwtree since we may want to do range queries (e.g. select all languages in a given namespace).

src/catalog/database_catalog.cpp
```diff
@@ -1623,8 +1687,18 @@ void DatabaseCatalog::BootstrapTypes(const common::ManagedPointer<transaction::T
...
+  InsertType(txn, postgres::VAR_ARRAY_OID, "var_array", postgres::NAMESPACE_CATALOG_NAMESPACE_OID, -1, false,
+             postgres::Type::COMPOSITE);
 }
 
+void DatabaseCatalog::BootstrapLanguages(const common::ManagedPointer<transaction::TransactionContext> txn) {
+  CreateLanguage(txn, "plpgsql", postgres::PLPGSQL_LANGUAGE_OID);
+  CreateLanguage(txn, "internal", postgres::INTERNAL_LANGUAGE_OID);
+}
+
+void BootstrapProcs(const common::ManagedPointer<transaction::TransactionContext> txn) { (void)txn; }
```
jrolli: Not called.  Should this be removed?

src/include/catalog/postgres/pg_proc.h
```diff
@@ -0,0 +1,63 @@
...
+#include "storage/sql_table.h"
+#include "storage/storage_defs.h"
+#include "transaction/transaction_context.h"
+#include "type/type_id.h"
+
+namespace terrier::catalog::postgres {
+
+constexpr table_oid_t PRO_TABLE_OID = table_oid_t(81);
+constexpr index_oid_t PRO_OID_INDEX_OID = index_oid_t(82);
+constexpr index_oid_t PRO_NAME_INDEX_OID = index_oid_t(83);
```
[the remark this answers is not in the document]
tanujnay112: This is consistent with the pg_proc Postgres documentation I believe.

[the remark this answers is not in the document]
jrolli: For internal C++ code Names such as pointers and table/index OIDs , we’ve taken some liberties from exact PG phrasing for clarity (classes vs rel* etc.).  It’s not a major concern, so I’m fine leaving it as well.

src/catalog/postgres/builder.cpp
tanujnay112: This seems to be an issue for the other builders as well in this file.

src/include/catalog/catalog_accessor.h
tanujnay112: I did this as I don't think we intend on using this beyond just as a reference table for pg_proc anytime soon.

jrolli: If that’s the case I’d leave it out of catalog accessor and leave a todo in database catalog.  That should (hopefully) force someone to think about the end API before using it (because they’ll have to implement it).

PR #677, opened 2020-01-08
Parser returns ParseResult on heap

mbutrovich:
The current parser returns the `ParseResult` on the stack. Due to wanting a bit more control over the object's life cycle, we want this on the heap.

Other refactors: remove `using std::` stuff from TrafficCop layer, and removed a redundant empty test callback.


mbutrovich:
We went from 2.6m NOOP/s to 2.2m NOOP/s. I can live with that.

mbutrovich:
Closing this for now since as I write the traffic cop it's unclear if this is what we want.

PR #680, opened 2020-01-10
Create/drop transformations + PlanGenerator

17zhangw:
This PR adds optimizer support for `CREATE/DROP` statements. Most of the logic is just passing values from logical -> physical ops (via new implementation rules) and then from physical ops -> plan nodes (via PlanGenerator); `CreateTable/CreateIndex` are slightly more involved since key/schema must also be derived.

Remaining work before review:
- [x] `CreateTable` transformation + plan
- [x] `CreateTable` test
- [x] `CreateDatabase` test
- [x] `CreateFunction` test
- [x] `CreateIndex` test
- [x] `CreateNamespace` test
- [x] `CreateTrigger` test
- [x] `CreateView` test
- [x] `DropDatabase` test
- [x] `DropIndex` test
- [x] `DropNamespaceIfExists` test, where it exists
- [x] `DropNamespaceIfExists` test, where it doesn't exist
- [x] `DropTable` test

Parser doesn't support these, neither does catalog, disabled for now
- [x] `DropTriggerIfExists` test, where it doesn't exist, there is no way to make triggers right now
- [x] `DropViewIfExists` test, where it doesn't exist, there is no way to make views right now

`plan_node_json_test` will be broken until we do a refactor of the FromJson for all plan nodes. This should probably go off the critical path of getting TPCC running.

The tests currently push through to plan generation, which is possible for creates and drops since the output columns are known to be empty. However, to invoke the optimizer, you need a QueryInfo object. This currently isn't being generated by any component, as far as I can tell. -- wan

Also, rule.Check doesn't seem to do anything useful right now ("return true"). -- wan


lmwnshn:
Finished with boilerplate tests. A little too much copypaste-type of logic, but it works as far as unit tests go. `plan_node_json_test` is broken because some fields in optimizer classes are ManagedPtr now. I don't think we want to try to fix it with this PR and vote that we disable the broken fields.

I can probably take a pass at actually doing JSON serialization of custom types right..? We're doing everything hackily right now.

mbutrovich:
I managed to execute all of these operations over psql, and also tested with TPC-C table and index creation. Seems to work! Nice! I did manage to break `PostgresParser` when trying to make a table inside of a namespace, but that's not this PR's problem.

PR #682, opened 2020-01-15
OutputSchema and assorted PlanGenerator fixes

17zhangw:
PR fixes the following set of bugs between the optimizer and execution engine:
- [x] OutputSchema of Seq/Index scans are now ColumnValueExpressions
- [x] IndexScanPlanNode's TableOid was not being set
- [x] Seq/Index scans predicate use ColumnValueExpressions

(actual index scans will be fixed in a separate PR)


mbutrovich:
I merged it into my `traffic_cop` branch and could do `SELECT * FROM "NEW_ORDER"` so LGTM!

PR #689, opened 2020-01-18
Analyze

rohanaggarwal7997:
This is starting of the work to build stats collecting and storing functionality to be merged into terrier so that the cost model can be used more effectively using real stats and the optimizer's effectiveness can be tested on Join Orderings. Also this functionality is required to have the optimizer support parameter value expressions such as while optimizing prepare statements before the first execute is called.
This PR merges analyze function, that given a table scans it and returns a vector of column stats collected for that table. A test case is provided.
Further work includes generating table stats, pg stats and storing the stats somewhere permanent and not in the heap.

### DIFF
--- a/src/optimizer/analyze.cpp
+++ b/src/optimizer/analyze.cpp
@@ -78,6 +78,7 @@ terrier::optimizer::ColumnStats CalculateColumnStats(catalog::db_oid_t database_
                                             static_cast<double>(num_null_values) / static_cast<double>(num_rows),
                                             most_common_values, most_common_frequencies, histogram_bounds, true);
 
+  delete[] buffer;
   return col_stats;
 }
 
--- a/src/optimizer/analyze.cpp
+++ b/src/optimizer/analyze.cpp
@@ -1,0 +1,143 @@
+#include <algorithm>
+#include <cstring>
+#include <functional>
+#include <limits>
+#include <memory>
+#include <random>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "optimizer/analyze.h"
+
+namespace terrier {
+
+template <typename T>
+terrier::optimizer::ColumnStats CalculateColumnStats(catalog::db_oid_t database_id, catalog::table_oid_t table_id,
+                                                     catalog::col_oid_t columnID, storage::SqlTable *sql_table,
+                                                     terrier::transaction::TransactionContext *scan_txn,
+                                                     unsigned mostFrequentNumber) {
+  // Stats collecting temporary data-structures
+  std::unordered_map<T, unsigned> counting_map;
+  unsigned num_null_values = 0;
+  unsigned num_rows = 0;
+  unsigned top_k = mostFrequentNumber;
+  std::vector<double> histogram_bounds;
+  histogram_bounds.push_back(DBL_MAX);
+  histogram_bounds.push_back(DBL_MIN);
+
+  std::vector<catalog::col_oid_t> current_col_oid_vector{columnID};
+  storage::ProjectionMap projection_list_indices = sql_table->ProjectionMapForOids(current_col_oid_vector);
+  auto initializer =
+      sql_table->InitializerForProjectedColumns(current_col_oid_vector, common::Constants::K_DEFAULT_VECTOR_SIZE);
+  auto *buffer = common::AllocationUtil::AllocateAligned(initializer.ProjectedColumnsSize());
+  storage::ProjectedColumns *projected_column = initializer.Initialize(buffer);
+
+  // scan the results
+  auto iterator_table = sql_table->begin();
+  while (iterator_table != sql_table->end()) {
+    sql_table->Scan(scan_txn, &iterator_table, projected_column);
+    common::RawBitmap *null_bit_map = projected_column->ColumnNullBitmap(projection_list_indices[columnID]);
+    T *col_pointer = reinterpret_cast<T *>(projected_column->ColumnStart(projection_list_indices[columnID]));
+    for (uint32_t i = 0; i < projected_column->NumTuples(); i++) {
+      if (null_bit_map->Test(i)) {
+        if (counting_map.find(col_pointer[i]) != counting_map.end())
+          counting_map[col_pointer[i]]++;
+        else
+          counting_map[col_pointer[i]] = 1;
+
+        histogram_bounds[0] = std::min(static_cast<double>(col_pointer[i]), histogram_bounds[0]);
+        histogram_bounds[1] = std::max(static_cast<double>(col_pointer[i]), histogram_bounds[1]);
+      } else {
+        num_null_values += 1;
+      }
+      num_rows += 1;
+    }
+  }
+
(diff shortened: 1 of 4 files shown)


src/optimizer/analyze.cpp
```diff
@@ -0,0 +1,144 @@
...
+#include <utility>
+#include <vector>
+
+#include "optimizer/analyze.h"
+
+namespace terrier {
+
+template <typename T>
+terrier::optimizer::ColumnStats CalculateColumnStats(catalog::db_oid_t database_id, catalog::table_oid_t table_id,
+                                                     catalog::col_oid_t columnID, storage::SqlTable *sql_table,
```
apavlo: Rename `columnID` to `column_oid`.

src/optimizer/analyze.cpp
```diff
@@ -0,0 +1,144 @@
...
+                                                     terrier::transaction::TransactionContext *scan_txn,
```
apavlo: The `TransactionContext` should always be the first parameter to a function.

src/optimizer/analyze.cpp
```diff
@@ -0,0 +1,144 @@
...
+    T *col_pointer = reinterpret_cast<T *>(projected_column->ColumnStart(projection_list_indices[columnID]));
+    for (uint32_t i = 0; i < projected_column->NumTuples(); i++) {
+      if (null_bit_map->Test(i)) {
+        if (counting_map.find(col_pointer[i]) != counting_map.end())
+          counting_map[col_pointer[i]]++;
+        else
+          counting_map[col_pointer[i]] = 1;
+
+        histogram_bounds[0] = std::min(static_cast<double>(col_pointer[i]), histogram_bounds[0]);
+        histogram_bounds[1] = std::max(static_cast<double>(col_pointer[i]), histogram_bounds[1]);
```
apavlo: Does this actually work??? What if it is a VARCHAR?

src/optimizer/analyze.cpp
```diff
@@ -0,0 +1,144 @@
...
+  while (iterator_table != sql_table->end()) {
+    sql_table->Scan(scan_txn, &iterator_table, projected_column);
+    common::RawBitmap *null_bit_map = projected_column->ColumnNullBitmap(projection_list_indices[columnID]);
+    T *col_pointer = reinterpret_cast<T *>(projected_column->ColumnStart(projection_list_indices[columnID]));
+    for (uint32_t i = 0; i < projected_column->NumTuples(); i++) {
+      if (null_bit_map->Test(i)) {
+        if (counting_map.find(col_pointer[i]) != counting_map.end())
+          counting_map[col_pointer[i]]++;
+        else
+          counting_map[col_pointer[i]] = 1;
```
apavlo: Are you storing every unique value in this map? What if I have 1 billion unique values?

src/optimizer/analyze.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+namespace terrier {
+
+template <typename T>
+terrier::optimizer::ColumnStats CalculateColumnStats(catalog::db_oid_t database_id, catalog::table_oid_t table_id,
+                                                     catalog::col_oid_t columnID, storage::SqlTable *sql_table,
+                                                     terrier::transaction::TransactionContext *scan_txn,
+                                                     unsigned mostFrequentNumber) {
+  // Stats collecting temporary data-structures
+  std::unordered_map<T, unsigned> counting_map;
```
apavlo: Why are we using this instead of [`terrier::optimizer::Histogram`](https://github.com/cmu-db/terrier/blob/master/src/include/optimizer/statistics/histogram.h)?

src/optimizer/analyze.cpp
```diff
@@ -0,0 +1,144 @@
...
+  std::vector<double> most_common_frequencies;
+
+  unsigned index = 0;
+  while (top_k--) {
+    most_common_values.push_back(intermediate_hist[index].second);
+    most_common_frequencies.push_back(intermediate_hist[index].first);
+    index++;
+  }
+
+  terrier::optimizer::ColumnStats col_stats(database_id, table_id, columnID, num_rows, counting_map.size(),
```
apavlo: Since this object has a lot of input arguments, switch to the builder programming pattern for the constructor.

src/optimizer/analyze.cpp
```diff
@@ -0,0 +1,144 @@
...
+                                            static_cast<double>(num_null_values) / static_cast<double>(num_rows),
+                                            most_common_values, most_common_frequencies, histogram_bounds, true);
+
+  delete[] buffer;
+  return col_stats;
+}
+
+std::vector<terrier::optimizer::ColumnStats> Analyze(catalog::db_oid_t database_id, catalog::table_oid_t table_id,
+                                                     const catalog::Schema &table_schema, storage::SqlTable *sql_table,
+                                                     terrier::transaction::TransactionContext *scan_txn,
```
apavlo: `TransactionContext` should always be the first argument.

test/optimizer/analyze_test.cpp
```diff
@@ -0,0 +1,118 @@
...
+  auto *const scan_txn = txn_manager_->BeginTransaction();
+
+  auto result_analyze =
+      Analyze(CatalogTestUtil::TEST_DB_OID, CatalogTestUtil::TEST_TABLE_OID, table_schema_, sql_table_, scan_txn, 10);
+
+  txn_manager_->Commit(scan_txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+
+  std::vector<double> result_most_common_values{198, 196, 194, 192, 190, 188, 186, 184, 182, 180};
+  std::vector<double> result_most_common_frequencies{99, 98, 97, 96, 95, 94, 93, 92, 91, 90};
+  std::vector<double> result_histogram_bounds{2, 198};
```
apavlo: Instead of hardcoding these, can you automatically derive them when you insert the tuples above? This is going to be a pain to maintain in the future.

test/optimizer/analyze_test.cpp
```diff
@@ -0,0 +1,118 @@
...
+      storage::ProjectedRowInitializer::Create(std::vector<uint16_t>{1}, std::vector<uint16_t>{1});
+
+ protected:
+  void SetUp() override {
+    db_main_ = terrier::DBMain::Builder().SetUseGC(true).SetUseGCThread(true).SetRecordBufferSegmentSize(1e6).Build();
+    txn_manager_ = db_main_->GetTransactionLayer()->GetTransactionManager();
+
+    auto col = catalog::Schema::Column(
+        "attribute", type::TypeId::INTEGER, false,
+        parser::ConstantValueExpression(type::TransientValueFactory::GetNull(type::TypeId::INTEGER)));
```
apavlo: This test is a good first start, but I would add varchars + floats to make sure that we support these types.

apavlo:
Do we still want this PR after #704?

rohanaggarwal7997:
> Do we still want this PR after #704?

I dont think so. But its better to not delete it right way, till we complete the analyze the codegen way. Maybe we run into some errors.

PR #690, opened 2020-01-18
Fix undefined behavior in data table benchmark

rohanaggarwal7997:
Lin and I figured out a small bug in Data Table Benchmark. He asked me to upload the change. 
common::AllocationUtil::AllocateAligned initializes an array while the benchmark does not deallocate it properly.


apavlo:
Fixed in #701

PR #693, opened 2020-01-20
Index Scan Fixes from the Optimizer

17zhangw:
PR fixes optimizer support with regards to `IndexScanPlanNode` for the exec engine.


mbutrovich:
I think we need to discuss with @apavlo the suggested API changes in this PR.

17zhangw:
@mbutrovich Ready for another review

mbutrovich:
So I was concerned about the performance change on `CompactIntsKey` from going from the `memcmp` comparator (which was believed to end up as a single SIMD instruction in O3) to your `PartialLessThan` which is effectively turning it into an interpreted comparator based on the `IndexKeySchema`. The benchmark is [here](https://github.com/mbutrovich/terrier/blob/0cf126188fda77443e07809907abc949c84852ba/benchmark/storage/index_key_benchmark.cpp).

Here are the results on dev4:
```
Running ./index_key_benchmark
Run on (40 X 3100 MHz CPU s)
CPU Caches:
  L1 Data 32K (x20)
  L1 Instruction 32K (x20)
  L2 Unified 256K (x20)
  L3 Unified 25600K (x2)
--------------------------------------------------------------------------------------
Benchmark                                               Time           CPU Iterations
--------------------------------------------------------------------------------------
IndexKeyBenchmark/CompactInts8LessThan                  4 ns          4 ns  185075699   2.08741GB/s
IndexKeyBenchmark/CompactInts16LessThan                 4 ns          4 ns  165934693   3.53248GB/s
IndexKeyBenchmark/CompactInts24LessThan                 4 ns          4 ns  179652072   5.11942GB/s
IndexKeyBenchmark/CompactInts32LessThan                 3 ns          3 ns  220674026   10.0995GB/s
IndexKeyBenchmark/CompactInts8PartialLessThan           2 ns          2 ns  307770721   3.30608GB/s
IndexKeyBenchmark/CompactInts16PartialLessThan          2 ns          2 ns  308812109   6.55158GB/s
IndexKeyBenchmark/CompactInts24PartialLessThan          2 ns          2 ns  308763097    9.8483GB/s
IndexKeyBenchmark/CompactInts32PartialLessThan          2 ns          2 ns  308707042   12.1139GB/s
```
I'm confused by the results and would probably need to see what the original comparator is doing to make sense of it.

PR #702, opened 2020-01-24
Configurable Multi-Threaded Benchmarks

apavlo:
1. New `BenchmarkConfig` class for global parameters.
2. All multi-threaded benchmarks now use `BenchmarkConfig::num_threads` to determine the number of threads to use (instead of hardcoding everything to four).
3. We pass in the # of threads to use as an environment variable with `run_micro_bench.py`.
4. Broke out `SlotIteratorBenchmark` into a separate file from `DataTableBenchmark`.

Props to @harsh141994 for starting this.

https://youtu.be/n-hfAPNKa1s


benchmark/catalog/catalog_benchmark.cpp
```diff
@@ -331,6 +331,11 @@ BENCHMARK_DEFINE_F(CatalogBenchmark, GetIndexObjects)(benchmark::State &state) {
   state.SetItemsProcessed(num_indexes * state.iterations());
 }
 
+// ----------------------------------------------------------------------------
```
mbutrovich: Not a fan of bringing this style of block comments over from Peloton but not a dealbreaker for me. Especially since it looks like they're made for 80-character lines (we use 120).

PR #705, opened 2020-01-25
Reconcile Optimizer and Codegen.

### LINKED ISSUE #719: SELECT With ORDER BY Produces Zero Results
Adding an ORDER BY clause causes a SELECT query to output zero results.

```
terrier=# CREATE TABLE xxx (id INT, val INT);
CREATE TABLE
terrier=# INSERT INTO xxx VALUES (1,2),(2,3),(4,5),(6,7),(8,9);
INSERT 0 5
terrier=# SELECT * FROM xxx;
 id | val 
----+-----
  1 |   2
  2 |   3
  4 |   5
  6 |   7
  8 |   9
(5 rows)

terrier=# SELECT * FROM xxx ORDER BY val;
 id | val 
----+-----
(0 rows)
```

### LINKED ISSUE #720: SELECT With Duplicate Columns Produces Zero Results
Adding the same column multiple times in the projection output for a SELECT list produces zero results.

```
terrier=# CREATE TABLE xxx (id INT, val INT);                  
CREATE TABLE
terrier=# INSERT INTO xxx VALUES (1,2),(2,3),(4,5),(6,7),(8,9);
INSERT 0 5
terrier=# SELECT * FROM xxx;
 id | val 
----+-----
  1 |   2
  2 |   3
  4 |   5
  6 |   7
  8 |   9
(5 rows)
terrier=# SELECT id, id FROM xxx;
 id | id 
----+----
(0 rows)
```

amlatyrngom:
This PR fixes several issues:
* William and I were indexing the aggregate expressions differently.
* `SELECT COUNT(*) FROM table` does not need to read any table column. But the `SqlTable` requires that at least one column is selected. The fix selects an arbitrary column to appease the `SqlTable`.
* Aggregates without Group By were not codegened correctly. The fix is not optimal, as we can technically avoid the `AggregationHashTable` altogether. I am working on that next.
* Inserts with non-inlined strings were using dangling pointers.
* Multi Way joins with long probes were not codegened correctly (a variable was not declared).

FIx #719. Fix #720.


src/include/planner/plannodes/seq_scan_plan_node.h
```diff
@@ -46,6 +46,11 @@ class SeqScanPlanNode : public AbstractScanPlanNode {
      */
     Builder &SetColumnOids(std::vector<catalog::col_oid_t> &&column_oids) {
       column_oids_ = std::move(column_oids);
+      if (column_oids_.empty()) {
+        // TODO(Amadou): In the SqlTable, at least one column must be selected. But Count(*) does not need any column
+        // in particular. Is there a better way than just selecting the first one?
+        column_oids_.emplace_back(1);
```
mbutrovich: Does this work if that's not a valid `col_oid` anymore? I'm imaging a schema change where maybe we've dropped that column.

src/main/terrier.cpp
```diff
@@ -96,6 +96,6 @@ int main(int argc, char *argv[]) {
   db_main->Run();
 
   terrier::LoggersUtil::ShutDown();
-
+  main_stat_reg->Shutdown(true);
```
mbutrovich: Thank you!

mbutrovich:
"Aggregates without Group By were not codegened correctly. The fix is not optimal, as we can technically avoid the AggregationHashTable altogether. I am working on that next."
After we merge this will you open an issue with a description of the intended fix just in case you don't have time to get to this?

src/execution/compiler/expression/star_translator.cpp
```diff
@@ -0,0 +1,15 @@
...
+#include "execution/sql/value.h"
+#include "parser/expression/star_expression.h"
+#include "type/transient_value_peeker.h"
+
+namespace terrier::execution::compiler {
+StarTranslator::StarTranslator(const terrier::parser::AbstractExpression *expression, CodeGen *codegen)
+    : ExpressionTranslator(expression, codegen) {}
+
+ast::Expr *StarTranslator::DeriveExpr(ExpressionEvaluator *evaluator) {
+  // TODO(Amadou): What we return here doesn't really matter. Should StarExpression be ConstantValueExpression?
```
pmenon: Return `1`, which is logically what a `COUNT(*)` returns on each tuple. Change comment to "Result technically isn't used. Can be replaced with a constant value expression.` Or something to that effect.

src/execution/compiler/operator/aggregate_translator.cpp
```diff
@@ -269,6 +269,11 @@ void AggregateBottomTranslator::GenHashCall(FunctionBuilder *builder) {
   for (uint32_t term_idx = 0; term_idx < op_->GetGroupByTerms().size(); term_idx++) {
     hash_args.emplace_back(GetGroupByTerm(agg_values_, term_idx));
   }
+  // TODO(Amadou): In case there is no group by term, we can actually bypass the hash table.
+  // For now, I am passing in a constant hash value.
+  if (hash_args.empty()) {
+    hash_args.emplace_back(codegen_->IntToSql(0));
+  }
```
pmenon: We talked about how to solve this. You'll have to undo this change in the new PR.

src/execution/compiler/operator/aggregate_translator.cpp
```diff
@@ -333,7 +338,7 @@ void AggregateTopTranslator::Abort(FunctionBuilder *builder) {
 }
 
 ast::Expr *AggregateTopTranslator::GetChildOutput(uint32_t child_idx, uint32_t attr_idx, terrier::type::TypeId type) {
-  return bottom_->GetOutput(attr_idx);
+  return bottom_->GetOutput(attr_idx + child_idx * bottom_->num_group_by_terms_);
```
pmenon: Why does this need to change?

src/execution/compiler/operator/hash_join_translator.cpp
```diff
@@ -14,7 +14,8 @@ HashJoinLeftTranslator::HashJoinLeftTranslator(const terrier::planner::HashJoinP
       hash_val_{codegen->NewIdentifier("hash_val")},
       build_struct_{codegen->NewIdentifier("BuildRow")},
       build_row_{codegen->NewIdentifier("build_row")},
-      join_ht_{codegen->NewIdentifier("join_ht")} {}
+      join_ht_{codegen->NewIdentifier("join_ht")},
+      left_semi_join_flag_{codegen->Context()->GetIdentifier("left_semi_join_flag")} {}
```
pmenon: This will be used for more than semi-joins. Suggest renaming to "mark" to support generic mark joins for semi, anti, left outer, etc.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -471,11 +471,8 @@ void BytecodeGenerator::VisitSqlConversionCall(ast::CallExpr *call, ast::Builtin
       auto dest = ExecutionResult()->GetOrCreateDestination(ast::BuiltinType::Get(ctx, ast::BuiltinType::StringVal));
       // Copy data into the execution context's buffer.
       auto input = call->Arguments()[0]->As<ast::LitExpr>()->RawStringVal();
-      auto input_length = input.Length();
-      auto *data = exec_ctx_->GetStringAllocator()->Allocate(input_length);
-      std::memcpy(data, input.Data(), input_length);
       // Assign the pointer to a local variable
-      Emitter()->EmitInitString(Bytecode::InitString, dest, input_length, reinterpret_cast<uintptr_t>(data));
+      Emitter()->EmitInitString(Bytecode::InitString, dest, input.Length(), reinterpret_cast<uintptr_t>(input.Data()));
```
pmenon: Oh boy, this is getting messier.

src/execution/compiler/operator/aggregate_translator.cpp
amlatyrngom: The optimizer passes in a `child_idx` of `1` for aggregate expressions, but the aggregate nodes only have one child. This expression creates an unambiguous attribute index.

PR #723, opened 2020-01-31
CREATE TABLE with a qualified namespace work as expected

### LINKED ISSUE #706: CREATE TABLE with a qualified namespace doesn't work as expected
```
terrier=# create schema foo;
CREATE SCHEMA
terrier=# create table foo.bar (id integer);
CREATE TABLE
terrier=# SELECT * from pg_catalog.pg_class;
```
The results show that the table ends up in the `public` namespace rather than `foo`, so there's a bug somewhere. Suggest stepping through each layer of the simple query execution in the `TrafficCop`. Basically: see if the parser result has the namespace name, then see if the bound parse result has the namespace oid, then the logical op, then the physical op. Somewhere in those translations we're losing it.

harsh141994:
Fix #706.
```
terrier=# create schema foo;
CREATE SCHEMA
terrier=# create table foo.bar (id integer);
CREATE TABLE
terrier=# SELECT * from pg_catalog.pg_class;
```

The results show that the table ends up in the public namespace rather than foo, so there's a bug somewhere. 

This PR: default namespace was being used and thats why the table was always created under the public namespace.

### DIFF
--- a/src/optimizer/query_to_operator_transformer.cpp
+++ b/src/optimizer/query_to_operator_transformer.cpp
@@ -295,7 +295,7 @@ void QueryToOperatorTransformer::Visit(parser::CreateStatement *op, parser::Pars
       break;
     case parser::CreateStatement::CreateType::kTable:
       create_expr = std::make_unique<OperatorExpression>(
-          LogicalCreateTable::Make(accessor_->GetDefaultNamespace(), op->GetTableName(), op->GetColumns(),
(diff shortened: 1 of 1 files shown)


mbutrovich:
So this fixes the first problem (creation), but it looks like binding this table's name after creation fails.
```
terrier=# create schema foo;
CREATE SCHEMA
terrier=# create table foo.bar (id int);
CREATE TABLE
terrier=# select * from foo.bar;
ERROR:  binding failed
```
I'm okay to merge this since it fixes an opened issue. But I'm going to open up an issue for the new problem. I suspect the fix will actually be very similar.

PR #725, opened 2020-01-31
Clang tidy diff

portablesounds:
Swapped out the original run-clang-tidy.py call in the check-clang-tidy macro in CMakeLists.txt with "git diff -U0 HEAD^ | clang-tidy-diff.py -p1", which only checks files changed between this commit and the previous one. Note that with merges, HEAD^ only checks the first commit in the merge. Also added a magic number into a random file, so this build *should* fail on Jenkins with a check-clang-tidy error for a magic number in src/binder/bind_node_visitor.cpp.


portablesounds:
Realized a) magic values are actually excepted from clang-tidy and b) printf doesn't want to compile well.

PR #726, opened 2020-02-01
Clang tidy diff

portablesounds:
Swapped out the original run-clang-tidy.py call in the check-clang-tidy macro in CMakeLists.txt with "git diff -U0 HEAD^ | clang-tidy-diff.py -p1", which only checks files changed between this commit and the previous one. Note that with merges, HEAD^ only checks the first commit in the merge. Also added a misformatted variable to a source file, so Jenkins should fail with a check-clang-tidy error.


CMakeLists.txt
```diff
@@ -310,12 +310,16 @@ add_custom_target(check-format ${BUILD_SUPPORT_DIR}/run_clang_format.py
 # "make check-clang-tidy" target
 ###########################################################
 if (${CLANG_TIDY_FOUND})
-    # runs clang-tidy and exits with a non-zero exit code if any errors are found.
+    set(CLANG_TIDY_THREADS 4)
```
lmwnshn: Is there a reason to set it to 4 instead of 0?
I think 0 will use all cores, [line 201](https://github.com/cmu-db/terrier/blob/8db01ce24341b6f9b593faf3b9d1e6453d8b7ee0/build-support/clang-tidy-diff.py#L201).

CMakeLists.txt
```diff
@@ -310,12 +310,16 @@ add_custom_target(check-format ${BUILD_SUPPORT_DIR}/run_clang_format.py
...
+    # runs clang-tidy-diff and exits with a non-zero exit code if any errors are found.
     # note that clang-tidy automatically looks for a .clang-tidy file in parent directories
     add_custom_target(check-clang-tidy
```
lmwnshn: Maybe have two separate commands, check-clang-tidy and check-clang-tidy-diff?

src/binder/bind_node_visitor.cpp
```diff
@@ -35,6 +35,12 @@ void BindNodeVisitor::BindNameToNode(common::ManagedPointer<parser::SQLStatement
 }
 
 void BindNodeVisitor::Visit(parser::SelectStatement *node, parser::ParseResult *parse_result) {
+  int UPPERCASEVARIABLE = 0;
```
lmwnshn: Please revert changes when you're done testing.

apavlo:
It did not run correctly: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-726/2/pipeline/116/#step-248

portablesounds:
I'm looking into it. Essentially part of the CMake logic isn't being triggered and command lists weren't getting compiled.

portablesounds:
Fixed the build error somehow. Might be me messing something up with Jenkins causing certain things not to be linked. Now testing if clang-tidy-diff differentiates between old files and new files (so commits fixing issues don't trigger clang-tidy-diff twice).

portablesounds:
I figured out why my implementation wasn't triggering! I think it has to do with the fact that, when Jenkins runs git, it's checking my local portablesounds/terrier commit against the main cmu-db/terrier master branch commit which isn't diffing properly. I'm not sure what kind of solution I have for this, beyond avoiding git diff entirely or somehow having a reference commit of the branch on Jenkins. @gonzalezjo This may also be of importance to you and #741 .

portablesounds:
In fact it's not checking against my local branch at all! It's checking the current master branch with the previous master branch! No wonder my stuff hasn't been triggering.

portablesounds:
Closed since #741 is more than efficient enough.

PR #735, opened 2020-02-03
IndexScan Table Alias Fix

17zhangw:
Prior to this commit, performing a query like `SELECT * from foo f where f.id = 1` that utilizes an index scan would trigger an assertion failure in the catalog. This PR replaces the table alias stored within the `IndexScan` with the actual table's oid.

### DIFF
--- a/src/optimizer/rules/implementation_rules.cpp
+++ b/src/optimizer/rules/implementation_rules.cpp
@@ -127,7 +127,7 @@ void LogicalGetToPhysicalIndexScan::Transform(common::ManagedPointer<OperatorExp
         if (IndexUtil::SatisfiesSortWithIndex(accessor, sort_prop, get->GetTableOid(), index)) {
           std::vector<AnnotatedExpression> preds = get->GetPredicates();
           auto op = std::make_unique<OperatorExpression>(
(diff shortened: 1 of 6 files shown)


src/optimizer/rules/implementation_rules.cpp
```diff
@@ -127,7 +127,7 @@ void LogicalGetToPhysicalIndexScan::Transform(common::ManagedPointer<OperatorExp
         if (IndexUtil::SatisfiesSortWithIndex(accessor, sort_prop, get->GetTableOid(), index)) {
           std::vector<AnnotatedExpression> preds = get->GetPredicates();
           auto op = std::make_unique<OperatorExpression>(
-              IndexScan::Make(db_oid, ns_oid, index, std::move(preds), get->GetTableAlias(), is_update,
+              IndexScan::Make(db_oid, ns_oid, get->GetTableOid(), index, std::move(preds), is_update,
```
mbutrovich: In theory we don't need ns_oid anymore if oids are unique to a database, right? I believe we can uniquely identify tables and indexes within a database with just their oid. It's not urgent, and likely just legacy behavior from Peloton when things had to be fully qualified.

PR #736, opened 2020-02-03
Fix bugs found with TATP.

amlatyrngom:
While testing TATP, we found that the system segfaults when the predicate columns are not selected.
For example: `SELECT colA FROM foo WHERE colB = 1` segfaults.
This PR fixes that issue by including the predicate columns in the list of input column oids.
We should be able to benchmark TATP with this branch.


mbutrovich:
Will merge after CI.

Issue #724, opened 2020-01-31
SELECT from table with a qualified namespace doesn't work as expected


mbutrovich:
@harsh141994 fixed #706, which then exposed a new issue:
```
terrier=# create schema foo;
CREATE SCHEMA
terrier=# create table foo.bar (id int);
CREATE TABLE
terrier=# select * from foo.bar;
ERROR:  binding failed
```
I suspect the fix will be similar.

mbutrovich:
DROP is broken too. Basically we should look at all the operator transformers that reference tables, and make sure we're not losing the namespace qualifier.

PR #737, opened 2020-02-03
Fix for Queries that use Namespace + TableName Together

harsh141994:
This pr is in reference to the issue : 
SELECT from the table with a qualified namespace doesn't work as expected #724

If my understanding is correct, the parser parses the namespace correctly so we need to use the namespace instead of the default namespace everywhere in the optimizer. I looked at all the operator transformers that reference tables, and make sure we're not losing the namespace qualifier.

Please refer to 
https://github.com/cmu-db/terrier/blob/master/src/optimizer/query_to_operator_transformer.cpp
specifically to lines 280, 367, 451, 527, 568
as these also use the default namespace parameter. I think we don't reference tables here but still need to verify.


mbutrovich:
We should have tests that fail on current master and are fixed with this branch.

harsh141994:
> We should have tests that fail on current master and are fixed with this branch.

Isnt Junit testing framework being setup? What else should I do in this pr?

apavlo:
@harsh141994 Can you write test cases for this?

PR #738, opened 2020-02-03
TerrierServer throws on socket already in use, DBMain catches and returns

### LINKED ISSUE #732: No error message when server tries to bind to port already being used
We don't throw an error message or stop the server if it fails to listen on its target network port number. It appears that the server started up correctly from the output message even though it did not.

To reproduce, we first will start three instances of the server. All will try to bind on port 15721:
 ```
$ for x in $(seq 1 3); do ./debug/terrier & sleep 2 ; done 
[1] 17615
[2020-02-01 14:44:15.242] [network_logger] [info] Listening on port 15721
[2] 17627
[2020-02-01 14:44:17.188] [network_logger] [info] Listening on port 15721
[3] 17639
[2020-02-01 14:44:19.210] [network_logger] [info] Listening on port 15721
```
From this we can see that there are three PIDs for the terrier server. The output messages all claim that they are listening on port 15721.

But according to the OS, only the first PID is actually listening to that port:
```
$ lsof -i:15721 -t
17615
```

### PR DESCRIPTION
**Pull request by mbutrovich**
System now properly handles failing to bind to a socket already in use:
```
[2020-02-03 15:07:45.951] [network_logger] [error] Failed to bind socket: Address already in use
```
Fix #732.

### DIFF
--- a/src/main/db_main.cpp
+++ b/src/main/db_main.cpp
@@ -10,7 +10,11 @@ namespace terrier {
 void DBMain::Run() {
   TERRIER_ASSERT(network_layer_ != DISABLED, "Trying to run without a NetworkLayer.");
   const auto server = network_layer_->GetServer();
-  server->RunServer();
+  try {
+    server->RunServer();
+  } catch (NetworkProcessException &e) {
+    return;
+  }
   {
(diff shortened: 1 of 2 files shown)


mbutrovich:
Will merge once it gets through CI.

PR #739, opened 2020-02-03
postgresql Extended Query protocol

mbutrovich:
- Support Postgres' [Extended Query protocol](https://www.postgresql.org/docs/10/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY).
- Significant binder refactor:
	- Consolidate binder logic in `BindNodeVisitor::Visit` methods rather than scattered between those and the nodes' `Accept` methods. 
	- `BinderSherpa` is now a `BindNodeVisitor` member rather than a field for all `SqlNodeVisitor::Visit` methods.
	- `BinderSherpa` can now carry parameters around through the binding process.
- `Portal` class matches Postgres' definition of the concept. See docs.
- `Statement` class matches Postgres' definition of the concept. See docs.
- `PostgresPacketUtil` helps with reading the wire protocol into internal types.
- Binary support added to `PostgresPacketWriter`.
	- `OutputWriter` now needs to know the field formats.
- Moved more Postgres-specific state into `PostgresProtocolInterpreter` instead of `ConnectionContext`.
- Moved more Postgres-specific logic up from `TrafficCop` into `PostgresNetworkCommands`.
- `ExecutionContext` doesn't own the parameters.
- Fixed several test cases for the binder and optimizer that weren't valid in Postgres.


mbutrovich:
Status update on this after trying TATP in Extended Query mode: the binder's got a bug with respect to binding parameter offsets. It tries to match the return type of the parameter offset (which is integral) to the expected, which could be whatever the statement was prepared to be. When the parameters are integral it seems to work, but a parameter of VARCHAR seemed to demonstrate the issue quickly. Need to dig into binder code a bit, but probably won't have time until later next week.

src/include/network/postgres/postgres_network_commands.h
```diff
@@ -50,7 +50,6 @@ DEFINE_POSTGRES_COMMAND(ExecuteCommand, true);
 DEFINE_POSTGRES_COMMAND(SyncCommand, true);
 DEFINE_POSTGRES_COMMAND(CloseCommand, true);
 DEFINE_POSTGRES_COMMAND(TerminateCommand, true);
-
-DEFINE_POSTGRES_COMMAND(EmptyCommand, true);
+DEFINE_POSTGRES_COMMAND(EmptyCommand, true);  // (Matt): This seems to be only for testing? Not a big fan of that.
```
tpan496: I believe so. Maybe we can move this to the network tests section?

mbutrovich:
We don't have much in the way of end-to-end benchmarking right now, but I at least confirmed that this is no slower (and possible marginally faster) at TATP with 10 terminals in Release mode.

src/include/network/postgres/portal.h
```diff
@@ -0,0 +1,76 @@
...
+
+namespace terrier::network {
+
+/**
+ * Portal is a postgres concept (see the Extended Query documentation:
+ * https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY)
+ * It encapsulates a reference to its originating statement, the parameters (if any), the output formats, and the
+ * optimized physical plan. It represents a query ready to be executed.
+ */
+class Portal {
```
tpan496: Is this just for Postgres? Maybe we can name it more specifically, something like postgres_portal?

src/include/network/postgres/statement.h
```diff
@@ -0,0 +1,91 @@
...
+
+namespace terrier::network {
+
+/**
+ * Statement is a postgres concept (see the Extended Query documentation:
+ * https://www.postgresql.org/docs/current/protocol-flow.html#PROTOCOL-FLOW-EXT-QUERY)
+ * It encapsulates a parsed statement, the parameter types (if any). It represents a statement ready to be turned into a
+ * Portal with a Bind message.
+ */
+class Statement {
```
tpan496: Maybe we can name it more specifically, like postgres_statement? Even though this is in the networking layer, I think it would be better to differentiate this from general statements.

mbutrovich:
The NULLs PR breaks this in the binder. I won't have time to look at this again for a several days.

mbutrovich:
Blocked by #831 to get parameters through the Binder, along with associated bug fixes.

src/include/network/postgres/postgres_network_commands.h
```diff
@@ -43,14 +43,13 @@ class PostgresNetworkCommand : public NetworkCommand {
 
 // Set all to force flush for now
```
tanujnay112: Remove/edit this comment

src/network/postgres/postgres_network_commands.cpp
```diff
@@ -2,36 +2,80 @@
...
+                          const common::ManagedPointer<trafficcop::TrafficCop> t_cop, const bool explicit_txn_block) {
+  trafficcop::TrafficCopResult result;
 
-  const auto parse_result = t_cop->ParseQuery(query, connection, out);
+  const auto query_type = portal->GetStatement()->GetQueryType();
+  const auto physical_plan = portal->PhysicalPlan();
 
-  if (parse_result == nullptr) {
+  // This logic relies on ordering of values in the enum's definition and is documented there as well.
+  if (query_type <= network::QueryType::QUERY_DELETE) {
```
tanujnay112: Nit: It'd be nice if this was turned into a function or a macro that made it clearer what this comparison represents like make a helper method like IsTransactionStatement(query_type) or something.

src/network/postgres/postgres_network_commands.cpp
```diff
@@ -2,36 +2,80 @@
...
+    // DML query to put through codegen
+    result = t_cop->CodegenAndRunPhysicalPlan(connection_ctx, out, portal);
+  } else if (query_type <= network::QueryType::QUERY_CREATE_VIEW) {
+    if (explicit_txn_block && query_type == network::QueryType::QUERY_CREATE_DB) {
+      out->WriteErrorResponse("ERROR:  CREATE DATABASE cannot run inside a transaction block");
+      connection_ctx->Transaction()->SetMustAbort();
+      return;
+    }
+    result = t_cop->ExecuteCreateStatement(connection_ctx, physical_plan, query_type);
+  } else if (query_type <= network::QueryType::QUERY_DROP_VIEW) {
```
tanujnay112: Same thing here as above comment

mbutrovich:
Squashing bugs with TATP.

mbutrovich:
Update on running oltpbench locally. tpcc, smallbank, ycsb good. smallbank broken and will investigate tomorrow.

mbutrovich:
tatp looks good.

PR #741, opened 2020-02-04
Improve the performance of check-clang-tidy 

gonzalezjo:
### Problem

check-clang-tidy is the slowest target on Jenkins by far. During any given Jenkins run, almost every file will go through the battery of clang-tidy checks, which is a process that can easily take over forty minutes to complete.

### Proposal

Leverage the existing infrastructure that's in place for excluding files from clang-tidy. To do so, use `git ls-files` to create a list of all of the files that _haven't_ changed, and add these files to the blacklist.

Though that essentially summarizes this PR, there have been additional changes (detailed by the strikethroughed text at the bottom) that give substantial performance on smaller changesets. Notably, the process of checking the blacklist has been optimized, and blacklisted files never make it into the work queue. 

### Difference vs. #726

#726 takes the approach of using LLVM's clang-tidy-diff system. While this approach has its advantages--notably, it attempts to filter out unchanged lines in addition to just unchanged files--it also introduces a significant amount of code and redundancy. 

The speed-gains from ignoring unchanged lines are likely to be modest, given that clang-tidy doesn't just look at lines in complete isolation. clang-tidy interprets line-number filters more as advice that it will often (by necessity) need to ignore. 

Nonetheless, it may also be the case that filtering out unchanged lines can lead to an even larger performance uplift. Given the potential for significant gains and the non-negligible overhead added by `make-check-clang-tidy` throughout the development cycle, #726 should still be investigated.

### Results

I removed all NOLINTS from tpl.cpp. I then found examples designed to raise clang-tidy warnings (from the clang-tidy docs) and added some to terrier_server.cpp. 

`make check-clang-tidy` was able to finish executing within 62 seconds on my relatively low-end laptop. **It was able to successfully flag the files**. 

I then reset both terrier.cpp and tpl.cpp, then re-ran `make check-clang-tidy`. **It successfully executed without raising any false-positives**, completing within 0.16 seconds.

### Where to go from here

#726 may be able to bring times down even further.


gonzalezjo:
This is ready for review.

portablesounds:
Nice work Jordi! Would you mind pushing another branch that passes on Jenkins so I can compare it to my working #726 implementation? Thanks.

gonzalezjo:
Whether your build passes Jenkins is essentially up to God. You can still compare via this Jenkins run: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-741/7/pipeline/115

gonzalezjo:
Actually, I think I noticed something that escaped testing. I'll put it through Jenkins again.

portablesounds:
I think that this is implying that your tests aren't currently running. It's at the end of your check clang tidy output.

fatal: ambiguous argument 'origin': unknown revision or path not in the working tree.

Use '--' to separate paths from revisions, like this:

'git <command> [<revision>...] -- [<file>...]'

gonzalezjo:
> I think that this is implying that your tests aren't currently running. It's at the end of your check clang tidy output.
> 
> fatal: ambiguous argument 'origin': unknown revision or path not in the working tree.
> 
> Use '--' to separate paths from revisions, like this:
> 
> 'git [...] -- [...]'

Yes, I've been troubleshooting the issue. I switched to a different approach and it unfortunately has the side effect of breaking on Jenkins' clones.

gonzalezjo:
The updated code seemingly did fine on Jenkins. I've removed all NOLINTs from tpl.cpp. Let's see if we get a failure.

gonzalezjo:
This is now ready for review. It seems to work fine on Jenkins.

PR #742, opened 2020-02-04
DatabaseCatalog registers and unregisters indexes for GC

### LINKED ISSUE #734: BwTree indexes aren't registered with the GarbageCollector
We had a temporary workaround for the TPC-C microbenchmark to prevent every BwTree spinning up its own GC, which is to have a registration mechanism with the GC to run on each registered index from a single thread. It was unclear if this was a short-term hack or something that'll stick around, so the Catalog does not register indexes with the GC. If we want that to happen, we need to inject a GC reference to the Catalog at construction (dependency ordering is fine for this) and then modify `DatabaseCatalog::SetIndexPointer` and `DatabaseCalog::DeleteIndex` to perform the registration and unregistration process respectively. We would also want to register the Catalog's one internal BwTree at bootstrap and deregister at teardown as well

mbutrovich:
This is a workaround for the fact that we need to register BwTrees to have GC invoked on them regularly. We don't want each BwTree with its own GC thread, so we have a mechanism to register and unregister indexes with the system-wide `GarbageCollector`. This PR injects a GC dependency to both the `Catalog` and `DatabaseCatalog` objects to use to register and unregister with `SetIndexPointer` and `DeleteIndex` respectively.

I also refactored a lot more ManagedPointers because I have a sickness.

Fix #734.


mbutrovich:
`terrier` binary exits with code 9 now instead of 0. Trying to track down why.

src/catalog/catalog.cpp
```diff
@@ -91,12 +95,13 @@ void Catalog::TearDown() {
...
                              const bool bootstrap, const catalog::db_oid_t db_oid) {
   // Instantiate the DatabaseCatalog
-  DatabaseCatalog *dbc = postgres::Builder::CreateDatabaseCatalog(catalog_block_store_, db_oid);
+  DatabaseCatalog *dbc = postgres::Builder::CreateDatabaseCatalog(catalog_block_store_, db_oid, garbage_collector_);
   txn->RegisterAbortAction([=](transaction::DeferredActionManager *deferred_action_manager) {
     dbc->TearDown(txn);
     delete dbc;
   });
   bool success = Catalog::CreateDatabaseEntry(common::ManagedPointer(txn), db_oid, name, dbc);
+  // TODO(Matt): should this be if (success && bootstrap), or should we bootstrap regardless of db name uniqueness?
```
lmwnshn: Would this clobber the existing database? Right now it looks like it would probably trip some asserts.

src/catalog/database_catalog.cpp
```diff
@@ -1129,14 +1129,17 @@ bool DatabaseCatalog::DeleteIndex(const common::ManagedPointer<transaction::Tran
...
   // manager. See base function comment.
-  txn->RegisterCommitAction([=](transaction::DeferredActionManager *deferred_action_manager) {
-    deferred_action_manager->RegisterDeferredAction([=]() {
-      deferred_action_manager->RegisterDeferredAction([=]() {
-        delete schema_ptr;
-        delete index_ptr;
+  txn->RegisterCommitAction(
+      [=, garbage_collector{garbage_collector_}](transaction::DeferredActionManager *deferred_action_manager) {
+        if (index_ptr->Type() == storage::index::IndexType::BWTREE)
+          garbage_collector->UnregisterIndexForGC(common::ManagedPointer(index_ptr));
```
lmwnshn: Perhaps braces and/or a newline to make it explicit that the double-deferral doesn't depend on the index type, though I could just be too sleepy to read carefully right now.

src/catalog/database_catalog.cpp
```diff
@@ -1193,15 +1196,20 @@ bool DatabaseCatalog::SetClassPointer(const common::ManagedPointer<transaction::
 }
 
 bool DatabaseCatalog::SetIndexPointer(const common::ManagedPointer<transaction::TransactionContext> txn,
-                                      const index_oid_t index, const storage::index::Index *const index_ptr) {
+                                      const index_oid_t index, storage::index::Index *const index_ptr) {
   TERRIER_ASSERT(write_lock_.load() == txn->FinishTime(),
                  "Setting the object's pointer should only be done after successful DDL change request. i.e. this txn "
                  "should already have the lock.");
+  if (index_ptr->Type() == storage::index::IndexType::BWTREE)
+    garbage_collector_->RegisterIndexForGC(common::ManagedPointer(index_ptr));
```
lmwnshn: Similar braces comment, up to you.

src/catalog/database_catalog.cpp
```diff
@@ -1193,15 +1196,20 @@ bool DatabaseCatalog::SetClassPointer(const common::ManagedPointer<transaction::
...
   // This needs to be deferred because if any items were subsequently inserted into this index, they will have deferred
   // abort actions that will be above this action on the abort stack.  The defer ensures we execute after them.
-  txn->RegisterAbortAction([=](transaction::DeferredActionManager *deferred_action_manager) {
-    deferred_action_manager->RegisterDeferredAction([=]() { delete index_ptr; });
-  });
+  txn->RegisterAbortAction(
+      [=, garbage_collector{garbage_collector_}](transaction::DeferredActionManager *deferred_action_manager) {
+        if (index_ptr->Type() == storage::index::IndexType::BWTREE)
+          garbage_collector->UnregisterIndexForGC(common::ManagedPointer(index_ptr));
```
lmwnshn: Another braces comment.

src/catalog/database_catalog.cpp
```diff
@@ -1372,11 +1380,16 @@ void DatabaseCatalog::TearDown(const common::ManagedPointer<transaction::Transac
...
-                   index_schemas{std::move(index_schemas)}, expressions{std::move(expressions)}]() {
+  auto dbc_nuke = [=, garbage_collector{garbage_collector_}, tables{std::move(tables)}, indexes{std::move(indexes)},
+                   table_schemas{std::move(table_schemas)}, index_schemas{std::move(index_schemas)},
+                   expressions{std::move(expressions)}]() {
     for (auto table : tables) delete table;
 
-    for (auto index : indexes) delete index;
+    for (auto index : indexes) {
+      if (index->Type() == storage::index::IndexType::BWTREE)
+        garbage_collector->UnregisterIndexForGC(common::ManagedPointer(index));
```
lmwnshn: Another braces comment.

src/include/catalog/catalog.h
```diff
@@ -36,11 +44,13 @@ class Catalog {
    * action framework in the destructor.
    * @param txn_manager for spawning read-only transactions in destructors
    * @param block_store to use to back catalog tables
+   * @param garbage_collector injected GC to register and deregister indexes. Temporary?
```
lmwnshn: What do you mean by temporary?

src/include/catalog/postgres/builder.h
```diff
@@ -37,9 +37,11 @@ class Builder {
    * Allocates a new database catalog that roughly conforms to PostgreSQL's catalog layout
    * @param block_store for backing the new catalog tables
    * @param oid of the database which is used for populating the field in redo records
+   * @param garbage_collector injected GC to register and deregister indexes. Temporary?
```
lmwnshn: Unsure what temporary refers to.

src/catalog/catalog.cpp
mbutrovich: We discussed this in the deferred events meeting last week, so I can remove this TODO.

mbutrovich:
Will merge after CI.

Issue #716, opened 2020-01-29
Add Support for Analyze Operators


apavlo:
We need to add support for taking the `ANALYZE` SQL query all the way through our binder + optimizer.

1. Add a new class `LogicalAnalyze` to [logical_operators.h](https://github.com/cmu-db/terrier/blob/master/src/include/optimizer/logical_operators.h). It should contain the same data members as [AnalyzeStatement](https://github.com/cmu-db/terrier/blob/master/src/include/parser/analyze_statement.h) and [AnalyzePlanNode](https://github.com/cmu-db/terrier/blob/master/src/include/planner/plannodes/analyze_plan_node.h).

2. Add a new class `Analyze` to [physical_operators.h](https://github.com/cmu-db/terrier/blob/master/src/include/optimizer/physical_operators.h). Again, it should have the same members as `LogicalAnalyze`.

3. Modify `BindNodeVisitor::Visit` to make sure that it checks all the data members from `AnalyzeStatement` -- Right now it looks like it only checks table names.

4. Modify `QueryToOperatorTransformer::Visit` to support converting `AnalyzeStatement` into `LogicalAnalyze`. The method is currently unimplemented: https://github.com/cmu-db/terrier/blob/master/src/optimizer/query_to_operator_transformer.cpp#L587-L590

5. Modify the query optimizer to transform `LogicalAnalyze` into `Analyze`. I don't know how we do this for CREATE operators, so you will need to check with @esargent28 / @17zhangw.

6. Modify `PlanGenerator` to add a new `Visit` method that transforms `Analyze` into `AnalyzePlanNode`.

malin1993ml:
Maybe we can directly transform the `LogicalAnalyze` to `Scan`+`Aggregate`, or `Scan`+`Aggregate`+`Analyze` (for saving the results)? That way we can probably reuse lots of the cogen logic for aggregation.

gengkev:
I think all of the items in this issue are basically complete thanks to #743, right? (Except for possibly 3: `BindNodeVisitor::Visit` doesn't look like it is checking the columns for validity, which maybe it should be...?)

rohanaggarwal7997:
Yup something like this should be present. I remember having some discussion with matt after which we decided not to change that right now. But I do not remember why. But I think adding a small code like this wont make any harm.
      for (const auto &col : *analyze_columns) {
        if (!BinderContext::ColumnInSchema(table_schema, col)) {
[...]

PR #704, opened 2020-01-24
Analyze exec engine support

rohanaggarwal7997:
This is starting of the work to build stats collecting and storing functionality to be merged into terrier so that the cost model can be used more effectively using real stats and the optimizer's effectiveness can be tested on Join Orderings. Also this functionality is required to have the optimizer support parameter value expressions such as while optimizing prepare statements before the first execute is called.

A decision was taken to merge the analyze function inside the execution engine and to do code generation for it. This PR adds aggregator support for topK elements in aggregators.h and provides 2(3 sub tests each) tests for it.

PR #743, opened 2020-02-05
Analyze Operators + Plan Nodes

rohanaggarwal7997:
This is the continuation of #704 which takes the Analyze operators through the entire pipeline. From parser , binder , optimizer and planner. The final step would be to add code generation and then analyze functionality would be good to go. Two tests are provided.

This is for #716.


17zhangw:
Was any conclusion ever reached about @malin1993ml's comment on #716?

rohanaggarwal7997:
I thought it will be unnecessary code duplication at 3 stages of making 3 different types of nodes under the analyse. It would be convinient calling three functions directly under codegen and return the result.

malin1993ml:
I was thinking of only adding those nodes at the last stage of query
optimization to reuse the aggregation logic during codegen. But I’m also
fine if it’s already convenient to directly call the functions. I don’t
think you’ve touched the codegen part yet, so you can revisit this when you
do that.

src/include/optimizer/physical_operators.h
```diff
@@ -2051,5 +2051,71 @@ class DropView : public OperatorNode<DropView> {
...
+   */
+  const catalog::namespace_oid_t &GetNamespaceOid() const { return namespace_oid_; }
+
+  /**
+   * @return OID of the table
+   */
+  const catalog::table_oid_t &GetTableOid() const { return table_oid_; }
+
+  /**
+   * @return trigger columns
```
apavlo: This is not a trigger.

rohanaggarwal7997: ya sorry for mistake, i copy pasted lot of documentation

src/include/optimizer/logical_operators.h
```diff
@@ -1951,4 +1951,70 @@ class LogicalDropView : public OperatorNode<LogicalDropView> {
...
+   */
+  const catalog::namespace_oid_t &GetNamespaceOid() const { return namespace_oid_; }
+
+  /**
+   * @return OID of the table
+   */
+  const catalog::table_oid_t &GetTableOid() const { return table_oid_; }
+
+  /**
+   * @return trigger columns
```
17zhangw: Should probably not be "trigger" but just "columns".

src/optimizer/logical_operators.cpp
```diff
@@ -1192,6 +1192,40 @@ bool LogicalDropView::operator==(const BaseOperatorNode &r) {
   return node.namespace_oid_ == namespace_oid_;
 }
 
+//===--------------------------------------------------------------------===//
+// LogicalCreateTrigger
```
17zhangw: Change from "LogicalCreateTrigger" to "LogicalAnalyze".

src/optimizer/physical_operators.cpp
```diff
@@ -1220,6 +1220,40 @@ bool DropView::operator==(const BaseOperatorNode &r) {
   return node.namespace_oid_ == namespace_oid_;
 }
 
+//===--------------------------------------------------------------------===//
+// LogicalCreateTrigger
```
17zhangw: Change from "LogicalCreateTrigger" to "Analyze"

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -584,9 +584,19 @@ void QueryToOperatorTransformer::Visit(parser::CopyStatement *op, parser::ParseR
   }
 }
 
-void QueryToOperatorTransformer::Visit(UNUSED_ATTRIBUTE parser::AnalyzeStatement *op,
+void QueryToOperatorTransformer::Visit(parser::AnalyzeStatement *op,
                                        UNUSED_ATTRIBUTE parser::ParseResult *parse_result) {
   OPTIMIZER_LOG_DEBUG("Transforming AnalyzeStatement to operators ...");
+  std::vector<catalog::col_oid_t> columns;
+  auto tb_oid = accessor_->GetTableOid(op->GetAnalyzeTable()->GetTableName());
```
17zhangw: If there is a namespace specified, shouldn't this instead use:
`CatalogAccessor::GetTableOid(namespace_oid_t ns, std::string name)`

rohanaggarwal7997: A similar thing is done in create trigger. Does it make a difference?

rohanaggarwal7997:
Addressed all comments

apavlo:
@rohanaggarwal7997 If you have addressed all the comments, mark them as 'resolved' on Github.

rohanaggarwal7997:
> @rohanaggarwal7997 If you have addressed all the comments, mark them as 'resolved' on Github.

done

src/include/optimizer/logical_operators.h
```diff
@@ -1951,4 +1951,70 @@ class LogicalDropView : public OperatorNode<LogicalDropView> {
...
+ private:
+  /**
+   * OID of the database
+   */
+  catalog::db_oid_t database_oid_;
+
+  /**
+   * OID of the namespace
+   */
+  catalog::namespace_oid_t namespace_oid_;
```
mbutrovich: Tables should be unique within a database so I don't think we need namespace oid here.

src/include/optimizer/physical_operators.h
```diff
@@ -2051,5 +2051,71 @@ class DropView : public OperatorNode<DropView> {
...
+ private:
+  /**
+   * OID of the database
+   */
+  catalog::db_oid_t database_oid_;
+
+  /**
+   * OID of the namespace
+   */
+  catalog::namespace_oid_t namespace_oid_;
```
mbutrovich: See previous namespace oid comment.

src/optimizer/query_to_operator_transformer.cpp
rohanaggarwal7997: Now there are no more namespace id

src/include/optimizer/physical_operators.h
rohanaggarwal7997: Addressed

src/include/optimizer/logical_operators.h
rohanaggarwal7997: Addressed

rohanaggarwal7997:
I dont know why but there were lot of changes in the optimizer for some reason of renamings. I had to rename almost all lines that I added. It works now. I also removed all namespace related things. Both the tests still pass. I think its good to merge.

apavlo:
@rohanaggarwal7997 Fix the linter error.

rohanaggarwal7997:
> @rohanaggarwal7997 Fix the linter error.

Its actually not a linter error. I did not knew that we have to check warnings.txt after running doxygen.

PR #746, opened 2020-02-05
Substitute GC with Deferred action framework

mush-zhang:
This PR modified the existing deferred action framework (DAF) and garbage collector, handing the functionalities of garbage collection to DAF. Now DAF takes care of memory deallocation, unlinking version chains, transaction context deletion, and index GC invocation. 

Changes:
- DAF: Process deferred actions and index GC. Use concurrent queue for the centralized deferred action queue to alleviate contention.
- GC: remove all functionalities of GC. The `PerformGarbageCollection()` remains as an interface to invoke the actual deferred action processing and memory management in DAF.
- DataTable: handles truncate version chain
- TxnContext: handles unlinking
- TxnManager: Upon commit/abort, instead of holding latch and moving commit/abort actions to a centralized queue commit/abort actions are invoked with DAF as parameter. Also register deferred action to clean up the transaction context upon commit/abort. As current DAF does not have any information on contents of the actions processed, we keep two atomic counters in transaction manager to fake the number of unlinked txns and number of deleted txns processed in each DAF invocation.
- Various signature changes.

Note:
- Currently faking the API (return type) for GC so that other parts of the system do not need to be changed too much. GC can be removed completely in the future.


src/include/storage/data_table.h
```diff
@@ -217,9 +217,18 @@ class DataTable {
    */
   const BlockLayout &GetBlockLayout() const { return accessor_.GetBlockLayout(); }
 
+  /**
+   * @return reference to this DataTable's accessor
+   */
+  const TupleAccessStrategy &GetAccessor() const { return accessor_; }
```
jrolli: I think this should probably be `private` since it easily allows leaking implementation details.  @mbutrovich, do you have any thoughts on this?

src/include/storage/storage_defs.h
```diff
@@ -257,7 +257,7 @@ class VarlenEntry {
    * @param content pointer to the varlen content
    * @param size length of the varlen content, in bytes (no C-style nul-terminator. Must be smaller than
    *             InlineThreshold())
-   * @return constructed VarlenEntry object
+   * @return constructed VarlenEntry objecprefix_t
```
jrolli: Did this change get mangled?

src/storage/data_table.cpp
```diff
@@ -354,9 +354,49 @@ void DataTable::AtomicallyWriteVersionPtr(const TupleSlot slot, const TupleAcces
...
+
+  // a version chain is guaranteed to not change when not at the head (assuming single-threaded GC), so we are safe
+  // to traverse and update pointers without CAS
+  UndoRecord *curr = version_ptr;
+  UndoRecord *next;
+  // Traverse until we find the earliest UndoRecord that can be unlinked.
+  while (true) {
+    next = curr->Next();
+    // This is a legitimate case where we truncated the version chain but had to restart because the previous head
+    // was aborted.
```
jrolli: I'm not sure this comment is still accurate.  I think the case is still needed, but that it protects against a race in the eventual multi-threaded GC.  @mbutrovich, does that sound right given the Hyper-race fix?

src/transaction/deferred_action_manager.cpp
```diff
@@ -0,0 +1,35 @@
+#include "transaction/deferred_action_manager.h"
+
+#include "common/macros.h"
+
+namespace terrier::transaction {
+
+// TODO(John:GC) Where should this live?
```
jrolli: Ideally, this remains solely in the existing garbage collector files with the goal of ultimately removing this logic by handling it in the index wrappers through deferrals.

src/transaction/deferred_action_manager.cpp
```diff
@@ -0,0 +1,35 @@
...
+void DeferredActionManager::RegisterIndexForGC(const common::ManagedPointer<storage::index::Index> index) {
+  TERRIER_ASSERT(index != nullptr, "Index cannot be nullptr.");
+  common::SharedLatch::ScopedExclusiveLatch guard(&indexes_latch_);
+  TERRIER_ASSERT(indexes_.count(index) == 0, "Trying to register an index that has already been registered.");
+  indexes_.insert(index);
+}
+
+// TODO(DeferredActionManager:GC) Where should this live?
```
jrolli: See above.

src/transaction/deferred_action_manager.cpp
```diff
@@ -0,0 +1,35 @@
...
+  indexes_.erase(index);
+}
+
+// TODO(John:GC) What is the right way to invoke this?  Should it be a periodic task in the deferred action queue or
+// special cased (which feels dirty)?  Frequency of call should not affect index performance just memory usage.  In
+// fact, keeping transaction processing regulare should be more important for throughput than calling this.  Also, this
+// potentially introduces a long pause in normal processing (could be bad) and could block (or be blocked by) DDL.
+//
+// @bug This should be exclusive because concurrent calls to PerformGarbageCollection for the same index may not be
+// thread safe (i.e. bwtrees...)
```
jrolli: I think this should be turned into a TODO to rip out this special casing and implement it in the index wrapper for the bwtree that defers an invocation every `x` number of write operations.  @apavlo, is this (part of) a possible class project?

src/include/transaction/deferred_action_manager.h
```diff
@@ -62,17 +63,21 @@ class DeferredActionManager {
...
     uint32_t processed = ClearBacklog(oldest_txn);
     // There is no point in draining new actions if we haven't cleared the backlog.
     // This leaves some mechanisms for the rest of the system to detect congestion
     // at the deferred action manager and potentially backoff
-    if (backlog_size != processed) return processed;
-    // Otherwise, ingest all the new actions
-    processed += ProcessNewActions(oldest_txn);
+    if (backlog_size == processed) {
+      // ingest all the new actions
+      processed += ProcessNewActions(oldest_txn);
```
jrolli: We should leave a TODO that this is now more conservative than it needs to be and can artificially delay garbage collection.  We should be able to query the cached oldest transaction (should be cheap) in between each event and more aggressively clear the backlog.

mush-zhang: I can try benchmarking that this weekend also.

src/include/transaction/deferred_action_manager.h
```diff
@@ -62,17 +63,24 @@ class DeferredActionManager {
    * Clear the queue and apply as many actions as possible
    * @return numbers of deferred actions processed
    */
-  uint32_t Process(transaction::timestamp_t oldest_txn) {
+  uint32_t Process() {
+    // TODO(John, Ling): this is now more conservative than it needs and can artificially delay garbage collection.
+    //  We should be able to query the cached oldest transaction (should be cheap) in between each event
+    //  and more aggressively clear the backlog abd the deferred event queue
+    timestamp_manager_->CheckOutTimestamp();
```
tli2: It should be worth a comment why we need to checkout a timestamp here. At least this is not immediately obvious to me.

src/transaction/transaction_manager.cpp
```diff
@@ -110,12 +111,12 @@ timestamp_t TransactionManager::Commit(TransactionContext *const txn, transactio
   LogCommit(txn, result, callback, callback_arg, oldest_active_txn);
 
   // We hand off txn to GC, however, it won't be GC'd until the LogManager marks it as serialized
-  if (gc_enabled_) {
-    common::SpinLatch::ScopedSpinLatch guard(&timestamp_manager_->curr_running_txns_latch_);
+  if (gc_enabled_ && deferred_action_manager_ != DISABLED) {
```
tli2: Eventually we might want to clean up these flags once the refactoring is done.

src/include/transaction/deferred_action_manager.h
```diff
@@ -25,7 +27,6 @@ class DeferredActionManager {
       : timestamp_manager_(timestamp_manager) {}
```
yash620, line 27: I would add a comment in the function description that says that if timestamp_manager is null then that means gc is off.

benchmark/benchmark_util/data_table_benchmark_util.cpp
```diff
@@ -78,8 +78,10 @@ LargeDataTableBenchmarkObject::LargeDataTableBenchmarkObject(const std::vector<u
       generator_(generator),
       layout_({attr_sizes}),
       table_(common::ManagedPointer(block_store), layout_, storage::layout_version_t(0)),
-      txn_manager_(common::ManagedPointer(&timestamp_manager_), DISABLED, common::ManagedPointer(buffer_pool), gc_on,
-                   common::ManagedPointer(log_manager)),
+      deferred_action_manager_(gc_on ? common::ManagedPointer(&timestamp_manager_) : nullptr),
```
yash620, line 81: instead of `nullptr`, can you use `DISABLED`?

src/catalog/database_catalog.cpp
```diff
@@ -760,7 +760,7 @@ bool DatabaseCatalog::DeleteTable(const common::ManagedPointer<transaction::Tran
                                   const table_oid_t table) {
   if (!TryLock(txn)) return false;
   // We should respect foreign key relations and attempt to delete the table's columns first
-  auto result = DeleteColumns<Schema::Column, table_oid_t>(txn, table);
+  auto result UNUSED_ATTRIBUTE = DeleteColumns<Schema::Column, table_oid_t>(txn, table);
```
yash620, line 763: why do you need `UNUSED_ATTRIBUTE` here? Isn't `result` being used in the next line?

src/include/storage/garbage_collector.h
```diff
@@ -24,32 +24,19 @@ class GarbageCollector {
...
-   * @param timestamp_manager source of timestamps in the system
    * @param deferred_action_manager pointer to deferred action manager of the system
    * @param txn_manager pointer to the TransactionManager
-   * @param observer the access observer attached to this GC. The GC reports every record gc-ed to the observer if
-   *                 it is not null. The observer can then gain insight invoke other components to perform actions.
-   *                 The observer's function implementation needs to be lightweight because it is called on the GC
-   *                 thread.
    */
   // TODO(Tianyu): Eventually the GC will be re-written to be purely on the deferred action manager. which will
   //  eliminate this perceived redundancy of taking in a transaction manager.
```
yash620, line 31: Can we get rid of this TODO now?

src/include/storage/garbage_collector.h
```diff
@@ -75,44 +62,10 @@ class GarbageCollector {
   void UnregisterIndexForGC(common::ManagedPointer<index::Index> index);
 
  private:
-  /**
```
yash620, line 78: I can't leave a comment higher, but why do we still need `PerformGarbageCollection()`? Wouldn't all of that be handled by the actions in the deferred actions framework?

src/include/transaction/deferred_action_manager.h
```diff
@@ -62,17 +64,28 @@ class DeferredActionManager {
    * Clear the queue and apply as many actions as possible
    * @return numbers of deferred actions processed
    */
-  uint32_t Process(transaction::timestamp_t oldest_txn) {
+  uint32_t Process() {
```
yash620, line 67: Ideally most of these functions should be in `deferred_action_manager.cpp`. Currently the .cpp doesn't really have much in it. Might as well get rid of it and move those functions here too if it's too much of a hassle to move all the implementations to the `.cpp`

src/include/transaction/deferred_action_manager.h
yash620: I think we still need this comment

src/include/transaction/deferred_action_manager.h
```diff
@@ -91,20 +104,51 @@ class DeferredActionManager {
...
   const common::ManagedPointer<TimestampManager> timestamp_manager_;
   // TODO(Tianyu): We might want to change this data structure to be more specialized than std::queue
-  std::queue<std::pair<timestamp_t, DeferredAction>> new_deferred_actions_, back_log_;
-  common::SpinLatch deferred_actions_latch_;
+  tbb::concurrent_queue<std::pair<timestamp_t, DeferredAction>> new_deferred_actions_;
+  std::queue<std::pair<timestamp_t, DeferredAction>> back_log_;
+  // It is sufficient to truncate each version chain once in a GC invocation because we only read the maximal safe
+  // timestamp once, and the version chain is sorted by timestamp. Here we keep a set of slots to truncate to avoid
+  // wasteful traversals of the version chain.
+  std::unordered_set<storage::TupleSlot> visited_slots_;
```
yash620, line 134: This seems like something that should belong in `garbage_collector.h`. I feel kind of iffy with having special cased gc logic in the deferred action manager. Ideally the deferred action manager is dumb and only processes the actions registered, so it wouldn't know if those actions are truncating an already visited_slot.

src/include/storage/garbage_collector.h
yash620: Nvm, I see that it just calls deferred action manager `Process` and there's a todo to replace `garbage_collector_thread`

src/transaction/transaction_context.cpp
```diff
@@ -0,0 +1,18 @@
+#include "transaction/transaction_context.h"
+
+namespace terrier::transaction {
+void TransactionContext::Unlink(timestamp_t oldest_txn, std::unordered_set<storage::TupleSlot> *const visited_slots) {
+  for (auto &undo_record : undo_buffer_) {
+    // It is possible for the table field to be null, for aborted transaction's last conflicting record
+    storage::DataTable *&table = undo_record.Table();
+    if (table != nullptr && visited_slots->insert(undo_record.Slot()).second)
+      table->TruncateVersionChain(undo_record.Slot(), oldest_txn);
```
yash620, line 9: Shouldn't this be a deferred action to unlink and truncate version chain?

src/transaction/transaction_manager.cpp
```diff
@@ -187,12 +204,9 @@ timestamp_t TransactionManager::Abort(TransactionContext *const txn) {
...
   // We hand off txn to GC, however, it won't be GC'd until the LogManager marks it as serialized
-  if (gc_enabled_) {
-    common::SpinLatch::ScopedSpinLatch guard(&timestamp_manager_->curr_running_txns_latch_);
-    // It is not necessary to have to GC process read-only transactions, but it's probably faster to call free off
-    // the critical path there anyway
-    // Also note here that GC will figure out what varlen entries to GC, as opposed to in the abort case.
-    completed_txns_.push_front(txn);
+  // TODO(Ling): eventually we will removed the gc_enabled flag after completely integrate the deferred action framework
+  if (gc_enabled_ && deferred_action_manager_ != DISABLED) {
+    CleanTransaction(txn);
```
yash620, line 209: Doesn't the unlink have to be deferred?

src/include/storage/garbage_collector.h
mush-zhang: let's leave it here for now that as we still keep fake gc there to keep the return format consistent.

src/include/transaction/deferred_action_manager.h
mush-zhang: Eventually the gc will be removed though. Where do you suggest to place visited_slots_?

src/transaction/transaction_manager.cpp
mush-zhang: Deferring unlinking is in CleanTransaction() function.

src/include/transaction/deferred_action_manager.h
yash620: I'm not sure if the gc should be removed. I think the garbage collector thread should be removed. But the `garbage_collector.cpp` should still have gc functions that execute gc specific logic. These functions would be called in deferred events but the logic of these functions shouldn't be a member or visible to deferred_action_manager. In the past that used to be `truncate_version_chain` and such but in this pr that's been moved to be a member of `DataTable` which is fine and make sense to me. So `garbage_collector.cpp` will probably end up not having much but I feel like gc specific logic should remain there.

src/include/storage/data_table.h
```diff
@@ -223,8 +223,12 @@ class DataTable {
  private:
   // The ArrowSerializer needs access to its blocks.
   friend class ArrowSerializer;
-  // The GarbageCollector needs to modify VersionPtrs when pruning version chains
+  // The TransactionContext needs to modify VersionPtrs when pruning version chains
+  friend class transaction::TransactionContext;
+  // TODO(Ling): remove this after deferred action is installed
```
mbutrovich, line 228: Is this TODO still relevant?

src/transaction/timestamp_manager.cpp
```diff
@@ -5,25 +5,34 @@
 namespace terrier::transaction {
 
 timestamp_t TimestampManager::OldestTransactionStartTime() {
-  common::SpinLatch::ScopedSpinLatch guard(&curr_running_txns_latch_);
-  const auto &oldest_txn = std::min_element(curr_running_txns_.cbegin(), curr_running_txns_.cend());
-  timestamp_t result = (oldest_txn != curr_running_txns_.end()) ? *oldest_txn : time_.load();
+  std::vector<timestamp_t> mins;
+  for (size_t i = 0; i < HASH_VAL; i++) {
```
mbutrovich, line 9: What is `HASH_VAL`?

src/include/transaction/timestamp_manager.h
```diff
@@ -62,23 +65,36 @@ class TimestampManager {
...
-      // There is a three-way race that needs to be prevented.  Specifically, we
-      // cannot allow both a transaction to commit and the GC to poll for the
-      // oldest running transaction in between this transaction acquiring its
-      // begin timestamp and getting inserted into the current running
-      // transactions list.  Using the current running transactions latch
-      // prevents the GC from polling and stops the race.  This allows us to
-      // replace acquiring a shared instance of the commit latch with a
-      // read-only spin-latch and move the allocation out of a critical section.
       start_time = time_++;
+      const auto idx = uint64_t(start_time) % HASH_VAL;
```
mbutrovich, line 80: How does this pass clang-tidy with a C-style cast?

src/include/transaction/timestamp_manager.h
```diff
@@ -95,6 +111,11 @@ class TimestampManager {
    */
   void RemoveTransactions(const std::vector<timestamp_t> &timestamps);
 
+  struct SpacedOutSpinLatch {
```
mbutrovich, line 114: This feels like it should be done with `alignas` instead.

src/include/transaction/timestamp_manager.h
```diff
@@ -95,6 +111,11 @@ class TimestampManager {
...
+    common::SpinLatch latch_;
+    byte padding_[common::Constants::CACHELINE_SIZE];
```
mbutrovich, line 116: Won't this cause memory fragmentation since each object is now larger than a cache line?

src/include/transaction/transaction_defs.h
```diff
@@ -7,6 +7,8 @@
 
 #include "common/strong_typedef.h"
 namespace terrier::transaction {
+static constexpr uint64_t HASH_VAL = 13;
```
mbutrovich, line 10: This is a magic number without a comment and a meaningless name.

src/include/storage/data_table.h
mush-zhang: Yes, but it means "remove this after gc is removed in the future". I will fix the todo comment

PR #749, opened 2020-02-05
Unique indexes for CREATE TABLE

### LINKED ISSUE #747: CREATE TABLE does not create non-pkey indexes
If I call `CREATE TABLE` and specify that a column is unique, then the system does not create an index for that column:

```sql
CREATE TABLE subscriber (
   s_id INTEGER NOT NULL PRIMARY KEY,
   sub_nbr VARCHAR(15) NOT NULL UNIQUE
);
```

The pkey index is in the catalog but not the unique index on sub_nbr:

```
terrier=# SELECT * FROM pg_catalog.pg_class;
 reloid |             relname              | relnamespace | relkind |     schema      |     pointer     | nextcoloid 
--------+----------------------------------+--------------+---------+-----------------+-----------------+------------
<SNIP>
   1002 | subscriber                       |           15 | 114     | 139898761851024 | 139898761861712 |         35
   1003 | subscriber_pk_s_id               |           15 | 105     | 139898761811440 | 139898761863776 |           
```

mbutrovich:
Fix #747.
**Changes**:
- Modify the DDL executor for `CreateTablePlanNode` to process the `UniqueInfo` constraints that get passed through the pipeline if a table's column is defined as `UNIQUE`.

**Limitations**:
- This supports the `UNIQUE` modifier on column definitions, i.e. `CREATE TABLE FOO (a INT UNIQUE, b INT UNIQUE, c INT UNIQUE)` will now generate 3 indexes. However, the general modifier to create composite keys doesn't work. For example, `CREATE TABLE FOO (a INT, b INT, c INT, UNIQUE (b,c))` explodes in `postgresparser.cpp`. It looks like `CreateStatement` doesn't even have anywhere to stash that metadata right now, and I'm unsure about the rest of the pipeline. The eventual result, `CreateTablePlanNode` seems like it could support it, but we need all the plumbing in between. These indexes can still be added with `CREATE INDEX` after table creation.
- This doesn't populate `pg_constraint`. The table currently exists in the `DatabaseCatalog`, but I don't think we have the `CatalogAccessor` and `DatabaseCatalog` logic to modify it yet.


mbutrovich:
Hold off on merging. Need to see if it makes sense to try to support composite keys in CREATE TABLE or just stick to UNIQUE columns right now. The latter works, but the former explodes pretty earlier in the pipeline (postgresparser.cpp).

PR #757, opened 2020-02-07
Operator expression/node class name changes

esargent28:
Change: New naming convention for the classes in the optimizer:
- OperatorExpression --> OperatorNode
- OperatorNode --> OperatorNodeContents

Once the rewriter is brought over, this will be expanded to the expression-based classes that the rewriter uses as well (ExpressionNode, ExpressionNodeContents), as well as the abstract classes that both of them inherit from (AbstractOptimizerNode, AbstractOptimizerNodeContents)


mbutrovich:
Looks like we just need a Doxygen cleanup and then I can merge it.

Issue #729, opened 2020-02-01
Inserting into table with out-of-order qualified columns is ignored


mbutrovich:
```
terrier=# CREATE TABLE tbl (c1 int not null primary key, c2 int, c3 int);
CREATE TABLE
terrier=# insert into tbl (c3, c1, c2) VALUES (3, 1, 2);
INSERT 0 1
terrier=# SELECT * FROM tbl;
 c1 | c2 | c3 
----+----+----
  3 |  1 |  2
(1 row)
terrier=# 
```

apavlo:
@17zhangw suspects that the problem is in the codegen layer.

apavlo:
Fixed in #758

PR #758, opened 2020-02-07
Support default value expressions, out-of-order inserts

### LINKED ISSUE #718: Binder Does Not Substitute DefaultValueExpression
If you create a table with a column that has a default value, then the binder does **not** convert the `DefaultValueExpression` object into the correct expression from the catalog. We are currently crashing in `TranslatorFactory::CreateExpressionTranslator` because it is getting a `DefaultValueExpression` that it shouldn't have to handle.

**Client**
```
terrier=# CREATE TABLE xxx (id INT, val INT DEFAULT 123);
CREATE TABLE
terrier=# INSERT INTO xxx VALUES (1, DEFAULT);
```

**Server**
```
Unsupported expression
UNREACHABLE executed at /home/pavlo/Documents/Peloton/Github/terrier/src/execution/compiler/translator_factory.cpp:145!
```

harsh141994:
Fix #718, #729.

Wan: This PR adds support for default value expressions and out of order inserts. Merged in master. We might want a JUnit test, otherwise LGTM. Will officially review once it passes CI.

---
Solution : In the bind_node_visitor, we are not checking if the value is default. Also, the order of insertion into columns is not correct too.

Solved it along with the other reordering issues 
terrier=# CREATE TABLE xxx (id INT, val INT DEFAULT 123);
CREATE TABLE
terrier=# INSERT INTO xxx VALUES (1, DEFAULT);
INSERT 0 1
terrier=# select * from xxx;
 id | val 
----+-----
  1 | 123
(1 row)

terrier=# CREATE TABLE x1 (c1 INT, c2 INT, c3 INT);
CREATE TABLE
terrier=# INSERT INTO x1  VALUES (1, 1, 1);
INSERT 0 1
terrier=# select * from x1
terrier-# ;
 c1 | c2 | c3 
----+----+----
  1 |  1 |  1
(1 row)

terrier=# INSERT INTO x1 (c2, c1, c3) VALUES (2, 3, 4);
INSERT 0 1
terrier=# select * from x1
;
 c1 | c2 | c3 
----+----+----
  1 |  1 |  1
  3 |  2 |  4
(2 rows)

terrier=# INSERT INTO x1 (c2, c1) VALUES (4, 5);
INSERT 0 1
terrier=# select * from x1
;
 c1 | c2 | c3 
----+----+----
  1 |  1 |  1
  3 |  2 |  4
  5 |  4 |   
(3 rows)

However, there are more issues to it when inserting partially,
terrier=# CREATE TABLE x2 (c1 INT, c2 INT DEFAULT 12, c3 INT DEFAULT 34);
CREATE TABLE
terrier=# INSERT INTO x2 (c2, c3) VALUES (DEFAULT, 7);
INSERT 0 1
terrier=# select * from x2;
 c1 | c2 | c3 
----+----+----
 12 |  7 |   
(1 row)

terrier=# INSERT INTO x2 (c1, c3) VALUES (1, DEFAULT);
INSERT 0 1
terrier=# select * from x2;
 c1 | c2 | c3 
----+----+----
 12 |  7 |   
  1 | 34 |   
(2 rows)
Schema order is assumed in the execution unit.


lmwnshn:
Let me know when it is ready for review and change the label accordingly, thanks. @harsh141994

mbutrovich:
Please don't tag PRs as `ready-for-review` until they can pass CI.

src/binder/bind_node_visitor.cpp
```diff
@@ -1,5 +1,6 @@
 #include "binder/bind_node_visitor.h"
 
+#include <term.h>
```
mbutrovich: Why?

src/binder/bind_node_visitor.cpp
```diff
@@ -354,25 +356,60 @@ void BindNodeVisitor::Visit(parser::InsertStatement *node, parser::ParseResult *
...
+
+        std::vector<std::pair<catalog::Schema::Column, common::ManagedPointer<parser::AbstractExpression>>> cols;
+        if (num_insert_columns == 0) {  //already ordered
+          for (int i = 0; i < num_values; i++) {
+            std::pair<catalog::Schema::Column, common::ManagedPointer<parser::AbstractExpression>> pair =
+                std::make_pair(table_schema.GetColumns()[i], values[i]);
+            cols.push_back(pair);
+          }
+        } else {
+          for (int i = 0; i < num_insert_columns; i++) {
```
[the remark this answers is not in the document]
lmwnshn: Let's do it as we discussed after meeting.

1. Go through the schema columns.
2. For each column, if the name of the column is in the InsertColumns, then insert the corresponding value.
3. If not, if the column has default value, insert that.
4. If not, if the column is nullable, insert the corresponding null value.
5. If not, throw a binder exception.

You also need to check if they do not give you insert columns.
If not, maybe you can copy the column names out in schema order from the schema.
It will be slower but the code will be cleaner and we can fix the slowness if we see it being a bottleneck.

mbutrovich:
Adding blocked tag since this needs #768.

script/testing/junit/InsertTest.java
```diff
@@ -210,4 +210,88 @@ public void testCastofIntegerString() throws SQLException {
...
+        assertNoMoreRows(rs);
+
+        String drop_SQL = "DROP TABLE xxx";
+        stmt = conn.createStatement();
+        stmt.execute(drop_SQL);
+    }
+
+    /**
+     * Support out of order inserts
+     * #718 fixed
```
lmwnshn: This should be #729.

PR #759, opened 2020-02-08
Refactor Optimizer Cost Model to Return PlanNode Cardinalities

xinzhu-cai:
Add cardinality estimates for plan nodes for self-driving models. 

- [x] Create a new strong typedef plan_node_id_t in optimizer_defs.h

- [x] Modify BaseOperatorNode to add a new member plan_node_id_.

- [x] Modify AbstractPlanNode to also add plan_node_id_

- [x] Create the CostModelEstimate in abstract_cost_model.h

- [x] Modify the optimizer to use a simple counter to generate new plan_node_id_t for each BaseOperatorNode that it generates.

- [x] Modify PlanGenerator to support passing the BaseOperatorNode id to their corresponding AbstractPlanNode.


apavlo:

PR #760, opened 2020-02-08
Race Condition Fix in ConnectionHandlerTask

tanujnay112:
Aims to close issue #413 where more connections than CONNECTION_THREAD_COUNT would overwrite an existing ConnectionHandlerTasks client_fd_ and protocol_handler_ fields before it is able to create a valid ConnectionHandle and start it. The fix was turning these fields into queues that are protected by a SpinLatch. The contention is expected to be low on this.

### DIFF
--- a/src/include/network/connection_handler_task.h
+++ b/src/include/network/connection_handler_task.h
@@ -5,7 +5,9 @@
 #include <event2/event.h>
 #include <event2/listener.h>
 #include <unistd.h>
+#include <deque>
 #include <memory>
+#include <utility>
 
 #include "common/exception.h"
 #include "common/notifiable_task.h"
@@ -51,19 +53,22 @@ class ConnectionHandlerTask : public common::NotifiableTask {
    * This method will create the necessary data structure for the client and
    * register its event base to receive updates with appropriate callbacks
    * when the client writes to the socket.
-   *[
+   *
    * @param new_conn_recv_fd the socket fd of the new connection
    * @param flags unused. For compliance with libevent callback interface.
    */
   void HandleDispatch(int new_conn_recv_fd, int16_t flags);
 
  private:
-  // using this instead of the Common::ConcurrentQueue as the overhead is not worth
-  // for the common case where there is no contention
+  /**
+   * Using this latch+deque instead of the Common::ConcurrentQueue as the overhead is not worth
+   * for the common case where there is no contention
+   */
   common::SpinLatch jobs_latch_;
+  /**
+   * each pair is represents <connection fd, ProtocolInterpreter>
+   */
   std::deque<std::pair<int, std::unique_ptr<ProtocolInterpreter>>> jobs_;
-  //  client_fd_;
-  //  std::unique_ptr<ProtocolInterpreter> protocol_interpreter_;
   event *notify_event_;
   common::ManagedPointer<ConnectionHandleFactory> connection_handle_factory_;
 };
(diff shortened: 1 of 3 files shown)


src/include/network/connection_handler_task.h
```diff
@@ -51,16 +51,19 @@ class ConnectionHandlerTask : public common::NotifiableTask {
    * This method will create the necessary data structure for the client and
    * register its event base to receive updates with appropriate callbacks
    * when the client writes to the socket.
-   *
+   *[
```
apavlo: ???

src/include/network/connection_handler_task.h
```diff
@@ -51,16 +51,19 @@ class ConnectionHandlerTask : public common::NotifiableTask {
...
-  int client_fd_;
-  std::unique_ptr<ProtocolInterpreter> protocol_interpreter_;
+  // using this instead of the Common::ConcurrentQueue as the overhead is not worth
+  // for the common case where there is no contention
+  common::SpinLatch jobs_latch_;
+  std::deque<std::pair<int, std::unique_ptr<ProtocolInterpreter>>> jobs_;
+  //  client_fd_;
+  //  std::unique_ptr<ProtocolInterpreter> protocol_interpreter_;
   event *notify_event_;
   common::ManagedPointer<ConnectionHandleFactory> connection_handle_factory_;
```
apavlo: Fix comments to follow doxygen format

test/network/network_test.cpp
```diff
@@ -245,6 +246,27 @@ TEST_F(NetworkTests, LargePacketsTest) {
   }
 }
 
+// NOLINTNEXTLINE
+TEST_F(NetworkTests, RacerTest) {
```
apavlo: Add a comment that explains what this test does and what bug is it addressing.

src/network/connection_handler_task.cpp
```diff
@@ -14,17 +14,22 @@ ConnectionHandlerTask::ConnectionHandlerTask(const int task_id,
 }
 
 void ConnectionHandlerTask::Notify(int conn_fd, std::unique_ptr<ProtocolInterpreter> protocol_interpreter) {
-  client_fd_ = conn_fd;
-  protocol_interpreter_ = std::move(protocol_interpreter);
+  {
```
apavlo: Add a comment that explains why we need the latch in case some braze soul tries to remove it in the future.

src/include/network/connection_handler_task.h
```diff
@@ -51,16 +51,19 @@ class ConnectionHandlerTask : public common::NotifiableTask {
...
   void HandleDispatch(int new_conn_recv_fd, int16_t flags);
 
  private:
-  // TODO(Tianyu): This is broken and needs to be fixed. See #413
-  int client_fd_;
-  std::unique_ptr<ProtocolInterpreter> protocol_interpreter_;
+  // using this instead of the Common::ConcurrentQueue as the overhead is not worth
+  // for the common case where there is no contention
+  common::SpinLatch jobs_latch_;
+  std::deque<std::pair<int, std::unique_ptr<ProtocolInterpreter>>> jobs_;
```
apavlo: What does the `int` in the pair represent? Should we strong typedef it?

tanujnay112: It represents a connection file descriptor to interact with a client on. We haven't typedeffed this anywhere else but I'm leaving a comment there on that field to make this clear.

PR #762, opened 2020-02-09
Fix Bugs found with TPCC

amlatyrngom:
This PR fixes these bugs:
* Cast expressions in updates (Reviewer: @lmwnshn)
* Offsets in Limit and OrderBy plan nodes (Reviewer: @17zhangw )
* Distinct and Static Aggregates (Reviewer: @pmenon).

For TPCC, I think the only bug left is the insert order one. This PR can run all queries except for New Order (it has out of order inserts).


mbutrovich:
Adding myself as reviewer just so I can push the merge conflict fix.

mbutrovich:
I fixed the merge conflict, but when running `check-tpl` locally I'm seeing static-agg.tpl seemingly run forever.

mbutrovich:
Scratch that, static-agg.tpl took forever but it finished.

src/execution/compiler/operator/limit_translator.cpp
```diff
@@ -0,0 +1,28 @@
...
+
+namespace terrier::execution::compiler {
+void LimitTranslator::Produce(FunctionBuilder *builder) {
+  builder->Append(codegen_->DeclareVariable(num_tuples_, nullptr, codegen_->IntLiteral(0)));
+  child_translator_->Produce(builder);
+}
+
+void LimitTranslator::Consume(FunctionBuilder *builder) {
+  // Check for offset and limit.
+  // if (num_tuples_ > offset && num_tuples_ < limit + offset)
```
pmenon: Fix comment to match logic.

src/execution/compiler/operator/sort_translator.cpp
```diff
@@ -10,6 +10,7 @@ namespace terrier::execution::compiler {
 SortBottomTranslator::SortBottomTranslator(const terrier::planner::OrderByPlanNode *op, CodeGen *codegen)
     : OperatorTranslator(codegen),
       op_(op),
+      use_top_k_(op->HasLimit() && op->GetOffset() == 0),
```
pmenon: 1. You don't need this assumption. Just issue a top-K sort on `op->GetOffset() + op->GetLimit()` if there's any limit.
2. In general, I'm against storing redundant data. It adds complexity and is a source of inconsistency. You just need `op->HasLimit()` so you can remove this member variable.

src/execution/compiler/operator/sort_translator.cpp
```diff
@@ -200,14 +225,36 @@ void SortTopTranslator::DeclareIterator(FunctionBuilder *builder) {
 }
 
 void SortTopTranslator::GenForLoop(FunctionBuilder *builder) {
-  // for (; @sorterIterHasNext(&sort_iter); @sorterIterNext(&sort_iter))
+  if (use_limit_) {
```
pmenon: You don't need a limit variable. The `Sorter` will handle this for you if you've performed a top-k sort. You just need to check `op->GetOffset()` to ensure sending tuples after `offset` tuples:

```
for (; @sorterIterHasNext(); @sorterIterNext()) {
  if (offset < num_tuples) {
    // Consume
  }
  num_tuples = num_tuples + 1
}
```

src/execution/sema/sema_builtin.cpp
```diff
@@ -1338,8 +1338,43 @@ void Sema::CheckBuiltinSorterInsert(ast::CallExpr *call) {
     return;
   }
 
-  // This call returns nothing
-  call->SetType(GetBuiltinType(ast::BuiltinType::Uint8)->PointerTo());
+  switch (builtin) {
```
pmenon: Please take this code from the TPL repo.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -1258,6 +1258,19 @@ void BytecodeGenerator::VisitBuiltinSorterCall(ast::CallExpr *call, ast::Builtin
       Emitter()->Emit(Bytecode::SorterAllocTuple, dest, sorter);
       break;
     }
+    case ast::Builtin::SorterInsertTopK: {
```
pmenon: Ditto.

src/include/execution/compiler/operator/index_scan_translator.h
```diff
@@ -46,7 +46,9 @@ class IndexScanTranslator : public OperatorTranslator {
   }
 
   // Return the projected row and its type
-  std::pair<ast::Identifier *, ast::Identifier *> GetMaterializedTuple() override { return {&table_pr_, &pr_type_}; }
+  std::pair<const ast::Identifier *, const ast::Identifier *> GetMaterializedTuple() override {
```
pmenon: `Identifiers` are already const, adding that there isn't necessary.

src/include/execution/compiler/operator/static_aggregate_translator.h
```diff
@@ -0,0 +1,123 @@
+#pragma once
+
+#include <unordered_map>
```
pmenon: Needed?

src/include/execution/compiler/operator/static_aggregate_translator.h
```diff
@@ -0,0 +1,123 @@
...
+  void InitializeStructs(util::RegionVector<ast::Decl *> *decls) override {}
+
+  // Does nothing
+  void InitializeHelperFunctions(util::RegionVector<ast::Decl *> *decls) override {}
+
+  // Does nothing
+  void InitializeSetup(util::RegionVector<ast::Stmt *> *setup_stmts) override {}
+
+  // Does nothing
+  void InitializeTeardown(util::RegionVector<ast::Stmt *> *teardown_stmts) override {}
```
pmenon: We don't have them now, but you should be destroying the aggregates here ...

src/execution/compiler/operator/sort_translator.cpp
```diff
@@ -200,14 +223,24 @@ void SortTopTranslator::DeclareIterator(FunctionBuilder *builder) {
 }
 
 void SortTopTranslator::GenForLoop(FunctionBuilder *builder) {
+  if (op_->HasLimit() && op_->GetOffset() != 0) {
```
pmenon: You still need a variable if there is **no** limit, but there **is** an offset.

src/include/execution/compiler/operator/limit_translator.h
```diff
@@ -0,0 +1,76 @@
...
+#include <vector>
+#include "execution/compiler/operator/operator_translator.h"
+#include "execution/compiler/translator_factory.h"
+#include "planner/plannodes/limit_plan_node.h"
+
+namespace terrier::execution::compiler {
+
+/**
+ * Limit Translator
+ * TODO(Amadou): The limit should be pushed down to the child operators. That's only being done with OrderBy.
```
pmenon: What's the purpose of the TODO? Limits/offsets cannot always be pushed down, so we'll always need this implementation.

test/execution/compiler_test.cpp
```diff
@@ -1,3 +1,4 @@
+#include <planner/plannodes/limit_plan_node.h>
```
pmenon: Fix include.

src/include/execution/compiler/operator/aggregate_util.h
```diff
@@ -0,0 +1,221 @@
...
+   */
+  uint32_t TermIdx() const {
+    TERRIER_ASSERT(is_distinct_, "Only Valid for distinct aggregates");
+    return term_idx_;
+  }
+
+  /**
+   * @return The hash table identifier
+   */
+  const ast::Identifier &HT() const { return ht_; }
```
pmenon: None of these need to be const references, `ast::Identifiers` are const by design. Copy by value.

src/include/execution/compiler/operator/limit_translator.h
lmwnshn: Removed.

src/include/execution/compiler/operator/aggregate_util.h
lmwnshn: By value.

test/execution/compiler_test.cpp
lmwnshn: Fixed.

src/execution/compiler/operator/sort_translator.cpp
lmwnshn: We have removed the HasLimit() check.

@17zhangw The op_'s GetOffset() assumes that for the offset to be valid, the limit exists. From the optimizer's perspective, is it safe to remove this check / will you always provide it?

17zhangw: This issue can probably be punted to a later time? The optimizer (as I am aware) does not push the limit/offset information down to the `OrderByPlanNode`.

mbutrovich: Is there a `TERRIER_ASSERT` to be made that'll document our assumptions right now that will yell at us to modify this code if the behavior changes in the future?

17zhangw: I think it would make sense to keep the `TERRIER_ASSERT` in the op_'s `GetOffset()` function for now? That way when we start pushing the limit/offset downwards, the assert will at least be tripped if the limit is missing.

malin1993ml:
It looks like this PR was always failing on `SelectTest::testSelectAlias` in our junit test, which seems to be a true positive? I don't see this failure showing up in other PRs, and the error still exists after merged into [master](http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/master/473/pipeline). @amlatyrngom @mbutrovich

PR #765, opened 2020-02-09
Improved distribution checking in packages.sh

apavlo:
Somebody in the class tried to install the system on Amazon's distro of Linux. Right now packages.sh only checks whether you are running Linux and not which distribution. This PR adds for checking that you are trying to install on Ubuntu. I tested on CentOS and the output format of the error report is correct.

This also adds support for Ubuntu 19.04/19.10, which we may not want to do right now.

https://youtu.be/an9Qqf9dxG4


mbutrovich:
Will merge once it gets through CI.

PR #768, opened 2020-02-11
Add support for NULL.

lmwnshn:
**PR Overview.**
1. Now supports inserting and selecting for NULLs.
2. ~~One TPL builtin per datatype, NullInt(), NullDate(), etc. This simplifies figuring out the type in sema. The proposed alternative was `int x; Null(&x)` or similar, but that seems less ergonomic for handwritten TPL and have a more complicated implementation.~~ We now have NullToSql(&expr) where expr is of the desired null type to instantiate. It was pointed out that we do not expect to handwrite TPL much and this would lead to less bytecodes.
2. Enable OPERATOR_NOT translation. It was already there, just not switched on.
3. Fixed PostgresPacketWriter not advancing to the next column when it saw a null value.
4. Fixed a couple of low-hanging fruit in the binder, ~~e.g. out of order inserts should be fixed, might already be fixed by other people.~~ Interestingly I thought I fixed this but apparently not. Will be fixed by #758.
5. Fix IGNORE_BUILTIN_TYPE macro.

**Details.**
```
terrier=# create table foo (a int, b varchar);
CREATE TABLE
terrier=# insert into foo values (1,'S1');
INSERT 0 1
terrier=# insert into foo values (2,'S2');
INSERT 0 1
terrier=# insert into foo values (3,null);
INSERT 0 1
terrier=# insert into foo values (null,'N4');
INSERT 0 1
terrier=# insert into foo values (null,null);
INSERT 0 1
terrier=# insert into foo values (6,'S6');
INSERT 0 1
terrier=# select * from foo;
 a | b
---+----
 1 | S1
 2 | S2
 3 |
   | N4
   |
 6 | S6
(6 rows)

terrier=# select * from foo where a is null;
 a | b
---+----
   | N4
   |
(2 rows)

terrier=# select * from foo where a is null and b is not null;;
 a | b
---+----
   | N4
(1 row)

terrier=# select * from foo where b is not null;;
 a | b
---+----
 1 | S1
 2 | S2
   | N4
 6 | S6
(4 rows)
```

The main difficulty is that NULL comes in with no type from the Postgres parser (T_Null). So we need to replace the NULL(Invalid Type) with NULL(Correct Type) in the binder. In Peloton this was fixed by having everything come in as NULL(Integer).


src/execution/vm/bytecode_generator.cpp
```diff
@@ -2273,7 +2331,7 @@ void BytecodeGenerator::VisitFile(ast::File *node) {
 }
 
 void BytecodeGenerator::VisitLitExpr(ast::LitExpr *node) {
-  TERRIER_ASSERT(ExecutionResult()->IsRValue(), "Literal expressions cannot be R-Values!");
+  TERRIER_ASSERT(ExecutionResult()->IsRValue(), "Literal expressions should be R-Values!");
```
lmwnshn: Checked with @amlatyrngom, this was a typo.

src/include/execution/compiler/translator_factory.h
```diff
@@ -77,7 +77,7 @@ class TranslatorFactory {
    */
   static bool IsUnaryOp(parser::ExpressionType type) {
     // TODO(Amadou): Support other unary operators.
-    return type == parser::ExpressionType::OPERATOR_UNARY_MINUS;
+    return type == parser::ExpressionType::OPERATOR_UNARY_MINUS || type == parser::ExpressionType::OPERATOR_NOT;
```
lmwnshn: Support was already implemented, just had to turn this on.

src/include/parser/expression/comparison_expression.h
```diff
@@ -48,7 +53,28 @@ class ComparisonExpression : public AbstractExpression {
     return expr;
   }
 
-  void Accept(SqlNodeVisitor *v, ParseResult *parse_result) override { v->Visit(this, parse_result); }
+  void Accept(SqlNodeVisitor *v, ParseResult *parse_result) override {
```
lmwnshn: I'm not happy with this hack, but this is both the fastest and cleanest way that I can see us reusing the binder infrastructure right now. A similar thing might need to be done for arithmetic expressions.

We'd ideally want to do this at the ConstantValueExpression containing the NULL itself, but by that point you've lost the context of "what type of null should this be?".

src/execution/vm/bytecode_generator.cpp
```diff
@@ -442,6 +442,55 @@ void BytecodeGenerator::VisitReturnStmt(ast::ReturnStmt *node) {
 void BytecodeGenerator::VisitSqlConversionCall(ast::CallExpr *call, ast::Builtin builtin) {
   ast::Context *ctx = call->GetType()->GetContext();
   switch (builtin) {
+    case ast::Builtin::IsNull: {
```
pmenon: These aren't really conversion calls. Can you pull it into a separate method?

src/execution/vm/bytecode_generator.cpp
```diff
@@ -442,6 +442,55 @@ void BytecodeGenerator::VisitReturnStmt(ast::ReturnStmt *node) {
...
+      break;
+    }
+    case ast::Builtin::IsNotNull: {
+      auto dest = ExecutionResult()->GetOrCreateDestination(ast::BuiltinType::Get(ctx, ast::BuiltinType::Bool));
+      auto input = VisitExpressionForRValue(call->Arguments()[0]);
+      Emitter()->Emit(Bytecode::IsNotNull, dest, input);
+      ExecutionResult()->SetDestination(dest.ValueOf());
+      break;
+    }
+    case ast::Builtin::NullBool: {
```
pmenon: The return types should already be set at this point, so you should just need: ` ExecutionResult()->GetOrCreateDestination(call->GetType()))`. 

Once you do that, these cases are effectively the same and can be macro-d out to make it clearer.

src/include/execution/ast/builtins.h
```diff
@@ -7,6 +7,16 @@ namespace terrier::execution::ast {
 // The list of all builtin functions
 // Args: internal name, function name
 #define BUILTINS_LIST(F)                                                \
+  /* SQL NULL. */                                                       \
+  F(IsNull, isNull)                                                     \
```
pmenon: Rename `isNull` -> `isSqlNull` to distinguish between pointer nulls and SQL nulls.

src/include/execution/ast/builtins.h
```diff
@@ -7,6 +7,16 @@ namespace terrier::execution::ast {
...
+  F(IsNotNull, isNotNull)                                               \
+  F(NullBool, nullBool)                                                 \
+  F(NullInt, nullInt)                                                   \
+  F(NullReal, nullReal)                                                 \
+  F(NullDecimal, nullDecimal)                                           \
+  F(NullString, nullString)                                             \
+  F(NullDate, nullDate)                                                 \
+  F(NullTimestamp, nullTimestamp)                                       \
```
pmenon: Can you modify these value creation builtins to indicate that they're SQL values with NULL? Similar to the `intToSql` conversion calls below.

src/include/execution/compiler/codegen.h
```diff
@@ -272,6 +272,52 @@ class CodeGen {
    */
   ast::Expr *MemberExpr(ast::Identifier lhs, ast::Identifier rhs);
 
+  /**
+   * @return The generated sql NULL bool.
+   */
+  ast::Expr *NullBool();
```
pmenon: Ditto here, modify names to include `Sql` to indicate SQL values.

src/include/execution/vm/bytecodes.h
```diff
@@ -172,6 +172,15 @@ namespace terrier::execution::vm {
                                                                                                                       \
   /* SQL type comparisons */                                                                                          \
   F(ForceBoolTruth, OperandType::Local, OperandType::Local)                                                           \
+  F(IsNull, OperandType::Local, OperandType::Local)                                                                   \
+  F(IsNotNull, OperandType::Local, OperandType::Local)                                                                \
```
pmenon: You can remove `IsNull` and `IsNotNull` and use `ValIsNull` and `ValIsNotNull` - they're meant to serve the same purpose. Tweak the latter two bytecodes to use primitive booleans.

src/execution/compiler/codegen.cpp
```diff
@@ -116,33 +122,43 @@ ast::Expr *CodeGen::TableIterInit(ast::Identifier tvi, uint32_t table_oid, ast::
 
 ast::Expr *CodeGen::PCIGet(ast::Identifier pci, type::TypeId type, bool nullable, uint32_t idx) {
   ast::Builtin builtin;
+  ast::Type *ast_type;
   switch (type) {
     case type::TypeId::BOOLEAN:
       builtin = nullable ? ast::Builtin::PCIGetBoolNull : ast::Builtin::PCIGetBool;
+      ast_type = ast::BuiltinType::Get(Context(), ast::BuiltinType::Bool);
```
pmenon: 1. This should be `ast::BuiltinType::Boolean`. Yes, its confusing.
2. While the change is necessary, this should be done in all the functions in `codegen.h`. I'd hold off on it for now, and include this in codegen V2 when it gets ported over.

src/execution/compiler/expression/null_check_translator.cpp
```diff
@@ -1,4 +1,6 @@
 #include "execution/compiler/expression/null_check_translator.h"
+
+#include "execution/ast/type.h"
```
pmenon: Is this needed?

src/include/execution/ast/builtins.h
pmenon: Try renaming `IsNull` to `IsNullVal` or `IsNullSQL` to indicate SQL-ness and to distinguish it from NULL pointer comparisons.

src/include/execution/compiler/codegen.h
pmenon: Try renaming these `Null*()` functions to `Null*Val()` to indicate these are SQL values?

src/include/execution/vm/bytecodes.h
pmenon: Reuse the `ValIsNull` and `ValIsNotNull` bytecodes rather than adding new ones - they're meant to do the same thing. Tweak the existing bytecodes to operate on primitive booleans.

src/include/execution/vm/bytecodes.h
```diff
@@ -172,6 +172,15 @@ namespace terrier::execution::vm {
...
+  F(InitNullBool, OperandType::Local)                                                                                 \
+  F(InitNullInt, OperandType::Local)                                                                                  \
+  F(InitNullReal, OperandType::Local)                                                                                 \
+  F(InitNullDecimal, OperandType::Local)                                                                              \
```
pmenon: For consistency, mimic the `Init*` bytecodes by adding the `Val` suffix for SQL values?

src/include/execution/vm/bytecode_handlers.h
```diff
@@ -1300,11 +1336,13 @@ VM_OP_WARM void OpPow(terrier::execution::sql::Real *result, const terrier::exec
 // ---------------------------------------------------------
 
 VM_OP_WARM void OpValIsNull(terrier::execution::sql::BoolVal *result, const terrier::execution::sql::Val *val) {
-  terrier::execution::sql::IsNullPredicate::IsNull(result, *val);
+  bool is_null = terrier::execution::sql::IsNullPredicate::IsNull(*val);
```
pmenon: Modify this to operate on primitive booleans and use these two bytecodes instead of adding new ones.

src/include/execution/compiler/translator_factory.h
```diff
@@ -77,7 +77,7 @@ class TranslatorFactory {
    */
   static bool IsUnaryOp(parser::ExpressionType type) {
     // TODO(Amadou): Support other unary operators.
```
pmenon: Remove `TODO`. Others will be added as they're supported.

mbutrovich:
Is there rough sketch or outline to address this in the future:
> I'm okay with adding one builtin per type for now. It'll get nasty later on.

I'm not actually sure where we would document it, but if you have something in mind it might be helpful for us to jot that down somewhere now while it's fresh.

src/execution/sema/sema_builtin.cpp
```diff
@@ -126,12 +126,45 @@ void Sema::CheckBuiltinSqlConversionCall(ast::CallExpr *call, ast::Builtin built
     return;
   }
 
+  // Null checks.
+  switch (builtin) {
+    case ast::Builtin::NullBool:
+      call->SetType(GetBuiltinType(ast::BuiltinType::Bool));
```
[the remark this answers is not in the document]
lmwnshn: Pulled out.

src/execution/vm/bytecode_generator.cpp
lmwnshn: Pulled out.

src/execution/vm/bytecode_generator.cpp
lmwnshn: Refactored to be one function.

src/include/execution/ast/builtins.h
lmwnshn: Renamed.

src/execution/compiler/codegen.cpp
lmwnshn: Added a TODO.

lmwnshn:
@mbutrovich Github was down earlier today, could you rerun checks? I don't have Jenkins permissions. Thanks!

src/include/execution/vm/bytecode_handlers.h
```diff
@@ -606,6 +607,11 @@ VM_OP_HOT void OpForceBoolTruth(bool *result, terrier::execution::sql::BoolVal *
   *result = input->ForceTruth();
 }
 
+VM_OP_HOT void OpInitSqlNull(terrier::execution::sql::Val *result) {
+  *result = terrier::execution::sql::Val::Null();
+  result->is_null_ = true;
```
pmenon: You can remove this line since the `is_null` flag should have been set in `Val::Null()`.

lmwnshn:
Test added, slight refactor. Good to go? @mbutrovich @pmenon

Issue #766, opened 2020-02-10
PostgresProtocolHandler does not allow for simultaneous connections most of the time


tanujnay112:
The way PostgresProtocolHandler handles new connections now is that it creates a temporary namespace in the catalog. In workloads with many concurrent clients connecting at the same time, this will lead to write conflicts, aborts and hence many denied connections. This will be annoying when we start benchmarking with many simultaneous client connections. Ideally, we want to get some retry mechanism going on this.
After adjusting RacerTest in NetworkTests to actually check if the connection was successful by adjusting the return value of ManualPacketUtil::StartConnection, it seems that RacerTest would fail for this reason.

Issue #731, opened 2020-02-01
Unable to support more than one simultaneous connection


apavlo:
If I run the fuzzer in one thread and then connect to the DBMS in another thread, then the first thread gets disconnected. There is no error on the server side and no useful error message in the client.

A even more simple to reproduce is to try connecting to the same server instance with separate psql instances. It doesn't always happen but once it does, only one client can connect at a time.

**Server**
```
$ ./debug/terrier 
[2020-02-01 14:31:46.082] [network_logger] [info] Listening on port 15721
```

**Client 1**
```
$ psql -h localhost -p 15721 -d terrier
psql (11.5 (Ubuntu 11.5-1), server 9.5devel)
Type "help" for help.

terrier=#
```

**Client 2**
```
$ psql -h localhost -p 15721 -d terrier
psql: server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
```

lmwnshn:
From the meeting, it looks like we have the following bug.

While no psql client has disconnected yet, you can keep adding new clients, no problem.
However, once the initial **set** of psql clients have disconnected, you can only connect one client at a time.

I suspect the network state machine just goes to the wrong state when you go from "have >0 clients" to "no clients".

apavlo:
Fixed in #760

PR #772, opened 2020-02-12
Network Layer Multiple Connection Fixes

tanujnay112:
This aims to close up on issue #766 and #731. The ConnectionHandle reuse logic was incomplete. I tried to take the reuse logic out but then realized that it might be less annoying to just deal with it in fact given that our current setup with libevent doesn't allow for cleaning up the ConnectionHandles neatly (libevent callbacks are all done in the context of a ConnectionHandle instance). So it's less annoying to just store these in the ConnectionHandleFactory and reuse them as it is now.

This PR also contains concurrency fixes for ConnectionHandlerFactory and PostgresProtocolHandler. As in issue #766, multiple simultaneous connections would bring about DDL write conflicts and aborts so this PR brought in exponential backoff from 2ms to 20ms by a factor of 2 in each round.

There were also issues with network layer testing code in that ManualPacketWriter did not test for the success of its system calls.


src/include/network/postgres/postgres_protocol_interpreter.h
```diff
@@ -15,6 +15,10 @@
 
 namespace terrier::network {
 
+#define INITIAL_BACKOFF_TIME 2
```
mbutrovich: prefer `constexpr` over `#define`. Gives us scoping and type checking.

src/network/postgres/postgres_protocol_interpreter.cpp
```diff
@@ -83,7 +83,17 @@ Transition PostgresProtocolInterpreter::ProcessStartup(const common::ManagedPoin
       NETWORK_LOG_TRACE(db_name);
     }
   }
+
   auto oids = t_cop->CreateTempNamespace(context->GetConnectionID(), db_name);
+
+  uint32_t sleep_time = INITIAL_BACKOFF_TIME;
+  while ((oids.first == catalog::INVALID_DATABASE_OID || oids.second == catalog::INVALID_NAMESPACE_OID) &&
+         sleep_time <= MAX_BACKOFF_TIME) {
+    sleep(sleep_time);
```
apavlo: Switch to `sleep_for`

https://en.cppreference.com/w/cpp/thread/sleep_for

src/network/postgres/postgres_protocol_interpreter.cpp
```diff
@@ -83,7 +83,17 @@ Transition PostgresProtocolInterpreter::ProcessStartup(const common::ManagedPoin
...
+    oids = t_cop->CreateTempNamespace(context->GetConnectionID(), db_name);
+    sleep_time *= BACKOFF_FACTOR;
+  }
+
   if (oids.first == catalog::INVALID_DATABASE_OID) {
```
apavlo: Add documentation that explains why this error can occur.

src/network/postgres/postgres_protocol_interpreter.cpp
```diff
@@ -83,7 +83,17 @@ Transition PostgresProtocolInterpreter::ProcessStartup(const common::ManagedPoin
       NETWORK_LOG_TRACE(db_name);
     }
   }
+
   auto oids = t_cop->CreateTempNamespace(context->GetConnectionID(), db_name);
+
+  uint32_t sleep_time = INITIAL_BACKOFF_TIME;
+  while ((oids.first == catalog::INVALID_DATABASE_OID || oids.second == catalog::INVALID_NAMESPACE_OID) &&
```
apavlo: Add documentation that explains what this loop is protecting from.

test/network/network_test.cpp
```diff
@@ -247,31 +247,87 @@ TEST_F(NetworkTests, LargePacketsTest) {
...
+    EXPECT_NE(io_socket_unique_ptr, nullptr);
+    io_sockets.emplace_back(std::move(io_socket_unique_ptr));
+  }
+  io_sockets.clear();
+
+  for (auto &socket : io_sockets) {
+    ManualPacketUtil::TerminateConnection(socket->GetSocketFd());
+    socket->Close();
+  }
+}
```
apavlo: Consolidate duplicated code.

test/network/network_test.cpp
```diff
@@ -247,31 +247,87 @@ TEST_F(NetworkTests, LargePacketsTest) {
...
+        } catch (const std::exception &e) {
+          EXPECT_TRUE(false);
+        }
+      });
+    }
+    for (auto &t : threads) {
+      t.join();
+    }
+    threads.clear();
+    EXPECT_EQ(successes, num_threads);
```
apavlo: This test code is *amazing*. I have never seen a test case like this before. It is mind blowing.

src/network/postgres/postgres_protocol_interpreter.cpp
```diff
@@ -83,7 +83,17 @@ Transition PostgresProtocolInterpreter::ProcessStartup(const common::ManagedPoin
...
+         sleep_time <= MAX_BACKOFF_TIME) {
+    sleep(sleep_time);
+    oids = t_cop->CreateTempNamespace(context->GetConnectionID(), db_name);
+    sleep_time *= BACKOFF_FACTOR;
+  }
```
apavlo: Combine this together into a `do...while` loop instead of calling `CreateTempNamespace` in multiple locations.

src/network/postgres/postgres_protocol_interpreter.cpp
mbutrovich: It shouldn't sit in a loop with an `or` on these conditions. The former is an invalid database name provided. We don't want to spin in that case. We should only retry in the latter case (failed to create namespace). You'll see a split this logic in Extended Query PR already, but I'll just resolve the changes since yours will be merged first:

https://github.com/cmu-db/terrier/blob/65034c5b22ab87d29dbf889af0635fbe4edfeb38/src/network/postgres/postgres_protocol_interpreter.cpp

Also those returning error messages don't really work since Terminating the connection prevent any packets from being sent back to the client. We should eventually fix that.

src/network/connection_handle_factory.cpp
```diff
@@ -9,6 +9,8 @@ namespace terrier::network {
 ConnectionHandle &ConnectionHandleFactory::NewConnectionHandle(int conn_fd,
                                                                std::unique_ptr<ProtocolInterpreter> interpreter,
                                                                common::ManagedPointer<ConnectionHandlerTask> handler) {
+  common::SpinLatch::ScopedSpinLatch guard(&reusable_handles_latch_);
```
mbutrovich: I realize we don't expect a ton of contention here, but are we holding the latch longer than we need to here? It's just needed to protect the data structure, right? If so we can manually reduce the scope with another set of braces.

src/network/postgres/postgres_protocol_interpreter.cpp
```diff
@@ -83,7 +84,25 @@ Transition PostgresProtocolInterpreter::ProcessStartup(const common::ManagedPoin
...
+  // at the same time, creating DDL conflicts when creating the temp namespace
+  do {
+    // this is so that we don't sleep on our first attempt
+    if (failed_once) {
+      std::this_thread::sleep_for(std::chrono::milliseconds{sleep_time});
+      sleep_time *= BACKOFF_FACTOR;
+    }
+    oids = t_cop->CreateTempNamespace(context->GetConnectionID(), db_name);
+    failed_once = true;
+  } while ((oids.first == catalog::INVALID_DATABASE_OID || oids.second == catalog::INVALID_NAMESPACE_OID) &&
```
mbutrovich: I still don't think we want INVALID_DATABASE_OID as a condition here. If the database name is wrong, we're gonna spin in this loop, right?

mbutrovich: Propose getting rid of `failed_once` and refactoring to:
```
 do {
    oids = t_cop->CreateTempNamespace(context->GetConnectionID(), db_name);
    if (oids.first == catalog::INVALID_DATABASE_OID || oids.second != catalog::INVALID_NAMESPACE_OID) break;
    std::this_thread::sleep_for(std::chrono::milliseconds{sleep_time});
    sleep_time *= BACKOFF_FACTOR;
  } while (sleep_time <= MAX_BACKOFF_TIME);
```
Does that logic seem right?

tanujnay112: Ya this works.

PR #774, opened 2020-02-12
OperatingUnit Instrumentation

17zhangw:
PR adds instrumentation for OperatingUnits during execution.

PR consists of following high level changes:
- Each pipeline begin with `StartResourceTracker()`, `EndPipelineTracker()`
- Added visitor pattern for plan nodes
- Added `PipelineMetric`
- Added `OperatingUnitFeatureType`, `OperatingUnitFeature`, `OperatingUnits`
- Added "static" analysis pass to extract relevant data from pipelines

Edited by Lin Ma:
Thanks for @17zhangw's work! I would review the part with the instrumentation for OperatingUnits.

In the mean time, this PR also contains the augmentation to the resource counters and the metrics system that I wrote. I would be helpful if @mbutrovich can review those parts to give some feedbacks.

These are the major parts of the additional logic:
1. a new class `ResourceTracker` that combines the `PerfMonitor` and `RusageMonitor` to provide a simple interface to track resources.
2. Add the ability to record the metrics in `GarbageCollector` and the execution engine. The metrics hooks for the execution engine are actually added as builtins within the `ExecutionContext`: `StartResourceTracker` and `EndResourceTracker`.
3. A new benchmark class `TPCHRunner` that runs a micro-benchmark style TPCH workload. The programs need to be loaded from TPL, which is not included in the source code. This also contains an extension to generate `ExecutableQuery` from TPL files.


17zhangw:
There is a significant regression on  `LoggingMetricsBenchmark` possibly due to high disk I/O caused by an `EnableMetrics()` invocation.

apavlo:
@17zhangw AFAIK all the benchmarks write to a RAM disk, so there shouldn't be any disk I/O

http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-774/11/pipeline#step-352-log-2

benchmark/metrics/logging_metrics_benchmark.cpp
```diff
@@ -9,7 +9,7 @@
 #include "storage/storage_defs.h"
 #include "storage/write_ahead_log/log_manager.h"
 
-#define LOG_FILE_NAME "/mnt/ramdisk/benchmark.txt"
+#define LOG_FILE_NAME "benchmark.txt"
```
malin1993ml: @17zhangw I in fact changed the logging dir for `LoggingMetricsBenchmark` from RAM disk to real disk to run experiments locally. You just need to change it back...

17zhangw:
Following @malin1993ml's comment above, the `LoggingMetricsBenchmark`'s `LOG_FILE_NAME` was changed back to the ramdisk so there are no longer significant regressions.

benchmark/metrics/logging_metrics_benchmark.cpp
17zhangw: Noticed and reverted!

src/brain/operating_unit_recorder.cpp
```diff
@@ -0,0 +1,240 @@
...
+      auto cardinality = 0.0;
+      auto itr = features.find(feature);
+      if (itr == features.end()) {
+        features.insert(std::make_pair(feature, OperatingUnitFeature(feature, num_rows, cardinality)));
+      } else {
+        // For now, we just record the max cardinality/max estimated rows
+        // for cases of duplicate fatures (i.e ADD). In the future, may
+        // want to change the "de-duplication" approach (i.e maybe store
+        // all or store the data associated with the plan that emits the
+        // most amount of tuples)
```
malin1993ml: For the duplicated features, we can just add all of them up for now (the cardinalities and the number of rows). Think about it as we're trying to record how much work in total this pipeline is going to do.

src/brain/operating_unit_recorder.cpp
```diff
@@ -0,0 +1,240 @@
...
+    plan_features_ = {};
+    plan->Accept(common::ManagedPointer<planner::PlanVisitor>(this));
+
+    for (auto feature : plan_features_) {
+      // TODO(wz2): Get these from cost model/plan?
+      size_t num_rows = 0;
+      auto cardinality = 0.0;
+      auto itr = features.find(feature);
+      if (itr == features.end()) {
+        features.insert(std::make_pair(feature, OperatingUnitFeature(feature, num_rows, cardinality)));
```
malin1993ml: Use `move` or `emplace` to avoid copying. Unless c++ will automatically optimizer that, which I don't know?

src/include/execution/executable_query.h
```diff
@@ -58,5 +100,13 @@ class ExecutableQuery {
   // together.
   std::unique_ptr<util::Region> region_;
   std::unique_ptr<ast::Context> ast_ctx_;
+  std::unique_ptr<brain::OperatingUnits> operating_units_;
+
+  // Used to specify the output for this query
+  std::unique_ptr<exec::SampleOutput> sample_output_;
+  std::unique_ptr<exec::OutputPrinter> printer_;
```
malin1993ml: These test outputs and printers actually should not be put here... **I (Lin)** will move them out since I actually wrote this part. Just remember to merge this part back when I'm done.

malin1993ml: I've done moving this. Please merge my branch in.

src/brain/operating_unit_recorder.cpp
```diff
@@ -0,0 +1,240 @@
...
+      return OperatingUnitFeatureType::OP_COMPARE_LT;
+    case parser::ExpressionType::COMPARE_GREATER_THAN:
+      return OperatingUnitFeatureType::OP_COMPARE_GT;
+    case parser::ExpressionType::COMPARE_LESS_THAN_OR_EQUAL_TO:
+      return OperatingUnitFeatureType::OP_COMPARE_LTE;
+    case parser::ExpressionType::COMPARE_GREATER_THAN_OR_EQUAL_TO:
+      return OperatingUnitFeatureType::OP_COMPARE_GTE;
+    default:
+      return OperatingUnitFeatureType::INVALID;
+  }
```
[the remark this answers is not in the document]
17zhangw: All type information should already be inferred at this stage. Would it make sense to add type specifier  only for `OP_PLUS_OR_MINUS` (integer/decimal), `OP_MULTIPLY` (integer/decimal), `OP_DIVIDE` (integer/decimal), and `OP_COMPARE` (bool/integer/decimal/timestamp/varchar)?

malin1993ml:
Yes! That makes sense. Except that for compare, I think we only need bool,
integer, and decimal. I think the timestamp comparison is basically done as
integer comparison (correct me if I’m wrong), and we can punt on the
varchar comparison because (1) we need to know the length, and (2) we
probably can just approximate the varchar comparison using multiple integer
comparisons. Assign integer for varchar compare for now and add a TODO for
that.

mbutrovich:
Reviewing this today, FYI.

cmake_modules/BuildUtils.cmake
```diff
@@ -228,6 +228,7 @@ function(ADD_TERRIER_BENCHMARK REL_BENCHMARK_NAME)
     set(BENCHMARK_PATH "${EXECUTABLE_OUTPUT_PATH}/${BENCHMARK_NAME}")
     add_executable(${BENCHMARK_NAME} "${REL_BENCHMARK_NAME}.cpp")
     target_link_libraries(${BENCHMARK_NAME} ${TERRIER_BENCHMARK_LINK_LIBS})
+    set_target_properties(${BENCHMARK_NAME} PROPERTIES ENABLE_EXPORTS 1)
```
mbutrovich: Just curious what the purpose of this is.

src/brain/operating_unit_recorder.cpp
```diff
@@ -0,0 +1,382 @@
+#include <utility>
+
+#include "brain/operating_unit.h"
```
mbutrovich: Are we all decided on `brain` as the directory name for this stuff?

src/execution/executable_query.cpp
```diff
@@ -3,18 +3,26 @@
...
+#include "execution/parsing/scanner.h"
+#include "execution/sema/sema.h"
 #include "execution/util/region.h"
 #include "execution/vm/bytecode_generator.h"
 #include "execution/vm/module.h"
 #include "loggers/execution_logger.h"
 
 namespace terrier::execution {
 
+std::atomic<query_id_t> ExecutableQuery::query_identifier{query_id_t{0}};
```
mbutrovich: So this is just a global identifier? I'm not thrilled with globals but struggling to think of a better home for this. Can we at least declare it as `static` or do we need its symbol exported outside of this translation unit?

src/include/brain/brain_defs.h
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
+namespace terrier::brain {
+
+enum class ExecutionOperatingUnitType : uint32_t {
```
mbutrovich: Is there an enum of operations we can leverage instead of possibility duplicating, or is this distinct?

cmake_modules/BuildUtils.cmake
malin1993ml: I think we can actually remove this. I probably have accidentally added this during one debugging, unless @17zhangw added this.

src/brain/operating_unit_recorder.cpp
malin1993ml: No, we haven't decided yet. I think we can probably devise a better name after we have more stuff and the functionalities of the new components become more clear.

src/execution/executable_query.cpp
malin1993ml: I think this is already `static`, so we probably don't need this here. @17zhangw I think you can directly initialize it during declaration, right?

cmake_modules/BuildUtils.cmake
17zhangw: I did not add this, so I will go ahead and remove it.

src/include/brain/brain_defs.h
malin1993ml: This is distinct. We've discussed with Prashanth about this. These are similar to the operations in the execution engine but different.

src/execution/executable_query.cpp
17zhangw: `query_identifier` is declared as static within `ExecutableQuery`. I believe the C++ standard does not actually allow for initialization during declaration (or at least compilation fails).

https://stackoverflow.com/questions/9656941/why-cant-i-initialize-non-const-static-member-or-static-array-in-class

17zhangw:
Responding to @mbutrovich's comments above, the metrics component is reasonably solid. I think my main complaint is that there is some amount of boilerplate code (but I do not anticipate that being too big of a concern).  Something that should probably be done at some point is actually enabling metrics during startup?

src/execution/executable_query.cpp
mbutrovich: Thanks, I missed the scoping on it.

PR #777, opened 2020-02-14
Allow Arithmetic Expressions in INSERT

### LINKED ISSUE #721: Unable to use arithmetic expression in INSERT VALUE clause
We are not able to insert tuples when there is an expression in the VALUE clause of an INSERT query

```
terrier=# CREATE TABLE xxx (id INT PRIMARY KEY, val INT);
CREATE TABLE
terrier=# INSERT INTO xxx VALUES (1+1,2);
ERROR:  syntax error
```

tpan496:
Fix #721, along with some tests
I am not sure about the behavior of ```DeriveReturnValueType```, so can @lmwnshn please take a look at this? Without calling this function, I would get binding error instead.


src/binder/bind_node_visitor.cpp
```diff
@@ -362,6 +362,7 @@ void BindNodeVisitor::Visit(parser::InsertStatement *node, parser::ParseResult *
           //  This appears to be a bad assumption. We should rename it to DeriveReturnValueTypeForAggregates()
           //  or else fix up any other codepaths. I've currently fixed it for ConstantValueExpression.
           auto expr = values[i];
+          expr->DeriveReturnValueType();
```
tpan496: @lmwnshn there is also a TODO above here

test/parser/parser_test.cpp
```diff
@@ -341,6 +341,12 @@ TEST_F(ParserTestBase, InsertTest) {
   EXPECT_EQ(insert_stmt->GetInsertionTable()->GetTableName(), "foo");
   EXPECT_EQ(insert_stmt->GetInsertColumns()->size(), 3);
   EXPECT_EQ((*insert_stmt->GetValues())[0][0]->GetExpressionType(), ExpressionType::VALUE_DEFAULT);
+
+  result = parser::PostgresParser::BuildParseTree("INSERT INTO foo VALUES (1, 1+1, 1+2);");
+  insert_stmt = result->GetStatement(0).CastManagedPointerTo<InsertStatement>();
+  EXPECT_EQ(insert_stmt->GetInsertionTable()->GetTableName(), "foo");
+  EXPECT_EQ(insert_stmt->GetInsertColumns()->size(), 0);
+  EXPECT_EQ((*insert_stmt->GetValues())[0][1]->GetExpressionType(), ExpressionType::OPERATOR_PLUS);
```
lmwnshn: It would be nice to check that the numbers are correct too, but I think this is alright.

src/binder/bind_node_visitor.cpp
lmwnshn: Do we need this call? If we don't, I would remove it for now. It re-walks the entire tree.

tpan496:
@lmwnshn If I don't call ```DeriveReturnValueType" I would get a binding error.

mbutrovich:
> @lmwnshn If I don't call `DeriveReturnValueType` I would get a binding error.

Is there an ordering of operations that we can adjust to fix that without multiple tree traversals?

lmwnshn:
Probably not in the scope of this PR. Sounds like it isn't being set properly, it should happen out of the parser or upon entry into the binder (I'd argue the latter makes more sense). I would merge it then. @mbutrovich

PR #790, opened 2020-02-24
Added LSAN Suppression for BwTree

apavlo:
This hopefully resolves the random failures with `tpcc_test`. I will run this overnight on my local machines. If we're okay, then I'll merge in the morning.

https://youtu.be/BPz9wiQV174


apavlo:
This doesn't work

PR #795, opened 2020-02-25
Alias Renaming Fix

### LINKED ISSUE #779: Alias Renaming (AS) Does Not Work for Output Columns
The DBMS currently ignores aliased column names in the output list. 
```
CREATE TABLE
INSERT 0 4
terrier=# SELECT id AS new_name FROM xxx;
 id 
----
  1
  2
  3
  4
(4 rows)
```

It also does not work if you give a constant as the output:
```
terrier=# SELECT 1 AS new_name FROM xxx;
 INTEGER 
---------
       1
       1
       1
       1
(4 rows)
```

wuwenw:
This PR fixes #779 Alias Renaming (AS) Does Not Work for Output Columns


src/include/network/postgres/postgres_packet_writer.h
```diff
@@ -153,7 +153,9 @@ class PostgresPacketWriter : public PacketWriter {
     for (const auto &col : columns) {
       const auto col_type = col.GetType();
       // TODO(Matt): Figure out how to get table oid and column oids in the OutputSchema (Optimizer's job?)
-      AppendString(col.GetName())
+      auto alias = col.GetExpr()->GetAlias();
```
mbutrovich: This is going to make an unnecessary copy before you check if you need it.

src/include/network/postgres/postgres_packet_writer.h
```diff
@@ -153,7 +153,9 @@ class PostgresPacketWriter : public PacketWriter {
...
+      auto name = alias.empty() ? col.GetName() : alias;
```
mbutrovich: This should be `const auto &`. We don't need a local copy of the string.

PR #796, opened 2020-02-25
Clang-format not returning non-zero exit on failure fix

### LINKED ISSUE #770: make check-format doesn't kill the build if it fails
e.g. the recently merged log on [Travis](https://api.travis-ci.org/v3/job/648731480/log.txt).

I think it used to work. Will look into it more later tonight.

portablesounds:
Fix #770. The only issue was the error flag was being overwritten for each file scanned, meaning if the last file succeeds it returns 0 regardless. I also put some obvious clang-format errors in this commit to see if the fix succeeded.


portablesounds:
Jenkins failed correctly (see [here](http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-796/1/pipeline)). Will fix issue and then should be ready to merge.

Issue #753, opened 2020-02-06
Modify cmake to Build Benchmarks without Building Tests


apavlo:
We currently cannot build the benchmarks without building the tests. This is because the benchmarks reference some TPC-C stuff in [test_util](https://github.com/cmu-db/terrier/tree/master/test/include/test_util).

If you try to build the benchmarks without the tests, you get the following error:

```
cmake -DCMAKE_BUILD_TYPE=debug -DTERRIER_BUILD_BENCHMARKS=ON -DTERRIER_BUILD_TESTS=OFF ..
<SNIP>
CMake Error at benchmark/CMakeLists.txt:9 (add_dependencies):
  The dependency target "test_util" of target "benchmark_util" does not
  exist.
```

We need to modify our cmake configuration to build `test_util` for the benchmarks.

Once this is done, modify our [Jenkins benchmark stage](https://github.com/cmu-db/terrier/blob/master/Jenkinsfile#L283) to only build the benchmarks. This should hopefully cut down the build time by 2 minutes.

PR #797, opened 2020-02-25
Attempt to decouple benchmark from test

portablesounds:
This is an attempted solution to #753 in order to slightly speed up build times. Instead of benchmark referring to tpcc in test, we refer to our own local copy in benchmark_util.


portablesounds:
That last merge was due to working on the wrong branch.

mbutrovich:
I don't think that this is the right approach. We don't want to duplicate the test_util code. We just need to modify CMake so that the test_util target gets built when tests aren't being built but benchmarks still are.

PR #798, opened 2020-02-26
Modify cmake to Build Benchmarks without Building Tests

### LINKED ISSUE #753: Modify cmake to Build Benchmarks without Building Tests

mbutrovich:
Fix #753


mbutrovich:
The benchmark build was already not that slow since it's a dedicated box. This dropped the build time from roughly ~4.5 minutes to ~3.

PR #802, opened 2020-02-28
Fix Table Qualifier With Star Expression

### LINKED ISSUE #780: Select Star with Table Qualifier Does Not Work
A query with `SELECT *` works on one or multiple tables:

```
CREATE TABLE
INSERT 0 4
terrier=# CREATE TABLE yyy (id INT);
CREATE TABLE
terrier=# INSERT INTO yyy VALUES (1),(2),(3),(4);
INSERT 0 4
terrier=# SELECT * FROM xxx JOIN yyy ON xxx.id = yyy.id;
 id | id 
----+----
  1 |  1
  2 |  2
  3 |  3
  4 |  4
(4 rows)
```

But if I add the table name as a qualifier in the output list, it does not work. This query should return the columns from table `xxx` and nothing from table `yyy`:
```
terrier=# SELECT xxx.* FROM xxx JOIN yyy ON xxx.id = yyy.id;
ERROR:  binding failed
```

tpan496:
Fix #780
This should fix any table qualifier that uses `*`.

Bug: even though I got rid of binding failures, there are false positives. e.g. `SELECT xxx.* FROM xxx JOIN yyy ON xxx.id = yyy.id;` returns both ids.
Update: this is now fixed


tpan496:
Not sure why `filter_manager_test` is failing, I don't think it's related.

src/binder/bind_node_visitor.cpp
```diff
@@ -61,10 +61,27 @@ void BindNodeVisitor::Visit(parser::SelectStatement *node, parser::ParseResult *
...
       continue;
     }
 
+    // Check if there is a table qualifier with a star expression
+    if (select_element->GetExpressionType() == parser::ExpressionType::COLUMN_VALUE) {
+      auto *expr = static_cast<parser::ColumnValueExpression *>(select_element.Get());
+      std::tuple<catalog::db_oid_t, catalog::table_oid_t, catalog::Schema> tuple;
+      std::string table_name = expr->GetTableName();
+      std::string col_name = expr->GetColumnName();
+      if (col_name == "*") {
```
lmwnshn: I am surprised, I thought we had a star expression.

src/parser/postgresparser.cpp
```diff
@@ -528,7 +528,11 @@ std::unique_ptr<AbstractExpression> PostgresParser::ColumnRefTransform(ParseResu
         table_name = "";
       } else {
         auto next_node = reinterpret_cast<Node *>(fields->head->next->data.ptr_value);
-        col_name = reinterpret_cast<value *>(next_node)->val_.str_;
+        if (next_node->type == T_A_Star) {
+          col_name = "*";
```
lmwnshn: Should this maybe generate a star expression instead? @tpan496

src/binder/bind_node_visitor.cpp
tpan496: There is a generic `StarExpression`, but not for table qualifiers. This case is simply not handled at all.

src/parser/postgresparser.cpp
tpan496: I can't directly return `StarExpression` in this case, because there is a table qualifier. We only want all the columns for the given table.

tpan496:
> Is there a reason to not use the existing StarExpression? If there is, please document it. Thanks!

I added them in postgresparser.cpp. Let me know if this makes sense. Thanks!

src/parser/postgresparser.cpp
apavlo: I agree with @lmwnshn. I don't think we should hardcode `col_name = "*"` here. That seems like a hack.

test/binder/binder_test.cpp
```diff
@@ -293,6 +293,50 @@ TEST_F(BinderCorrectnessTest, SelectStatementStarTest) {
   EXPECT_TRUE(b2_exists);
 }
 
+// NOLINTNEXTLINE
+TEST_F(BinderCorrectnessTest, SelectStatementQualifierStarTest) {
+  // Check if star expression is correctly processed
+  BINDER_LOG_DEBUG("Checking STAR expression in select and sub-select");
+
+  std::string select_sql = "SELECT A.* FROM A JOIN B ON A.A1 = B.B1";
```
apavlo: I would also test multiple tables.

apavlo:
@lmwnshn Is this PR still valid after #831?

lmwnshn:
Sorry, I didn't see the ping. This PR is still valid. If it doesn't merge soon, it will probably conflict with the changes me and Matt are doing because I reorganized the bind_node_visitor file -- if this happens please ping me if you need help reapplying changes.

test/binder/binder_test.cpp
lmwnshn: Marking as resolved in the interests of getting it merged.

apavlo:
@lmwnshn Why not extend `StarExpression` to support a table name?

lmwnshn:
@apavlo If we want to cache entire expression trees as part of a binder rewrite, and pull out parameter and context stuff to the side, I think it helps to avoid storing state in StarExpression. AFAIK StarExpression does not [0] come in with any table information and so this has to be figured out during binding. I'm generally not in favor of annotating expressions. Thoughts?

[0] https://github.com/cmu-db/terrier/blob/cac92a274a7ae031534e6092c36cc79f4bdaef89/third_party/libpg_query/src/postgres/src_backend_nodes_copyfuncs.c#L2618

apavlo:
It just seems super hacky to name the column "*". I'm not comfortable with that.

apavlo:

Issue #744, opened 2020-02-05
Execution of FunctionExpressions on Builtin Procs


tanujnay112:
After #658 gets merged we have a working pg_proc table in the catalog. There is also added functionality in the binder to bind a proc_oid_t to the FunctionExpression node. As of #658, there are bootstrapped procedures in the catalog such as cosine, uppercase, and lowercase functions. Their implementations are also written in the execution part of the codebase. The link between the binder and the execution engine for these is what's missing.

To begin supporting UDF functionality, we need to get the execution of these builtin procs that are bound to the FunctionExpression nodes going at the TPL/CodeGen level. This will involve adding another column to pg_proc to store function pointers and generating their execution at the TPL level.

PR #803, opened 2020-03-01
Getting builtin procs to execute end to end

tanujnay112:
Works on closing issue #744. Creates a UDF_context object whose pointers are added as an internal column in pg_proc. This object holds metadata about a proc that is to be executed. For now this just includes info about a particular builtin.

EDIT: Also added TPL AST support for the LOWER builtin.


script/testing/junit/BuiltinUDFTest.java
```diff
@@ -0,0 +1,95 @@
...
+    private void initDatabase() throws SQLException {
+        Statement stmt = conn.createStatement();
+        stmt.execute(SQL_DROP_TABLE);
+        stmt.execute(SQL_CREATE_TABLE);
+    }
+
+    /**
+     * Get columns, for follow on value checking
+     */
+    private void getResults() throws SQLException {
```
rohanaggarwal7997: I may be wrong But I dont see this being used anywhere. Is it required, When you are checking you are doing select function() from table , right?

script/testing/junit/BuiltinUDFTest.java
```diff
@@ -0,0 +1,95 @@
...
+        ResultSet rs = stmt.executeQuery(select_SQL);
+        rs.next();
+        checkRow(rs, new String [] {"cos(theta)"}, new double [] {1.0}, 0.0001);
+        assertNoMoreRows(rs);
+
+        select_SQL = "SELECT sin(theta) FROM tbl;";
+        stmt = conn.createStatement();
+        rs = stmt.executeQuery(select_SQL);
+        rs.next();
+        checkRow(rs, new String [] {"sin(theta)"}, new double [] {0.0}, 0.0001);
```
rohanaggarwal7997: I think we need to have an assert checking there is actually a row present in result set as checkrow assumes that a row is present. Ow there might be a segmentation fault if nothing gets returned.

src/execution/sema/sema_builtin.cpp
```diff
@@ -2201,6 +2201,23 @@ void Sema::CheckBuiltinParamCall(ast::CallExpr *call, ast::Builtin builtin) {
   call->SetType(ast::BuiltinType::Get(GetContext(), sql_type));
 }
 
+void Sema::CheckBuiltinStringCall(ast::CallExpr *call, ast::Builtin builtin) {
+  ast::BuiltinType::Kind sql_type;
+  switch (builtin) {
+    case ast::Builtin::Lower: {
+      if (!CheckArgCount(call, 1)) {
+        return;
```
rohanaggarwal7997: I feel we should call ReportIncorrectCallArg to report that it was called without parameters. Otherwise tpcc just crashes.

tanujnay112: Over here, we're not complaining about incorrect argument. Rather we are complaining about the wrong number of arguments right. It seems like on such instances like over here
https://github.com/cmu-db/terrier/blob/c3061172ee6037bbad82bfdc94c9ca9fa5a3d95f/src/execution/sema/sema_builtin.cpp#L1754

We just return as is here.

src/include/catalog/postgres/pg_proc.h
```diff
@@ -53,14 +53,16 @@ constexpr col_oid_t PROSRC_COL_OID = col_oid_t(21);  // VARCHAR (skey)
...
 
 constexpr std::array<col_oid_t, NUM_PG_PROC_COLS> PG_PRO_ALL_COL_OIDS = {
     PROOID_COL_OID,      PRONAME_COL_OID,        PRONAMESPACE_COL_OID, PROLANG_COL_OID,         PROCOST_COL_OID,
     PROROWS_COL_OID,     PROVARIADIC_COL_OID,    PROISAGG_COL_OID,     PROISWINDOW_COL_OID,     PROISSTRICT_COL_OID,
     PRORETSET_COL_OID,   PROVOLATILE_COL_OID,    PRONARGS_COL_OID,     PRONARGDEFAULTS_COL_OID, PRORETTYPE_COL_OID,
     PROARGTYPES_COL_OID, PROALLARGTYPES_COL_OID, PROARGMODES_COL_OID,  PROARGDEFAULTS_COL_OID,  PROARGNAMES_COL_OID,
-    PROSRC_COL_OID,      PROCONFIG_COL_OID};
+    PROSRC_COL_OID,      PROCONFIG_COL_OID,      PRO_CTX_PTR_COL_OID};
 
 constexpr proc_oid_t ATAN2_PRO_OID = proc_oid_t(84);
```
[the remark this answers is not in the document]
tanujnay112: Can you generate the id values on the RHS in macros?

src/include/execution/udf/udf_context.h
```diff
@@ -0,0 +1,76 @@
...
+#include "common/managed_pointer.h"
+#include "execution/ast/builtins.h"
+#include "type/type_id.h"
+
+namespace terrier::execution::udf {
+
+/**
+ * @brief Stores execution and type information about a stored procedure
+ */
+class UDFContext {
```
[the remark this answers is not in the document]
tanujnay112: Is there a way to specify to just recompute this after recovery? I feel like that would be a better idea especially when this stores dynamically added UDF's where you would potentially also have to serialize ast::File's.

mbutrovich:
Tagging as in progress due to merge conflict, but once that's resolved and passes CI this will be ready to merge and not need another review.

Issue #793, opened 2020-02-25
HAVING fails to handle contraints


wuwenw:
The following query of HAVING doesn't work. There are only 4 rows so the query should return 0 rows as expected.

```
terrier=# CREATE TABLE xxx (id INT);
CREATE TABLE
terrier=# INSERT INTO xxx VALUES (1),(2),(3),(4);
INSERT 0 4
terrier=# SELECT COUNT(*) AS cnt FROM xxx HAVING COUNT(*) > 10;
```
But in terrier it returns 4 rows.
 ```
COUNT STAR
------------
          4
```

VenkatDatta:
Hi, 
I was going through source code for debugging this issue.
I observed that there was a hard check on the group_by operation in the postgresparser.cpp. Due to this, the HAVING operation was not getting added to the parse_result in cases where group by is missing. After rectifying this code, I got stuck with optimizer. In this, operation type is getting resolved to Aggregate type instead of HashGroupBy type. 
I'm new to this project. I'm not sure whether I'm headed in the right path please let me know. 
Also, Can I work on this task?

apavlo:
Fixed in #823

Issue #821, opened 2020-03-07
Remove Travis Depedency


apavlo:
Travis is timing out because our builds take longer than 50 minutes. We are at the point where it is not helping us so we can just remove it.

crd477:
As part of this, remember to remove the Travis badge from `README.md`.

crd477:
Is there anything more that is blocking this?  It's not a problem but I'm occasionally reminded when I get error emails for my own branches.

PR #823, opened 2020-03-08
 Fix Having Clause (#793)

VenkatDatta:
Fix for #793 
Test:
```
CREATE TABLE
INSERT 0 4
terrier=# SELECT COUNT(*) AS cnt FROM xxx HAVING COUNT(*) > 10;

 cnt 
-----
(0 rows)
```

### DIFF
--- a/script/testing/junit/src/SelectTest.java
+++ b/script/testing/junit/src/SelectTest.java
@@ -130,11 +130,9 @@ public class SelectTest extends TestUtility {
         assertNoMoreRows(rs);
 
         /* Wrong count, wait for Having fix */
-        select_SQL = "SELECT COUNT(*) AS cnt FROM tbl HAVING COUNT() > 10;";
+        select_SQL = "SELECT COUNT(*) AS cnt FROM tbl HAVING COUNT(*) > 10;";
         stmt = conn.createStatement();
         rs = stmt.executeQuery(select_SQL);
-        rs.next();
-        checkRow(rs, new String [] {"cnt"}, new int [] {2});
         assertNoMoreRows(rs);
 
         /* self name */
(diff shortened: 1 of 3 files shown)


apavlo:
@VenkatDatta Thanks for submitting this. @wuwenw will review it for you. You can ignore the failed/timed-out build on Travis. We plan on disabling it in the near future (#821).

script/testing/junit/src/SelectTest.java
```diff
@@ -130,11 +130,9 @@ public void testSelectAlias() throws SQLException {
         assertNoMoreRows(rs);
 
         /* Wrong count, wait for Having fix */
-        select_SQL = "SELECT COUNT(*) AS cnt FROM tbl HAVING COUNT() > 10;";
+        select_SQL = "SELECT COUNT(*) AS cnt FROM tbl HAVING COUNT(*) > 10;";
         stmt = conn.createStatement();
         rs = stmt.executeQuery(select_SQL);
```
wuwenw: Keep checkRow function call to verify the result, but change it to 0

VenkatDatta: Query: SELECT COUNT(*) AS cnt FROM tbl HAVING COUNT(*) > 10;
In this query, cnt value will always be 2. Since we have a check of having count(*) > 10, no rows will be returned. I'm validating this with help of assertNoMoreRows. I don't think we can use checkRow function in this case.
Please let me know if I'm wrong.

wuwenw: Ok then it should be good to go.

apavlo:
@VenkatDatta Please fix the conflicts and we can merge.

PR #826, opened 2020-03-09
Cast SQL Integers to Reals for arithmetic operations.

amlatyrngom:
This PR adds cast from `sql::Integer` to `sql::Real` to perform arithmetic operations between the two types. This is needed by the NewOrder query.

For example, the following now works:
```sql
CREATE TABLE xxx (a int, b float);
INSERT INTO xxx (1, 1.37);
SELECT a+b FROM xxx;
```


script/testing/junit/src/SelectTest.java
```diff
@@ -171,8 +171,29 @@ public void testSelectAlias() throws SQLException {
...
+      stmt.execute(drop_SQL);
+      stmt = conn.createStatement();
+      stmt.execute(create_SQL);
+      stmt = conn.createStatement();
+      stmt.execute(insert_SQL);
+      stmt = conn.createStatement();
+      rs = stmt.executeQuery(select_SQL);
+      rs.next();
+      checkDoubleRow(rs, new String [] {"AB"}, new double[] {2.37});
+      assertNoMoreRows(rs);
```
lmwnshn: Consider dropping tbl here?

src/main/terrier.cpp
```diff
@@ -79,9 +79,9 @@ int main(int argc, char *argv[]) {
   auto db_main = terrier::DBMain::Builder()
                      .SetSettingsParameterMap(std::move(param_map))
                      .SetUseSettingsManager(true)
-                     .SetUseMetrics(true)
```
lmwnshn: Undo these changes?

src/network/postgres/postgres_network_commands.cpp
```diff
@@ -27,7 +27,7 @@ Transition SimpleQueryCommand::Exec(common::ManagedPointer<ProtocolInterpreter>
                                     common::ManagedPointer<trafficcop::TrafficCop> t_cop,
                                     common::ManagedPointer<ConnectionContext> connection) {
   const std::string query = in_.ReadString();
-  NETWORK_LOG_TRACE("Execute SimpleQuery: {0}", query.c_str());
+  NETWORK_LOG_INFO("Execute SimpleQuery: {0}", query.c_str());
```
lmwnshn: Is this intended?

PR #830, opened 2020-03-10
Enable ccache in Jenkins

crd477:
@portablesounds pinged me recently about using ccache in the CI environment.  This PR does that.
Here's a sample of the time savings for a single stage, `-DCMAKE_BUILD_TYPE=Debug -DTERRIER_USE_ASAN=ON && time make -j$(nproc)`:
First run:
```
...
[100%] Built target recovery_benchmark

real    7m11.187s
user    116m16.729s
sys     10m35.875s
```

Second run in a new container on the same Docker host:
```
...
[100%] Built target tpch_runner

real    1m56.294s
user    26m8.217s
sys     4m22.803s
```

Running on the full pipeline saved 30 minutes (comparing branches master to ccache):
http://jenkins.db.cs.cmu.edu:8080/job/crd/job/terrier-crd/

It should get a little better as more things wind up in the cache but I was surprised how much of the total time is consumed by the tests; ccache won't help those run faster.

The key details are as follows:

1. since terrier already uses CMake, it turned out to be trivial to invoke ccache.  All I had to do was add `-DCMAKE_CXX_COMPILER_LAUNCHER=ccache` to the cmake invocation and it just works.  CMake is a true gem...
1. the `ccache_dir` is shared on the local filesystem of each worker. That means that there is no need to do anything special from the Jenkins pipeline in order to take advantage of the ccache - no rsync, no `archiveArtifacts`, nothing.  The downside is that the cache will only contain object files for build types that have already occurred on that particular jenkins worker node.  While this is not quite as ideal as having a completely centralized cache directory, it make things much simpler to implement and it will improve over time.
1. the way that the docker containers share the cache on the host is by using the `-v /jenkins/ccache:/home/jenkins/.ccache` bind mount.
1. you might notice that I just arbitrarily picked 250GB for the size of the cache directory based on knowledge of the resources for the workers.  I'm not sure what the retention policy should be and how that policy should be enforced (e.g., crontab, etc.).  That's a point for discussion, I guess.
1. this does not enable ccache on the macOS workers.  It can be done but those nodes have much less local storage and the configuration would have to be a little different.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -113,14 +113,15 @@ pipeline {
                     agent {
                         docker {
                             image 'ubuntu:bionic'
-                            args '--cap-add sys_ptrace'
+                            args '--cap-add sys_ptrace -v /jenkins/ccache:/home/jenkins/.ccache'
                         }
                     }
                     steps {
                         sh 'echo $NODE_NAME'
                         sh 'echo y | sudo ./script/installation/packages.sh'
+                        sh 'sudo apt-get -y install ccache'
                         sh 'mkdir build'
-                        sh 'cd build && cmake -DCMAKE_BUILD_TYPE=Debug -DTERRIER_USE_ASAN=ON .. && make -j$(nproc)'
+                        sh 'cd build && cmake -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Debug -DTERRIER_USE_ASAN=ON .. && make -j$(nproc)'
                         sh 'cd build && make check-clang-tidy'
                         sh 'cd build && timeout 1h make unittest'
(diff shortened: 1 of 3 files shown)


crd477:
The `sys_ptrace` capability is not required for docker stages that do not build and run with address sanitizer.  deacbde363ae7886ba803d34a9d32aedd02f191b just removes it from a few places where it is unnecessary.  I suspect it was just copy/pasted from other stages out of convenience.

That commit is not strictly required for this PR.  Sorry, that's probably poor form.  You can leave it out if you want.

crd477:
I don't think that commit will have any effect on the build times.  It is only a change to the security privileges that are granted to a running Docker container.

apavlo:
@crd477 Thank you! We are going to look into switching to dynamic linking so this should reduce the storage size used for each build

PR #833, opened 2020-03-13
Minor Fix for Junit Python Script

wuwenw:
Handles error if DBMS is not ready for junit


script/testing/junit/run_junit.py
```diff
@@ -79,6 +79,8 @@ def _wait_for_db_server(self):
             except:
                 time.sleep(0.1)
                 continue
+        self._stop_db_server()
```
apavlo: Do you need to do this if it failed to start? Probably doesn't matter.

wuwenw: We need to terminate the child process for DB server in case of timeout (not failure though), otherwise the next time we will get a complaint of address already in use

wuwenw:
Experiencing a timeout in jenkins now

script/testing/junit/run_junit.py
```diff
@@ -69,16 +69,21 @@ def _wait_for_db_server(self):
...
                 print ("connected to server in {} seconds".format(i*0.1))
+                print("--- %s seconds ---" % (time.time() - start_time))
                 return
             except:
                 time.sleep(0.1)
                 continue
+        # self._stop_db_server()
+        print("--- %s seconds ---" % (time.time() - start_time))
+        print('DBMS fails to start up in 15s')
         return
```
apavlo: Shouldn't this return an error to stop the script?

script/testing/junit/src/InsertTest.java
```diff
@@ -61,8 +61,15 @@ public void setup() throws SQLException {
...
-        Statement stmt = conn.createStatement();
-        stmt.execute(SQL_DROP_TABLE);
+        try {
+            Statement stmt = conn.createStatement();
+            stmt.execute(SQL_DROP_TABLE);
+        } catch (SQLException e) {
+            System.out.print("============ Java SQL Exception below ================");
+            e.printStackTrace();
+            System.out.print("======================================================");
+            throw new SQLException("Drop fail");
```
apavlo: You can make this a `RuntimeException` and wrap `e`

script/testing/junit/run_junit.py
wuwenw: Yeah now an exception is raised

PR #835, opened 2020-03-17
Inheritance structure for Optimizer Node / Contents classes

esargent28:
This PR introduces the class inheritance structure for operator / expression nodes in the optimizer to prep for the core rewriter logic. The existing operator-based node & node contents classes now inherit from AbstractOptimizerNode and AbstractOptimizerNodeContents, and expression-based ExpressionNode and ExpressionNodeContents are introduced that also inherit from those abstract classes.

Many classes in the optimizer that had operator member variables / method return values have those switched over to the abstract versions for the rewriter.

UPDATE: Memory management for operators & operator nodes created on-the-fly is done through the deferred events manager. Some methods have had a transaction context parameter added to them, so that when they create objects, they register their deletion as a commit / abort action.


src/include/optimizer/expression_node.h
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+  /**
+   * @return A string representation of this node's information (currently, all nodes
+   * have the empty string as their info)
+   */
+  std::string GetInfo() const override {
+    // TODO(esargent): create proper info statement
+    return "";
+  }
```
apavlo: I don't think we use `GetInfo` anymore. I think we can get rid of this.

src/optimizer/optimizer_task.cpp
```diff
@@ -453,14 +455,14 @@ void BottomUpRewrite::Execute() {
...
-      std::vector<std::unique_ptr<OperatorNode>> after;
+      std::vector<std::unique_ptr<AbstractOptimizerNode>> after;
       rule->Transform(common::ManagedPointer(before.get()), &after, context_);
 
       // Rewrite rule should provide at most 1 expression
       TERRIER_ASSERT(after.size() <= 1, "rule generated too many transformations");
       // If a rule is applied, we replace the old expression and optimize this
       // group again, this will ensure that we apply rule for this level until
-      // saturated, also childs are already been rewritten
+      // saturated, also children are already been rewritten
```
apavlo: :+1:

src/include/optimizer/operator_node.h
```diff
@@ -67,39 +105,55 @@ class OperatorNode {
...
   /**
    * @return vector of children
    */
-  std::vector<common::ManagedPointer<OperatorNode>> GetChildren() const {
-    std::vector<common::ManagedPointer<OperatorNode>> result;
+  std::vector<common::ManagedPointer<AbstractOptimizerNode>> GetChildren() const override {
+    std::vector<common::ManagedPointer<AbstractOptimizerNode>> result;
     result.reserve(children_.size());
-    for (auto &i : children_) result.emplace_back(i);
+    int count = 0;
```
mush-zhang, line 116: What is `count` tracking?

esargent28: ah shoot, I think I had that for debugging purposes when working out the memory leak, but it doesn't do anything for the functionality. I'll get rid of it

esargent28:
> Generally LGTM. Some minor comments/questions.
> 
> Structurally, it might not be avoidable since the `TransactionContext` is needed in order to register a deferred action against commit or abort. But as of now it seems like the `TransactionContext` is being passed around to a lot of places which I am not personally a big fan of. Particularly with regards to `Make` of the logical/physical operators, can the function that invokes `Make` handle registering deferred event?

I could add a method to nodes / node contents / operators like "registerWithTxnContext" that would go in and register the internal pointer with the transaction context, meaning we could call Make without the txn context and then register it in the method that's calling Make. It feels a bit weird to have the object register its memory management during its lifetime, since when I see the managed pointer field I assume it's externally managed from the get-go, but maybe that's just me overthinking it. That would lower the amount of passing transaction contexts around

17zhangw:
> I could add a method to nodes / node contents / operators like "registerWithTxnContext" that would go in and register the internal pointer with the transaction context, meaning we could call Make without the txn context and then register it in the method that's calling Make. It feels a bit weird to have the object register its memory management during its lifetime, since when I see the managed pointer field I assume it's externally managed from the get-go, but maybe that's just me overthinking it. That would lower the amount of passing transaction contexts around

I think if it's not too much of a hassle, it might be worth taking this step if it will make things a bit cleaner? Just my 2 cents on this matter though.

mbutrovich:
> UPDATE: Memory management for operators & operator nodes created on-the-fly is done through the deferred events manager. Some methods have had a transaction context parameter added to them, so that when they create objects, they register their deletion as a commit / abort action.

I'm a bit uncomfortable with this design since it's sticking stuff on the critical path of query execution, which does not seem like the appropriate place for `new` and `delete` calls if avoidable.

src/include/optimizer/expression_node.h
```diff
@@ -0,0 +1,100 @@
...
+   * @return This ExpressionNode's child nodes.
+   */
+  std::vector<common::ManagedPointer<AbstractOptimizerNode>> GetChildren() const override {
+    std::vector<common::ManagedPointer<AbstractOptimizerNode>> result;
+    result.reserve(children_.size());
+    for (auto &i : children_) {
+      ExpressionNode *copy_node = reinterpret_cast<ExpressionNode *>(i->Copy().release());
+      if (txn_ != nullptr) {
+        txn_->RegisterCommitAction([=]() { delete copy_node; });
+        txn_->RegisterAbortAction([=]() { delete copy_node; });
```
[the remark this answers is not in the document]
esargent28: You're right, GetChildren() doesn't need to copy the children. It no longer copies the nodes and just returns ManagedPointers referencing the original unique_ptr children. So the deallocation of the children is grouped in with the deallocation of the parent

src/optimizer/expression_node_contents.cpp
```diff
@@ -0,0 +1,116 @@
...
+#include "parser/expression/aggregate_expression.h"
+#include "parser/expression/comparison_expression.h"
+#include "parser/expression/conjunction_expression.h"
+#include "parser/expression/operator_expression.h"
+#include "parser/expression/star_expression.h"
+#include "parser/expression_defs.h"
+
+namespace terrier::optimizer {
+
+common::ManagedPointer<parser::AbstractExpression> ExpressionNodeContents::CopyWithChildren(
```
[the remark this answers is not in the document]
esargent28: Yeah, you're right. Now it registers the pointer with the txn context before returning

PR #844, opened 2020-03-24
Modeling based on operating units

malin1993ml:
This PR mainly contains the Python scripts that build models of the database based on the operating units. This is for @17zhangw to combine with his earlier implementation of the mini runners which provide the data. @17zhangw please review the code and let me know your feedback.

For the coding style, I tried to follow the [Google Python Style Guide](http://google.github.io/styleguide/pyguide.html) as much as possible. I think our future modifications can also be based on this style guide. (P.S., I also find the static Python code analysis from CLion very helpful to keep the coding style consistent.)

A few main components of the code:
1. `mini_trainer.py` is for training the mini models, `global_trainer.py` is for training the global models, and `endtoend_estimator.py` is for testing all the models and their estimations end-to-end.
2. Modules in the `data_class` directory contain different classes that we need to hold the training data.
3. Modules in the `info` directory contain information about what's in the data, what the inputs/outputs are for the models, and what operating units that the test queries have.
4. Modules in the `training_util` directory contain the utility code for model training, e.g., constructing the data objects.
5. Scripts in the `execution_engine_runner` folder are rather independent. They generate the mini runner code for the execution engines by directly printing TPL. We may want to retire them after we completely vetting the new mini runners. But now they're still useful.

And a few random things: A tpcc runner implemented by @hjhhsy120 that we will also use to collect training data after the system can support TPCC; added support for the `extractYear` builtin function in TPL for a TPCH query; extend the TPCH runner to add more setting variations.


test/test_util/tpcc/workload_cached.cpp
```diff
@@ -0,0 +1,239 @@
...
+  vec.emplace_back("SELECT D_NEXT_O_ID, D_TAX FROM DISTRICT WHERE D_W_ID=1 AND D_ID=2");
+  vec.emplace_back("INSERT INTO \"NEW ORDER\" (NO_O_ID, NO_D_ID, NO_W_ID) VALUES (1,2,3)");
+  vec.emplace_back("UPDATE DISTRICT SET D_NEXT_O_ID = D_NEXT_O_ID + 1 WHERE D_W_ID = 1 AND D_ID = 2");
+  vec.emplace_back(
+      "INSERT INTO \"ORDER\" (O_ID, O_D_ID, O_W_ID, O_C_ID, O_ENTRY_D, O_OL_CNT, O_ALL_LOCAL) VALUES (1,2,3,4,0,6,7)");
+  vec.emplace_back("SELECT I_PRICE, I_NAME, I_DATA FROM ITEM WHERE I_ID=1");
+  vec.emplace_back(
+      "SELECT S_QUANTITY, S_DATA, S_DIST_01, S_DIST_02, S_DIST_03, S_DIST_04, S_DIST_05, S_DIST_06, S_DIST_07, "
+      "S_DIST_08, S_DIST_09, S_DIST_10 FROM STOCK WHERE S_I_ID=1 AND S_W_ID=2");
+  // vec.emplace_back("UPDATE STOCK SET S_QUANTITY = 1, S_YTD = S_YTD + 1, S_ORDER_CNT = S_ORDER_CNT + 1, S_REMOTE_CNT =
```
17zhangw: Is there a particular reason why this query is commented out? Just curious.

test/test_util/tpcc/workload_cached.cpp
```diff
@@ -0,0 +1,239 @@
...
+          db_main_->GetStatsStorage(), std::make_unique<optimizer::TrivialCostModel>(), optimizer_timeout);
+
+      auto exec_ctx = std::make_unique<execution::exec::ExecutionContext>(db_oid_, common::ManagedPointer(txn), nullptr,
+                                                                          nullptr, common::ManagedPointer(accessor));
+
+      // generate executable query and emplace it into the vector; break down here
+      execution::ExecutableQuery executable{common::ManagedPointer(plan_node), common::ManagedPointer(exec_ctx)};
+      curr.first->second.emplace_back(std::move(executable));
+
+      txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
```
17zhangw: Does running this cause any heap-use-after-freed errors? My understanding is that `txn_manager_->Commit` could end up causing certain plan components (i.e scan predicate) to be freed by deferred events.

src/storage/data_table.cpp
```diff
@@ -60,6 +60,8 @@ void DataTable::Scan(const common::ManagedPointer<transaction::TransactionContex
 }
 
 DataTable::SlotIterator &DataTable::SlotIterator::operator++() {
+  // TODO(Lin): We need to temporarily comment out this latch for the concurrent TPCH experiments. Should be replaced
```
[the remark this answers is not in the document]
malin1993ml: I prefer to keep the comment there because (1) it is a real TODO that we need to address at some point (2) if we're going to run TPCH benchmark, this will be a pointer to comment out the latches. I intentionally revert latch commenting because I don't want to break the correctness of the system.

cmake_modules/BuildUtils.cmake
```diff
@@ -228,6 +228,7 @@ function(ADD_TERRIER_BENCHMARK REL_BENCHMARK_NAME)
     set(BENCHMARK_PATH "${EXECUTABLE_OUTPUT_PATH}/${BENCHMARK_NAME}")
     add_executable(${BENCHMARK_NAME} "${REL_BENCHMARK_NAME}.cpp")
     target_link_libraries(${BENCHMARK_NAME} ${TERRIER_BENCHMARK_LINK_LIBS})
+    set_target_properties(${BENCHMARK_NAME} PROPERTIES ENABLE_EXPORTS 1)
```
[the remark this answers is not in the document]
malin1993ml: Yes. We have to add this to expose the bytecode handler symbols to the benchmarks (`tpch_runner.cpp`). Otherwise we cannot execute TPL program from the benchmarks.

test/test_util/tpcc/workload_cached.cpp
malin1993ml: I actually don't know... I'll just revert this.

test/test_util/tpcc/workload_cached.cpp
malin1993ml: I'm not sure about the predicate freeing part. But regardless, what would be used after being freed that you're referring to? The plan is only used to compile the TPL modules in the `ExecutableQuery`. We do not save it anyway.

17zhangw: Ah ok. Thanks for the clarification.

PR #850, opened 2020-04-02
Rewrite Slot Iterator

eppingere:
Rewrote to allow for concurrent insert and iteration in latch free mannor

Also, this is my first PR. I tried to follow the instructions on the wiki but lmk if anything is wrong!! 😃


mbutrovich:
Assigning to Wan because this relates to #944.

lmwnshn:
I took a cursory pass over this PR. I am inclined towards merging #944 first because it is simpler to understand, especially in the context of implementing features like slot reuse. The data structure in #850 could be useful at a future date (specifically, if profiling shows that the slot iterator sucks on "typical" workloads).

eppingere:
> I took a cursory pass over this PR. I am inclined towards merging #944 first because it is simpler to understand, especially in the context of implementing features like slot reuse. The data structure in #850 could be useful at a future date (specifically, if profiling shows that the slot iterator sucks on "typical" workloads). It shouldn't really ever fall too far behind master because we can just pull the ConcurrentPointerVector out.

Ok sounds good.

eppingere:
Reopening because of #1071

src/include/common/concurrent_pointer_vector.h
```diff
@@ -0,0 +1,280 @@
...
+ */
+template <class T>
+class alignas(Constants::CACHELINE_SIZE) ConcurrentPointerVector {
+ public:
+  /**
+   * ConcurrentPointerVector constuctor
+   *
+   * @param start_size optional initial allocation size for the vector. Default is START_SIZE
+   */
+  explicit ConcurrentPointerVector(uint64_t start_size = START_SIZE)
```
lmwnshn, line 24: Andy is generally not in favor of default arguments, and in this case the default argument seems to be a little fishy.

Could you TERRIER_ASSERT that start size shouldn't be 0 instead?

src/include/common/concurrent_pointer_vector.h
```diff
@@ -0,0 +1,280 @@
...
+  uint64_t Insert(T *item) {
+    // claims index in vector
+    uint64_t my_index;
+    do {
+      my_index = claimable_index_;
+    } while (!claimable_index_.compare_exchange_strong(my_index, my_index + 1));
+
+    // resize vector
+    while (my_index >= capacity_) {
+      if (UNLIKELY(my_index == capacity_)) {
```
lmwnshn, line 53: Could you add a comment on why this is safe? (Something about the compare xchg strong above perhaps)

src/include/common/concurrent_pointer_vector.h
```diff
@@ -0,0 +1,280 @@
...
+
+  void SetNotReadable(T **array, uint64_t i) {
+    array[i] = reinterpret_cast<T *>(reinterpret_cast<uint64_t>(array[i]) & ~static_cast<uint64_t>(READABLE_FLAG));
+  }
+
+  bool GetReadability(T **array, uint64_t i) const {
+    return static_cast<bool>(reinterpret_cast<uint64_t>(array[i]) >> SHIFT_AMOUNT);
+  }
+
+  std::atomic<uint64_t> capacity_ = 0;
```
lmwnshn, line 267: Could you add a comment saying that this is protected by resize_mutex_?

src/include/common/concurrent_pointer_vector.h
```diff
@@ -0,0 +1,280 @@
...
+  char buffer1_[Constants::CACHELINE_SIZE - sizeof(uint64_t)] = {};
```
lmwnshn, line 268: Could you add a comment saying that these buffers are for padding?

src/include/storage/data_table.h
```diff
@@ -88,31 +104,29 @@ class DataTable {
...
-     */
-    SlotIterator(const DataTable *table, uint32_t block_index, int32_t num_advances, uint32_t offset_in_block)
-        : table_(table), block_index_(block_index), num_advances_(num_advances) {
-      current_slot_ = {block_index >= table_->blocks_.size() ? nullptr : table->blocks_[block_index], offset_in_block};
-      TERRIER_ASSERT((current_slot_.GetBlock() == nullptr && current_slot_.GetOffset() == 0) ||
-                         current_slot_.GetBlock() != nullptr,
-                     "Offset should be 0 when block is nullptr.");
+    SlotIterator(const DataTable *table) : table_(table), i_(0), is_end_(false) {  // NOLINT
+      uint64_t num_blocks = table->blocks_.size();
+      TERRIER_ASSERT(num_blocks >= 1, "there should allways be at least one block");
```
lmwnshn, line 111: typo, allways

_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-minimal
```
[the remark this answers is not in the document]
eppingere: oops this is a thing that i put in my fork and forgot to take out

benchmark/benchmark_util/data_table_benchmark_util.cpp
```diff
@@ -77,7 +77,7 @@ LargeDataTableBenchmarkObject::LargeDataTableBenchmarkObject(const std::vector<u
       operation_ratio_(std::move(operation_ratio)),
       generator_(generator),
       layout_({attr_sizes}),
-      table_(common::ManagedPointer(block_store), layout_, storage::layout_version_t(0)),
+      table_(common::ManagedPointer<storage::BlockStore>(block_store), layout_, storage::layout_version_t(0)),
```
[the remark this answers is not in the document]
eppingere: i've had issues compiling on linux without those in the past? idk its really necessary but it used to fail CI back before this  was closed. I can try it without it if you want?

src/include/common/concurrent_pointer_vector.h
```diff
@@ -0,0 +1,280 @@
...
+#include <common/shared_latch.h>
+#include <execution/util/execution_common.h>
+#include <tbb/mutex.h>
+#include <atomic>              // NOLINT
+#include <condition_variable>  // NOLINT
+
+namespace terrier::common {
+
+/**
+ * A concurrent Vector of pointers that supports inserts and accesses at an index
```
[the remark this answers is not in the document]
eppingere: done

benchmark/benchmark_util/data_table_benchmark_util.cpp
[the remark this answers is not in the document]
mbutrovich: It should only fail if there's a single argument to the constructor (in which case you need an explicit cast because single argument constructors must be declared `explicit` in our codebase).

src/include/common/concurrent_pointer_vector.h
eppingere: done

src/include/common/concurrent_pointer_vector.h
```diff
@@ -0,0 +1,280 @@
...
+    bool operator==(const Iterator &other) const {
+      if (LIKELY(other.is_end_)) {
+        return current_index_ >= end_index_;
+      }
+      if (LIKELY(is_end_)) {
+        return other.current_index_ >= other.end_index_;
+      }
+      TERRIER_ASSERT(vector_ == other.vector_, "should only compare iterators for the same vector");
+      return current_index_ == other.current_index_;
+    }
```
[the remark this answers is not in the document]
eppingere: done

src/include/common/concurrent_pointer_vector.h
```diff
@@ -0,0 +1,280 @@
...
+
+   private:
+    uint64_t current_index_ = 0, end_index_ = 0;
+    ConcurrentPointerVector *vector_ = nullptr;
+    T **local_value_ = nullptr;
+    bool is_end_ = true;
+  };
+
+  /**
+   * begin
```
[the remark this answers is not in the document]
eppingere: done

src/include/common/concurrent_pointer_vector.h
eppingere: i added one let me know if it makes things more clear

src/include/common/concurrent_pointer_vector.h
eppingere: done

src/include/common/concurrent_pointer_vector.h
eppingere: i renamed them to padding?

src/include/storage/data_table.h
eppingere: done

benchmark/benchmark_util/data_table_benchmark_util.cpp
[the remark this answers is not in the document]
eppingere: done

eppingere:
So I had some trouble converting from `ConcurrentPointerVector` to `tbb::concurrent_vector`. Unclear why that is. 

`ConcurrentPointerVector`'s makes most concurrent inserts nonblocking. If we use a `std::vector` with a `SharedLatch` instead, we get all the performance from the rest of the `SlotItorator` except when a tread wants to insert a new block which is blocking.

I've re-requested reviews due to these changes.

src/include/common/concurrent_pointer_vector.h
```diff
@@ -0,0 +1,293 @@
+#pragma once
```
lmwnshn, line 1: If I'm not mistaken, this means ConcurrentPointerVector and its associated test are not used; could those be removed for a smaller diff? Thanks!

lmwnshn (changes requested):
Quick question before I take a deeper stab at it

src/include/common/concurrent_pointer_vector.h
eppingere: Yes thats correct. I shall remove them

src/include/storage/data_table.h
```diff
@@ -74,8 +85,14 @@ class DataTable {
      * @return if the two iterators point to the same slot
      */
     bool operator==(const SlotIterator &other) const {
-      // TODO(Tianyu): I believe this is enough?
-      return current_slot_ == other.current_slot_;
+      if (LIKELY(other.is_end_)) {
```
mbutrovich, line 88: Do we believe that `LIKELY` does anything without an else clause (adjust instruction ordering) or being in a loop (adjust unrolling)? I don't think this one or the one after it are necessary.

src/include/common/concurrent_vector.h
```diff
@@ -125,6 +125,18 @@ class ConcurrentVector {
    */
   Iterator end() { return Iterator(vector_.end()); }  // NOLINT for STL name compability
 
+  /**
+   * Returns size of the vector.
+   * @return size of the vector.
+   */
+  uint32_t size() { return vector_.size(); }  // NOLINT for STL name compability
```
mbutrovich, line 132: Redundant.

benchmark/benchmark_util/data_table_benchmark_util.cpp
```diff
@@ -73,15 +73,15 @@ LargeDataTableBenchmarkObject::LargeDataTableBenchmarkObject(const std::vector<u
                                                              storage::RecordBufferSegmentPool *buffer_pool,
                                                              std::default_random_engine *generator, bool gc_on,
                                                              storage::LogManager *log_manager)
-    : txn_length_(txn_length),
-      operation_ratio_(std::move(operation_ratio)),
-      generator_(generator),
-      layout_({attr_sizes}),
+    : layout_({attr_sizes}),
```
mbutrovich, line 76: Why was this class changed?

src/include/common/concurrent_vector.h
```diff
@@ -125,6 +125,18 @@ class ConcurrentVector {
    */
```
mbutrovich, line 125: This class should stay in common/container since it's a container.

src/include/storage/data_table.h
eppingere: I think it might do something. I imagine the most common use case is going to be comparing the iterator to the end of the data table. So I wanted to make sure that is the case that the compiler optimizes for. I put the likely in the other case so that if someone does something like `for (auto it = table.begin(); table.end() != it; it++)` that will also be the second optimized for case. Do you think that's reasonable?

src/include/common/concurrent_vector.h
eppingere: done

benchmark/benchmark_util/data_table_benchmark_util.cpp
eppingere: Check-clang-tidy would like this order since there is padding in the data_table

src/include/common/concurrent_vector.h
eppingere: done

mbutrovich:
Perhaps this is a side effect of the SlotIteratorBenchmark only iterating, but it looks like the heavy synchronization was punted from the `SlotIterator`'s `++` operator to the `*` operator. I extended the benchmarks to have one that also reads the iterator, and this branch gets crushed. Note that reading the iterator after an iteration is how it's used in `TableVectorIterator`.
**Master**:
```
SlotIteratorBenchmark/ConcurrentSlotIterators/manual_time             126 ms          0 ms          7   75.6027M items/s
SlotIteratorBenchmark/ConcurrentSlotIteratorsReads/manual_time        131 ms          0 ms          6   72.6148M items/s
```
**PR #850**:
```
SlotIteratorBenchmark/ConcurrentSlotIterators/manual_time              33 ms          0 ms         27   285.152M items/s
SlotIteratorBenchmark/ConcurrentSlotIteratorsReads/manual_time        533 ms          0 ms          1   17.8926M items/s
```

eppingere:
> Perhaps this is a side effect of the SlotIteratorBenchmark only iterating, but it looks like the heavy synchronization was punted from the `SlotIterator`'s `++` operator to the `*` operator. I extended the benchmarks to have one that also reads the iterator, and this branch gets crushed. Note that reading the iterator after an iteration is how it's used in `TableVectorIterator`.
> **Master + a new benchmark**:
> 
> ```
> SlotIteratorBenchmark/ConcurrentSlotIterators/manual_time             126 ms          0 ms          7   75.6027M items/s
> SlotIteratorBenchmark/ConcurrentSlotIteratorsReads/manual_time        131 ms          0 ms          6   72.6148M items/s
> ```
> 
> **PR #850 + a new benchmark**:
> 
> ```
> SlotIteratorBenchmark/ConcurrentSlotIterators/manual_time              33 ms          0 ms         27   285.152M items/s
> SlotIteratorBenchmark/ConcurrentSlotIteratorsReads/manual_time        533 ms          0 ms          1   17.8926M items/s
> ```

Yeah i agree. I could be smarter about how I calculate the current slot? Like just update the current slot manually and only go to the table's blocks vector when actually needed?

Also would you mind sending me your reads benchmark?

mbutrovich:
Benchmark is here:
```
#include <memory>
#include <vector>

#include "benchmark/benchmark.h"
#include "benchmark_util/benchmark_config.h"
#include "common/scoped_timer.h"
#include "storage/storage_util.h"
#include "test_util/multithread_test_util.h"
#include "test_util/storage_test_util.h"

namespace terrier {

/**
 * This benchmark measures how fast the
 */
class SlotIteratorBenchmark : public benchmark::Fixture {
 public:
  void SetUp(const benchmark::State &state) final {
    // generate a random redo ProjectedRow to Insert
    redo_buffer_ = common::AllocationUtil::AllocateAligned(initializer_.ProjectedRowSize());
    redo_ = initializer_.InitializeRow(redo_buffer_);
    StorageTestUtil::PopulateRandomRow(redo_, layout_, 0, &generator_);

    // generate a ProjectedRow buffer to Read
    read_buffer_ = common::AllocationUtil::AllocateAligned(initializer_.ProjectedRowSize());
    read_ = initializer_.InitializeRow(read_buffer_);

    // generate a vector of ProjectedRow buffers for concurrent reads
    for (uint32_t i = 0; i < BenchmarkConfig::num_threads; ++i) {
      // Create read buffer
      byte *read_buffer = common::AllocationUtil::AllocateAligned(initializer_.ProjectedRowSize());
      storage::ProjectedRow *read = initializer_.InitializeRow(read_buffer);
      read_buffers_.emplace_back(read_buffer);
      reads_.emplace_back(read);
    }
  }

  void TearDown(const benchmark::State &state) final {
    delete[] redo_buffer_;
    delete[] read_buffer_;
    for (uint32_t i = 0; i < BenchmarkConfig::num_threads; ++i) delete[] read_buffers_[i];
    // google benchmark might run benchmark several iterations. We need to clear vectors.
    read_buffers_.clear();
    reads_.clear();
  }

  // Tuple layout
  const uint8_t column_size_ = 8;
  const storage::BlockLayout layout_{{column_size_, column_size_, column_size_}};

  // Tuple properties
  const storage::ProjectedRowInitializer initializer_ =
      storage::ProjectedRowInitializer::Create(layout_, StorageTestUtil::ProjectionListAllColumns(layout_));

  // Workload
  const uint32_t num_reads_ = 10000000;
  const uint64_t buffer_pool_reuse_limit_ = 10000000;

  // Test infrastructure
  storage::BlockStore block_store_{1000, 1000};
  storage::RecordBufferSegmentPool buffer_pool_{num_reads_, buffer_pool_reuse_limit_};

  // Insert buffer pointers
  byte *redo_buffer_;
  storage::ProjectedRow *redo_;

  // Read buffer pointers;
  byte *read_buffer_;
  storage::ProjectedRow *read_;

  // Read buffers pointers for concurrent reads
  std::vector<byte *> read_buffers_;
  std::vector<storage::ProjectedRow *> reads_;
};

// Iterate the num_reads_ of tuples in the sequential  order from a DataTable concurrently
// NOLINTNEXTLINE
BENCHMARK_DEFINE_F(SlotIteratorBenchmark, ConcurrentSlotIterators)(benchmark::State &state) {
  storage::DataTable read_table(common::ManagedPointer<storage::BlockStore>(&block_store_), layout_,
                                storage::layout_version_t(0));

  // populate read_table_ by inserting tuples
  // We can use dummy timestamps here since we're not invoking concurrency control
  transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0),
                                      common::ManagedPointer(&buffer_pool_), DISABLED);
  for (uint32_t i = 0; i < num_reads_; ++i) {
    read_table.Insert(common::ManagedPointer(&txn), *redo_);
  }

  auto workload = [&]() {
    auto it = read_table.begin();
    uint32_t num_reads = 0;
    while (it != read_table.end()) {
      num_reads++;
      it++;
    }
    EXPECT_EQ(num_reads, num_reads_);
  };

  common::WorkerPool thread_pool(BenchmarkConfig::num_threads, {});
  thread_pool.Startup();

  // NOLINTNEXTLINE
  for (auto _ : state) {
    uint64_t elapsed_ms;
    {
      common::ScopedTimer<std::chrono::milliseconds> timer(&elapsed_ms);
      for (uint32_t j = 0; j < BenchmarkConfig::num_threads; j++) {
        thread_pool.SubmitTask([&workload] { workload(); });
      }
      thread_pool.WaitUntilAllFinished();
    }
    state.SetIterationTime(static_cast<double>(elapsed_ms) / 1000.0);
  }
  state.SetItemsProcessed(state.iterations() * num_reads_ * BenchmarkConfig::num_threads);
}

// Iterate the num_reads_ of tuples in the sequential  order from a DataTable concurrently
// NOLINTNEXTLINE
BENCHMARK_DEFINE_F(SlotIteratorBenchmark, ConcurrentSlotIteratorsReads)(benchmark::State &state) {
  storage::DataTable read_table(common::ManagedPointer<storage::BlockStore>(&block_store_), layout_,
                                storage::layout_version_t(0));

  // populate read_table_ by inserting tuples
  // We can use dummy timestamps here since we're not invoking concurrency control
  transaction::TransactionContext txn(transaction::timestamp_t(0), transaction::timestamp_t(0),
                                      common::ManagedPointer(&buffer_pool_), DISABLED);
  for (uint32_t i = 0; i < num_reads_; ++i) {
    read_table.Insert(common::ManagedPointer(&txn), *redo_);
  }

  std::vector<std::vector<storage::TupleSlot>> reads(BenchmarkConfig::num_threads);

  for (auto &i : reads) i.resize(num_reads_);

  auto workload = [&](const uint32_t worker_id) {
    auto it = read_table.begin();
    uint32_t num_reads = 0;
    while (it != read_table.end()) {
      reads[worker_id][num_reads++] = *it++;
    }
    EXPECT_EQ(num_reads, num_reads_);
  };

  common::WorkerPool thread_pool(BenchmarkConfig::num_threads, {});
  thread_pool.Startup();

  // NOLINTNEXTLINE
  for (auto _ : state) {
    uint64_t elapsed_ms;
    {
      common::ScopedTimer<std::chrono::milliseconds> timer(&elapsed_ms);
      for (uint32_t j = 0; j < BenchmarkConfig::num_threads; j++) {
        thread_pool.SubmitTask([&workload, j] { workload(j); });
      }
      thread_pool.WaitUntilAllFinished();
    }
    state.SetIterationTime(static_cast<double>(elapsed_ms) / 1000.0);
  }
  state.SetItemsProcessed(state.iterations() * num_reads_ * BenchmarkConfig::num_threads);
}

// ----------------------------------------------------------------------------
// BENCHMARK REGISTRATION
// ----------------------------------------------------------------------------
// clang-format off
BENCHMARK_REGISTER_F(SlotIteratorBenchmark, ConcurrentSlotIterators)
    ->Unit(benchmark::kMillisecond)
    ->UseRealTime()
    ->UseManualTime();
BENCHMARK_REGISTER_F(SlotIteratorBenchmark, ConcurrentSlotIteratorsReads)
    ->Unit(benchmark::kMillisecond)
    ->UseRealTime()
    ->UseManualTime();
// clang-format on

}  // namespace terrier

```

eppingere:
@mbutrovich I updated the pr let me know if you have any other thoughts

src/storage/data_table.cpp
```diff
@@ -425,7 +343,7 @@ bool DataTable::CompareAndSwapVersionPtr(const TupleSlot slot, const TupleAccess
 }
 
 RawBlock *DataTable::NewBlock() {
-  RawBlock *new_block = block_store_->Get();
+  RawBlock *new_block = block_store_.operator->()->Get();
```
jrolli, line 346: Why did you make this change?  If the previous code was no longer working, then we have a problem with the block store code and should fix that rather than make this change.

src/storage/data_table.cpp
```diff
@@ -174,60 +120,28 @@ bool DataTable::Update(const common::ManagedPointer<transaction::TransactionCont
...
+    uint64_t size = blocks_size_;
+    if (current_insert_idx >= size) {
+      block = NewBlock();
+      common::SharedLatch::ScopedExclusiveLatch latch(&blocks_latch_);
+      blocks_.push_back(block);
+      current_insert_idx = blocks_.size() - 1;
+      blocks_size_ = blocks_.size();
+    } else {
+      common::SharedLatch::ScopedSharedLatch latch(&blocks_latch_);
+      block = const_cast<RawBlock *>(blocks_[current_insert_idx]);
```
jrolli, line 143: Why did you add `const_cast`?  As far as I can tell, it is unnecessary because neither the vector nor the lvalue are const qualified.

src/storage/data_table.cpp
```diff
@@ -11,28 +11,26 @@
...
 DataTable::~DataTable() {
-  common::SpinLatch::ScopedSpinLatch guard(&blocks_latch_);
-  for (RawBlock *block : blocks_) {
+  common::SharedLatch::ScopedExclusiveLatch latch(&blocks_latch_);
+  for (auto block : blocks_) {
     StorageUtil::DeallocateVarlens(block, accessor_);
     for (col_id_t i : accessor_.GetBlockLayout().Varlens())
       accessor_.GetArrowBlockMetadata(block).GetColumnInfo(accessor_.GetBlockLayout(), i).Deallocate();
-    block_store_->Release(block);
+    block_store_.operator->()->Release(block);
```
jrolli, line 33: Why did you change this to explicitly invoke the arrow operator rather than how it was?

src/storage/arrow_serializer.cpp
```diff
@@ -43,9 +43,8 @@ void ArrowSerializer::AssembleMetadataBuffer(std::ofstream &outfile, flatbuf::Me
 void ArrowSerializer::WriteSchemaMessage(std::ofstream &outfile, std::unordered_map<col_id_t, int64_t> *dictionary_ids,
                                          std::vector<type::TypeId> *col_types,
                                          flatbuffers::FlatBufferBuilder *flatbuf_builder) {
-  RawBlock *block = data_table_.blocks_.front();
   const BlockLayout &layout = data_table_.accessor_.GetBlockLayout();
-  ArrowBlockMetadata &metadata = data_table_.accessor_.GetArrowBlockMetadata(block);
+  ArrowBlockMetadata &metadata = data_table_.accessor_.GetArrowBlockMetadata(*(data_table_.GetBlocks().begin()));
```
jrolli, line 47: nit:  It's fine the way you have it now, but I would generally prefer the original form because it is easier to read and digest the different steps when reading it and is also more friendly for debugger sessions.

src/include/storage/data_table.h
```diff
@@ -243,28 +306,27 @@ class DataTable {
...
+  uint32_t GetNumBlocks() const { return blocks_size_; }
 
   /** @return Maximum number of blocks in the data table. */
   static uint32_t GetMaxBlocks() { return std::numeric_limits<uint32_t>::max(); }
 
   /**
    * @return a coarse estimation on the number of tuples in this table
    */
-  uint64_t GetNumTuple() const { return GetBlockLayout().NumSlots() * blocks_.size(); }
+  uint64_t GetNumTuple() const { return accessor_.GetBlockLayout().NumSlots() * blocks_size_; }
```
jrolli, line 317: Use the `DataTable` method rather than hardcoding the reference to `accessor_`.

src/include/storage/data_table.h
```diff
@@ -88,31 +114,39 @@ class DataTable {
...
-    // The remaining number of times that this iterator will advance to the next block,
-    // or ADVANCE_TO_THE_END to advance to the end.
-    int32_t num_advances_;
-    TupleSlot current_slot_;
+    void UpdateFromNextBlock() {
+      TERRIER_ASSERT(table_->blocks_size_ >= 1, "there should always be at least one block");
+      RawBlock *b;
+      {
+        common::SharedLatch::ScopedSharedLatch latch(&table_->blocks_latch_);
+        b = const_cast<DataTable *>(table_)->blocks_[block_index_];
```
jrolli, line 139: I don't think you need the `const_cast`

src/include/common/container/concurrent_vector.h
```diff
@@ -125,6 +125,11 @@ class ConcurrentVector {
    */
   Iterator end() { return Iterator(vector_.end()); }  // NOLINT for STL name compability
 
+  /**
+   * @return size of the vector.
+   */
+  uint32_t size() const { return vector_.size(); }  // NOLINT for STL name compability
+
```
jrolli, line 132: Not sure why this is added in this PR when you don't appear to use `concurrent_vector` at all.

jrolli (changes requested):
The core logic of the iterator seems OK, but I think an overly aggressive linter with a different style got run against the code as there are numerous quirks that are inconsistent with the style used in the rest of the project.

In particular, `const` has generally been an effective tool for preventing new code from violating invariants in the storage layer and this PR seems to relax many of those uses without clear indicators of it being necessary or generally advantageous.

src/include/storage/data_table.h
```diff
@@ -38,25 +42,42 @@ class HashIndex;
...
      * @return reference to the underlying tuple slot
      */
-    const TupleSlot &operator*() const { return current_slot_; }
+    TupleSlot &operator*() { return current_slot_; }
 
     /**
      * @return pointer to the underlying tuple slot
      */
-    const TupleSlot *operator->() const { return &current_slot_; }
+    TupleSlot *operator->() { return &current_slot_; }
```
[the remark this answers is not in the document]
eppingere: done

src/include/storage/data_table.h
```diff
@@ -74,7 +95,12 @@ class DataTable {
...
      */
     bool operator==(const SlotIterator &other) const {
-      // TODO(Tianyu): I believe this is enough?
+      if (LIKELY(other.is_end_)) {
+        return block_index_ >= end_index_;
+      }
+      if (LIKELY(is_end_)) {
+        return other.block_index_ >= other.end_index_;
+      }
       return current_slot_ == other.current_slot_;
```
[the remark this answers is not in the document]
eppingere: So I should definitely add a comment but this is done to avoid computation when comparing to end(). Since the data table only needs to provide snap shot isolation, when the `SlotIterator` is created, it only needs to iterate through all the slots that are available then. So the end index is kept and maintained within each iterator that tells it when its end is. This essentially means that we don't need to do anything in end to see when the iterator has reached the end. This is why we have the magic `end_` constant. Whenever an iterator is being compared to `end_` it knows to just check if it has reached the end of the `DataTable` when it was created.

src/include/storage/data_table.h
```diff
@@ -88,31 +114,39 @@ class DataTable {
    private:
     friend class DataTable;
 
-    /** Indicates that the iterator should advance to the end of the table. */
-    static constexpr int32_t ADVANCE_TO_THE_END = -1;
-    /** An invalid TupleSlot. */
-    static TupleSlot InvalidTupleSlot() { return TupleSlot(nullptr, 0); }
+    SlotIterator() : is_end_(true) {}
```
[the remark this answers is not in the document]
eppingere: https://github.com/cmu-db/noisepage/pull/850#discussion_r500703460

src/include/storage/data_table.h
```diff
@@ -88,31 +114,39 @@ class DataTable {
...
-     */
-    SlotIterator(const DataTable *table, uint32_t block_index, int32_t num_advances, uint32_t offset_in_block)
-        : table_(table), block_index_(block_index), num_advances_(num_advances) {
-      current_slot_ = {block_index >= table_->blocks_.size() ? nullptr : table->blocks_[block_index], offset_in_block};
-      TERRIER_ASSERT((current_slot_.GetBlock() == nullptr && current_slot_.GetOffset() == 0) ||
-                         current_slot_.GetBlock() != nullptr,
-                     "Offset should be 0 when block is nullptr.");
+    SlotIterator(const DataTable *table) : table_(table), block_index_(0), is_end_(false) {  // NOLINT
+      end_index_ = table_->blocks_size_;
+      TERRIER_ASSERT(end_index_ >= 1, "there should always be at least one block");
```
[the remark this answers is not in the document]
eppingere: It's not necessary per se but if the `DataTable`'s accessor is valid then there is one block added to the table by default. I suppose that we can check against it and early return and updating the slot? In that case we know that we will always have an empty table. Im fine either way

src/include/storage/data_table.h
```diff
@@ -88,31 +114,39 @@ class DataTable {
...
+
+      // only 1 empty block
+      if (end_index_ == 1) {
+        common::SharedLatch::ScopedSharedLatch latch(&table_->blocks_latch_);
+        if (table_->blocks_[0]->GetInsertHead() == 0) {
+          end_index_ = 0;
+        }
+      }
```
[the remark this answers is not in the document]
eppingere: Similar to [before](https://github.com/cmu-db/noisepage/pull/850#discussion_r500705055) the table by default initializes 1 block to start, but there are no tuples in that block. So we need to be able to identify that the slot was created on an empty table with 1 block. So to deal with that we do, otherwise its not equal to end. I thought about maybe putting this logic in `==` but that is called very frequently so I figured its best to keep it streamlined?

src/include/storage/data_table.h
```diff
@@ -279,22 +341,15 @@ class DataTable {
...
-  mutable common::SpinLatch blocks_latch_;
-  // latch used to protect insertion_head_
-  mutable common::SpinLatch header_latch_;
-  std::atomic<uint32_t> insertion_head_;
-  // Check if we need to advance the insertion_head_
-  // This function uses header_latch_ to ensure correctness
-  void CheckMoveHead(uint32_t block_index);
+  mutable common::SharedLatch blocks_latch_;
+  const layout_version_t layout_version_;
+  const SlotIterator end_ = {};
```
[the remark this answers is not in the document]
eppingere: Yeah this is the same as before. An iterator knows when it has reached the end of its range. So it checks that when compared to end. Im am willing to change that design tho?

src/include/storage/data_table.h
```diff
@@ -279,22 +341,15 @@ class DataTable {
   // needs raw access to the underlying table.
   friend class BlockCompactor;
 
-  const common::ManagedPointer<BlockStore> block_store_;
-  const layout_version_t layout_version_;
-  const TupleAccessStrategy accessor_;
+  alignas(common::Constants::CACHELINE_SIZE) std::atomic<uint64_t> blocks_size_ = 0;
+  alignas(common::Constants::CACHELINE_SIZE) std::atomic<uint64_t> insert_index_ = 0;
```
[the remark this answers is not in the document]
eppingere: yeah we only update them when we hold the lock but we do access them without holding the lock. I haven't determined that they're hot enough to warrant that. I could invest some time into investigating that if you feel its warranted?

src/storage/data_table.cpp
```diff
@@ -174,60 +120,28 @@ bool DataTable::Update(const common::ManagedPointer<transaction::TransactionCont
...
                  "The input buffer never changes the version pointer column, so it should have  exactly 1 fewer "
                  "attribute than the DataTable's layout.");
 
-  // Insertion header points to the first block that has free tuple slots
-  // Once a txn arrives, it will start from the insertion header to find the first
-  // idle (no other txn is trying to get tuple slots in that block) and non-full block.
-  // If no such block is found, the txn will create a new block.
-  // Before the txn writes to the block, it will set block status to busy.
-  // The first bit of block insert_head_ is used to indicate if the block is busy
-  // If the first bit is 1, it indicates one txn is writing to the block.
```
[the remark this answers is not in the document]
eppingere: done

src/storage/data_table.cpp
```diff
@@ -174,60 +120,28 @@ bool DataTable::Update(const common::ManagedPointer<transaction::TransactionCont
...
-    {
-      common::SpinLatch::ScopedSpinLatch guard(&blocks_latch_);
-      block = blocks_[block_index];
+    uint64_t size = blocks_size_;
+    if (current_insert_idx >= size) {
+      block = NewBlock();
+      common::SharedLatch::ScopedExclusiveLatch latch(&blocks_latch_);
+      blocks_.push_back(block);
+      current_insert_idx = blocks_.size() - 1;
+      blocks_size_ = blocks_.size();
```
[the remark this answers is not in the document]
eppingere: done

src/storage/data_table.cpp
```diff
@@ -238,10 +152,14 @@ TupleSlot DataTable::Insert(const common::ManagedPointer<transaction::Transactio
       accessor_.ClearBlockBusyStatus(block);
       // if the full block is the insertion_header, move the insertion_header
       // Next insert txn will search from the new insertion_header
-      CheckMoveHead(block_index);
+      if (current_insert_idx == insert_index_.load()) {
+        // if we fail, that's ok because that means that someone else incremented insert_index_
+        // so we retry on the next index
+        insert_index_.compare_exchange_strong(current_insert_idx, current_insert_idx + 1);
```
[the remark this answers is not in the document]
eppingere: done

src/storage/data_table.cpp
```diff
@@ -77,60 +75,8 @@ void DataTable::Scan(const common::ManagedPointer<transaction::TransactionContex
...
-}
-
-DataTable::SlotIterator DataTable::GetBlockedSlotIterator(uint32_t start, uint32_t end) const {
-  TERRIER_ASSERT(start <= end, "Start index should come before ending index.");
-  TERRIER_ASSERT(static_cast<int32_t>(end - start - 1) >= 0, "Too many blocks or sign issue.");
-
-  common::SpinLatch::ScopedSpinLatch guard(&blocks_latch_);
-  TERRIER_ASSERT(start <= blocks_.size() && end <= blocks_.size(), "Indexes must be within bounds.");
-  return {this, start, static_cast<int32_t>(end - start - 1), 0};
+  return end_;
```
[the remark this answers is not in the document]
eppingere: done

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -217,7 +217,7 @@ uint64_t LogSerializerTask::SerializeRecord(const terrier::storage::LogRecord &r
 
       // Write out the attr sizes boundaries, this way we can deserialize the records without the need of the block
       // layout
-      const auto &block_layout = record_body->GetTupleSlot().GetBlock()->data_table_->GetBlockLayout();
+      const auto &block_layout = record_body->GetTupleSlot().GetBlock()->data_table_->accessor_.GetBlockLayout();
```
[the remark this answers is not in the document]
eppingere: done

test/include/test_util/storage_test_util.h
```diff
@@ -187,7 +187,7 @@ class StorageTestUtil {
   template <class Random>
   static std::unordered_map<storage::TupleSlot, storage::ProjectedRow *> PopulateBlockRandomly(
       storage::DataTable *table, storage::RawBlock *block, double empty_ratio, Random *const generator) {
-    const storage::BlockLayout &layout = table->GetBlockLayout();
+    const storage::BlockLayout layout = table->accessor_.GetBlockLayout();
```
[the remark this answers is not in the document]
eppingere: done

test/include/test_util/storage_test_util.h
```diff
@@ -223,8 +223,8 @@ class StorageTestUtil {
   // performance. Returns the number of non-empty slots in the block after population.
   template <class Random>
   static uint32_t PopulateBlockRandomlyNoBookkeeping(storage::DataTable *table, storage::RawBlock *block,
-                                                     double empty_ratio, Random *const generator) {
-    const storage::BlockLayout &layout = table->GetBlockLayout();
+                                                     double empty_ratio, Random *generator) {
+    const storage::BlockLayout layout = table->accessor_.GetBlockLayout();
```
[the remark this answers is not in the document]
eppingere: done

test/storage/data_table_test.cpp
```diff
@@ -376,4 +376,38 @@ TEST_F(DataTableTests, InsertNoWrap) {
...
+    for (uint64_t i = 0; i < max_tuples_inserted; i++) {
+      // check that table is correct size
+      size = 0;
+      for (auto UNUSED_ATTRIBUTE _ : tested.GetTable()) size++;
+      EXPECT_EQ(size, i);
+
+      // add new table and make sure that expected element is in table
+      auto slot = tested.InsertRandomTuple(txn, &generator_, &buffer_pool_);
+      auto it = tested.GetTable().begin();
+      for (uint64_t ith = 0; it != tested.GetTable().end() && ith < size; ith++) it++;
```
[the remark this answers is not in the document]
eppingere: done

test/storage/data_table_test.cpp
```diff
@@ -376,4 +376,38 @@ TEST_F(DataTableTests, InsertNoWrap) {
...
+  const uint64_t max_tuples_inserted = 10000;
+  for (uint32_t iteration = 0; iteration < num_iterations; ++iteration) {
+    RandomDataTableTestObject tested(&block_store_, max_columns, null_ratio_(generator_), &generator_);
+    transaction::timestamp_t timestamp(0);
+    auto *txn =
+        new transaction::TransactionContext(timestamp, timestamp, common::ManagedPointer(&buffer_pool_), DISABLED);
+    uint64_t size;
+    // number of tuples in the table
+    for (uint64_t i = 0; i < max_tuples_inserted; i++) {
+      // check that table is correct size
```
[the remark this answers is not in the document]
eppingere: done

src/include/storage/data_table.h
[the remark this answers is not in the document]
jrolli: So the magic `end_` constant is a static value common to all `DataTable`s and is never updated.  I don't have an issue with that concept (it matches STL iterators in practice).   My point is that we should have a single canonical value for the tupleslot of an invalid/end iterator (matching the STL) which makes equality comparisons cheap (a single, unconditional comparison).

src/include/storage/data_table.h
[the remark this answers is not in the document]
jrolli: This is special casing only one instance of a much larger case.  I think the right way to do this would be record the insert head for the block when you first look at it and only iterate to that point since any concurrent insertions are logically invisible to this thread.  As long as we check the stored tupleslot validity on initialization and advancement and set to `end` when appropriate, then we should be able to consolidate this logic into a single location.

src/include/storage/data_table.h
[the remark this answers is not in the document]
jrolli: Just have a global constant (e.g. `SlotIterator::EndIterator`) that defines it exactly once and is referenced in the appropriate places instead.  The compiler has a much greater chance of leveraging constant propagation and inlining with that setup.

src/include/storage/data_table.h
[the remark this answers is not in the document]
jrolli: Unless we have evidence that false sharing or cache thrashing is a problem, we tend to not worry about these micro-optimizations.  I think pairs of hot, but unrelated, locks are the only place we've really seen this be a problem so far.

src/storage/data_table.cpp
eppingere: done

src/storage/data_table.cpp
eppingere: done

src/include/storage/data_table.h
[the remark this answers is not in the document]
eppingere: Im not sure I understand the difference?

src/storage/data_table.cpp
eppingere: done

src/storage/arrow_serializer.cpp
eppingere: fixed

src/include/storage/data_table.h
eppingere: done

src/include/storage/data_table.h
eppingere: done

src/include/common/container/concurrent_vector.h
eppingere: oops that is a rellic of days long past

src/include/storage/data_table.h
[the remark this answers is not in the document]
eppingere: done

src/include/storage/data_table.h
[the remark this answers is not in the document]
eppingere: this is no longer the case with our updated value

src/include/storage/data_table.h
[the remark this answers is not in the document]
eppingere: I couldnt get `SlotIterator::EndIterator` to compile because it contains the current slot tuple slot.

src/include/storage/data_table.h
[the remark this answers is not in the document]
eppingere: ok removed

src/storage/arrow_serializer.cpp
eppingere: done

src/include/storage/data_table.h
```diff
@@ -114,39 +101,46 @@ class DataTable {
...
+        }
+        slot_num_ = 0;
+        max_slot_num_ = b->GetInsertHead();
+        current_slot_ = {b, slot_num_};
+
+        if (max_slot_num_ == 0) {
+          block_index_++;
+          continue;
+        }
+        return;
```
jrolli, line 135: Nit: I would generally prefer writing the conditional so that the return is in if-block for two reasons:  1) mirrors the conditional structure of the loop, and 2) makes the loop clearer by eliding the (now) unnecessary `continue`.

src/include/storage/data_table.h
```diff
@@ -224,7 +218,10 @@ class DataTable {
    *
    * @return one past the last tuple slot contained in the data table.
    */
-  SlotIterator end() const;  // NOLINT for STL name compatibility
+
+  SlotIterator end() const {  // NOLINT for STL name compability
+    return SlotIterator();
```
jrolli, line 223: Changing this should allow us to remove the constructor without a `DataTable` pointer.
```suggestion
    return InvalidTupleSlot();
```

jrolli (changes requested):
A couple of nits, but otherwise looks good.

src/include/storage/data_table.h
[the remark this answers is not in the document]
jrolli: Thanks for making the change.  If you still want to discuss why I asked for this, hit me up on Slack and we figure out a time to chat (I think it would be easier to hash out in real time).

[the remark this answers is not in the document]
eppingere: No it makes a lot of sense. I think it might also make it faster since it makes equal comparisons much faster probably.

test/storage/data_table_test.cpp
```diff
@@ -376,4 +376,38 @@ TEST_F(DataTableTests, InsertNoWrap) {
...
+// NOLINTNEXTLINE
+TEST_F(DataTableTests, SlotIteraterSingleThreadedTest) {
+  const uint32_t num_iterations = 10;
+  const uint16_t max_columns = 10;
+  const uint64_t max_tuples_inserted = 10000;
+  for (uint32_t iteration = 0; iteration < num_iterations; ++iteration) {
+    RandomDataTableTestObject tested(&block_store_, max_columns, null_ratio_(generator_), &generator_);
+    transaction::timestamp_t timestamp(0);
+    auto *txn =
+        new transaction::TransactionContext(timestamp, timestamp, common::ManagedPointer(&buffer_pool_), DISABLED);
```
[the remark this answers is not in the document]
eppingere: I couldnt get this to work so I left it if that's ok?

src/include/storage/data_table.h
eppingere: done

src/include/storage/data_table.h
eppingere: Im not sure this works. i dont think it type checks?

jrolli: Ok, now I get the issue (I was type confuesd...).  I'm fine with this code if you add a comment above the `SlotIterator() = default;` explaining that it is only for constructing an "end" iterator.

eppingere: done

jrolli (approved):
LGTM

test/storage/export_table_test.cpp
```diff
@@ -223,7 +223,7 @@ TEST_F(ExportTableTest, ExportDictionaryCompressedTableTest) {
   // Technically, the block above is not "in" the table, but since we don't sequential scan that does not matter
   storage::DataTable table(common::ManagedPointer<storage::BlockStore>(&block_store_), layout,
                            storage::layout_version_t(0));
-  storage::RawBlock *block = table.begin()->GetBlock();
+  storage::RawBlock *block = *table.GetBlocks().begin();
```
mbutrovich, line 226: Unintended change?

test/storage/export_table_test.cpp
```diff
@@ -301,7 +301,7 @@ TEST_F(ExportTableTest, ExportVarlenTableTest) {
...
```
mbutrovich, line 304: Unintended change?

test/include/test_util/storage_test_util.h
```diff
@@ -223,8 +223,8 @@ class StorageTestUtil {
   // performance. Returns the number of non-empty slots in the block after population.
   template <class Random>
   static uint32_t PopulateBlockRandomlyNoBookkeeping(storage::DataTable *table, storage::RawBlock *block,
-                                                     double empty_ratio, Random *const generator) {
-    const storage::BlockLayout &layout = table->GetBlockLayout();
+                                                     double empty_ratio, Random *generator) {
```
mbutrovich, line 226: Unintended change?

src/include/storage/data_table.h
```diff
@@ -122,7 +148,8 @@ class DataTable {
    * @param layout the initial layout of this DataTable. First 2 columns must be 8 bytes.
    * @param layout_version the layout version of this DataTable
    */
-  DataTable(common::ManagedPointer<BlockStore> store, const BlockLayout &layout, layout_version_t layout_version);
+  DataTable(common::ManagedPointer<BlockStore> store, const BlockLayout &layout,  // NOLINT
```
mbutrovich, line 151: Why were these needed?

mbutrovich (changes requested):
Nits.

test/storage/export_table_test.cpp
eppingere: No, in the empty table we have one block with no tuples. In the iterator now, when the `SlotIterator` is equal to `end()`, the `current_slot_` is the `InvalidTupleSlot`. So the old code will yield `nullptr` since it points to an invalid tuple slot. Getting the blocks instead returns all the blocks in the table even if some of them are currently empty. So we use that instead here.

test/storage/export_table_test.cpp
eppingere: same

src/include/storage/data_table.h
jrolli: I didn't catch this one (side effect of a change made and reverted elsewhere).  This should be reverted as well.

eppingere: 👍

eppingere: done

test/include/test_util/storage_test_util.h
eppingere: done

mbutrovich (approved):
LGTM

src/include/storage/data_table.h
eppingere: Apparently gcc doesnt like it: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-850/116/pipeline#step-231-log-214

eppingere: so i reverted it back

lmwnshn (approved):
Rubber stamping.

PR #857, opened 2020-04-08
Checkpoint Recovery

tpan496:
Current Goal: realize checkpoint recovery using a single checkpoint
Things to examine:
- We currently support recovering data tables with fixed-length columns only; we will support var-length and compressed data in the future. Let us know if we missed any other important data types.
- We are not very familiar with the catalog code yet. For recovering the catalog, it is important to know how we can properly retrieve its metadata. If you think we misused some of the functions, please let us know.
- We are currently working on supporting var-length data. We might be changing files in the `block_compactor`, which is related to GC and compression. This is a part that we are not yet familiar with as well.
- While we haven't made any changes, we might modify the indices in the future. In particular, if we do not save the indices, we would need to rebuild our index on restart. On the other hand, if we decide to save the indices, we need to inject the raw bytes back into the indices.


src/include/storage/data_table.h
```diff
@@ -120,6 +120,9 @@ class DataTable {
    */
   DataTable(common::ManagedPointer<BlockStore> store, const BlockLayout &layout, layout_version_t layout_version);
 
+  DataTable(common::ManagedPointer<BlockStore> store, const BlockLayout &layout, layout_version_t layout_version,
```
arvindsaik, line 123: Could explain why the extra constructor was introduced

src/include/storage/data_table.h
```diff
@@ -236,6 +239,8 @@ class DataTable {
   // The block compactor elides transactional protection in the gather/compression phase and
   // needs raw access to the underlying table.
   friend class BlockCompactor;
+  friend class Checkpoint;
```
arvindsaik, line 242: Comments on why this was made a friend class would be nice (just like for the BlockCompactor)

src/include/storage/recovery/recovery_manager.h
```diff
@@ -107,8 +110,11 @@ class RecoveryManager : public common::DedicatedThreadOwner {
     }
   }
 
+  void RecoverFromCheckpoint(const std::string &path, catalog::db_oid_t db);
```
vilas897, line 113: Function comment missing

src/include/storage/recovery/recovery_manager.h
```diff
@@ -91,10 +94,10 @@ class RecoveryManager : public common::DedicatedThreadOwner {
   /**
    * Starts a background recovery task. Recovery will fully recover until the log provider stops providing logs.
    */
-  void StartRecovery() {
+  void StartRecovery(bool catalog_only = false) {
```
arvindsaik, line 97: Explanation for the function argument is missing

src/include/storage/sql_table.h
```diff
@@ -49,6 +49,9 @@ class SqlTable {
    */
   SqlTable(common::ManagedPointer<BlockStore> store, const catalog::Schema &schema);
 
+  SqlTable(common::ManagedPointer<BlockStore> store, const catalog::Schema &schema,
```
vilas897, line 52: Comment explaining the new constructor is missing

src/include/storage/recovery/recovery_manager.h
```diff
@@ -159,19 +165,25 @@ class RecoveryManager : public common::DedicatedThreadOwner {
    * Recovers the databases using the provided log provider
    * @return number of committed transactions replayed
    */
-  void Recover() { RecoverFromLogs(); }
+  void Recover(bool catalog_only = false) { RecoverFromLogs(catalog_only); }
```
arvindsaik, line 168: Comments to explain function argument

src/include/storage/sql_table.h
arvindsaik, line 52: Comments for why this new constructor was introduced will be helpful

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    std::string out_file = GenFileName(db_oid, curr_table_oid);
+    std::ofstream f;
+
+    auto curr_data_table = curr_table->table_.data_table_;
+    const BlockLayout &layout = curr_data_table->GetBlockLayout();
+    const std::vector<col_id_t> varlens = layout.Varlens();
+    auto column_ids = layout.AllColumns();
+    size_t col_num = column_ids.size();
+    std::cout << "col num: " << col_num << std::endl;
```
arvindsaik, line 57: Probably want to remove print statements?

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,144 @@
...
+      ArrowBlockMetadata &metadata = curr_data_table->accessor_.GetArrowBlockMetadata(blocks.front());
+      // count dict_col num
+      for (auto i = 0u; i < col_num; i++) {
+        col_id_t col_id = column_ids[i];
+        ArrowColumnInfo &col_info = metadata.GetColumnInfo(layout, col_id);
+        if (col_info.Type() == ArrowColumnType::DICTIONARY_COMPRESSED) {
+          dict_col_num += 1;
+        }
+      }
+      std::cout << "dict col num: " << dict_col_num <<std::endl;
```
arvindsaik, line 83: print statement should probably be removed.

src/storage/data_table.cpp
```diff
@@ -24,6 +24,58 @@ DataTable::DataTable(const common::ManagedPointer<BlockStore> store, const Block
   insertion_head_ = blocks_.begin();
 }
 
+DataTable::DataTable(const common::ManagedPointer<BlockStore> store, const BlockLayout &layout,
```
arvindsaik, line 27: Again, comments for the purpose of this constructor would be nice

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,144 @@
...
+    curr_data_table->blocks_latch_.Unlock();
+    f.open(path + out_file, std::ios::binary);
+
+    // record # of blocks in block list, # of varlen col and # of dict col in the table
+    if (f.is_open()) {
+      unsigned long block_num = blocks.size();
+      unsigned long var_col_num = varlens.size();
+      unsigned long dict_col_num = 0;
+      unsigned long insertion_head_index = std::distance(curr_data_table->blocks_.begin(), curr_data_table->insertion_head_);
+      std::cout << "block num: " << block_num << std::endl;
```
vilas897, line 70: Maybe the debug logging statements used in terrier could replace all the std::cout statements?

src/storage/recovery/recovery_manager.cpp
```diff
@@ -75,7 +76,113 @@ void RecoveryManager::RecoverFromLogs() {
...
+    // Now we have all the blocks and metadata, so we can recover the table
+    // Create the new table
+    auto sql_table = accessor->GetTable(table_oid);
+    sql_table->table_.data_table_->blocks_ = blocks;
+    for (auto i = 0; i < insertion_head_index; i++) {
+      sql_table->table_.data_table_->insertion_head_++;
+    }
+  }
+  txn_manager_->Commit(recovery_txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+  std::cout << "ok" << std::endl;
```
arvindsaik, line 182: Should probably not be using cout

src/storage/recovery/recovery_manager.cpp
```diff
@@ -75,7 +76,113 @@ void RecoveryManager::RecoverFromLogs() {
...
+  // Get all table oids
+  auto table_oids = accessor->GetAllTableOids();
+  for (auto file_entry : std::filesystem::directory_iterator(path)) {
+    // Find the table
+    catalog::table_oid_t table_oid;
+    std::string in_file = file_entry.path();
+    Checkpoint::GenOidFromFileName(in_file.substr(path.length(), in_file.length()), db_oid, table_oid);
+    common::ManagedPointer<storage::SqlTable> table = accessor->GetTable(table_oid);
+    auto &data_table = table->table_.data_table_;
+    // const auto &layout = data_table->GetBlockLayout();
```
vilas897, line 93: Deadcode?

src/storage/recovery/recovery_manager.cpp
```diff
@@ -75,7 +76,113 @@ void RecoveryManager::RecoverFromLogs() {
...
+       */
+      RawBlock *block = new RawBlock();
+      blocks.push_back(block);
+
+      // Set datatable
+      block->data_table_ = data_table;
+
+      // Set padding
+      uint16_t padding;
+//      f.seekg(sizeof(void *), std::ios::cur);  // skip data table
```
vilas897, line 155: Dead code

src/storage/recovery/recovery_manager.cpp
```diff
@@ -103,6 +224,14 @@ void RecoveryManager::ProcessCommittedTransaction(terrier::transaction::timestam
   txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
 }
 
+bool RecoveryManager::IsCatalogTable(catalog::table_oid_t table_oid) {
```
arvindsaik, line 227: Comments for this function would really help.

src/storage/sql_table.cpp
```diff
@@ -57,6 +57,54 @@ SqlTable::SqlTable(const common::ManagedPointer<BlockStore> store, const catalog
...
+        col_oid_to_id[column.Oid()] = col_id_t(offsets[2]++);
+        break;
+      case 2:
+        col_oid_to_id[column.Oid()] = col_id_t(offsets[3]++);
+        break;
+      case 1:
+        col_oid_to_id[column.Oid()] = col_id_t(offsets[4]++);
+        break;
+      default:
+        throw std::runtime_error("unexpected switch case value");
```
arvindsaik, line 100: Can give more context in the message printed to make error tracking easier

src/storage/data_table.cpp
```diff
@@ -24,6 +24,58 @@ DataTable::DataTable(const common::ManagedPointer<BlockStore> store, const Block
...
+    : block_store_(store), layout_version_(layout_version), accessor_(layout) {
+  TERRIER_ASSERT(layout.AttrSize(VERSION_POINTER_COLUMN_ID) == 8,
+                 "First column must have size 8 for the version chain.");
+  TERRIER_ASSERT(layout.NumColumns() > NUM_RESERVED_COLUMNS,
+                 "First column is reserved for version info, second column is reserved for logical delete.");
+  // Copy all the blocks
+  for (auto &block : blocks) {
+    blocks_.push_back(block);
+  }
+  insertion_head_ = blocks_.begin();
```
jrolli, line 38: This needs significantly more documentation about necessity and performance as it deep copies an entire `DataTable` object which could potentially be huge (10 GiB+ in a production system) and run the risk of exhausting memory.

What is the use case for this constructor and is it really the correct choice?

src/storage/recovery/recovery_manager.cpp
```diff
@@ -103,6 +224,14 @@ void RecoveryManager::ProcessCommittedTransaction(terrier::transaction::timestam
...
+  return !table_oid == !catalog::postgres::DATABASE_TABLE_OID ||
+         !table_oid == !catalog::postgres::NAMESPACE_TABLE_OID || !table_oid == !catalog::postgres::CLASS_TABLE_OID ||
+         !table_oid == !catalog::postgres::COLUMN_TABLE_OID || !table_oid == !catalog::postgres::CONSTRAINT_TABLE_OID ||
+         !table_oid == !catalog::postgres::INDEX_TABLE_OID || !table_oid == !catalog::postgres::TYPE_TABLE_OID ||
+         !table_oid == !catalog::postgres::LANGUAGE_TABLE_OID || !table_oid == !catalog::postgres::PRO_TABLE_OID;
```
jrolli, line 232: There is a *much* easier check for this.  All catalog OIDs are below `START_OID` and all user-defined OIDs are greater than or equal to `START_OID`.

src/storage/recovery/recovery_manager.cpp
jrolli: If this function is needed, it belongs under the `catalog` namespace and not under `RecoveryManager`.

src/storage/sql_table.cpp
```diff
@@ -57,6 +57,54 @@ SqlTable::SqlTable(const common::ManagedPointer<BlockStore> store, const catalog
   table_ = {new DataTable(block_store_, layout, layout_version_t(0)), layout, col_oid_to_id};
 }
 
+SqlTable::SqlTable(const common::ManagedPointer<BlockStore> store, const catalog::Schema &schema,
+                   const std::list<RawBlock *> &blocks)
```
jrolli, line 61: Two things: (1) this is 90% a copy paste job from the existing constructor, and (2) this is essentially a deep-copy constructor for a `SqlTable` object.  For (1), it would be more readable and more maintainable to have factor the common code of the two constructors into a single function (or have one constructor call the other).  For (2), I highly doubt that a deep-copy is appropriate.

README.md
```diff
@@ -1,32 +1,65 @@
...
 
-   Yes, our goal is to have this new system to support autonomous operation and optimization as a first-class design principle.
 
-5. **Will the new system still be PostgreSQL compatiable?**
-   
-   Yes. The DBMS supports the PostgresSQL network protocol (both simple and extended) and emulates PostgresSQL's catalog layout. We are working on support for PL/pgSQL UDFs.
+## Design Rationale
+On a high level, we seperate the recovery of catalog tables from recovery from data tables. For catalog information and catalog table data and indices, we make use of the existing write-ahead logs to recover those portions. For all other user created data tables, we recover the content of the data tables with our checkpoint mechanism. This approach simplifies the implementation of checkpointing to stay in the scope of user-defined data-table. It also provides us with the assumption that catalog tables are fully recovered before we recover from the checkpoint. We believe this approach will still take the advantage of using checkpoint, since in a larger database the majority of data that requires recovery will lie within user defined data tables.
```
jrolli, line 41: I think you have a serious flaw in the value of a checkpoint that excludes the catalog.

Currently, both DDL and DML are logged in a single stream.  This means in order to reconstruct the catalog you have to scan the entire log (with disk IO being the expensive part of log replay).  Additionally, your assumption further assumes that no DDL has occurred since the most recent checkpoint.  If it has, then either you must either retain all logs forever (defeats the main advantage of checkpointing) or risk significant data loss if a crash occurs after DDL but before completion of next checkpoint.

README.md
```diff
@@ -1,32 +1,65 @@
...
+
+Another major design decision we made is to deploy our own file format instead of using the existing arrow serializer. The reasoning is that we believe the current arrow format includes a lot of unnecessary information not needed for recovery. As we already performed all catalog recovery through the existing recovery from logging, recovery of the data tables’ content will require very little metadata about table columns. Therefore, we have created our own file format for this part.
+
+Finally, we design the recovery of table index to be performed in a tuple by tuple fashion. After we recover the table content, we iterate over each tuple available in the data table, and update the corresponding table index as an insert operation. We believe this approach makes full use of the efficiency of our checkpoints by eliminating the cost of performing deletes on indexes, and at the same time minimizes the design cost of hard coding updates into table indexes.
+
+## Testing Plan
```
jrolli, line 47: Missing tests:
- Detecting a failed/incomplete checkpoint and reverting to logs and/or previous checkpoint.
- Recovery when there is DML in the log after the most recent checkpoint (appears to be in-scope for your project)
- Recovery when there is DDL and DML in the log after the most recent checkpoitn (appears to be out-of-scope for your project, but should probably be at least documented as an important, but missing, test).

README.md
```diff
@@ -1,32 +1,65 @@
-# terrier
+# Checkpoint Recovery
```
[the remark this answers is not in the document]
apavlo: @abhijithanilkumar is correct. Please fix.

src/catalog/catalog_accessor.cpp
```diff
@@ -62,15 +62,19 @@ table_oid_t CatalogAccessor::GetTableOid(namespace_oid_t ns, std::string name) c
 
 table_oid_t CatalogAccessor::CreateTable(namespace_oid_t ns, std::string name, const Schema &schema) const {
   NormalizeObjectName(&name);
-  return dbc_->CreateTable(txn_, ns, name, schema);
+  auto oid = dbc_->CreateTable(txn_, ns, name, schema);
```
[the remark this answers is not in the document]
apavlo: @DUKaige why did you change this? This increases the footprint of your branch and can lead to additional conflicts.

src/include/storage/checkpoints/checkpoint.h
```diff
@@ -0,0 +1,96 @@
...
+#include "catalog/postgres/pg_attribute.h"
+#include "catalog/postgres/pg_constraint.h"
+#include "catalog/postgres/pg_database.h"
+#include "catalog/postgres/pg_index.h"
+#include "storage/sql_table.h"
+#include "transaction/transaction_manager.h"
+
+namespace terrier::storage {
+
+class Checkpoint {
```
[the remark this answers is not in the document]
apavlo: Absolutely. We can't introduce a new class without documentation. Did Doxygen not throw an error for you? It should have caught this.

src/storage/checkpoints/checkpoint.cpp
apavlo: Yes! Our code check tools should have thrown an error when if you have `std::cout`. Did it not catch this for you?

src/include/storage/recovery/recovery_manager.h
apavlo: Actually, there should be **no** default argument values in any function. Did clang not catch this and throw an error?

src/include/storage/recovery/recovery_manager.h
apavlo: See comment from above. No default argument values in any function

src/include/storage/checkpoints/checkpoint.h
```diff
@@ -0,0 +1,96 @@
...
+    tb_oid = (catalog::table_oid_t)std::stoi(file_name.substr(sep_ind + 1, file_name.length()));
+  }
+
+  static const std::vector<std::string> StringSplit(const std::string &s, const char &c) {
+    std::string buff{""};
+    std::vector<std::string> v;
+
+    for (auto n : s) {
+      if (n != c)
+        buff += n;
```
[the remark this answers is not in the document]
apavlo: Clang should have caught this and thrown an error. Did it not work?

src/storage/block_compactor.cpp
```diff
@@ -35,8 +35,8 @@ void BlockCompactor::ProcessCompactionQueue(transaction::DeferredActionManager *
           // If no compaction was performed, we still need to shut out any potentially racey transactions that
           // are alive at the same time as us flipping the block status flag to cooling. However, we must manually
           // ask the GC to enqueue this block, because no access will be observed from the empty compaction transaction.
-          if (cg.txn_->IsReadOnly())
-            deferred_action_manager->RegisterDeferredAction([this, block]() { PutInQueue(block); });
+          // if (cg.txn_->IsReadOnly())
+          //   deferred_action_manager->RegisterDeferredAction([this, block]() { PutInQueue(block); });
```
pdaviso, line 39: I don't think commenting these lines out is a long-term solution to being able to use the block compaction for the checkpointing process in addition to being able to use it for the compaction process.

src/storage/block_compactor.cpp
```diff
@@ -78,8 +78,8 @@ bool BlockCompactor::EliminateGaps(CompactionGroup *cg) {
   for (auto &entry : cg->blocks_to_compact_) {
     RawBlock *block = entry.first;
     std::vector<uint32_t> &empty_slots = entry.second;
-    TERRIER_ASSERT(block->GetInsertHead() == layout.NumSlots(),
-                   "The block should be full to stop inserts from coming in");
+    //    TERRIER_ASSERT(block->GetInsertHead() == layout.NumSlots(),
```
pdaviso, line 81: Same comment as above: I don't think commenting these lines out is a long-term solution to being able to use the block compaction for the checkpointing process in addition to being able to use it for the compaction process.

src/include/storage/checkpoints/checkpoint_thread.h
```diff
@@ -0,0 +1,57 @@
+#pragma once
+
```
pdaviso, line 2: I'm curious if there is a thread (or maybe it is called a task?) that is used by the general recovery/log manager that could be used for checkpoints, since it seems that checkpointing would be a fluid part of the recovery process, and  not in temporal conflict with logs being written to disk.

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,156 @@
...
+    // copy contents of table to disk
+    WriteToDisk(path, accessor, db);
+  };
+  for (auto i = 0u; i < num_threads; i++) {
+    thread_pool_->SubmitTask([i, &workload] { workload(i); });
+  }
+  thread_pool_->WaitUntilAllFinished();
+  txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+  if (queue.size() > 0) {
+    // the table oid that failed to be backup
```
pdaviso, line 49: I don't understand this comment because usually I would take it to understand that you are returning the table oid that failed to be backed up, but here you are returning true/false. So do you mean a table oid failed to be backed up? (not a specific table?)

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,156 @@
...
+    if (queue.size() <= 0) {
+      queue_latch.unlock();
+      return;
+    }
+    catalog::table_oid_t curr_table_oid = queue[0].first;
+    storage::DataTable *data_table = queue[0].second;
+    queue.erase(queue.begin());
+    queue_latch.unlock();
+
+    // get daga table
```
pdaviso, line 68: daga -> data. This section (next 15-ish lines) could benefit from more comments

src/storage/data_table.cpp
```diff
@@ -169,14 +169,27 @@ TupleSlot DataTable::Insert(const common::ManagedPointer<transaction::Transactio
...
+
+  InsertInto(txn, redo, result);
+
+  data_table_counter_.IncrementNumInsert(1);
+  return result;
+}
+
+void DataTable::UpdateInsertionHead(TupleSlot *result) {
   auto block = insertion_head_;
+  if (!result) block = blocks_.begin();
```
pdaviso, line 183: Add braces {} around the inside of the if statement

src/storage/data_table.cpp
```diff
@@ -187,9 +200,14 @@ TupleSlot DataTable::Insert(const common::ManagedPointer<transaction::Transactio
 
     if (accessor_.SetBlockBusyStatus(*block)) {
       // No one is inserting into this block
-      if (accessor_.Allocate(*block, &result)) {
+      if (result && accessor_.Allocate(*block, result)) {
```
pdaviso, line 203: Why this change? Not clear to me because of no change in commenting

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,156 @@
...
+        if ((uint32_t)(record_body->GetTableOid()) < catalog::START_OID) {
+          outfile.write(reinterpret_cast<char *>(buf), size);
+//          o << "REDO " << record_body->GetDatabaseOid() << ' ' << log_record->Size()<<' '<<record_body->GetTableOid()<<std::endl;
+        }
+      } else if (log_record->RecordType() == LogRecordType::DELETE) {
+        auto *record_body = log_record->GetUnderlyingRecordBodyAs<DeleteRecord>();
+        infile.read(reinterpret_cast<char *>(buf), size);
+        if ((uint32_t)(record_body->GetTableOid()) < catalog::START_OID) {
+          outfile.write(reinterpret_cast<char *>(buf), size);
+//          o << "DELETE " << record_body->GetDatabaseOid() << ' ' << record_body->GetTableOid() <<' '<< log_record->Size()<<std::endl;
```
arvindsaik, line 143: Dead code

src/storage/recovery/recovery_manager.cpp
```diff
@@ -152,7 +288,7 @@ void RecoveryManager::ReplayRedoRecord(transaction::TransactionContext *txn, Log
     // Insert will always succeed
     auto new_tuple_slot = sql_table_ptr->Insert(common::ManagedPointer(txn), staged_record);
     UpdateIndexesOnTable(txn, staged_record->GetDatabaseOid(), staged_record->GetTableOid(), sql_table_ptr,
-                         new_tuple_slot, staged_record->Delta(), true /* insert */);
+                         new_tuple_slot, staged_record->Delta(), true /* inser */);
```
arvindsaik, line 291: Seems like the t in "insert" was removed by mistake

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+void Checkpoint::WriteToDisk(const std::string &path, const std::unique_ptr<catalog::CatalogAccessor> &accessor,
+                             catalog::db_oid_t db_oid) {
+  while (queue.size() > 0) {
+    queue_latch.lock();
+    if (queue.size() <= 0) {
+      queue_latch.unlock();
+      return;
+    }
+    catalog::table_oid_t curr_table_oid = queue[0];
```
vilas897: While pulling a table oid to checkpoint from a queue whenever a thread is free, is certainly a good approach, I feel a better approach (if you are sticking with a thread handling writetodisk for multiple tables) is to sort the list of tables according to the size and then partitioning them for each thread. This way, you also avoid having a mutex for the queue as the partitioning is already done for each thread. I do realize there's a chance that a table oid is pushed back to the queue, but to handle that you can have multiple iterations of partitioning and checkpointing. 

A simple example of how this strategy can be beneficial: 
Say there are 4 tables, table sizes are:  1G, 1G, 1G, 10G. Let's say time taken to write 1G = 10s. 
Say you had 3 threads/cores. According to the current logic, time taken to write to disk = 110s (the last thread handles 11G). According to the partitioning logic, the first thread handles 10G, second thread handles 2G, third one handles 1G. Time taken = 100s. 

I'm not very sure what exactly is the design that you would go with in future, but this is just a thought that you may want to consider.

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,144 @@
...
+    // traverse each block in the block list
+    for (RawBlock *block : blocks) {
+      ArrowBlockMetadata &metadata = curr_data_table->accessor_.GetArrowBlockMetadata(block);
+
+      // first pass: record var len data
+      for (auto i = 0u; i < col_num; i++) {
+        col_id_t col_id = column_ids[i];
+        ArrowColumnInfo &col_info = metadata.GetColumnInfo(layout, col_id);
+        if (layout.IsVarlen(col_id) && col_info.Type() == ArrowColumnType::GATHERED_VARLEN) {
+          std::cout << "varchar" << std::endl;
```
vilas897: Comment above. Might want to check the entire code for print statements before final code drop.

src/storage/recovery/recovery_manager.cpp
vilas897: Still needs to be addressed.

src/include/storage/checkpoints/checkpoint.h
```diff
@@ -0,0 +1,96 @@
...
+  // Catalog to fetch table pointers
+  const common::ManagedPointer<catalog::Catalog> catalog_;
+  //  const common::ManagedPointer<BlockStore> block_store_;
+  const common::ManagedPointer<transaction::TransactionContext> txn_;
+  std::unordered_map<catalog::table_oid_t, catalog::Schema> catalog_table_schemas_;
+  std::vector<catalog::table_oid_t> queue;  // for multithreading
+  std::mutex queue_latch;
+
+  /**
+   * Write the data of a database to disk in parallel, called by TakeCheckpoint()
```
vilas897, line 113: A bit more information on what you mean by "parallel" might help. Like a general comment that describes how the work is divided among threads because it's something that future developers may want to have a look at and change. The comment may belong either here or in the TakeCheckpoint code.

README.md
```diff
@@ -1,32 +1,65 @@
...
+
+Finally, current checkpoints record the content field of each block in a data table for the convenience of software engineering, which includes repeated block headers. If given many giant tables, the redundant headers might consume relatively large space on the disk. 
+
+
+## Future Work
+As mentioned in the previous section, the current design of checkpoints taking creates one file for each table with the filename db_oid-table_oid. Future work can modify the write_to_disk function in current checkpoint class to record all the tables in one large file, which ensures the order correctness of the tables to be recovered. Also, in the current design of checkpoints taking, each table is processed by one thread. People can consider a better multithreading plan such as allowing threads which finish early to process a part of work from those process large tables to reduce the total time for taking checkpoints.
```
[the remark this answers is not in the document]
vilas897: Have added comments with some more suggestions along the lines of what @pdaviso mentioned above for the checkpoint.cpp file. 

Also, this design doc needs to be updated because I don't think a thread handles a single table as of now.

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,156 @@
...
+  auto workload = [&](uint32_t worker_id) {
+    // copy contents of table to disk
+    WriteToDisk(path, accessor, db);
+  };
+  for (auto i = 0u; i < num_threads; i++) {
+    thread_pool_->SubmitTask([i, &workload] { workload(i); });
+  }
+  thread_pool_->WaitUntilAllFinished();
+  txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+  if (queue.size() > 0) {
```
abhijithanilkumar, line 48: Maybe use `empty()` function?

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,156 @@
...
+    // the table oid that failed to be backup
+    return false;
+  }
+  return true;
+}
+
+void Checkpoint::WriteToDisk(const std::string &path, const std::unique_ptr<catalog::CatalogAccessor> &accessor,
+                             catalog::db_oid_t db_oid) {
+  while (queue.size() > 0) {
```
abhijithanilkumar, line 57: Same comment as above, use `empty`?

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,156 @@
...
+    auto cur_pos = log_provider.in_.read_head_;
+//    record_size = log_record->Size();
+    uint32_t size = cur_pos - pos;
+    if (cur_pos < pos) size = (cur_pos + common::Constants::LOG_BUFFER_SIZE) - pos;
+    auto buf = new char[size];
+
+    if (log_record->RecordType() == LogRecordType::ABORT || log_record->RecordType() == LogRecordType::COMMIT) {
+      infile.read(reinterpret_cast<char *>(buf), size);
+      outfile.write(reinterpret_cast<char *>(buf), size);
+//      o << "A/C " << log_record->Size()<<std::endl;
```
abhijithanilkumar, line 129: Dead code?

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,156 @@
...
+    } else {
+      if (log_record->RecordType() == LogRecordType::REDO) {
+        auto *record_body = log_record->GetUnderlyingRecordBodyAs<RedoRecord>();
+        infile.read(reinterpret_cast<char *>(buf), size);
+        if ((uint32_t)(record_body->GetTableOid()) < catalog::START_OID) {
+          outfile.write(reinterpret_cast<char *>(buf), size);
+//          o << "REDO " << record_body->GetDatabaseOid() << ' ' << log_record->Size()<<' '<<record_body->GetTableOid()<<std::endl;
```
abhijithanilkumar, line 136: Dead code?

src/storage/checkpoints/checkpoint.cpp
```diff
@@ -0,0 +1,156 @@
...
+//          o << "DELETE " << record_body->GetDatabaseOid() << ' ' << record_body->GetTableOid() <<' '<< log_record->Size()<<std::endl;
+        }
+      }
+    }
+    delete[] buf;
+  }
+
+  infile.close();
+  outfile.close();
+//  o.close();
```
abhijithanilkumar, line 152: Dead code?

src/include/storage/checkpoints/checkpoint.h
```diff
@@ -0,0 +1,126 @@
...
+   * Splits a string into a vector
+   * @param s to split
+   * @param delimiter that separates the string
+   * @return vector of strings
+   */
+  static const std::vector<std::string> StringSplit(const std::string &s, const char &delimiter) {
+    std::string buff{""};
+    std::vector<std::string> str_vec;
+
+    for (auto c : s) {
```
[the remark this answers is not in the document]
DUKaige: Thanks for the suggestion! The find function is less efficient than a for loop in that it requires multiple iterations of the same string, while the for loop is one pass.

src/include/storage/checkpoints/checkpoint.h
```diff
@@ -0,0 +1,96 @@
...
+  bool TakeCheckpoint(const std::string &path, catalog::db_oid_t db);
+
+  /**
+   * Generate a file name for a table
+   * @tparam db_oid the oid of the database the table belongs to
+   * @param tb_oid the oid of the table
+   * @return a file name in the format db_oid-tb_oid.txt
+   */
+  static std::string GenFileName(catalog::db_oid_t db_oid, catalog::table_oid_t tb_oid) {
+    return std::to_string((uint32_t)db_oid) + "-" + std::to_string((uint32_t)tb_oid);
```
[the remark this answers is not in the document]
DUKaige: Thanks for the suggestion, but we don't think that's necessary.

src/include/storage/checkpoints/checkpoint_thread.h
DUKaige: We have changed the CheckpointThread class to use an existing thread pool

apavlo:

PR #860, opened 2020-04-09
Multi-Threaded Queries

chufu2013:
The current terrier system’s execution engine only supports single-threaded versions of operators like sequential scan, hash join, aggregate, etc. The project works on providing multi-threaded functionalities starting with the sequential scan. After integrating the project, the execution engine will be able to generate tpl codes which invokes multiple threads to concurrently produce outputs for the upper-level operator (output operator in this case).

This project can be mainly divided into two components:

- Parallel Scan:
The `ParallelScan `function inside `TableVectorIterator` will invoke tbb to create and distribute work to multiple threads. Each worker thread have its own `TableVectorIterator` and buffer to perform the scan on a partition of the table.

- Codegen:
We followed along the whole codegen and compilation process of a sequential scan query. To get to the goal of building an executable_query, the system lets the compiler produce the pipeline function, generate the main tpl function, and then compile the bytecode. 

Notice that currently we’re just modifying other operators in order to comply with the function signature in the parent class. The only operator we’re working on is `seq_scan_translator`.


src/execution/sema/sema_builtin.cpp
```diff
@@ -1094,47 +1094,55 @@ void Sema::CheckBuiltinTableIterCall(ast::CallExpr *call, ast::Builtin builtin)
...
-  // Third argument is the thread state container
-  const auto tls_kind = ast::BuiltinType::ThreadStateContainer;
-  if (!IsPointerToSpecificBuiltin(call_args[2]->GetType(), tls_kind)) {
-    ReportIncorrectCallArg(call, 2, GetBuiltinType(tls_kind)->PointerTo());
-    return;
-  }
+  // TODO(Ron): Update type checks here when the argument types are finalized
+  //  // Second argument is an opaque query state. For now, check it's a pointer.
+  //  const auto void_kind = ast::BuiltinType::Nil;
+  //  if (!call_args[1]->GetType()->IsPointerType()) {
```
thepinetree, line 1112: It looks like this definition of the arguments is not being used, in which case, it would be cleaner to delete it

src/execution/sema/sema_builtin.cpp
```diff
@@ -1094,47 +1094,55 @@ void Sema::CheckBuiltinTableIterCall(ast::CallExpr *call, ast::Builtin builtin)
...
+  //  }
+  //
+  //  // Third argument is the thread state container
+  //  const auto tls_kind = ast::BuiltinType::ThreadStateContainer;
+  //  if (!IsPointerToSpecificBuiltin(call_args[2]->GetType(), tls_kind)) {
+  //    ReportIncorrectCallArg(call, 2, GetBuiltinType(tls_kind)->PointerTo());
+  //    return;
+  //  }
 
   // Third argument is scanner function
```
thepinetree, line 1124: Now the second argument

src/execution/sema/sema_builtin.cpp
```diff
@@ -1094,47 +1094,55 @@ void Sema::CheckBuiltinTableIterCall(ast::CallExpr *call, ast::Builtin builtin)
...
-  }
+  // TODO(Ron): Update type checks here when the argument types are finalized
+  //  // Second argument is an opaque query state. For now, check it's a pointer.
+  //  const auto void_kind = ast::BuiltinType::Nil;
+  //  if (!call_args[1]->GetType()->IsPointerType()) {
+  //    ReportIncorrectCallArg(call, 1, GetBuiltinType(void_kind)->PointerTo());
+  //    return;
+  //  }
+  //
+  //  // Third argument is the thread state container
```
thepinetree, line 1117: It looks like this definition of the arguments is not being used, in which case, it would be cleaner to delete it

src/execution/sema/sema_builtin.cpp
```diff
@@ -1094,47 +1094,55 @@ void Sema::CheckBuiltinTableIterCall(ast::CallExpr *call, ast::Builtin builtin)
...
-    GetErrorReporter()->Report(call->Position(), ErrorMessages::kBadParallelScanFunction, call_args[3]->GetType());
+
+  auto exec_ctx_kind = ast::BuiltinType::ExecutionContext;
+  if (!IsPointerToSpecificBuiltin(call_args[2]->GetType(), exec_ctx_kind)) {
+    ReportIncorrectCallArg(call, 2, GetBuiltinType(exec_ctx_kind)->PointerTo());
     return;
   }
 
+  // Check type
+  //  const auto tvi_kind = ast::BuiltinType::TableVectorIterator;
```
thepinetree, line 1138: It might be a good idea to preserve this check for the scan_fn_type to check the parameters of the function call are of appropriate types

src/execution/vm/vm.cpp
```diff
@@ -577,14 +577,14 @@ void VM::Interpret(const uint8_t *ip, Frame *frame) {
   }
 
   OP(ParallelScanTable) : {
-    auto db_oid = READ_UIMM4();
+    // TODO(Yuhong): might need to change function arguments
     auto table_oid = READ_UIMM4();
-    auto query_state = frame->LocalAt<void *>(READ_LOCAL_ID());
-    auto thread_state_container = frame->LocalAt<sql::ThreadStateContainer *>(READ_LOCAL_ID());
+    //    auto query_state = frame->LocalAt<void *>(READ_LOCAL_ID());
```
thepinetree, line 582: It looks like these can now be removed safely based on the arguments in place

src/execution/sql/table_vector_iterator.cpp
```diff
@@ -23,41 +31,99 @@ bool TableVectorIterator::Init() {
   // Find the table
   table_ = exec_ctx_->GetAccessor()->GetTable(table_oid_);
   TERRIER_ASSERT(table_ != nullptr, "Table must exist!!");
-
+  if (col_oids_.empty()) {
+    table_->GetAllColOid(&col_oids_);
+  }
   // Initialize the projected column
-  TERRIER_ASSERT(!col_oids_.empty(), "There must be at least one col oid!");
+  // TERRIER_ASSERT(!col_oids_.empty(), "There must be at least one col oid!");
```
thepinetree, line 38: Does this assert fail?

src/execution/sql/table_vector_iterator.cpp
```diff
@@ -23,41 +31,99 @@ bool TableVectorIterator::Init() {
...
 
-  // Begin iterating
-  iter_ = std::make_unique<storage::DataTable::SlotIterator>(table_->begin());
+  // Create the start and end iterator
+  iter_ = std::make_unique<storage::DataTable::SlotIterator>(table_->beginAt(start_block_idx_));
+  iter_end_ = std::make_unique<storage::DataTable::SlotIterator>(table_->endAt(end_block_idx_));
   return true;
 }
 
 bool TableVectorIterator::Advance() {
```
thepinetree, line 50: This function now seems more like an `AdvanceToEnd` function

src/execution/sql/table_vector_iterator.cpp
```diff
@@ -23,41 +31,99 @@ bool TableVectorIterator::Init() {
...
+  common::ManagedPointer<storage::SqlTable> table =
+      exec_ctx->GetAccessor()->GetTable(static_cast<catalog::table_oid_t>(table_oid));
+  if (table == nullptr) {
+    return false;
+  }
+  // get the number of blocks in the table
+  auto block_count = table->GetBlockListSize();
+
+  // TODO(Ron): min_grain_size = num_blocks / num_threads
+  size_t min_grain_size = 3;
```
thepinetree, line 117: Running with different grain sizes may give more insight in tests, as well

test/execution/compiler_test.cpp
```diff
@@ -68,6 +68,68 @@ class CompilerTest : public SqlBasedTest {
   static constexpr vm::ExecutionMode MODE = vm::ExecutionMode::Interpret;
 };
 
+// NOLINTNEXTLINE
+// TEST_F(CompilerTest, SimplerSeqScanTest) {
```
thepinetree, line 72: Is there a reason this test is commented out?

YuhongZhang98: We originally wanted to test the implementation without thread local state argument but got some problem with the test case. We'll delete this part in the next commit

src/include/execution/compiler/pipeline.h
```diff
@@ -40,6 +41,35 @@ class Pipeline {
...
+      result += "_" + func_name;
+    }
+    return result;
+  }
+
+  /**
+   * Get the identifier of the pipeline function
+   * @return the identifier made from the unique pipeline function name
+   */
+  ast::Identifier GetWorkFunctionName() const {
```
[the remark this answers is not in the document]
apavlo: We should use the naming convention that Prashanth uses in his branch.

src/execution/sql/table_vector_iterator.cpp
apavlo: Yes, why was this assert disabled? We don't want to do this carelessly.

src/execution/sql/table_vector_iterator.cpp
apavlo: No magic constants!

test/execution/compiler_test.cpp
apavlo: @YuhongZhang98 Is this test case no longer valid?

apavlo:
Please update the PR description to explain what is in this branch instead of just listing the files that you modified (this info is redundant as Github already lists the modified files).

src/execution/sql/table_vector_iterator.cpp
chufu2013: This assertion should never fail because even if the column id vector is empty, it will process all columns instead. But I guess there's no harm to do a double-check here. I'll uncomment it later.

src/execution/sql/table_vector_iterator.cpp
chufu2013: > This function now seems more like an `AdvanceToEnd` function
The advance() function is still possible to be executed multiple times if the buffer is full during the previous round of scan. So it might not reach the end in this case.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -637,7 +637,23 @@ void BytecodeGenerator::VisitBuiltinTableIterCall(ast::CallExpr *call, ast::Buil
...
+
+  // Then the scan function as an identifier
+  const auto scan_fn_name = call->Arguments()[1]->As<ast::IdentifierExpr>()->Name().Data();
+  const FunctionId scan_fn_id = LookupFuncIdByName(scan_fn_name);
+
+  // Execution context
+  LocalVar exec_ctx = VisitExpressionForRValue(call->Arguments()[2]);
+
+  //  // Next is the thread state container
+  //  LocalVar thread_state_container = VisitExpressionForRValue(call->Arguments()[2]);
```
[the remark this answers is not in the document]
YuhongZhang98: we have fixed the arguments later on and now the comments should be correct

src/execution/compiler/compiler.cpp
```diff
@@ -176,6 +176,12 @@ ast::Decl *Compiler::GenMainFunction() {
   // Make the function
   FunctionBuilder builder{codegen_, fn_name, std::move(params), ret_type};
 
+  // TODO(Ron): Enabled this when we have isParallel global variable
+  // Step 0.0: Call Launch Work
```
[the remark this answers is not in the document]
zechengh09: We have fixed these comments in later commits.

src/execution/compiler/operator/seq_scan_translator.cpp
```diff
@@ -85,6 +91,18 @@ void SeqScanTranslator::Consume(FunctionBuilder *builder) {
   GenTVIReset(builder);
 }
 
+util::RegionVector<ast::FieldDecl *> SeqScanTranslator::GetWorkerParams() {
+  auto tvi_type = codegen_->PointerType(ast::BuiltinType::TableVectorIterator);
+  return codegen_->MakeFieldList({codegen_->MakeField(tvi_, tvi_type)});
+}
+
+void SeqScanTranslator::LaunchWork(FunctionBuilder *builder, ast::Identifier work_func) {
```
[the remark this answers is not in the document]
zechengh09: We use the same naming convention as Prashanth's branch

src/execution/compiler/operator/seq_scan_translator.cpp
```diff
@@ -27,7 +27,11 @@ SeqScanTranslator::SeqScanTranslator(const terrier::planner::SeqScanPlanNode *op
 
 void SeqScanTranslator::Produce(FunctionBuilder *builder) {
   SetOids(builder);
+  // TODO(Ron): Declare the tvi only for serial scan
```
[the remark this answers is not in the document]
zechengh09: Yes and we updated this with IsParallel condition

src/execution/sema/sema_builtin.cpp
```diff
@@ -1094,47 +1094,55 @@ void Sema::CheckBuiltinTableIterCall(ast::CallExpr *call, ast::Builtin builtin)
 }
 
 void Sema::CheckBuiltinTableIterParCall(ast::CallExpr *call) {
-  if (!CheckArgCount(call, 4)) {
+  if (!CheckArgCount(call, 3)) {
```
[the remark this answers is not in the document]
zechengh09: Good point! We'll add more comments here explaining what's going on

src/execution/sema/sema_builtin.cpp
```diff
@@ -1094,47 +1094,55 @@ void Sema::CheckBuiltinTableIterCall(ast::CallExpr *call, ast::Builtin builtin)
...
     return;
   }
 
   const auto &call_args = call->Arguments();
 
   // First argument is table name as a string literal
-  if (!call_args[0]->IsStringLiteral()) {
+  if (!call_args[0]->IsIntegerLiteral()) {
```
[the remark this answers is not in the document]
zechengh09: Good find, we did not update the comment when updating from string table name to table oid

src/execution/sema/sema_builtin.cpp
zechengh09: We reenabled it in later commits.

src/execution/sql/table_vector_iterator.cpp
zechengh09: We updated the min_grain_size - number of blocks divided by the number of threads we have

src/execution/vm/vm.cpp
YuhongZhang98: Updated

chufu2013:
> Is there a reason why `iter_` and `iter_end_` needs to be a std::unique_ptr. I know that iter_ was one, but I don't immediately see a reason why it needs to be one? I think that this could be solved by having a default value for SlotIterator?
We made iter_end_ a std::unique_ptr since in the original code iter_  was also a std::unique_ptr, although I'm not sure about why it's implemented in this way. But I guess there's no harm to use a smart pointer here?
> Can you combine `begin` and `beginAt` into one function with a default argument?
> Similarly for `end` and `endAt`? Idk if this will mess up STL iteration but I think it would clean up the code base
Because `beginAt `and `endAt `does include additional branchings that are unnecessary for the default scan but necessary for range scan, we want to keep the logic separate so the default scan can save time on it.

benchmark/storage/parallel_scan_benchmark.cpp
```diff
@@ -0,0 +1,193 @@
...
+                               nullptr,          // The thread state destruction function
+                               nullptr);         // Context passed to init/destroy functions
+  // NOLINTNEXTLINE
+  for (auto _ : state) {
+    uint64_t elapsed_ms = 0;
+    {
+      common::ScopedTimer<std::chrono::milliseconds> timer(&elapsed_ms);
+      execution::sql::TableVectorIterator::ParallelScan(static_cast<uint32_t>(table_oid),  // ID of table to scan
+                                                        col_oids,  // Query state to pass to scan threads
+                                                        2,         // Container for thread states
```
eppingere, line 95: Did the function definition change since this comment? Isn't `2` the `num_oids`?

src/execution/sql/table_vector_iterator.cpp
```diff
@@ -23,41 +31,112 @@ bool TableVectorIterator::Init() {
...
+  uint32_t *col_oids_;
+  uint32_t num_oids_;
+  void *const query_state_;
+  TableVectorIterator::ScanFn scanner_;
+};
+}  // namespace
+
+bool TableVectorIterator::ParallelScan(uint32_t table_oid, uint32_t *col_oids, uint32_t num_oids,
+                                       void *query_state, const ScanFn scan_fn, exec::ExecutionContext *exec_ctx,
+                                       unsigned int num_threads) {
```
eppingere, line 106: Is there a reason why `num_threads` is a `unsigned int` but `num_oids` is `uint32_t`?

src/include/execution/exec/output.h
```diff
@@ -45,23 +45,50 @@ class EXPORT OutputBuffer {
...
       : memory_pool_(memory_pool),
-        num_tuples_(0),
         tuple_size_(tuple_size),
-        tuples_(
-            reinterpret_cast<byte *>(memory_pool->AllocateAligned(BATCH_SIZE * tuple_size, alignof(uint64_t), true))),
-        callback_(std::move(callback)) {}
+        id_(0),
+        callback_(std::move(callback)) {
+          max_thread_ = std::thread::hardware_concurrency();
+          if (max_thread_ == 0) {
```
eppingere, line 52: I noticed that you did this in a few places. `std::thread::hardware_concurrency()` returns `0` only "If this value is not computable or well defined" whatever that means. I googled around and couldn't find any info on what would cause it to return 0. 🤷

test/execution/table_vector_iterator_test.cpp
```diff
@@ -129,4 +129,55 @@ TEST_F(TableVectorIteratorTest, IteratorColOidsTest) {
...
+  for (unsigned int i = 1; i <= num_cores; i *= 2) {
+
+    thread_state_container.Reset(sizeof(Counter),  // The type of each thread state structure
+                                init_count,       // The thread state initialization function
+                                nullptr,          // The thread state destruction function
+                                nullptr);         // Context passed to init/destroy functions
+    auto table_oid = exec_ctx_->GetAccessor()->GetTableOid(NSOid(), "test_1");
+    TableVectorIterator::ParallelScan(static_cast<uint32_t>(table_oid),  // ID of table to scan
+                                      col_oids,
+                                      1,  // Query state to pass to scan threads
```
eppingere, line 171: same as other test

src/include/execution/compiler/operator/aggregate_translator.h
```diff
@@ -42,6 +43,19 @@ class AggregateBottomTranslator : public OperatorTranslator {
   void Abort(FunctionBuilder *builder) override;
   void Consume(FunctionBuilder *builder) override;
 
+  /**
+   * @return The pipeline work function parameters.
+   */
+  util::RegionVector<ast::FieldDecl *> GetWorkerParams() override { UNREACHABLE("Not implemented yet"); }
```
eppingere, line 49: Can this be abstracted into an interface? Like the `GetWorkerParams` is by default unreachable?

src/include/execution/compiler/operator/operator_translator.h
```diff
@@ -122,6 +125,22 @@ class OperatorTranslator : public ExpressionEvaluator {
    */
   virtual ast::Expr *GetOutput(uint32_t attr_idx) = 0;
 
+  /**
+   * @return The list of extra fields added to the main work function. This is only called on
+   *         parallel pipelines and only for the source/root of the pipeline that's responsible for
+   *         launching the worker function. By default, the first two argument of a parallel work
+   *         function are the query state and thread-local pipeline state.
+   */
+  virtual util::RegionVector<ast::FieldDecl *> GetWorkerParams() = 0;
```
eppingere, line 134: yeah like here it's default is `UNREACHABLE`?

src/include/execution/compiler/operator/operator_translator.h
```diff
@@ -122,6 +125,22 @@ class OperatorTranslator : public ExpressionEvaluator {
...
+
+  /**
+   * This is called on the source/root of parallel pipelines to launch the provided worker function
+   * in parallel across a set of threads.
+   * @param function The caller function
+   * @param work_func_name The name of the work function that implements the pipeline logic.
+   */
+  virtual void LaunchWork(FunctionBuilder *function, ast::Identifier work_func_name) = 0;
```
eppingere, line 142: Same for `LaunchWork`?

src/include/execution/exec/output.h
```diff
@@ -45,23 +45,50 @@ class EXPORT OutputBuffer {
...
+          num_tuples_ = reinterpret_cast<uint32_t *>(
+          memory_pool->AllocateAligned(max_thread_ * sizeof(uint32_t), alignof(uint32_t), true));
+        }
+
+  /**
+   * Insert the corresponding buffer of a thread to the map if not created yet
+   * @param id The thread ID
+   */
+  void InsertIfAbsent(std::thread::id id) {
+    if (buffer_map_.find(id) == buffer_map_.end()) {
```
eppingere, line 65: I believe there is a race condition between here and `AllocOutputSlot` if we're running inside a thread pool or something where thread ids are reused?

benchmark/storage/parallel_scan_benchmark.cpp
```diff
@@ -0,0 +1,193 @@
...
+#include "execution/sql/value.h"
+#include "execution/vm/llvm_engine.h"
+#include "execution/vm/module.h"
+#include "planner/plannodes/seq_scan_plan_node.h"
+#include "type/transient_value.h"
+#include "type/transient_value_factory.h"
+#include "type/type_id.h"
+
+namespace terrier {
+class ParalleScanBenchmark : public benchmark::Fixture {};
```
thepinetree, line 40: It might be helpful to include a setup function that sets a bunch of class variables due to the repetition of creating the benchmark variables (db_main, catalog, txn_manager, etc.)

YuhongZhang98: We encountered a weird bug when taking the setup code out in a separate function so we jut put them in here. We'll give it another shot and will update it in the next commit.

benchmark/storage/parallel_scan_benchmark.cpp
YuhongZhang98: Nice catch! We will fix it

src/include/execution/compiler/operator/aggregate_translator.h
YuhongZhang98: We were also talking about this but since we were told that others will work on these operators during the summer, we decided to have these signatures here "to be implemented"

src/include/execution/compiler/operator/operator_translator.h
YuhongZhang98: Right this method should only be used by parallel pipelines so it's default to be unreachable for others

src/include/execution/compiler/operator/operator_translator.h
YuhongZhang98: Same reason as above 👍

src/include/execution/compiler/pipeline.h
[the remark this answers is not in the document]
YuhongZhang98: Yes we kept the naming from Prashanth's branch

src/include/execution/exec/output.h
zechengh09: Yea I'm not sure in what cases this function would return 0, but it wouldn't hurt to be extra safe.

test/execution/table_vector_iterator_test.cpp
```diff
@@ -129,4 +129,55 @@ TEST_F(TableVectorIteratorTest, IteratorColOidsTest) {
...
+      }
+    }
+  };
+
+  uint32_t col_oids[] = {1};
+  // Setup thread states
+  ThreadStateContainer thread_state_container(common::ManagedPointer<MemoryPool>(exec_ctx_->GetMemoryPool()));
+
+  unsigned int num_cores = std::thread::hardware_concurrency();
+  if (num_cores == 0) {
```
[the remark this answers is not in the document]
zechengh09: Addressed.

test/execution/table_vector_iterator_test.cpp
```diff
@@ -129,4 +129,55 @@ TEST_F(TableVectorIteratorTest, IteratorColOidsTest) {
...
+    num_cores = 1;
+  }
+
+  for (unsigned int i = 1; i <= num_cores; i *= 2) {
```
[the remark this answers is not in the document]
zechengh09: Good point! I will add comment here

test/execution/table_vector_iterator_test.cpp
```diff
@@ -129,4 +129,55 @@ TEST_F(TableVectorIteratorTest, IteratorColOidsTest) {
...
+    TableVectorIterator::ParallelScan(static_cast<uint32_t>(table_oid),  // ID of table to scan
+                                      col_oids,
+                                      1,  // Query state to pass to scan threads
+                                      thread_state_container.AccessThreadStateOfCurrentThread(),  // Thread states
+                                      scanner,                                                    // Scan function
+                                      exec_ctx_.get(), i);
+
+    // Count total aggregate tuple count seen by all threads
+    uint32_t aggregate_tuple_count = 0;
+    thread_state_container.ForEach<Counter>([&](Counter *counter) { aggregate_tuple_count += counter->c_; });
```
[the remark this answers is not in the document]
chufu2013: For this test case, there's only one counter for each thread worker.

src/include/execution/exec/output.h
zechengh09: Could you explain more about the race condition? I think threads could be reused across different output buffers, but then they would be different instances of output buffers. Buffer map itself is a concurrent map, which guarantees thread-safety for one output buffer.

apavlo:
Closing for now to reduce Jenkins contention.

PR #862, opened 2020-04-09
Schema Change

mush-zhang:
Major files we modified: 
- sql_table.h
- sql_table.cpp


githubxxcc:
Major files modified:
- sql_table_test.cpp
- storage_test_util.cpp

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### Tuple Transformation
+
+We need to translate column_ids in the storage version to column_ids of the intended version. For any column present in the storage version but not present in the intended version we set the column_id to IGNORE_COLUMN_ID, and within DataTable we will skip over any columns with IGNORE_COLUMN_ID . For any column present in the intended version but not present in the storage version, we fill in default values for columns that were not present in the storage version.
+
+## Design Rationale
+
+#### Map or vector of datatables? 
+
+Within Sqltable, we can either use a ordered concurrent map of version number to datatable, or use a vector of datatables (index i of vector corresponds to datatable with version number i) augmented with locks.  We conjecture that using a vector will be faster than a map. One potential problem with vectors is how to make it work with GC of old versions, and recovery. One potential problem with using Terrier's concurrent_map.h, which is a wrapper around tbb_concurrent_unordered_map, is that it is unordered (we need order for scan) and unsafe for concurrent deletes (useful for GC of old versions).
```
alex-stanescu, line 80: Using a concurrent vector seems to have large potential downsides wrt deletion of arbitrary elements (e.g. with GC of old versions). Something to consider is that if you expect the number of concurrently existing datatable versions to be small, then it may be worth using an unordered_map, and using a scan over the map to populate a (local) vector which can then be sorted.

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### When to perform background migration and GC of old versions?
+
+When we are sure that no transactions will ever access the schema version represented by an old datatable(inferring from low_watermark of transaction timestamps), we can safely GC the outdated datatable. However, since we do lazy migration, if most tuples in a datatable is not updated after a schema change, they will still be in the outdated datatable, and we can read and access them from the outdated datatable. In this case, it might be quite expensive to GC the outdated datatable, since we need to first migrate all its tuples to newer datatables. It seems that doing background migrations conservatively is good for read-heavy workloads, and doing background migrations aggresively might be good for update-heavy workloads interleaved with schema changes.  
+
+Our current decision is to use a threshold to decide whether to do background migration: once the amount of tuples in an outdated datatable drops below the threshold, we start a background thread to migrate its tuples and GC the datatable when we're done.
```
alex-stanescu, line 88: I am not sure a threshold is going to sufficiently capture the situations where background migrations are good. An analysis of previous accesses to this table might be better (e.g. composition of reads vs writes), but that might be out of scope for this project

src/include/storage/data_table.h
```diff
@@ -143,10 +144,13 @@ class DataTable {
   // Alternatively, we can provide an easy wrapper that takes in a const SlotIterator & and returns a SlotIterator,
   // just like the ++i and i++ dichotomy.
   /**
-   * Sequentially scans the table starting from the given iterator(inclusive) and materializes as many tuples as would
+   * Sequentially scans the table starting from the given iterator(inclusive) and materializes as many tuples as
+   would
```
alex-stanescu, line 148: Why did this get split into 2 lines?

src/include/storage/data_table.h
```diff
@@ -143,10 +144,13 @@ class DataTable {
...
    * fit into the given buffer, as visible to the transaction given, according to the format described by the given
-   * output buffer. The tuples materialized are guaranteed to be visible and valid, and the function makes best effort
-   * to fill the buffer, unless there are no more tuples. The given iterator is mutated to point to one slot passed the
+   * output buffer. The tuples materialized are guaranteed to be visible and valid, and the function makes best
+   effort
+   * to fill the buffer, unless there are no more tuples. The given iterator is mutated to point to one slot passed
+   the
```
alex-stanescu, line 153: Ditto here

src/include/storage/data_table.h
```diff
@@ -236,11 +240,19 @@ class DataTable {
...
   friend class BlockCompactor;
+  // The SqlTable needs to access the layout version
+  friend class SqlTable;
 
   const common::ManagedPointer<BlockStore> block_store_;
   const layout_version_t layout_version_;
   const TupleAccessStrategy accessor_;
 
+  // TODO(Schema-Change)
+  // Singly-linked list for tables with different version?
```
alex-stanescu, line 251: Is this something that has yet to be decided? This seems a bit odd to do both

src/include/storage/sql_table.h
```diff
@@ -74,49 +84,34 @@ class SqlTable {
...
-                   "immediately before?");
-    const auto result = table_.data_table_->Update(txn, redo->GetTupleSlot(), *(redo->Delta()));
-    if (!result) {
-      // For MVCC correctness, this txn must now abort for the GC to clean up the version chain in the DataTable
-      // correctly.
-      txn->SetMustAbort();
-    }
-    return result;
-  }
+  std::pair<bool, TupleSlot> Update(common::ManagedPointer <transaction::TransactionContext> txn, RedoRecord *const redo,
```
alex-stanescu, line 90: Why return a pair here?

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+
+We will not change the API of Datatable, and make only minor additions to it. For convenience, within each Datatable we store its corresponding version, default value map, column id to oid map, and column oid to id map (explained below). Similarly, we will not be touching Tupleslots, Rawblocks, or Index.
+
+We will also make minor changes to upper levels of Terrier, such the execution layer (e.g. storage_interface.cpp). 
+
+## Architectural Design
+
+To manage multiple datatables, each Sqltable keeps a ordered map (tables_) from version number to a struct containing the metadata of each datatable (called DataTableVersion). 
+
+Note that in the vast majority of cases, there will be no schema changes, and each Sqltable manages only one Datatable. If we need to go through tables_, which is a concurrent map (or vector), on each call to Sqltable, that will be a huge overhead. Therefore, we will cache the DataTable in the SqlTable when there is only one DataTable to bypass the map/vector lookup.
```
alex-stanescu, line 33: This appears not to be implemented. Is this a TODO or did something change

src/storage/sql_table.cpp
```diff
@@ -12,9 +11,226 @@ namespace terrier::storage {
...
+  TERRIER_ASSERT(tuple_version <= layout_version,
+                 "The iterator should not go to data tables with more recent version than the current transaction.");
+
+  bool result;
+  if (tuple_version == layout_version) {
+    result = tables_.at(layout_version).data_table_->Update(txn, curr_tuple, *(redo->Delta()));
+  } else {
+    // tuple in an older version, check if all modified columns are in the datatable version where the tuple is in
+
+    col_id_t ori_header[redo->Delta()->NumColumns()];
```
alex-stanescu, line 75: would suggest renaming this to orig_header for readability

src/storage/sql_table.cpp
```diff
@@ -12,9 +11,226 @@ namespace terrier::storage {
...
+    if (out_buffer->AccessWithNullCheck(i) == nullptr && default_val != nullptr) {
+      auto col_oid = desired_version.column_id_to_oid_map_.at(col_ids[i]);
+      // FIXME(all): proper way to use AbstractExpression
+      StorageUtil::CopyWithNullCheck(default_val.template CastManagedPointerTo<const byte>().Get(), out_buffer,
+                                     desired_version.schema_->GetColumn(col_oid).AttrSize(), i);
+    }
+  }
+}
+
+template void SqlTable::FillMissingColumns<ProjectedRow>(ProjectedRow *out_buffer,
```
alex-stanescu, line 191: Are these specializations necessary? C++ should be able to infer them

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+
+There is no *storage_version* in this case. We directly insert to the table corresponding to *intended_version*.
+
+#### Sqltable::delete
+
+Whether *storage_version < intended_version* or *storage_version = intended_version*, we always logically delete the tuple from the datatable it is currently stored in.
+
+#### Sqltable::update
+
+If *storage_version < intended_version* we will read the tuple from the old datatable, do tuple transformation to new schema and update the tuple, then logically delete it from the old datatable  (corresponding to
```
kjobanputra, line 55: Wouldn't this always be the case since the version number is incremented?

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+Whether *storage_version < intended_version* or *storage_version = intended_version*, we always logically delete the tuple from the datatable it is currently stored in.
+
+#### Sqltable::update
+
+If *storage_version < intended_version* we will read the tuple from the old datatable, do tuple transformation to new schema and update the tuple, then logically delete it from the old datatable  (corresponding to 
+*storage_version*) and then insert it to the new datatable (corresponding to *intended_version*).  We will return the old tupleslot and the new tupleslot to the execution layer, which will update the index as described above.
+
+#### Sqltable::select
+
+Since we don't do migration on reads for now, we will not migrate tuples even in the case of *storage_version < intended_version*. We will read the tuple, transform it to the new schema, select its columns and return.
```
kjobanputra, line 60: Why don't you migrate? Seems like it could be better if storage_version < intended_version for a long time.

src/storage/sql_table.cpp
```diff
@@ -12,9 +11,226 @@ namespace terrier::storage {
...
+      }
+      delete[] buffer;
+    }
+  }
+  if (!result) {
+    // For MVCC correctness, this txn must now abort for the GC to clean up the version chain in the DataTable
+    // correctly.
+    txn->SetMustAbort();
+  }
+  return std::make_pair(result, returned_slot);
```
kjobanputra, line 116: I believe that make_pair has very bad performance. Try using a struct instead.

src/include/storage/sql_table.h
```diff
@@ -74,49 +84,34 @@ class SqlTable {
...
   /**
    * Deletes the given TupleSlot. StageDelete must have been called as well in order for the operation to be logged.
    * @param txn the calling transaction
    * @param slot the slot of the tuple to delete
+   * @param layout_version Schema version the current querying transaction should see
    * @return true if successful, false otherwise
    */
-  bool Delete(const common::ManagedPointer<transaction::TransactionContext> txn, const TupleSlot slot) {
+  bool Delete(const common::ManagedPointer<transaction::TransactionContext> txn, const TupleSlot slot,
+              layout_version_t layout_version = layout_version_t{0}) {
```
kjobanputra, line 114: Why is this needed?

src/include/storage/sql_table.h
```diff
@@ -140,67 +139,127 @@ class SqlTable {
...
   }
 
+  // TODO(Schema-Change): Do we retain the begin() and end(), or implement begin and end function with version number?
+
   /**
    * @return the first tuple slot contained in the underlying DataTable
    */
-  DataTable::SlotIterator begin() const { return table_.data_table_->begin(); }  // NOLINT for STL name compability
+  // NOLINTNEXTLINE for STL name compability
+  DataTable::SlotIterator begin() const { return tables_.begin()->second.data_table_->begin(); }
```
kjobanputra, line 177: Shouldn't this be the first tuple where storage_version == intended_version?

src/storage/data_table.cpp
```diff
@@ -259,9 +262,9 @@ bool DataTable::Delete(const common::ManagedPointer<transaction::TransactionCont
 template <class RowType>
 bool DataTable::SelectIntoBuffer(const common::ManagedPointer<transaction::TransactionContext> txn,
                                  const TupleSlot slot, RowType *const out_buffer) const {
-  TERRIER_ASSERT(out_buffer->NumColumns() <= accessor_.GetBlockLayout().NumColumns() - NUM_RESERVED_COLUMNS,
-                 "The output buffer never returns the version pointer columns, so it should have "
-                 "fewer attributes.");
+  //  TERRIER_ASSERT(out_buffer->NumColumns() <= accessor_.GetBlockLayout().NumColumns() - NUM_RESERVED_COLUMNS,
```
kjobanputra, line 265: Why is this assertion no longer valid? Also, if it's dead code, remove.

src/storage/sql_table.cpp
```diff
@@ -1,20 +1,314 @@
...
+
+  // copy back the original header
+  std::memcpy(out_buffer->ColumnIds(), ori_header, sizeof(col_id_t) * out_buffer->NumColumns());
+
+  // fill in missing columns and default values
+  FillMissingColumns(out_buffer, desired_v);
+  return result;
+}
+
+std::pair<bool, TupleSlot> SqlTable::Update(const common::ManagedPointer<transaction::TransactionContext> txn,
```
CalLavicka, line 53: Maybe take in a pointer to a TupleSlot rather than returning a pair? Should be cleaner code, since a lot of times you are just using the bool.

src/storage/sql_table.cpp
```diff
@@ -1,20 +1,314 @@
...
+
+  if (missing) {
+    for (uint32_t idx = 0; idx < out_buffer->NumTuples(); idx++) {
+      ProjectedColumns::RowView row = out_buffer->InterpretAsRow(idx);
+      FillMissingColumns(&row, desired_v);
+    }
+  }
+}
+
+// return false if it is null
```
CalLavicka, line 175: Add a more detailed function header

src/storage/sql_table.cpp
```diff
@@ -54,43 +361,58 @@ SqlTable::SqlTable(const common::ManagedPointer<BlockStore> store, const catalog
...
+  TERRIER_ASSERT(!col_oids.empty(), "Should be used to access at least one column.");
+  // Build the input to the initializer constructor
+  for (const catalog::col_oid_t col_oid : col_oids) {
+    TERRIER_ASSERT(tables_.at(layout_version).column_oid_to_id_map_.count(col_oid) > 0,
+                   "Provided col_oid does not exist in the table.");
+    const col_id_t col_id = tables_.at(layout_version).column_oid_to_id_map_.at(col_oid);
+    inverse_map[col_id] = col_oid;
+  }
+
+  // for (uint16_t i = 0; i < col_oids.size(); i++) inverse_map[col_ids[i]] = col_oids[i];
```
CalLavicka, line 400: Dead code

test/include/test_util/storage_test_util.h
```diff
@@ -268,6 +270,54 @@ class StorageTestUtil {
...
+
+        // Check that the two have the same content bit-wise
+        uint8_t attr_size = layout1.AttrSize(one_id);
+        const byte *one_content = one->AccessWithNullCheck(idx1);
+        const byte *other_content = other->AccessWithNullCheck(idx2);
+        // Either both are null or neither is null.
+        if (one_content == nullptr || other_content == nullptr) {
+          EXPECT_EQ(one_content, other_content);
+          if (one_content == other_content) continue;
+          return false;
```
CalLavicka, line 295: If they should be equal, why are you checking twice?

test/include/test_util/storage_test_util.h
```diff
@@ -268,6 +270,54 @@ class StorageTestUtil {
...
+        }
+        // Otherwise, they should be bit-wise identical.
+        if (memcmp(one_content, other_content, attr_size) != 0) {
+          return false;
+        }
+      } else {  // Column is dropped
+        // a dropped column should not exists in the map
+        EXPECT_EQ(oid_map2.find(oid1), oid_map2.end());
+        if (oid_map2.find(oid1) != oid_map2.end()) {
+          return false;
```
CalLavicka, line 305: Again, why check twice?

src/storage/sql_table.cpp
```diff
@@ -1,20 +1,314 @@
...
+  if (missing) {
+    for (uint32_t idx = 0; idx < out_buffer->NumTuples(); idx++) {
+      ProjectedColumns::RowView row = out_buffer->InterpretAsRow(idx);
+      FillMissingColumns(&row, desired_v);
+    }
+  }
+}
+
+// return false if it is null
+static bool PeekValue(const type::TransientValue &transient_val, byte *value_output) {
```
CalLavicka, line 176: This should probably live in type::TransientValuePeeker, so it can be used elsewhere in the codebase

src/include/storage/sql_table.h
```diff
@@ -140,67 +140,147 @@ class SqlTable {
...
+   * Since the catalog table prevents write-write conflict with version pointer, calling UpdateSchema here is always
+   * thread-safe.
+   * @param txn the calling transaction
+   * @param layout_version Version number for the new schema
+   * @param schema the updated schema of this SqlTable
    */
-  void Scan(const common::ManagedPointer<transaction::TransactionContext> txn, DataTable::SlotIterator *const start_pos,
-            ProjectedColumns *const out_buffer) const {
-    return table_.data_table_->Scan(txn, start_pos, out_buffer);
+  void UpdateSchema(const common::ManagedPointer<transaction::TransactionContext> txn, const catalog::Schema &schema,
```
CalLavicka, line 163: Take in a managedpointer<schema>, instead of creating one here (no reason to take in a reference)

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+
+Each Sqltable has a separate version number counter, managed by the catalog. Each Sqltable starts out with version number 0, and manages a single datatable. When UpdateSchema() is called, the version number is incremented, and we create a new datatable via SqlTable::CreateTable(), which is the intended location for all tuples under the new schema. Note that within a Sqltable, each datatable uniquely corresponds to a version number and a schema, both of which never change. Therefore, in each datatable, we store its version number.  Moreover, for each datatable within Sqltable,  we store in DataTableVersion its schema, a default value map, and two maps from column oid to id and from id to oid.
+
+To do an SQL query such as updating a tuple, a transaction first uses the index on Sqltable to find the tupleslot of the tuple (note that when we update the index when migrating tupeslots). We can get the raw block from the top 44 bits of the tupleslot, and the raw block has reference to the datatable of the tupleslot.   Thus we know the version number of the datatable the tuple is current stored in, call this the *storage_version*.
```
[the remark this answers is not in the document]
githubxxcc: There is a one-to-one mapping from the transaction timestamp to the database version, which will be acquired by each txn from the catalog before accessing the sqltable. It is correct that Select and Scan might work on multiple tuples, but they will only look at one datatable version.

src/include/storage/data_table.h
githubxxcc: right, we are going with the sqltable pointer approach now. good catch

DESIGN.md
shengxu1: yes it is a TODO

DESIGN.md
shengxu1: migrating on reads requires taking write locks for reads, which is expensive

DESIGN.md
shengxu1: if the tuple is updated multiple times, only the first time will be this case

DESIGN.md
```diff
@@ -0,0 +1,111 @@
...
+
+#### Sqltable::select
+
+Since we don't do migration on reads for now, we will not migrate tuples even in the case of *storage_version < intended_version*. We will read the tuple, transform it to the new schema, select its columns and return.
+
+#### Sqltable::scan
+
+Note that each tuple might physically exist in multiple datatables under a single Sqltable. However, for each transaction, only one copy of each tuple is visible across all datatables (we guarantee this by always logically deleting a tuple before migration).
+
+Therefore, scan works by scanning all datatables in order. Recall that tables_ is an ordered map, so we can simply scan from tables_.begin() to tables_.end(). We will scan each datatable fully, can jump to the next datatable when we reach the end. It is possible that we need to revise the order to scan from tables_.end() to tables_.begin(), because in the special case that a single transaction interleaves scanning with updating, we might scan the same tuple twice (it is first scanned, then migrated to the end of tables_, then scanned again).
```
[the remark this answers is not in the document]
shengxu1: that depends on the workloads. same as above, need to take write locks

DESIGN.md
shengxu1: hmm in that case it might be better to just use a vector

DESIGN.md
shengxu1: sure

src/storage/sql_table.cpp
shengxu1: ok

src/include/storage/sql_table.h
apavlo: Furthermore, what does the pair represent?

src/storage/sql_table.cpp
apavlo: Agreed. What are you trying to do with `template` here?

src/storage/sql_table.cpp
apavlo: @kjobkjobanputra citation? Not that I don't believe you but I would interested to see what makes you think that.

src/storage/data_table.cpp
apavlo: Removing asserts should not be done lightly.

src/storage/sql_table.cpp
```diff
@@ -131,109 +140,219 @@ TupleSlot SqlTable::Insert(const common::ManagedPointer<transaction::Transaction
...
 
 template <class RowType>
-bool SqlTable::AlignHeaderToVersion(RowType *const out_buffer, const DataTableVersion &tuple_version,
-                                    const DataTableVersion &desired_version, col_id_t *cached_ori_header) const {
-  bool missing_col = false;
-  // reserve the original header, aka intended column ids'
-  std::memcpy(cached_ori_header, out_buffer->ColumnIds(), sizeof(col_id_t) * out_buffer->NumColumns());
-
-  // for each column id in the intended version of datatable, change it to match the current schema version
+std::vector<std::pair<size_t, catalog::col_oid_t>> SqlTable::AlignHeaderToVersion(
```
alex-stanescu, line 300: Consider returning a std::vector<size_t> and then getting the col_oid_t later as needed in this class

DESIGN.md
```diff
@@ -0,0 +1,133 @@
...
+
+Each SQL query also comes with an argument, version_number, which is the schema version of the Sqltable that should be seen by the current transaction. Let us call these versions *intended_version*. 
+
+To migrate a tuple with primary key *K* from an older datatable to a newer datatable, we first logically delete the tuple from the older datatable with the current transaction's timestamp, then insert the tuple to the newer datatable. Note that the old tupleslot for *K* has been deleted and a new tupleslot was created. We will update the index by logically deleting *K* then inserting *K* with the new tupleslot.
+
+Note that *storage_version* might be smaller than *intended_version*, which occurs if a transaction accesses a tuple of a newer schema version, but the tuple has not yet been migrated over from old datatables. Different Sqltable operations deal with this special case of  *storage_version < intended_version* differently, as we will  describe below.
+
+#### Sqltable::insert 
+
+There is no *storage_version* in this case. We directly insert to the table corresponding to *intended_version*.
```
kjobanputra, line 49: I think you mean that this new tuple wouldn't exist in storage_version, but the wording implies that storage_version doesn't exist, which can be confusing.

DESIGN.md
```diff
@@ -0,0 +1,133 @@
...
+
+Each SQL query also comes with an argument, version_number, which is the schema version of the Sqltable that should be seen by the current transaction. Let us call these versions *intended_version*. 
+
+To migrate a tuple with primary key *K* from an older datatable to a newer datatable, we first logically delete the tuple from the older datatable with the current transaction's timestamp, then insert the tuple to the newer datatable. Note that the old tupleslot for *K* has been deleted and a new tupleslot was created. We will update the index by logically deleting *K* then inserting *K* with the new tupleslot.
+
+Note that *storage_version* might be smaller than *intended_version*, which occurs if a transaction accesses a tuple of a newer schema version, but the tuple has not yet been migrated over from old datatables. Different Sqltable operations deal with this special case of  *storage_version < intended_version* differently, as we will  describe below.
```
kjobanputra, line 45: For unique transactions with timestamps t1, t2, t2 > t1, does the transaction with timestamp t2 have an intended_version greater than the transaction with timestamp t1? If so, wouldn't the case of having storage_version > intended_version need to be considered?

kjobanputra: For example, if the transaction with timestamp t2 commits before the transaction with timestamp t1, storage_version would be updated to a version number higher than the intended_version associated with the transaction with timestamp t1.

DESIGN.md
```diff
@@ -0,0 +1,133 @@
...
+
+We decided to use a fixed-size vector, where the datatable with version i is stored in the i-th position of the vector. We use a MAX_NUM_OF_VERSIONS as the size of the vector, which is the max allowed number of versions for a single datatable.
+
+We use a fixed-size vector because of low synchronization overhead: there is no need for locking the vector since different versions access different indices. Also, we need not make a special case for tables with a single version, because accessing an index in the vector is as fast as reading a pointer. Using a vector is thus much faster than a concurrent map.
+
+One potential issue with fixed-size vector is dealing with datatables with more than MAX_NUM_OF_VERSIONS. This should be extremely rare if we set MAX_NUM_OF_VERSIONS, since schema changes generally don't happen too often on a datatable. When we support GC of stale versions, we can implement recycling segments of the vector with stale versions, and adding offset to allow using this vector in a cyclic fashion,
```
kjobanputra, line 101: Something to consider is that you can reserve MAX_NUM_OF_VERSIONS, which means that the vector is at least of this size. Then, if you do need to dynamically resize, it would be a constant operation amortized, which could be reasonable until there is GC of stale versions.

src/catalog/database_catalog.cpp
```diff
@@ -940,12 +920,129 @@ bool DatabaseCatalog::RenameTable(const common::ManagedPointer<transaction::Tran
...
+          auto *const update_pr = update_redo->Delta();
+
+          update_redo->SetTupleSlot(tuple_slot);
+
+          *reinterpret_cast<col_oid_t *>(update_pr->AccessForceNotNull(0)) = next_oid + 1;
+          storage::TupleSlot new_slot;
+          auto UNUSED_ATTRIBUTE res = classes_->Update(txn, update_redo, storage::layout_version_t(0), &new_slot);
+          TERRIER_ASSERT(res, "Updating the next_col_oid should not fail");
+          TERRIER_ASSERT(update_redo->GetTupleSlot() == new_slot, "Updating should not move the tuple slot");
+        } break;
```
kjobanputra, line 972: Small thing, but putting break outside the brackets seems odd.

src/catalog/database_catalog.cpp
```diff
@@ -1359,11 +1457,10 @@ void DatabaseCatalog::TearDown(const common::ManagedPointer<transaction::Transac
       TERRIER_ASSERT(schemas[i] != nullptr, "Pointer to schemas in pg_class should not be nullptr");
       switch (classes[i]) {
         case postgres::ClassKind::REGULAR_TABLE:
-          table_schemas.emplace_back(reinterpret_cast<Schema *>(schemas[i]));
           tables.emplace_back(reinterpret_cast<storage::SqlTable *>(objects[i]));
           break;
         case postgres::ClassKind::INDEX:
-          index_schemas.emplace_back(reinterpret_cast<IndexSchema *>(schemas[i]));
+          //          index_schemas.emplace_back(reinterpret_cast<IndexSchema *>(schemas[i]));
```
kjobanputra, line 1463: Take out dead code

src/include/parser/sql_statement.h
```diff
@@ -215,5 +218,335 @@ class TableRefStatement : public SQLStatement {
...
+
+      case DataType::DECIMAL:
+      case DataType::DOUBLE:
+      case DataType::FLOAT:
+        return type::TypeId::DECIMAL;
+
+      case DataType::BOOLEAN:
+        return type::TypeId::BOOLEAN;
+
+        // case ADDRESS:
```
kjobanputra, line 401: Delete dead code.

src/include/planner/plannodes/alter_plan_node.h
```diff
@@ -0,0 +1,415 @@
...
+    parser::AlterTableStatement::AlterType GetType() const override {
+      return parser::AlterTableStatement::AlterType::AddColumn;
+    }
+
+    /**
+     * @return new column to be added
+     */
+    const catalog::Schema::Column &GetColumn() const { return col_; }
+
+    // TODO(SC)
```
kjobanputra, line 118: Still a todo?

src/execution/sql/alter_executors.cpp
```diff
@@ -0,0 +1,55 @@
...
+    }
+  }
+  // Generate new schema
+  std::unique_ptr<catalog::Schema> tmp_schema(new catalog::Schema(cols));
+  (*schema).swap(tmp_schema);
+
+  // Record the change
+  (*change_map)[drop_col_cmd->GetName()].push_back(ChangeType::DropNoCascade);
+
+  // TODO(Schema-Change): where to handle cascade dropping?
```
kjobanputra, line 52: Also need to think about cascading updates

src/execution/sql/ddl_executors.cpp
```diff
@@ -156,4 +158,60 @@ bool DDLExecutors::CreateIndex(const common::ManagedPointer<catalog::CatalogAcce
...
+  // Schema to accumulate the changes from various actions
+  std::unique_ptr<catalog::Schema> update_schema(nullptr);
+
+  // Map to indicate the changes to columns
+  std::unordered_map<std::string, std::vector<ChangeType>> change_map;
+  // Get the current schema
+  if (update_schema == nullptr) {
+    const auto &schema = accessor->GetSchema(table_oid);
+    const auto &cols = schema.GetColumns();
+    // NOLINTNEXTLINE
```
kjobanputra, line 185: What's the reason for not linting here?

src/include/parser/sql_statement.h
```diff
@@ -215,5 +218,335 @@ class TableRefStatement : public SQLStatement {
...
+   * @param name column name
+   * @param type column type
+   * @param is_primary is primary key
+   * @param is_not_null is not nullable
+   * @param is_unique is unique
+   * @param default_expr default expression
+   * @param check_expr check expression
+   * @param varlen size of column if varlen
+   */
+  ColumnDefinition(std::string name, DataType type, bool is_primary, bool is_not_null, bool is_unique,
```
kjobanputra, line 289: Shouldn't parameters be const?

src/include/storage/sql_table.h
```diff
@@ -140,93 +137,229 @@ class SqlTable {
...
    * @param col_oids set of col_oids, they must be in the table's ColumnMap
    * @return vector of col_ids for these col_oids
    */
-  std::vector<col_id_t> ColIdsForOids(const std::vector<catalog::col_oid_t> &col_oids) const;
+  std::vector<col_id_t> ColIdsForOids(const std::vector<catalog::col_oid_t> &col_oids,
+                                      layout_version_t layout_version = layout_version_t{0}) const;
 
   /**
    * @warning This function is expensive to call and should be used with caution and sparingly.
+   * (TODO: no longer expensive?)
```
kjobanputra, line 358: Is this to say it's not expensive anymore or that you're thinking of how to not make it expensive?

src/optimizer/plan_generator.cpp
```diff
@@ -891,4 +892,72 @@ void PlanGenerator::Visit(const Analyze *analyze) {
...
+        auto if_exist = cmd_ref->IsIfExists();
+        auto drop_cascade = cmd_ref->IsDropCascade();
+
+        // Add to the cmd queues
+        auto drop_col_cmd =
+            std::make_unique<planner::AlterPlanNode::DropColumnCmd>(col_name, if_exist, drop_cascade, col_oid);
+        cmds.push_back(std::move(drop_col_cmd));
+      } break;
+      default:
+        TERRIER_ASSERT(false, "Not implemented");
```
kjobanputra, line 952: Might want to be more specific about what cases are not considered in this implementation

src/parser/postgresparser.cpp
```diff
@@ -1661,6 +1665,72 @@ std::unique_ptr<DeleteStatement> PostgresParser::DeleteTransform(ParseResult *pa
...
+        if (expr != nullptr) parse_result->AddExpression(std::move(expr));
+        cmds.emplace_back(col_name, default_val, cmd->behavior_ == DropBehavior::DROP_CASCADE);
+        break;
+      }
+      case AT_DropColumn:
+        col_name = std::string(cmd->name_);
+        cmds.emplace_back(col_name, cmd->missing_ok_, cmd->behavior_ == DropBehavior::DROP_CASCADE);
+        break;
+      case AT_AlterColumnType:
+        // FIXME(xc): why cmd->def->colname_ might be 0x0 here???
```
kjobanputra, line 1715: An assert here might be useful.

src/planner/plannodes/alter_plan_node.cpp
```diff
@@ -0,0 +1,57 @@
...
+  if (cmds_ != other.cmds_) return false;
+
+  return true;
+}
+
+nlohmann::json AlterPlanNode::ToJson() const {
+  nlohmann::json j = AbstractPlanNode::ToJson();
+  j["table_oid"] = table_oid_;
+  j["column_oids"] = column_oids_;
+  // j["cmds"] = cmds_;
```
kjobanputra, line 44: Delete dead code.

src/planner/plannodes/alter_plan_node.cpp
```diff
@@ -0,0 +1,57 @@
...
+  return j;
+}
+
+std::vector<std::unique_ptr<parser::AbstractExpression>> AlterPlanNode::FromJson(const nlohmann::json &j) {
+  std::vector<std::unique_ptr<parser::AbstractExpression>> exprs;
+  auto e1 = AbstractPlanNode::FromJson(j);
+  exprs.insert(exprs.end(), std::make_move_iterator(e1.begin()), std::make_move_iterator(e1.end()));
+  table_oid_ = j.at("table_oid").get<catalog::table_oid_t>();
+  column_oids_ = j.at("column_oids").get<std::vector<catalog::col_oid_t>>();
+  // cmds_ = j.at("cmds").get<std::vector<AlterCmdBase>>();
```
kjobanputra, line 54: Delete dead code

src/storage/sql_table.cpp
```diff
@@ -1,96 +1,457 @@
...
+      break;
+    }
+    case type::TypeId::BIGINT: {
+      auto val = type::TransientValuePeeker::PeekBigInt(transient_val);
+      memcpy(value_output, &val, type::TypeUtil::GetTypeSize(transient_val.Type()));
+      break;
+    }
+    case type::TypeId::DATE: {
+      // TODO(Schema-Change): find a way to handle it without codegen
+      //      auto val = type::TransientValuePeeker::PeekDate(transient_val);
```
kjobanputra, line 220: dead code?

src/include/storage/sql_table.h
```diff
@@ -140,93 +137,229 @@ class SqlTable {
...
+   * SqlTable::begin(), the results will always contain the same tuples
    * @param txn the calling transaction
-   * @param start_pos iterator to the starting location for the sequential scan
+   * @param start_pos iterator to the starting location for the sequential scan.
    * @param out_buffer output buffer. The object should already contain projection list information. This buffer is
    *                   always cleared of old values.
+   * @param layout_version Schema version the current querying transaction should see
+   */
+  void Scan(common::ManagedPointer<transaction::TransactionContext> txn, DataTable::SlotIterator *start_pos,
+            ProjectedColumns *out_buffer, layout_version_t layout_version = layout_version_t{0}) const;
```
kjobanputra, line 150: Shouldn't layout_version be a const?

DESIGN.md
shengxu1: This would be hard to pull off, because vector.resize() is not thread safe, it might move the vector to another memory location, so concurrent accesses on vector locations will be unsafe

DESIGN.md
shengxu1: I think you are mixing up versioning with transaction timestamps. They are 2 completely different things.

Version only updates when updateschema is called. For any tupleslot, it will either be placed in the datatable with correct version, or be in a datatable with older version. We will never more a tupleslot to a datatable with version newer then its intended version.

DESIGN.md
shengxu1: it actually does mean that storage_version doesn't exist. The tupleslot is NOT in the table.

src/include/storage/sql_table.h
```diff
@@ -251,43 +251,81 @@ class SqlTable {
...
-   * @param out_buffer a buffer that corresponds to an old schema version
+   * @param out_buffer a buffer initially with desired version, and translated to tuple version
    * @param tuple_version the old schema version
-   * @param desired_version desired schema version
-   * @param cached_ori_header original header cached
-   * @return
+   * @param desired_version the desired schema version
+   * @param cached_orig_header original header cached
+   * @param size_map columns which have size mismatch
+   * @return whether there are columns in the desired schema version but not in the tuple version.
```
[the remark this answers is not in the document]
shengxu1: will fix

DESIGN.md
kjobanputra: Resize doesn't need to be called. Reserve just ensures that the size of the vector is at least of a certain size. As for data races with reallocation, "if reallocation happens, the container and all its contained elements are modified. Otherwise, the container is accessed, but not the contained elements: concurrently accessing or modifying them is safe." Source: http://www.cplusplus.com/reference/vector/vector/reserve/

shengxu1: well reserving space for the vector is not that much different from our current approach, which basically allocates vector of size MAX_NUM_VERSIONS upfront. I don't see why this is necessary.

src/include/storage/sql_table.h
shengxu1: const is pointless when are passing arguments as values

src/include/parser/sql_statement.h
shengxu1: const is pointless when are passing arguments as values

src/include/catalog/database_catalog.h
```diff
@@ -378,6 +382,25 @@ class DatabaseCatalog {
...
   type_oid_t GetTypeOidForType(type::TypeId type);
 
+  /**
+   * Get entries from pg_attribute
+   * @tparam Column type of columns
+   * @param txn txn to use
+   * @param class_oid oid of table or index
+   * @return the column from pg_attribute
+   */
+  template <typename Column, typename ClassOid, typename ColOid>
```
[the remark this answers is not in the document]
shengxu1: it is, see databasecatalog.cpp line 662

src/include/catalog/catalog_accessor.h
```diff
@@ -360,6 +367,25 @@ class CatalogAccessor {
    */
   common::ManagedPointer<storage::BlockStore> GetBlockStore() const;
 
+  /**
+   * Get the table columns given a table oid
+   * @param table table_oid
+   * @return vector of columns of the table
+   */
+  std::vector<Schema::Column> GetColumns(table_oid_t table);
```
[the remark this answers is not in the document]
shengxu1: why does it matter? we are return a value, not a reference

src/storage/sql_table.cpp
shengxu1: I don't see why this is necessary. The current design works fine

test/include/test_util/storage_test_util.h
githubxxcc: The first `EXPECT_EQ` check is so that the `one_content != other_content` is reported when test fails. 
The second return false is so that the invocation will be reported as false. 

As a result, you have better test output where you know 2 projected rows across schema versions do not match, and you know what exactly causes them not to match. 
And in fact, other test utility functions are have this "double check".

src/include/parser/sql_statement.h
```diff
@@ -215,5 +218,335 @@ class TableRefStatement : public SQLStatement {
...
+        default_expr_(default_expr),
+        check_expr_(check_expr),
+        varlen_(varlen) {}
+
+  /**
+   * @param str type string
+   * @return data type
+   */
+  static DataType StrToDataType(const char *str) {
+    DataType data_type;
```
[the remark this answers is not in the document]
githubxxcc: the parser results from that particular parser node does ensure it. Check postgres parser.cpp:1490 and its corresponding struct definition.

githubxxcc:
Ready for final code drop 🚀

apavlo:
Closing for now. We will come back to this later. Too many jenkins triggers.

PR #863, opened 2020-04-09
Add/Drop Index support

CalLavicka:
Right now, there is a basic blocking mechanism in place using condition variables. Design document is at addDropIndexDesign.md. There is a TODO for updating the index catalogue, but besides that it should be functional, albeit inefficient at this point.


addDropIndexDesign.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Design Rationale
+>*Explain the goals of this design and how the design achieves these goals. Present alternatives considered and document why they are not chosen.*
+
+The goals of the above design are to ensure that the system knows when to block and when to not, through the utilization of transaction timestamps.
+
+## Testing Plan
+>*How should the component be tested?*
+
+The component will be tested with unit tests to make sure the create index successfully adds existing entries. Multi-threaded tests will be included to make sure there are no bugs with reads/writes happening at the same time as the index is being built. Additionally, the drop functionality will need to be tested.
```
shengxu1, line 37: probably want to measure throughput of operations on index while creating index

addDropIndexDesign.md
```diff
@@ -0,0 +1,48 @@
...
+
+At the start of the next phase (which starts at t2), a separate flag is set in the catalog - indicating that any transactions with a timestamp after t2 that wish to modify the column the index is being built on must block. The transaction that is building the index then blocks - waiting for all transactions with timestamp less than t1 to commit. Once that is done, the index building transaction scans through the delta store and updates the index with any changes.
+
+The output of this component is the index created
+
+## Design Rationale
+>*Explain the goals of this design and how the design achieves these goals. Present alternatives considered and document why they are not chosen.*
+
+The goals of the above design are to ensure that the system knows when to block and when to not, through the utilization of transaction timestamps.
```
shengxu1, line 32: are there other alternative designs and what are the tradeoffs? You guys only describe one architecture, and don't really explain why you choose this architecture

src/catalog/catalog_accessor.cpp
```diff
@@ -90,8 +90,8 @@ std::vector<constraint_oid_t> CatalogAccessor::GetConstraints(table_oid_t table)
   return dbc_->GetConstraints(txn_, table);
 }
 
-std::vector<index_oid_t> CatalogAccessor::GetIndexOids(table_oid_t table) const {
-  return dbc_->GetIndexOids(txn_, table);
+std::vector<index_oid_t> CatalogAccessor::GetIndexOids(table_oid_t table, bool only_live) const {
+  return dbc_->GetIndexOids(txn_, table, only_live);
```
shengxu1, line 94: what does "only_live" mean? Probably add a comment. Same for below.

src/execution/sql/ddl_executors.cpp
```diff
@@ -150,10 +152,20 @@ bool DDLExecutors::CreateIndex(const common::ManagedPointer<catalog::CatalogAcce
...
   auto *const index = index_builder.Build();
+  index->SetNotLive();
+
   bool result UNUSED_ATTRIBUTE = accessor->SetIndexPointer(index_oid, index);
   TERRIER_ASSERT(result, "CreateIndex succeeded, SetIndexPointer must also succeed.");
+
+  // Now, populate the index
+  index_builder.BulkInsert(index);
+
+  // Make the index live, with the later transaction
```
shengxu1, line 166: what does "with the later transaction" mean?

src/include/storage/index/index.h
```diff
@@ -41,6 +43,21 @@ class Index {
    */
   const IndexMetadata metadata_;
 
+  /**
+   * Whether or not the index is live yet and ready for updates
+   */
+  bool is_live_ = true;
```
shengxu1, line 49: is it safer to initialize this to false?

addDropIndexDesign.md
```diff
@@ -0,0 +1,48 @@
...
+ - index_schema.\*
+ - codegen.\*
+
+
+## Architectural Design
+>*Explain the input and output of the component, describe interactions and breakdown the smaller components if any. Include diagrams if appropriate.*
+
+There are two phases to index creation. In the first phase, the index is created and added to the catalog (with a flag indicating it is not yet ready for use). Then, that transaction will populate the index as if it were a snapshot of the table at the timestamp of the transaction. _Are we using the standard delta store here or are we maintaining a separate delta store_?
+
+At the start of the next phase (which starts at t2), a separate flag is set in the catalog - indicating that any transactions with a timestamp after t2 that wish to modify the column the index is being built on must block. The transaction that is building the index then blocks - waiting for all transactions with timestamp less than t1 to commit. Once that is done, the index building transaction scans through the delta store and updates the index with any changes.
```
shengxu1, line 25: what about transactions with timestamp > t2 that access the column but do not modify the column? Should they be blocked?

shengxu1: Are there other possibilities apart from blocking? Is it possible to fall back to a sequential scan?

shengxu1: Also, I though indexes can be built on any collection of tupleslots, not necessarily just columns? For example, I can have an index on a table.

src/include/storage/index/index.h
```diff
@@ -169,6 +186,35 @@ class Index {
...
+      std::lock_guard<std::mutex> lock(is_live_mutex_);
+      is_live_ = true;
+    }
+    is_live_condvar_.notify_all();
+  }
+
+  /**
+   * Sets the index as not being live yet. Used while populating
+   */
+  void SetNotLive() {
```
shengxu1, line 215: why doesn't this require locking?

src/include/storage/index/index_builder.h
```diff
@@ -67,7 +72,39 @@ class IndexBuilder {
...
   }
 
+ /**
+  * @param sql_table, transaction context
+  * @return the builder object
+  */
+  IndexBuilder &SetSqlTableAndTransactionContext(common::ManagedPointer<storage::SqlTable> sql_table,
+                                                 const common::ManagedPointer<transaction::TransactionContext> txn) {
+    TERRIER_ASSERT(sql_table == nullptr && txn == nullptr || sql_table != nullptr && txn != nullptr,
+                   "sql_table / txn is null and txn / sql_table is not.");
```
shengxu1, line 82: better phrased as:   sqltable and txn are both null, or sqltable and txn are both not null

src/include/storage/index/index_builder.h
```diff
@@ -67,7 +72,39 @@ class IndexBuilder {
...
+    sql_table_ = sql_table;
```
shengxu1, line 83: can you use sql_table_(sql_table) syntax here?

src/traffic_cop/traffic_cop.cpp
```diff
@@ -146,8 +146,23 @@ void TrafficCop::ExecuteCreateStatement(const common::ManagedPointer<network::Co
...
     }
     case network::QueryType::QUERY_CREATE_INDEX: {
+      // Create a transaction for the index builder
+      const auto populate_txn = txn_manager_->BeginTransaction();
       if (execution::sql::DDLExecutors::CreateIndexExecutor(
-              physical_plan.CastManagedPointerTo<planner::CreateIndexPlanNode>(), connection_ctx->Accessor())) {
+              physical_plan.CastManagedPointerTo<planner::CreateIndexPlanNode>(), connection_ctx->Accessor(),
+              common::ManagedPointer(populate_txn))) {
+        if (populate_txn->MustAbort()) {
+          out->WriteErrorResponse("ERROR:  failed to execute CREATE INDEX");
```
shengxu1, line 155: what causes CREATE INDEX to fail? add a comment maybe

src/execution/sql/ddl_executors.cpp
```diff
@@ -150,10 +152,20 @@ bool DDLExecutors::CreateIndex(const common::ManagedPointer<catalog::CatalogAcce
...
+  index_builder.SetSqlTableAndTransactionContext(accessor->GetTable(table), populate_txn);
   index_builder.SetKeySchema(schema);
   auto *const index = index_builder.Build();
+  index->SetNotLive();
+
   bool result UNUSED_ATTRIBUTE = accessor->SetIndexPointer(index_oid, index);
   TERRIER_ASSERT(result, "CreateIndex succeeded, SetIndexPointer must also succeed.");
+
+  // Now, populate the index
+  index_builder.BulkInsert(index);
```
shengxu1, line 164: is it better to bulkinsert everything in one giant transaction (if index is large), or divide it up into multiple smaller transactions?

addDropIndexDesign.md
```diff
@@ -0,0 +1,48 @@
+# Add/Drop Index
+
+## Overview
+
+>*What motivates this to be implemented? What will this component achieve?*
+
+Currently, the system just creates the index without populating it. This component will support actually adding the index (including populating the index) and/or dropping in a transactionally consistent manner. This will be achieved by halting any ongoing transactions and maintaining a delta storage while the index is being built, applying the delta changes after the index is done being built. The bonus task would include index building in a multi-threaded fashion.
```
shengxu1, line 7: where is the delta storage implemented right now?

src/execution/sql/ddl_executors.cpp
```diff
@@ -139,7 +140,8 @@ bool DDLExecutors::DropIndexExecutor(const common::ManagedPointer<planner::DropI
 
 bool DDLExecutors::CreateIndex(const common::ManagedPointer<catalog::CatalogAccessor> accessor,
                                const catalog::namespace_oid_t ns, const std::string &name,
-                               const catalog::table_oid_t table, const catalog::IndexSchema &input_schema) {
+                               const catalog::table_oid_t table, const catalog::IndexSchema &input_schema,
+                               const common::ManagedPointer<terrier::transaction::TransactionContext> populate_txn) {
   // Request permission from the Catalog to see if this a valid namespace and table name
   const auto index_oid = accessor->CreateIndex(ns, table, name, input_schema);
```
githubxxcc, line 146: How about txns that start after t1 (starting time of this populate_index txn)?
How will those txns be able to run? 
Are those txns being blocked here as well? If not, are they operating on another index?

src/include/storage/index/index_builder.h
```diff
@@ -67,7 +72,39 @@ class IndexBuilder {
...
+
+    ProjectedRow *index_pr = index->GetProjectedRowInitializer().InitializeRow(index_pr_buffer);
+
+    for (auto it = sql_table_->begin(); it != sql_table_->end(); ++it) {
+      const TupleSlot slot = *it;
+      index->Insert(txn_, *index_pr, slot);
+    }
+
+    delete[] index_pr_buffer;
+  }
```
mush-zhang, line 106: maybe in the future we can implement more efficient way to build indexed rather than one-by-one insertion

src/include/storage/index/index.h
apavlo: The name `SetNotLive` is awkward. Why not just make it `SetIsLive(bool)`?

addDropIndexDesign.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Scope
+>*Which parts of the system will this feature rely on or modify? Write down specifics so people involved can review the design doc*
+
+The parts of the system this feature will modify include the following files:
```
shengxu1, line 12: this list needs to be updated

addDropIndexDesign.md
```diff
@@ -0,0 +1,48 @@
...
+ - ddl_executors.\*
+ - catalog_accessor.\*
+ - index_schema.\*
+ - codegen.\*
+
+
+## Architectural Design
+>*Explain the input and output of the component, describe interactions and breakdown the smaller components if any. Include diagrams if appropriate.*
+
+There are two phases to index creation. In the first phase, the index is created and added to the catalog (with a flag indicating it is not yet ready for use). Then, that transaction will populate the index as if it were a snapshot of the table at the timestamp of the transaction. _Are we using the standard delta store here or are we maintaining a separate delta store_?
```
shengxu1, line 23: The design md can be updated to answer the question in the end

addDropIndexDesign.md
```diff
@@ -0,0 +1,48 @@
...
+
+## Trade-offs and Potential Problems
+>*Write down any conscious trade-off you made that can be problematic in the future, or any problems discovered during the design process that remain unaddressed (technical debts).*
+
+## Future Work
+>*Write down future work to fix known problems or otherwise improve the component.*
+
+If the bonus task is not accomplished, this can be put as a future work.
+
+
```
shengxu1, line 48: please update design md to reflect your latest design choices

src/execution/sql/ddl_executors.cpp
```diff
@@ -152,8 +153,21 @@ bool DDLExecutors::CreateIndex(const common::ManagedPointer<catalog::CatalogAcce
   storage::index::IndexBuilder index_builder;
   index_builder.SetKeySchema(schema);
   auto *const index = index_builder.Build();
-  bool result UNUSED_ATTRIBUTE = accessor->SetIndexPointer(index_oid, index);
-  TERRIER_ASSERT(result, "CreateIndex succeeded, SetIndexPointer must also succeed.");
+
+  if (concurrent) {
+
```
shengxu1, line 158: hmm is the concurrent case not implemented as of now?

src/catalog/database_catalog.cpp
```diff
@@ -988,7 +988,10 @@ std::vector<index_oid_t> DatabaseCatalog::GetIndexOids(
   for (auto &slot : index_scan_results) {
     const auto result UNUSED_ATTRIBUTE = indexes_->Select(txn, slot, select_pr);
     TERRIER_ASSERT(result, "Index already verified visibility. This shouldn't fail.");
-    index_oids.emplace_back(*(reinterpret_cast<index_oid_t *>(select_pr->AccessForceNotNull(0))));
+    if (!only_live || *(reinterpret_cast<bool *>(select_pr->AccessForceNotNull(
```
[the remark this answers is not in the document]
shengxu1: I agree with Ricky. Maybe you guys can add more comments to explain these.

src/catalog/database_catalog.cpp
```diff
@@ -977,7 +986,8 @@ std::vector<index_oid_t> DatabaseCatalog::GetIndexOids(
   auto *key_pr = oid_pri.InitializeRow(buffer);
   *(reinterpret_cast<table_oid_t *>(key_pr->AccessForceNotNull(0))) = table;
   std::vector<storage::TupleSlot> index_scan_results;
-  indexes_table_index_->ScanKey(*txn, *key_pr, &index_scan_results);
+  transaction::TransactionContext fake_txn(transaction::timestamp_t{UINT64_MAX}, transaction::timestamp_t{UINT64_MAX});
```
shengxu1, line 989: why do you use a fake txn here with max timestamp instead of txn? Is this safe?

src/execution/sql/storage_interface.cpp
```diff
@@ -2,6 +2,8 @@
 
 #include <algorithm>
 #include <vector>
+#include <mutex>
+#include <condition_variable>
```
shengxu1, line 6: these includes are unnecessary?

src/include/storage/index/index_builder.h
shengxu1: the phrasing still seems logically incorrect

src/include/storage/data_table.h
```diff
@@ -138,6 +138,13 @@ class DataTable {
   bool Select(common::ManagedPointer<transaction::TransactionContext> txn, TupleSlot slot,
               ProjectedRow *out_buffer) const;
 
+  enum class VersionChainType {
+    PRE_UPDATE, POST_UPDATE, INSERT, DELETE, VISIBLE, INVISIBLE
```
shengxu1, line 142: What is PRE_UPDATE and POST_UPDATE, what is VISIBLE and INVISIBLE? Please add some comments to explain

src/include/storage/index/index_builder.h
```diff
@@ -67,7 +72,85 @@ class IndexBuilder {
...
+          }
+        }
+
+        switch (type) {
+          case DataTable::VersionChainType::VISIBLE:
+            printf("Inserting into index");
+            index->Insert(txn_, *index_pr, slot);
+            break;
+          case DataTable::VersionChainType::INVISIBLE:
+          case DataTable::VersionChainType::PRE_UPDATE:
```
shengxu1, line 136: what's the point of inserting then deleting?

src/include/storage/index/index_builder.h
```diff
@@ -67,7 +72,85 @@ class IndexBuilder {
...
+          if (table_pr->IsNull(pr_map[table_col_oid])) {
+            index_pr->SetNull(index->GetKeyOidToOffsetMap().at(index_col_oid));
+          } else {
+            auto size = AttrSizeBytes(col.AttrSize());
+            std::memcpy(index_pr->AccessForceNotNull(index->GetKeyOidToOffsetMap().at(index_col_oid)),
+                        table_pr->AccessWithNullCheck(pr_map[table_col_oid]), size);
+          }
+        }
+
+        switch (type) {
```
shengxu1, line 130: of the 6 types you define, you are only using 3 types here?

src/include/storage/sql_table.h
```diff
@@ -105,6 +113,8 @@ class SqlTable {
                                ->LogRecord::GetUnderlyingRecordBodyAs<RedoRecord>(),
                    "This RedoRecord is not the most recent entry in the txn's RedoBuffer. Was StageWrite called "
                    "immediately before?");
+
+    //
```
shengxu1, line 117: delete this

src/include/storage/sql_table.h
```diff
@@ -40,6 +41,8 @@ class SqlTable {
   };
 
  public:
+  transaction::TransactionContext::DebugLock modify_mutex_;
```
shengxu1, line 44: delete this after debugging

src/include/transaction/transaction_context.h
```diff
@@ -30,6 +31,28 @@ namespace terrier::transaction {
  */
 class TransactionContext {
  public:
+  class DebugLock : public std::shared_mutex {
```
shengxu1, line 34: is this only used for debugging? If so, you can delete thiis after debugging.

src/include/transaction/transaction_context.h
```diff
@@ -30,6 +31,28 @@ namespace terrier::transaction {
...
+
+    void debug_lock_shared(const common::ManagedPointer<transaction::TransactionContext> &txn) {
+      TERRIER_ASSERT(std::count(readers.begin(), readers.end(), txn) == 0, "Locking a lock I already have!");
+      readers.push_back(txn);
+      lockers.insert(txn);
+      std::shared_mutex::lock_shared();
+    }
+
+    void debug_unlock_shared(const common::ManagedPointer<transaction::TransactionContext> &txn) {
+      int UNUSED_ATTRIBUTE cnt = lockers.count(txn);
```
shengxu1, line 45: dead code

src/include/transaction/transaction_context.h
```diff
@@ -203,6 +246,21 @@ class TransactionContext {
    */
   void SetMustAbort() { must_abort_ = true; }
 
+  storage::UndoBuffer GetUndoBuffer() { return undo_buffer_; }
+  storage::RedoBuffer GetRedoBuffer() { return redo_buffer_; }
+
+  void LockIfNotLocked(catalog::table_oid_t table_oid, common::ManagedPointer<std::shared_mutex> table_lock) {
+    if (!IsTableLocked(table_oid)) {
+      reinterpret_cast<DebugLock*>(table_lock.Get())->debug_lock_shared(common::ManagedPointer(this));
+      held_table_oids_.insert(table_oid);
```
shengxu1, line 255: this looks like a race condition to me. For example, thread 1 can acquire the tablelock, and before thread 1 inserts to held_table_oids_, thread 2 checks IsTableLocked, which returns false. Then both threads get the lock

src/include/transaction/transaction_context.h
```diff
@@ -51,6 +74,26 @@ class TransactionContext {
...
+
+  /**
+   * Constructs a new transaction context.
+   *
+   * @warning In the src/ folder this should only be called in TransactionManager::BeginTransaction to adhere to MVCC
+   * semantics. Tests are allowed to deterministically construct them in ways that violate the current MVCC semantics.
+   * @warning Beware that the buffer pool given must be the same one the log manager uses,
+   * if logging is enabled.
+   * @param start the start timestamp of the transaction. Should be unique within the system.
+   * @param finish in HyPer parlance this is txn id. Should be larger than all start times and commit times in current
```
shengxu1, line 86: what is HyPer parlance? Maybe its bad practice to refer to other DBMSs in comment

test/storage/index_builder_test.cpp
```diff
@@ -0,0 +1,142 @@
...
+  }
+  void TearDown() override {
+    thread_pool_.Shutdown();
+
+    db_main_->GetTransactionLayer()->GetDeferredActionManager()->RegisterDeferredAction([=]() { delete sql_table_; });
+  }
+};
+
+/**
+ * This test creates multiple worker threads that all try to insert [0,num_inserts) as tuples in the table and into the
```
shengxu1, line 80: wait really? seems like you're just scanning an empty table

test/storage/index_builder_test.cpp
```diff
@@ -0,0 +1,142 @@
...
+  EXPECT_NE(index, nullptr);
+
+  std::vector<TupleSlot> values;
+  index->ScanAscending(*txn, storage::index::ScanType::OpenBoth, 1, nullptr, nullptr, 0, &values);
+  EXPECT_EQ(values.size(), 0);
+  txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+}
+
+// NOLINTNEXTLINE
+TEST_F(IndexBuilderTests, OneTxnFullTable) {
```
shengxu1, line 99: what about add more tests, with more txns, tables, and operations

src/include/transaction/transaction_context.h
CalLavicka: IsTableLocked only applies to whether the current txn has the lock, so we don't double grab a lock. We are assuming that txns are single threaded wrt this function

src/execution/sql/ddl_executors.cpp
```diff
@@ -152,20 +151,23 @@ bool DDLExecutors::CreateIndex(const common::ManagedPointer<catalog::CatalogAcce
...
   storage::index::IndexBuilder index_builder;
-  index_builder.SetSqlTableAndTransactionContext(accessor->GetTable(table), populate_txn);
   index_builder.SetKeySchema(schema);
   auto *const index = index_builder.Build();
-  index->SetNotLive();
 
-  bool result UNUSED_ATTRIBUTE = accessor->SetIndexPointer(index_oid, index);
-  TERRIER_ASSERT(result, "CreateIndex succeeded, SetIndexPointer must also succeed.");
+  if (concurrent) {
+
```
githubxxcc, line 158: I guess here is a TODO for the concurrent case?

src/include/storage/index/index_builder.h
```diff
@@ -78,33 +78,79 @@ class IndexBuilder {
...
+          case DataTable::VersionChainType::VISIBLE:
+            printf("Inserting into index");
+            index->Insert(txn_, *index_pr, slot);
+            break;
+          case DataTable::VersionChainType::INVISIBLE:
+          case DataTable::VersionChainType::PRE_UPDATE:
+            printf("Inserting into index");
+            index->Insert(txn_, *index_pr, slot);
+            printf("Deleting from index");
```
githubxxcc, line 139: At the end of the bulk inserts, what are the guarantees? (E.g., does the index contain valid entries for all txns that have committed at a particular timestamp, or does the index contain dirty entries which need to be rollbacked?)
Also, have you considered cases when transaction aborts and the version chains get rollback?

alex-stanescu:
@githubxxcc @shengxu1 Ive added some documentation and removed some old features that arent useful anymore that we missed removing earlier

src/execution/sql/ddl_executors.cpp
alex-stanescu: We'd like to keep the concurrent variables since we're using them to tell the user CREATE INDEX CONCURRENTLY is not implemented. Plus, putting them in is an annoying process and they dont do any harm sitting there waiting for someone to implement CONCURRENTLY

addDropIndexDesign.md
alex-stanescu: We're working on updating the design markdown

src/catalog/database_catalog.cpp
alex-stanescu: This will also be communicated in the design doc. Essentially, we want to use a fake txn in order to get any indices that are logically invisible to us (e.g. whose commit timestamp is after our start timestamp) so that we can still insert into them and keep them up-to-date.

Is this safe? Yes, from an MVCC viewpoint indices are performance hints anyways and are not going to affect the result. From a logical viewpoint we're only ever reading from live indices which are guaranteed to be correct (even if the commit timestamp is later than our sstart timestamp)

src/execution/sql/storage_interface.cpp
alex-stanescu: Yep

src/include/storage/data_table.h
alex-stanescu: Comments added

src/include/storage/index/index_builder.h
alex-stanescu: That's how updates are handled on an index - you insert the new value and then delete the old value

src/include/storage/index/index_builder.h
alex-stanescu: There are 6 types defined for future-proofing the method we added in sql_table. We only need two types for this specific application, but one can easily imagine an application where differentiating between an INSERT and an UPDATE is necessary

src/include/storage/sql_table.h
alex-stanescu: done

src/include/transaction/transaction_context.h
alex-stanescu: Yes - we will

src/include/transaction/transaction_context.h
alex-stanescu: Moot point - this is a debugging class anyways

src/include/transaction/transaction_context.h
alex-stanescu: This is a copy of another TransactionContext constructor comment - I'm wary of removing it in both places since we didn't write it

test/storage/index_builder_test.cpp
alex-stanescu: This was an old comment from some copied code. It will be deleted in a future update to this PR

test/storage/index_builder_test.cpp
alex-stanescu: IndexBuilder is too low-level to care about other transactions. We have JDBC tests to verify transactional semantics

src/catalog/database_catalog.cpp
```diff
@@ -1251,7 +1284,8 @@ index_oid_t DatabaseCatalog::GetIndexOid(const common::ManagedPointer<transactio
 
 const IndexSchema &DatabaseCatalog::GetIndexSchema(const common::ManagedPointer<transaction::TransactionContext> txn,
                                                    index_oid_t index) {
-  auto ptr_pair = GetClassSchemaPtrKind(txn, static_cast<uint32_t>(index));
+  transaction::TransactionContext fake_txn(transaction::timestamp_t{UINT64_MAX}, transaction::timestamp_t{UINT64_MAX});
```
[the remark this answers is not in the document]
alex-stanescu: This will also be communicated in the design doc. Essentially, we want to use a fake txn in order to get any indices that are logically invisible to us (e.g. whose commit timestamp is after our start timestamp) so that we can still insert into them and keep them up-to-date.

Is this safe? Yes, from an MVCC viewpoint indices are performance hints anyways and are not going to affect the result. From a logical viewpoint we're only ever reading from live indices which are guaranteed to be correct (even if the commit timestamp is later than our sstart timestamp)

src/catalog/database_catalog.cpp
```diff
@@ -2150,6 +2184,26 @@ Column DatabaseCatalog::MakeColumn(storage::ProjectedRow *const pr, const storag
...
 
+bool DatabaseCatalog::TransferLock(common::ManagedPointer<transaction::TransactionContext> from, common::ManagedPointer<transaction::TransactionContext> to) {
+  bool locked = TryLock(from);
+  if(!locked) {
+    return false;
+  }
+  const transaction::timestamp_t txn_id = to->FinishTime();
+  auto current_val = write_lock_.load();
+
+  if (write_lock_.compare_exchange_strong(current_val, txn_id)) {
```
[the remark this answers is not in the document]
alex-stanescu: This was an old idea we had and we missed removing this method. It's been deleted

src/catalog/database_catalog.cpp
```diff
@@ -2150,6 +2184,26 @@ Column DatabaseCatalog::MakeColumn(storage::ProjectedRow *const pr, const storag
...
+    // acquired the lock
+    auto *const write_lock = &write_lock_;
+    to->RegisterCommitAction([=]() -> void { write_lock->store(to->FinishTime()); });
+    to->RegisterAbortAction([=]() -> void { write_lock->store(current_val); });
```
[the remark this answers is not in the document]
alex-stanescu: This was an old idea we had and we missed removing this method. It's been deleted

src/execution/sql/ddl_executors.cpp
alex-stanescu: This is now an assert

src/include/storage/index/index_builder.h
alex-stanescu: These will be addressed in the design doc as well, but I'll respond here first

> At the end of the bulk inserts, what are the guarantees? 
The index contains valid entries for all transactions that may be in flight. The index cannot contain dirty entries which need to be rollbacked as we wait until any pre-existing modifying transactions are committed.

Any modifying transactions that want to perform modifications after we grab the lock on the table are forced to wait until the index is fully built after which point they will insert into the table using normal semantics (e.g. rolling back as necessary)

> Also, have you considered cases when transaction aborts and the version chains get rollback?
Yes - see the previous comment wrt waiting until in-flight transactions commit.

src/include/storage/index/index_builder.h
CalLavicka: Specifically, this registers the old value in the index and then registers a delete, which what the index does is defer this delete until later. This means currently running transactions timestamped before this update will see it in the index, but it will be removed once current transactions are finished.

alex-stanescu:
@githubxxcc @shengxu1 The design doc has been updated to reflect our current design.

shengxu1:
> @githubxxcc @shengxu1 The design doc has been updated to reflect our current design.

Thanks! lgtm

script/testing/junit/src/CreateIndexTest.java
```diff
@@ -0,0 +1,160 @@
...
+            checkIntRow(rs, new String [] {"c1", "c2", "c3"}, new int [] {3, 4, 200});
+        }
+        for (int i = 0; i < num_rows; i++) {
+            rs.next();
+            checkIntRow(rs, new String [] {"c1", "c2", "c3"}, new int [] {5, 6, 100});
+        }
+        assertNoMoreRows(rs);
+    }
+
+}
```
mush-zhang, line 160: Does this select statement go through the index?

addDropIndexDesign.md
```diff
@@ -0,0 +1,68 @@
...
+ - transaction_manager.\*
+ - catalog_accessor.\*
+ - database_catalog.\*
+
+
+## Architectural Design
+>*Explain the input and output of the component, describe interactions and breakdown the smaller components if any. Include diagrams if appropriate.*
+
+In broad strokes, this component works by first waiting until all uncommitted transactions that have already modified the table either commit or abort. During this time and until index creation is completed, any transactions that have not already modified the table and wish to do so are blocked and must wait until index creation is complete. After the transactions all complete, index creation proceeds by scanning through the table and inserting all visible versions of each tuple into the index. Once index creation is complete, waiting transactions are allowed to proceed.
+
```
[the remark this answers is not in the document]
alex-stanescu: > Is it possible to not blocking the transactions after index creation started

It is not possible to do no blocking of transactions after index creation has started as you run into too many concurrency issues that way. It is however possible to only block them while running through a smaller delta store (rather than while running through the whole table). This is what CREATE INDEX CONCURRENTLY should do (see §Future Work).

> We could put a flag to suggest if one table is being built index on. Modification by those txns are not visible to the index creation txn during scan, and the modifying txns can defer/stash the index update of the modified tuples and let those run before the index become visible.

This is similar to what our original proposal did. There's two classes of issues here:
  - Transactions that have started running but have not yet made modifications to the table at time of index creation:
           These transactions wont have access to knowledge of the index being built unless we provide a lot more information to the codegen'ed code (which would require massive infrastructure changes). This is still solvable though.
  - Transactions that have started running and have already made modifications to the table before index creations:
            These transactions have no way of finding out the index is being built. In order to solve the problem of these transactions committing (for delete) or aborting (for insert), you would have to be able to go from transaction timestamp to transaction pointer. See the first paragraph of §Trade-offs and Potential Problems

src/catalog/catalog_accessor.cpp
```diff
@@ -80,6 +80,10 @@ common::ManagedPointer<storage::SqlTable> CatalogAccessor::GetTable(table_oid_t
   return dbc_->GetTable(txn_, table);
 }
 
+common::ManagedPointer<std::shared_mutex> CatalogAccessor::GetTableLock(table_oid_t table) const {
```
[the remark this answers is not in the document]
alex-stanescu: We were unaware of `shared_latch.h`. We'll take a look at it

addDropIndexDesign.md
```diff
@@ -0,0 +1,68 @@
...
+Additionally, we wait for the writing transactions to fully commit/abort before we start building the index. Originally, we just waited for that individual insert to complete, but we ran into problems when handling aborted transactions, and needed to do messy stuff like add abort callbacks to the uncommitted transactions. This is much cleaner (even if it's a little bit less convenient). This did lead to one small issue with a CREATE INDEX after additions to the table in the same transactions, as that would cause deadlock potentially. We decided to disallow this for now, as it is not a common use case at all (you can just build the index first before additions).
+
+Once we are building the index, we have to scan the version chain such that transactions which are in progress can still see old versions in the index, but these versions get correctly deleted later. We do this using a scan function which takes in a lambda, as that provides clean code that could potentially be used in other systems. We originally were going to write a low-level specialized function inside data_table.cpp, but felt it was too specialized and instead went for a higher-order function.
+
```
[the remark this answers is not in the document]
alex-stanescu: No - what is mentioned in that sentence is that there may be old versions of the _indexed key_ in the index (e.g. if an UPDATE transaction ran within this same epoch and updated the indexed value from 2 to 3, then both 2 and 3 would appear in the index)

script/testing/junit/src/CreateIndexTest.java
alex-stanescu: This has been verified manually - however without an EXPLAIN command implemented in the database we have unfortunately no way of programmatically verifying this

src/catalog/database_catalog.cpp
alex-stanescu: Alive is not necessarily congruent with visible - e.g. a transaction may be alive and still invisible to another transaction, so you would still need to do this

alex-stanescu: > I think you can define the fake_txn_timestamp in some file like txn_def and document it.

We considered this, but weren't sure where best to put it so we put it off. We'll take another look at it

PR #871, opened 2020-04-17
Compiled query caching added to PostgresProtocolInterpreter

mbutrovich:
At a high level, this implements a mechanism to cache physical plans and `ExecutableQuery` (TPL module, AST context, etc.) from code generation in `PostgresProtocolInterpreter`. It relies on libpgquery's pg_fingerprint functionality to identify statements. It builds on top of the Postgres extended query protocol, and doesn't change anything for simple query.

Changes:
- `ExecutableQuery` stashes the main function as a member.
- `Portal` no longer owns the physical plan, but holds a `ManagedPointer` to it.
- `Statement` owns the physical plan and `ExecutableQuery` of its bound `Portal`s. A `Portal` just describes the parameters and output format for that invocation of the `Statement`.
- A new map in `PostgresProtocolInterpreter` from fingerprint to owned `Statement`. The old `Statement` map from name to statement is now not owning.
` After first code generation of a `Statement` at the execute phase, we stash the physical plan and `ExecutableQuery` upstream to be reused at later invocations.

Limitations:
- We have no cache invalidation scheme. This will introduce bugs with DDL changes.
- We should figure out how to do code generation without an `ExecutionContext`. It would allow us to perform that step at the bind phase, rather than execution, which seems more natural to me. Perhaps that goes against adaptive execution though.

### DIFF
--- a/src/include/network/postgres/postgres_protocol_interpreter.h
+++ b/src/include/network/postgres/postgres_protocol_interpreter.h
@@ -158,11 +158,25 @@ class PostgresProtocolInterpreter : public ProtocolInterpreter {
   }
 
   /**
+   * @param fingerprint key
+   * @param statement statement to take ownership of
+   */
+  void AddStatementToCache(const std::string &fingerprint, std::unique_ptr<network::Statement> &&statement) {
+    statement_cache_[fingerprint] = std::move(statement);
+  }
+
+  common::ManagedPointer<network::Statement> LookupStatementInCache(const std::string &fingerprint) const {
+    const auto it = statement_cache_.find(fingerprint);
+    if (it != statement_cache_.end()) return common::ManagedPointer(it->second);
+    return nullptr;
+  }
+
+  /**
    * @param name key
    * @param statement statement to take ownership of
    */
-  void SetStatement(const std::string &name, std::unique_ptr<network::Statement> &&statement) {
-    statements_[name] = std::move(statement);
+  void SetStatement(const std::string &name, const common::ManagedPointer<network::Statement> statement) {
+    statements_[name] = statement;
   }
 
   /**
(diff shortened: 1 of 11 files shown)


src/network/postgres/postgres_network_commands.cpp
```diff
@@ -210,7 +216,16 @@ Transition ParseCommand::Exec(const common::ManagedPointer<ProtocolInterpreter>
     out->WriteNoticeResponse("NOTICE:  we don't yet support that query type.");
   }
 
-  postgres_interpreter->SetStatement(statement_name, std::move(statement));
+  const auto fingerprint_result UNUSED_ATTRIBUTE = pg_query_fingerprint(query.c_str());
+  auto cached_statement = postgres_interpreter->LookupStatementInCache(fingerprint_result.hexdigest);
```
tanujnay112, line 220: Why is this lookup being done after parsing?

src/include/network/postgres/postgres_protocol_interpreter.h
```diff
@@ -157,12 +157,30 @@ class PostgresProtocolInterpreter : public ProtocolInterpreter {
...
+   */
+  void AddStatementToCache(const std::string &fingerprint, std::unique_ptr<network::Statement> &&statement) {
+    statement_cache_[fingerprint] = std::move(statement);
+  }
+
+  /**
+   * @param fingerprint key
+   * @return Statement if it exists in the cache, otherwise nullptr
+   */
+  common::ManagedPointer<network::Statement> LookupStatementInCache(const std::string &fingerprint) const {
```
tanujnay112, line 172: Shouldn't there be a way to remove something from the cache as well to account for things like invalidation or eviction?

mbutrovich: We haven't defined what the invalidation policies are yet (besides a DDL change, but we don't know how to orchestrate that yet). I can add a placeholder API if it makes sense, but I'm not sure we should without thinking through all of the safeguards and constraints.

src/network/postgres/postgres_network_commands.cpp
mbutrovich: My thinking was to make sure it parses properly before you try to fingerprint it, otherwise fingerprinting will fail and I'm not sure we want to rely on that as our parse validity first;

mbutrovich:
For documentation reasons, this now adds a `use_query_cache` command line argument to enable or disable this feature. The default is on, but if you're developing stuff with DDL changes, you might want to turn it off.

Issue #829, opened 2020-03-10
Move implementations of expressions, plan nodes, and operators from header to cpp files.


lmwnshn:
Right now, most of the expressions, plan nodes, and operators have their implementations in the .h files.

They get included in many places. It could help compilation time (significantly?) to move them into cpp files instead.

Related to #623.

apavlo:
Example: https://github.com/cmu-db/terrier/blob/master/src/include/parser/expression/operator_expression.h#L52-L68

PR #873, opened 2020-04-20
Move Expression Functions out of Headers

abalakumar083:
I moved the methods from the expression header files to implementation files, under the expectation that it may improve compilation time. 

I recorded 4 compilation times before the change and 1 afterwards which is at this link:
https://drive.google.com/open?id=1bwOF_FXKPAZ_SD-_Tnk86sbuFXERfM-T

In terms of compilation time I think that the compilation time of the other files did decrease. However, the new cpp files took additional time to compile. So it's difficult for me to gauge the overall effect.

This is part of #829


apavlo:
This looks good. @abalakumar083 is going to measure how much this improves compilation before we merge.

PR #878, opened 2020-04-22
Update LLVM_DIR to reflect new llvm@8 path in Homebrew (macOS)

mbutrovich:
They changed the path from `8.0.1_1` to `8.0.1_3`.


mbutrovich:
This may not be necessary yet until we run a `brew upgrade` on our macOS machines.

mbutrovich:
We'll reopen it when it's necessary.

PR #880, opened 2020-04-27
Keep query string in Statement object

mbutrovich:
Statement now takes ownership of the query string rather than allowing it to be discarded after parsing. Also refactored some classes (Statement, Portal, BinderSherpa) to have .cpp files to adhere to codebase best practices.

The main change is the change to Statement's constructor and the addition of GetQueryText on that class. The rest is just refactoring.


src/network/postgres/statement.cpp
```diff
@@ -0,0 +1,70 @@
...
+#include "type/type_id.h"
+
+namespace terrier::network {
+
+Statement::Statement(std::string &&query_text, std::unique_ptr<parser::ParseResult> &&parse_result)
+    : Statement(std::move(query_text), std::move(parse_result), {}) {}
+
+Statement::Statement(std::string &&query_text, std::unique_ptr<parser::ParseResult> &&parse_result,
+                     std::vector<type::TypeId> &&param_types)
+    : parse_result_(std::move(parse_result)), param_types_(std::move(param_types)) {
```
malin1993ml, line 21: Seems that you need to move `query_text` into `query_text_` here?

src/include/binder/binder_sherpa.h
```diff
@@ -31,41 +32,32 @@ class BinderSherpa {
...
    */
-  common::ManagedPointer<parser::ParseResult> GetParseResult() const { return parse_result_; }
+  common::ManagedPointer<parser::ParseResult> GetParseResult() const;
 
   /**
    * @return parameters for the query being bound
    * @warning can be nullptr if there are no parameters
    */
-  common::ManagedPointer<std::vector<type::TransientValue>> GetParameters() const { return parameters_; }
+  common::ManagedPointer<std::vector<type::TransientValue>> GetParameters() const;
```
malin1993ml, line 47: If I understand gcc correctly, the compiler cannot inline these functions if you put them in the cpp file. Just curious is there a reason to move these one-liner functions into the cpp file?

src/network/postgres/statement.cpp
mbutrovich: Whoops. Good catch.

src/include/binder/binder_sherpa.h
mbutrovich: You're right. For the one-liners I should just leave them in the headers.

PR #884, opened 2020-05-04
New locking

CalLavicka:
Should be 75% (read-write lock controls access to the table by writers while an index is being built)

Adds a java test to test this behavior
Still need to add more tests and cleanup the existing tests that we may have broken.


CalLavicka:
Ignore this sorry

PR #885, opened 2020-05-05
Adding CTE Scan Support

preetansh:


src/brain/operating_unit_recorder.cpp
```diff
@@ -88,7 +89,7 @@ size_t OperatingUnitRecorder::ComputeKeySize(catalog::table_oid_t tbl_oid,
     }
   }
 
-  TERRIER_ASSERT(key_size > 0, "KeySize must be greater than 0");
+  // TERRIER_ASSERT(key_size > 0, "KeySize must be greater than 0");
```
Orankarl, line 92: Is this assertion still necessary here? I think it can be modified or just removed.

src/execution/ast/ast_pretty_print.cpp
```diff
@@ -0,0 +1,360 @@
...
+    case CastKind::SqlTimestampToTimestamp:
+      return "SqlTimestampToTimestamp";
+
+      // Conversion of a primitive timestamp valueinto a SQL timestamp
+    case CastKind::TimestampToSqlTimestamp:
+      return "TimestampToSqlTimestamp";
+
+      // Convert a SQL integer into a SQL real
+    case CastKind::SqlIntToSqlReal:
+      return "SqlIntToSqlReal";
```
Orankarl, line 60: I think the comments here should have the same indent.

src/execution/compiler/operator/cte_scan_translator.cpp
```diff
@@ -0,0 +1,245 @@
...
+
+  // Use std::map to effectively sort OIDs by their corresponding ID
+  std::map<storage::col_id_t, catalog::col_oid_t> inverse_map;
+
+  // Notice the change in the inverse map argument different from sql_table get projection map function
+  for (auto col_oid : col_oids_) inverse_map[col_oid_to_id[col_oid]] = col_oid;
+
+  // Populate the projection map using the in-order iterator on std::map
+  uint16_t i = 0;
+  for (auto &iter : inverse_map) projection_map_[iter.second] = i++;
```
Orankarl, line 98: Can we use std::sort here? From my own test of sorting 100,000 std::pair<int, int>, std::sort is 10%~20% faster than using std::map.

src/brain/operating_unit_recorder.cpp
rohanaggarwal7997: Fixed in latest commit, We have to discuss this more with william as this assert is not entirely correct.

src/execution/ast/ast_pretty_print.cpp
rohanaggarwal7997: Fixed in latest commit

src/execution/compiler/operator/cte_scan_translator.cpp
rohanaggarwal7997: It doesnt matter as you wont have 100,000 columns in a table. Also this is done to have similar api as SQL Table so that in future when it is replaced by a temp memory space, its easier to change.

src/execution/sema/sema_builtin.cpp
```diff
@@ -2606,7 +2739,7 @@ void Sema::CheckBuiltinCall(ast::CallExpr *call) {
       break;
     }
     default: {
-      UNREACHABLE("Unhandled builtin!");
```
tanujnay112, line 2609: Why was this taken out?

src/execution/executable_query.cpp
```diff
@@ -24,6 +25,8 @@ ExecutableQuery::ExecutableQuery(const common::ManagedPointer<planner::AbstractP
   compiler::CodeGen codegen(exec_ctx.Get());
   compiler::Compiler compiler(query_id_, &codegen, physical_plan.Get());
   auto root = compiler.Compile();
+  //  terrier::execution::ast::AstPrettyPrint::Dump(std::cout, root);
```
tanujnay112, line 28: Remove this

rohanaggarwal7997: this was kept specificallly so that people can uncomment this and print their tpl.

src/execution/sema/sema_builtin.cpp
rohanaggarwal7997: i believe this is a mistake

src/execution/executable_query.cpp
tanujnay112: oh I see ok

src/include/execution/sql/cte_scan_iterator.h
```diff
@@ -0,0 +1,66 @@
...
+
+  /**
+   * Returns the slot which was inserted in the table using the projected row
+   */
+
+  storage::TupleSlot TableInsert();
+
+  /**
+   * Return the iterator that will scan the temp table
+   */
```
tanujnay112, line 46: Extra documentation here?

src/execution/sql/cte_scan_iterator.cpp
```diff
@@ -0,0 +1,69 @@
...
+
+namespace terrier::execution::sql {
+
+parser::ConstantValueExpression DummyCVE() {
+  return parser::ConstantValueExpression(type::TransientValueFactory::GetInteger(0));
+}
+
+CteScanIterator::CteScanIterator(terrier::execution::exec::ExecutionContext *exec_ctx, uint32_t *schema_cols_type,
+                                 uint32_t num_schema_cols)
+    : exec_ctx_(exec_ctx), cte_table_oid_(static_cast<catalog::table_oid_t>(999)), table_redo_(nullptr) {
```
tanujnay112, line 14: Magic number? Can you make this a macro?

tanujnay112: How do you know 999 won't conflict with any other table_oid? Will that cause issues?

test/execution/cte_scan_test.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+  // The index iterator gives us the slots to update.
+  IndexIterator index_iter1{
+      exec_ctx_.get(), 1, !table_oid1, !index_oid1, col_oids.data(), static_cast<uint32_t>(col_oids.size())};
+  index_iter1.Init();
+
+  // Create cte_table
+  uint32_t cte_table_col_type[1] = {4};  // {INTEGER}
+
+  auto cte_scan = new terrier::execution::sql::CteScanIterator(exec_ctx_.get(), cte_table_col_type, 1);
```
tanujnay112, line 82: allocate these on the stack instead

apavlo:

PR #889, opened 2020-05-07
Add pg_statistic catalog

gengkev:
This creates a new catalog, `pg_statistic`, which contains some of the same basic information as stored in the Postgres catalog of the same name: https://www.postgresql.org/docs/9.5/catalog-pg-statistic.html.

The catalog contains a tuple for every column in a table. For now, only columns in actual tables (as opposed to indexes) are tracked. Whenever a table is created or dropped, a corresponding tuple in the pg_statistic catalog is added or removed accordingly.

A new method `GetTableStats` is added to the `DatabaseCatalog` class, which pulls statistics from the `pg_statistic` table. This method returns a `TableStats` object, so it is somewhat compatible with the existing `StatsStorage` interface.

However, the new API is not actually used in the codebase anywhere — the users of the `StatsStorage` API have not yet been updated at this time.

Questions for reviewers:
* Is `database_catalog.cpp` the right location to add functions inserting into, deleting from, and updating the `pg_statistic` catalog?
* Is `optimizer/stats_catalog_test.cpp` the right place to add a test for this? Maybe it should go under `catalog/`?
* Does an API returning a `TableStats` object make sense? (Is the `TableStats` object used immutably, or is it expected that updates to the object will be reflected in the actual storage?)
* The code for "find all `TupleSlot`s in `pg_statistics` matching a table OID" is verbose, and is sort of duplicated in multiple places. How can this be resolved? I've written a helper function in `stats_catalog_test` for ease of use inside that file, but what about inside `database_catalog.cpp`? Adding private helper functions to that class for this purpose doesn't seem to fit the type of code in that file.


src/catalog/catalog_accessor.cpp
```diff
@@ -178,6 +178,10 @@ common::ManagedPointer<execution::udf::UDFContext> CatalogAccessor::GetUDFContex
   return dbc_->GetUDFContext(txn_, proc_oid);
 }
 
+std::unique_ptr<optimizer::TableStats> CatalogAccessor::GetTableStats(table_oid_t table_id) {
```
rohanaggarwal7997, line 181: I do not understand why this is a unique pointer? Does the stats object is deleted as soon as it is fetched? Why isn't it a managed pointer owned by someone else? We should not create this object every time while doing query optimisation for every table.

test/optimizer/stats_catalog_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+    db_main_ = terrier::DBMain::Builder().SetUseGC(true).SetUseCatalog(true).Build();
+    txn_manager_ = db_main_->GetTransactionLayer()->GetTransactionManager();
+    catalog_ = db_main_->GetCatalogLayer()->GetCatalog();
+
+    auto *txn = txn_manager_->BeginTransaction();
+    db_ = catalog_->GetDatabaseOid(common::ManagedPointer(txn), catalog::DEFAULT_DATABASE);
+
+    // Ensure pg_statistic table exists with correct OID
+    auto accessor = catalog_->GetAccessor(common::ManagedPointer(txn), db_);
+    EXPECT_NE(accessor, nullptr);
```
rohanaggarwal7997, line 24: Why is this check required?

gengkev: It probably isn't, I just wrote this code after catalog_test which does the same check

src/catalog/catalog_accessor.cpp
gengkev: Thanks for the comments! This is definitely a good point in terms of efficiency, but I am not sure I can think of a significantly better solution.

In the existing interface, the StatsStorage object owns an unordered_map of containing TableStats objects, and thus looking up the statistics for a table is as simple as returning a `common::ManagedPointer<TableStats>`. However, since TableStats are only kept in memory, they are not persisted and will need to be recomputed each time after the database is restarted.

Instead, this commit stores the statistics data in the pg_statistics table. However, in order to work with the cost model code (which currently expects a TableStats object), this function will need to create a copy of the table data to fit into the TableStats object.

You're right that this will lead to more allocations than the existing StatsStorage API. However, I am not sure that there are better alternatives:

* The returned TableStats storage objects could be cached at some layer. However, this means that the ANALYZE command would need to invalidate that cache, which leads to tricky design issues, including lifetime issues that might require reference counting to resolve. For example, what if ANALYZE is running at the same time that a query is being executed on the same table — how can we ensure the statistics are read correctly? Solving these issues would lead to essentially re-implementing the storage layer: with the current approach, the guarantees of MVCC prevent us from having to worry about this.
* The cost model code could be changed to use a different API that requires less allocations. I am not familiar enough to say whether this is the case or not, but I am not sure it will be possible to make an API with significantly better performance characteristics in terms of allocations.

Given these options, I think it wouldn't be worth making things more complicated than they already are. Please let me know what you think, or if there's any other way the allocation cost can be reduced!

gengkev: (And in fact, I think some of those lifetime issues could possibly exist in the existing StatsStorage API, if the same TableStats object is being updated by one thread at the same time that it is being read by another...?)

src/catalog/postgres/builder.cpp
```diff
@@ -87,6 +88,7 @@ DatabaseCatalog *Builder::CreateDatabaseCatalog(
   dbc->constraints_ = new storage::SqlTable(block_store, Builder::GetConstraintTableSchema());
   dbc->languages_ = new storage::SqlTable(block_store, Builder::GetLanguageTableSchema());
   dbc->procs_ = new storage::SqlTable(block_store, Builder::GetProcTableSchema());
+  dbc->statistics_ = new storage::SqlTable(block_store, Builder::GetStatisticTableSchema());
```
rohanaggarwal7997, line 91: Use std::move in the second argument

test/optimizer/stats_catalog_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+  // Check that pg_statistic tuples are deleted when table is dropped
+  {
+    auto txn = txn_manager_->BeginTransaction();
+    auto accessor = catalog_->GetAccessor(common::ManagedPointer(txn), db_);
+    EXPECT_NE(accessor, nullptr);
+
+    // Drop table from storage
+    accessor->DropTable(table_oid);
+
+    // Search the statistics index for matching tuples
```
rohanaggarwal7997, line 209: This comment can be made far better. Plus you can tell what exactly does the function do?

test/optimizer/stats_catalog_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+  }
+
+  void VerifyStatsCatalog(const catalog::CatalogAccessor &accessor) {
+    auto ns_oid = accessor.GetNamespaceOid("pg_catalog");
+    EXPECT_EQ(ns_oid, catalog::postgres::NAMESPACE_CATALOG_NAMESPACE_OID);
+    auto table_oid = accessor.GetTableOid(ns_oid, "pg_statistic");
+    EXPECT_EQ(table_oid, catalog::postgres::STATISTIC_TABLE_OID);
+  }
+
+  void GetStatTuplesForTable(common::ManagedPointer<transaction::TransactionContext> txn,
```
rohanaggarwal7997, line 41: It would be helpful to desribe what exactly does the function do and why you are passing invalid column id in the second column of the projected row. I understand it after reading it, but the API should be explained

test/optimizer/stats_catalog_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+    // High key (class + 1, INVALID_COLUMN_OID)
+    *(reinterpret_cast<catalog::table_oid_t *>(pr_hi->AccessForceNotNull(oid_prm.at(catalog::indexkeycol_oid_t(1))))) =
+        next_oid;
+    *(reinterpret_cast<catalog::col_oid_t *>(pr_hi->AccessForceNotNull(oid_prm.at(catalog::indexkeycol_oid_t(2))))) =
+        catalog::col_oid_t(0);
+
+    statistics_oid_index_->ScanAscending(*txn, storage::index::ScanType::Closed, 2, pr_lo, pr_hi, 0, index_results);
+  }
+
+  std::unique_ptr<DBMain> db_main_;
```
rohanaggarwal7997, line 69: Some comments here about the class variables can be helpful

test/optimizer/stats_catalog_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+  catalog::db_oid_t db_;
+  common::ManagedPointer<storage::SqlTable> statistics_;
+  common::ManagedPointer<storage::index::Index> statistics_oid_index_;
+};
+
+/*
+ * Create and delete a user table.
+ */
+// NOLINTNEXTLINE
+TEST_F(StatsCatalogTests, StatisticTest) {
```
rohanaggarwal7997, line 81: You should also describe exactly what you are doing in the test.

test/optimizer/stats_catalog_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+  // Create a new table in the catalog, backed by a SqlTable
+  catalog::table_oid_t table_oid;
+  std::set<catalog::col_oid_t> col_oids;
+  {
+    auto txn = txn_manager_->BeginTransaction();
+    auto accessor = catalog_->GetAccessor(common::ManagedPointer(txn), db_);
+    EXPECT_NE(accessor, nullptr);
+
+    // Create the column definition (no OIDs)
```
rohanaggarwal7997, line 90: What does "no OIDs" mean?

test/optimizer/stats_catalog_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+    std::vector<catalog::Schema::Column> cols;
+    cols.emplace_back("id", type::TypeId::INTEGER, false,
+                      parser::ConstantValueExpression(type::TransientValueFactory::GetNull(type::TypeId::INTEGER)));
```
rohanaggarwal7997, line 93: I believe these should be columnValueExpressions not constant value expressions.

test/optimizer/stats_catalog_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+    cols.emplace_back("user_col_1", type::TypeId::INTEGER, false,
+                      parser::ConstantValueExpression(type::TransientValueFactory::GetNull(type::TypeId::INTEGER)));
+    auto tmp_schema = catalog::Schema(cols);
+
+    table_oid = accessor->CreateTable(accessor->GetDefaultNamespace(), "test_table", tmp_schema);
+    EXPECT_NE(table_oid, catalog::INVALID_TABLE_OID);
```
rohanaggarwal7997, line 99: Again this is a useless check. The table will always get created. Is this code copy-pasted?

test/optimizer/stats_catalog_test.cpp
```diff
@@ -0,0 +1,220 @@
...
+    auto schema = accessor->GetSchema(table_oid);
+
+    // Verify we can instantiate a storage object with the generated schema
+    auto table = new storage::SqlTable(db_main_->GetStorageLayer()->GetBlockStore(), schema);
```
rohanaggarwal7997, line 103: Where did you deallocate this table. This will remain on the heap. Does this test work without an ASAN?

rohanaggarwal7997: Ok I understand you are manually creating the table in the catalog. Why are you not using the executor in DDL executors? This is literally adding another codepath to create a table.

src/catalog/database_catalog.cpp
```diff
@@ -2557,6 +2589,200 @@ proc_oid_t DatabaseCatalog::GetProcOid(common::ManagedPointer<transaction::Trans
   return ret;
 }
 
+template <typename Column, typename ClassOid, typename ColOid>
+void DatabaseCatalog::CreateColumnStatistic(const common::ManagedPointer<transaction::TransactionContext> txn,
```
rohanaggarwal7997, line 2593: Why dont we have tests of all these functions. This is supposed to be the heart of the PR, while we never see these running in action. You created a separate table in the test and added things to it manually. You should use these functions, so in the future someone adding functionality in the optimizer can see the test and know how to do things.

apavlo:

PR #894, opened 2020-05-12
Query Rewriter

esargent28:
This PR contains the query rewriter with some rules implemented. The rules currently implemented are:

- Symmetric Reordering: switching order of arguments to symmetric operations so that match patterns don't depend on the ordering of symmetric operations' children
- Transitive Closure with Constant Values: "A = 1 AND A = B" --> "A = 1 AND 1 = B"
- Comparison Intersection: When comparing the same two expressions in two different branches of an AND, intersect them into one (example: "A <= B AND A != B" --> "A < B")

This builds off of the class inheritance structure in #835 and adds the core rewriter logic and some rules. As a result, #835 should be merged before this.


lmwnshn:
RIP query rewriter.

PR #897, opened 2020-05-15
Sequences (Final Code Drop)

zianke:
## Overview

This project implements the sequences feature, which works as global counters that can generate unique sequence numbers.

The sequences feature works similarly as the one in PostgreSQL.

* Support functionalities:
  * Create and drop sequences
  * `nextval` function
  * `currval` function
  * Parameters (`INCREMENT`, `START`, `MINVALUE`, `MAXVALUE`, `CYCLE`)
* Example usage:
```
CREATE SEQUENCE seq INCREMENT BY 2 MINVALUE 3 MAXVALUE 10 START WITH 4 CYCLE;
SELECT nextval('seq') FROM pg_sequence;  -- The FROM clause is required by the current DBMS, where pg_sequence can be replaced with any dummy table
SELECT currval('seq') FROM pg_sequence;
DROP SEQUENCE seq;
```

## Scope
The implementation has made changes in modules including parser, binder, optimizer, planner, execution, catalog, traffic_cop, network, recovery, junit.

We have updated these files:
* Parser:
  * `src/include/parser/parsenodes.h`
  * `src/include/parser/create_statement.h`
  * `src/include/parser/drop_statement.h`
  * `src/include/parser/postgresparser.h`
  * `src/parser/postgresparser.cpp`
  * `test/parser/parser_test.cpp`
* Binder:
  * `src/binder/bind_node_visitor.cpp`
  * `test/binder/binder_test.cpp`
* Optimizer:
  * `src/include/optimizer/optimizer_defs.h`
  * `src/optimizer/query_to_operator_transformer.cpp`
  * `src/include/optimizer/logical_operators.h`
  * `src/optimizer/logical_operators.cpp`
  * `src/include/optimizer/physical_operators.h`
  * `src/optimizer/physical_operators.cpp`
  * `src/include/optimizer/operator_visitor.h`
  * `src/include/optimizer/child_property_deriver.h`
  * `src/optimizer/child_property_deriver.cpp`
  * `src/include/optimizer/rule.h`
  * `src/optimizer/rule.cpp`
  * `src/include/optimizer/rules/implementation_rules.h`
  * `src/optimizer/rules/implementation_rules.cpp`
  * `src/include/optimizer/plan_generator.h`
  * `src/optimizer/plan_generator.cpp`
  * `test/optimizer/logical_operator_test.cpp`
  * `test/optimizer/physical_operator_test.cpp`
  * `test/optimizer/operator_transformer_test.cpp`
* Planner:
  * `src/include/planner/plannodes/plan_node_defs.h`
  * `src/include/planner/plannodes/create_sequence_plan_node.h`
  * `src/planner/plannodes/create_sequence_plan_node.cpp`
  * `src/include/planner/plannodes/drop_sequence_plan_node.h`
  * `src/planner/plannodes/drop_sequence_plan_node.cpp`
  * `src/include/planner/plannodes/plan_visitor.h`
* Catalog:
  * `src/include/catalog/postgres/pg_sequence.h`
  * `src/include/catalog/catalog_defs.h`
  * `src/include/catalog/catalog_accessor.h`
  * `src/catalog/catalog_accessor.cpp`
  * `src/include/catalog/postgres/builder.h`
  * `src/catalog/postgres/builder.cpp`
  * `src/include/catalog/database_catalog.h`
  * `src/catalog/database_catalog.cpp`
  * `test/catalog/catalog_test.cpp`
  * `src/include/catalog/postgres/pg_proc.h`
* Execution:
  * `src/include/execution/ast/builtins.h`
  * `src/execution/sema/sema_builtin.cpp`
  * `src/include/execution/sql/functions/string_functions.h`
  * `src/execution/sql/functions/string_functions.cpp`
  * `src/include/execution/vm/bytecodes.h`
  * `src/include/execution/vm/bytecode_handlers.h`
  * `src/execution/vm/bytecode_generator.cpp`
  * `src/execution/vm/vm.cpp`
  * `src/include/execution/exec/execution_context.h`
  * `src/include/execution/sql/ddl_executors.h`
  * `src/execution/sql/ddl_executors.cpp`
  * `test/execution/ddl_executors_test.cpp`
* TrafficCop:
  * `src/include/traffic_cop/traffic_cop_defs.h`
  * `src/traffic_cop/traffic_cop_util.cpp`
  * `src/traffic_cop/traffic_cop.cpp`
* Network
  * `src/include/network/network_defs.h`
  * `src/include/network/postgres/postgres_packet_writer.h`
  * `src/include/network/connection_context.h`
  * `src/network/postgres/postgres_protocol_interpreter.cpp`
* Recovery
  * `src/include/storage/recovery/recovery_manager.h`
  * `src/storage/recovery/recovery_manager.cpp`
* JUnit
  * `script/testing/junit/src/SequenceTest.java`

## Architectural Design
When initializing the database, a new catalog named `pg_sequence` is created, which stores the sequence metadata, including the value of sequence and its parameters. A temp table is created for each new connection under the temp namespace, which is used to store the session-local sequence values.

The `CREATE SEQUENCE` pipeline starts with parsing query into a statement, by reusing the Postgres parser which can parse the sequence name and parameters. After binding it, we send the parsed results into optimizer and generate a plan node. The plan node is then executed and insert a new entry into `pg_class`, as well as a new entry in `pg_sequence` with a foreign key referencing the `pg_class` entry.

The `DROP SEQUENCE` pipeline will go through the same parsing and execution process, and delete the corresponding entries from `pg_class` and `pg_sequence`.

The built-in function `nextval` takes a sequence name as input, and output the next sequence value. It first retrieves the corresponding entry from `pg_sequence`. Based on the value of `seqlastval` field, and other parameters such as `seqincrement`, it will increment the sequence. The new sequence value will be updated into the `pg_sequence` entry, and also updated into the corresponding entry in the session-local temp table.

The built-in function `currval` takes a sequence name as input, and output the last value returned by `nextval` in the current session. If `nextval` has never been called, then the session-local temp table does not contain an entry for the sequence, and calling `currval` will throw an exception. If `nextval` has already been called, then `currval` will directly read the latest sequence value from the temp table.

Both `nextval` and `currval` will commit a mini transaction during its execution. As a result, even if the query calling `nextval` is aborted, the sequence increment will not rollback, and `currval` will still return the updated value.
 
## Design Rationale
The goal of this design is to keep the create and drop pipeline similar to other classes, such as table and index. A sequence’s OID and name are stored in the `pg_class` catalog, just as table and index. Its metadata are stored in `pg_sequence`, with a foreign key referencing the OID in `pg_class`. We have reused some code of the exisiting create/drop pipelines in our implementation. The design of sequence unit tests also takes those of index as a reference.

For `nextval` and `currval` function, in order to support mini transactions, we have passed the transaction manager into execution context (as an optional argument of its constructor, so the existing code base won't be affected), and the execution context is then passed to the built-in functions. During the execution of `nextval` and `currval`, we begin a new transaction and commit it once the function logics are completed.

To support session-local `currval`, we decide to use a temp table for each session. It's created under the existing temp namespace, which is created when a new connection is established and deleted when the connection is closed. Therefore, the temp table will also be dropped during the session tear down. For each function call of `nextval`, the new sequence value is updated into both `pg_sequence` as a global latest value, and temp table as the result of next `currval` call. 

## Testing Plan
We have added unit tests for each part of the pipeline, which currently include `parser_test`, `binder_test`, `logical_operator_test`, `physical_operator_test`, `operator_transformer_test`, `ddl_executors_test`, `catalog_test`.

We have also added JUnit integration tests in `SequenceTest.java`, which include the basic usage of sequences, sequences with parameters, behaviors with multiple connections, and behaviors with transaction rollback.

## Trade-offs and Potential Problems
1. Our current implementation creates the temp table immediately after the temp namespace is created. In other words, a temp table is created for each connection, even if it won't use sequences in its session. This problem can affect the performance of DBMS when a large number of connections need to be created. A better approach could be to create a temp table when `nextval` is called for the first time. However, we find this hard to achieve due to the limited context a built-in function can access.

2. In `currval` function we need to scan through the temp table to get the desired value. We have assumed that for each session, there won't be a large number of different sequences, so it would be sufficient to do a sequential scan. However, it might be better to build an index for the temp table to speed up the searching process when the number of sequences grows larger.

## Future Work
1. Implement other sequence parameters like `cache` to achieve the same behavior as in the PostgreSQL [documentation](https://www.postgresql.org/docs/12/sql-createsequence.html).


apavlo:

PR #899, opened 2020-05-15
Cost Model (final)

vivsy:
Final code drop for cost model project.

Design doc: https://github.com/vivsy/terrier/wiki/Statistics-&-Cost-Model-Design-Doc

PR #900, opened 2020-05-15
Add translator for ANALYZE

gengkev:
Important caveats:

* This PR builds off of #889, which should arguably be considered first since it is fairly self-contained, has gone through (some) review, and is noticeably more complete.
* This PR is incomplete, and is more of a "minimum viable product". _ANALYZE is currently only supported for one table and one column at a time._ The code quality could certainly use improvement. And the test coverage is really not there.
* For the final design document, see #899 (part of the same 15-721 group).

Summary of notable changes:
* **Plan generation:**
  * AnalyzeStatement is converted to a LogicalAnalyze node with two LogicalGet children.
  * The left child gets tuples from the source table, likely a sequential scan.
  * The right child gets tuples from `pg_statistic` for update, likely an index scan.
* **Translation:**
  * The AnalyzeTranslator is split into a "bottom" and "top" translator.
  * The bottom is almost exactly equivalent to StaticAggregateBottomTranslator, which computes aggregations on the tuples from the left child (source table).
  * The top is similar to UpdateTranslator, as it inserts tuples into the pg_statistic table based on the aggregation results.
* **TopKAggregate:**
  * The design taken in #704 (squashed into two separate commits in this PR) is significantly modified; the commit message for "Redo TopKAggregate to behave like other aggregates" explains why.
  * The aggregate now has the same interface as other aggregates, with a Result() method that returns one value.
  * The value returned is a serialized TopKElements object, that can be directly inserted into the pg_statistic table without further processing.
  * The `GetTableStats()` function now populates the `common_values` and `common_freqs` fields in `ColumnStats` using the serialized TopKElements object.
* **Other "collateral damage":**
  * `ColumnStats` API is changed to reflect the view that it is an immutable object generated from the `pg_statistics` table. This may affect tests.
  * All builtin `@aggResult` functions now have `*ExecutionContext` as its first argument. This was necessary for the TopKAggregate to allocate a `sql::StringVal` to store serialized data. This breaks several TPL tests that use this builtin.
  * CountStarAggregate is now usable, at least internally. This may need more tests.
  * Some sanity checks in the form of TERRIER_ASSERT were added to various strange portions of the code, e.g. bwtree.cpp, optimizer.cpp that are not really related to this PR. Hopefully, they will be helpful, rather than incorrect.

TODOs:

- [ ] Get ANALYZE to be runnable from psql (something in the network layer is blocking this)
- [ ] Write actual tests!!!
- [ ] Fix tests that were broken due to "collateral damage" above.
- [ ] Add support for running ANALYZE on multiple (or all) columns in a table.
- [ ] Move more logic into the binder: e.g. stop creating a large number of `ColumnValueExpression`s inside of `InputColumnDeriver` instead of in the binder.
- [ ] Update `AggregateHelper` to take in aggregate and group-by expressions separately, so we don't have to construct a fake `AggregatePlanNode`.
- [ ] Add support for running ANALYZE on all tables in the database (no idea how to do this)


apavlo:

Issue #813, opened 2020-03-04
Switch to Dynamic Linking


apavlo:
We are using static linking for all of the tests + benchmarks. That makes the linking super long. IIRC, there was a problem with using ASAN on OSX with dynamic linked binaries, but I don't think it is an issue any more.

I think it's just a matter of making a switch in our cmake files to enable dynamic linking.

lmwnshn:
Will be addressed by #1224 in two ways: lld for faster linking, and actual proper dynamic linking made possible by splitting up dependencies.

It was a _little_ more than just making a switch.

PR #916, opened 2020-05-24
Add support for compiling shared library

gengkev:
This is actually my 3rd attempt at getting Terrier to compile as a shared library, borne out of getting very angry at compile/link times during 15-721, and it seems to work! The process of linking the final executable is now almost instant, and binaries like `./terrier` and `catalog_test` are now an order of magnitude smaller, which is a blessing for my tiny SSD.

The main issue was that .o files were being compiled with `-fvisibility=hidden`, which sets the default visibility of all symbols to be hidden. This caused those symbols to disappear when being linked into the `.so` file. That flag might makes sense for a true library where the number of symbols that are actually exported is small and can all be tagged manually. But it doesn't make sense for libterrier, where the `./terrier` binary, tests, etc. all need to call various functions all over the codebase.

**How was this investigated?** `nm -C libterrier.so.1.0.0` showed the function `DBMain::ForceShutdown()` as `t` (local symbol), whereas compiling a toy project on my own would show it as `T` instead (global symbol). So I was able to narrow down the difference to the `-fvisibility` command-line flag.

**Why did the static library work?** Because it seems that `-fvisibility` only affects the symbols that are visible when linked into a shared library, whereas an archive file would be unaffected.

**Caveat:** I tested this on Ubuntu 18.04 only (I do not use Mac OS). I am sure there is a reason the `-fvisibility=hidden` flag was introduced (in 4e1835b67b97ee3b4031b03b456fd54a46ba69a8). So it might be worth looking into why that is, for example if this breaks the Mac OS build for some reason.

**Caveat:** Is it necessary to change additional files (e.g. use of `STATIC_LINK_LIBS` in [cmake_modules/BuildUtils.cmake](https://github.com/cmu-db/terrier/tree/master/cmake_modules/BuildUtils.cmake#L285))? I don't know if I understand the build system well enough to say. It could also be possible to compile other helper libraries as static (test_util, etc).


gengkev:
related to #813 probably

mbutrovich:
This does indeed break the macOS build. I believe you need explicit symbol exporting to fix it which is a non-trivial refactor to keep unit tests working.

See https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/CppRuntimeEnv/Articles/SymbolVisibility.html

mbutrovich:
This does seem to break the macOS build with warnings blowing up from the linker like crazy, which happened the first time @amlatyrngom ported TPL and resulted in the flag change:
```
ld: warning: direct access in function '(anonymous namespace)::ValueTable::lookupOrAdd(llvm::Value*)' from file '/usr/local/Cellar/llvm@8/8.0.1_1/lib/libLLVMScalarOpts.a(GVNSink.cpp.o)' to global weak symbol 'guard variable for llvm::hashing::detail::get_execution_seed()::seed' from file 'CMakeFiles/terrier_objlib.dir/execution/ast/context.cpp.o' means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.
```
The system still compiles and runs, but it's unclear if we're always getting the behavior we expect.

lmwnshn:
This seems to be a small (but tricky) patch. Ideally it gets fixed for Mac as well, but if not, I wonder if it is worth having different linking behavior on Mac and Ubuntu? The advertised changes are very tempting. I might develop locally with this on Ubuntu, right now linking is a 50/50 on whether it causes my computer to stop responding.

lmwnshn:
Conversation with @gengkev 

```
We think we want our final goal to be statically link the terrier binary (for performance) and dynamically link tests (save space, save linking time)

A possible stopgap measure is to statically link on mac, dynamically link on ubuntu, though we're curious about what it would take to go from there to the envisioned solution.
```

In case someone else wants to pick it up.

apavlo:
Replaced with #1017

PR #920, opened 2020-05-26
Add a virtual destructor to Val

mbutrovich:
Fix ASAN issue on GCC when running the terrier binary.


mbutrovich:
Closed since it sounds like we'll change to a variant in CVE instead.

PR #919, opened 2020-05-26
Fixes for Jenkinsfile (May 2020)

apavlo:
* Reenable ASAN correctly for the 'Test' stage.
* Added timeouts for the junit + oltpbench tests
* Reenable tests in the 'Test' stage

https://youtu.be/rtNlJ24QV5s

PR #925, opened 2020-05-28
Change CVE's underlying value to be a std::variant

mbutrovich:
Based on @pmenon's feedback.

**Major changes:**
- CVE now holds a `std::variant` to an execution engine value, rather than a `std::unique_ptr` This removes the need for a virtual destructor to clean up the value properly, which we don't want because we'd like to keep the value.h classes POD.
- Added a `Peek<>()` method to CVE to quickly extract C++ types.
- Lots of mindless signature refactors.
- Jenkinsfile from #919 that re-enables some missing checks.

This fixes an issue where we couldn't start the server when compiled with ASAN on due to a destructor mismatch on the held value.

### DIFF
--- a/src/parser/expression/constant_value_expression.cpp
+++ b/src/parser/expression/constant_value_expression.cpp
@@ -67,22 +67,28 @@ template ConstantValueExpression::ConstantValueExpression(const type::TypeId typ
 
 template <typename T>
 T ConstantValueExpression::Peek() const {
+  // NOLINTNEXTLINE: bugprone-suspicious-semicolon: seems like a false positive because of constexpr
   if constexpr (std::is_same_v<T, bool>) {
     return static_cast<T>(GetBoolVal().val_);
   }
+  // NOLINTNEXTLINE: bugprone-suspicious-semicolon: seems like a false positive because of constexpr
   if constexpr (std::is_same_v<T, int8_t> || std::is_same_v<T, int16_t> || std::is_same_v<T, int32_t> ||
                 std::is_same_v<T, int64_t>) {
     return GetInteger().val_;
   }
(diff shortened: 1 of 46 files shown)


src/parser/expression/constant_value_expression.cpp
```diff
@@ -12,24 +12,77 @@
...
+  }
+  // NOLINTNEXTLINE: bugprone-suspicious-semicolon: seems like a false positive because of constexpr
+  if constexpr (std::is_same_v<T, execution::sql::Timestamp>) {
+    return GetTimestampVal().val_;
+  }
+  // NOLINTNEXTLINE: bugprone-suspicious-semicolon: seems like a false positive because of constexpr
+  if constexpr (std::is_same_v<T, std::string_view>) {
+    return std::get<execution::sql::StringVal>(value_).StringView();
+  }
+  UNREACHABLE("Invalid type for GetAs.");
```
lmwnshn, line 85: for Peek?

src/parser/expression/constant_value_expression.cpp
```diff
@@ -42,60 +95,32 @@ ConstantValueExpression &ConstantValueExpression::operator=(const ConstantValueE
...
 }
 
+ConstantValueExpression::ConstantValueExpression(const ConstantValueExpression &other) : AbstractExpression(other) {
+  if (std::holds_alternative<execution::sql::StringVal>(other.value_)) {
+    auto string_val = execution::sql::ValueUtil::CreateStringVal(other.GetStringVal());
+
+    value_ = string_val.first;
+    buffer_ = std::move(string_val.second);
+  } else {
+    value_ = other.value_;
```
lmwnshn, line 119: buffer_ = nullptr here?

lmwnshn: I just realized this is the constructor.

PR #929, opened 2020-05-29
Change the default log serialization and persist intervals.

malin1993ml:
Making them 1ms for now.


src/include/settings/settings_defs.h
```diff
@@ -120,8 +120,8 @@ SETTING_int(
 // Log file persisting interval
 SETTING_int(
     log_persist_interval,
-    "Log file persisiting interval (ms) (default: 10)",
-    10,
+    "Log file persisiting interval (ms) (default: 1)",
```
lmwnshn, line 123: Typo.. ah, we can fix it in the future.

PR #930, opened 2020-05-30
[issue-822] save CTest output in Jenkins builds

### LINKED ISSUE #822: Retain CTest Output Files
After we run `make unittest` in our Jenkins pipeline, we need to save the output files so that we can go back and see why things are failing. Right now I think we are just blowing all the files away.

crd477:
Fix #822

Save CTest output as artifacts per #822


script/jenkins/benchmark-agent-setup.sh
```diff
@@ -50,8 +51,8 @@ chmod 1777 /mnt/ramdisk
 
 # setup nat
 route delete default
-route add default gw 10.92.0.4
-route add -net 128.2.0.0/16 gw 10.92.0.1
+route add default gw 10.111.0.4
```
apavlo, line 54: I'm assuming this is internal stuff that is needed?

test/CTestCustom.cmake
```diff
@@ -0,0 +1 @@
+set(CTEST_CUSTOM_MAXIMUM_PASSED_TEST_OUTPUT_SIZE 8192)
```
apavlo, line 1: What does this need to be in a separate file? Why not put this one line in `test/CMakeLists.txt`?

If we need to have this second file, then you should add a comment at the top of it that explains what it does and when would anyone ever need to modify it.

script/jenkins/benchmark-agent-setup.sh
crd477: Yes, it's actually unrelated to this PR but is an intended commit anyway.  I think I had it in place from a previous branch.

test/CTestCustom.cmake
crd477: My interpretation of the relevant [CMake documentation](https://cmake.org/cmake/help/v3.4/variable/CTEST_CUSTOM_MAXIMUM_PASSED_TEST_OUTPUT_SIZE.html) was that it needed to be there.
I'll test setting it in `test/CMakeLists.txt` and comment it in either case.

crd477: Confirmed that it cannot be set in `test/CMakeLists.txt`.  I'm glad you called attention to this though because I discovered that I missed two tests that generated more than 8kB of output.  I'll update accordingly in a few moments.

PR #933, opened 2020-06-01
Extended query fixes

### LINKED ISSUE #926: Junit tests failed with extended query mode
for the [pr-921](https://github.com/cmu-db/terrier/pull/921), after allowing command line option for query mode for the junit testing, following error log showed up if to `run python3 run_junit.py --query-mode extended` or `python3 run_junit.py`  with no query mode specified
```
BUILD FAILED
/repo/script/testing/junit/build.xml:40: Java returned: 1
Total time: 7 seconds
[2020-05-27 00:10:30.233] [network_logger] [info] Listening on port 15721
terrier: /repo/src/catalog/database_catalog.cpp:686: bool terrier::catalog::DatabaseCatalog::DeleteColumns(terrier::common::ManagedPointer<terrier::transaction::TransactionContext>, ClassOid) [with Column = terrier::catalog::Schema::Column; ClassOid = terrier::common::StrongTypeAlias<terrier::catalog::tags::table_oid_t_typedef_tag, unsigned int>]: Assertion `(!index_results.empty()) && ("Incorrect number of results from index scan. empty() implies that function was called with an oid " "that doesn't exist in the Catalog, but binding somehow succeeded. That doesn't make sense.")' failed.
Exception trying to run junit tests
```

mbutrovich:
Fix #926.

Fixes:
- Change to use query text hashing instead of pg_fingerprint for the key in the query cache. Different queries could have the same fingerprint with different values, which resulted in DML queries with different values (but not as params) being reused. That's not right.
- Doubles as parameters were being read incorrectly.

Note we can now pass all of the JUnit tests in extended query mode again. I do this via modifying TestUtility.java for now:
```
    public static Connection makeConnection(String host, int port, String username) throws SQLException {
        String url = String.format("jdbc:postgresql://%s:%d/", host, port);
        Properties props = new Properties();
        props.setProperty("user", username);
//         props.setProperty("preferQueryMode", "simple"); // force SimpleQuery protocol for now
//         props.setProperty("prepareThreshold", "0"); // suppress switchover to binary protocol
        Connection conn = DriverManager.getConnection(url, props);
        return conn;
    }

```
I also changed the WireTest JUnit test to insert more tuples to encourage binary format.

### DIFF
--- a/src/include/network/network_io_utils.h
+++ b/src/include/network/network_io_utils.h
@@ -142,7 +142,7 @@ class ReadBufferView {
    * to the caller to ensure that there are enough bytes available in the read
    * buffer at this point.
    * @tparam T type of value to read off. Has to be size 1, 2, 4, or 8.
-   * @return value of integer switched from network byte order
+   * @return value of numeric switched from network byte order
    */
   template <typename T>
   T ReadValue() {
@@ -164,7 +164,7 @@ class ReadBufferView {
         }
           // Will never be here due to compiler optimization
         default:
-          throw NETWORK_PROCESS_EXCEPTION("invalid size for floating point");
+          throw NETWORK_PROCESS_EXCEPTION("Invalid size for floating point.");
       }
     } else {  // NOLINT: false positive on indentation with clang-tidy, fixed in upstream check-clang-tidy
       const auto val = ReadRawValue<T>();
@@ -179,7 +179,7 @@ class ReadBufferView {
           return static_cast<T>(be64toh(static_cast<uint64_t>(val)));
           // Will never be here due to compiler optimization
         default:
-          throw NETWORK_PROCESS_EXCEPTION("invalid size for integer");
+          throw NETWORK_PROCESS_EXCEPTION("Invalid size for integer.");
       }
     }
   }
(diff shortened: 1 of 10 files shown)


src/include/network/network_io_utils.h
```diff
@@ -149,7 +149,7 @@ class ReadBufferView {
     // We only want to allow for certain type sizes to be used
     // After the static assert, the compiler should be smart enough to throw
     // away the other cases and only leave the relevant return statement.
-    static_assert(sizeof(T) == 1 || sizeof(T) == 2 || sizeof(T) == 4 || sizeof(T) == 8, "Invalid size for integer");
+    static_assert(sizeof(T) == 1 || sizeof(T) == 2 || sizeof(T) == 4 || sizeof(T) == 8, "Invalid size for numeric.");
```
thepinetree, line 152: Ah, sorry, I meant up in the documentation for the function with @return too, couldn't select that line

jazzplato:
One small question, for the junit side code below
```java
public static Connection makeConnection(String host, int port, String username) throws SQLException {
        String url = String.format("jdbc:postgresql://%s:%d/", host, port);
        Properties props = new Properties();
        props.setProperty("user", username);
//         props.setProperty("preferQueryMode", "simple"); // force SimpleQuery protocol for now
//         props.setProperty("prepareThreshold", "0"); // suppress switchover to binary protocol
        Connection conn = DriverManager.getConnection(url, props);
        return conn;
    }
```
does it mean the `prepareThreshold` is also optional (or should not be set to any value by default) like `preferQueryMode`?

mbutrovich:
`prepareThreshold` is optional and only applies under extended query mode.

jazzplato:
great, thx!

PR #934, opened 2020-06-01
Abort running txn on disconnect

mbutrovich:
If the client disconnects and there is an open txn, abort it. Otherwise the system could end up in a bad state where that txn is never GC'd and version chains never get cleaned up.

**Major changes:**
- Extended `Teardown` logic in `PostgresProtocolInterpreter` to abort any running txn belonging to its `ConnectionContext`.
- JUnit test case that fails on current master to show the issue is fixed

### DIFF
--- a/script/testing/junit/src/TrafficCopTest.java
+++ b/script/testing/junit/src/TrafficCopTest.java
@@ -37,25 +37,32 @@ public class TrafficCopTest extends TestUtility {
   @Test
   public void test_DisconnectAbort() throws SQLException {
 
+   // create another connection that will take the write lock on a tuple, forcing an about on the default connection's
+   // createStatement
    Connection second_conn = makeDefaultConnection();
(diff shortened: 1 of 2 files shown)


script/testing/junit/src/TrafficCopTest.java
```diff
@@ -31,6 +31,45 @@ public void test_EmptyStatement() throws SQLException {
   stmt.execute("");
  }
 
+ /**
+   * DisconnectAbortTest
+   */
+  @Test
+  public void test_DisconnectAbort() throws SQLException {
+
+   // create another connection that will take the write lock on a tuple, forcing an about on the default connection's
```
lmwnshn: typo: abort

lmwnshn: Nvm. xd

PR #940, opened 2020-06-04
Improving compilation time

abalakumar083:
Remove misc. header dependencies to improve compilation time. 

* Remove arrow_serializer.h from data_table.h, also remove flatbuf namespace

* Remove database_catalog.h from catalog.h 

* Remove <functional> from managed_pointer.h 

* Remove exception.h from expression_defs.h 

* Seperate gate.h into cpp file and header file in order to not include immintrin.h

* Remove date/date.h from all header files - this requires splitting  src/include/util/time_util.h  and test/include/execution/compiler/expression_util.h into header and cpp files

* Removed disk_log_consumer_task.h, log_serializer_task.h, and transaction_defs.h from log_manager.h

* Remove storage/storage_defs.h from files which don't use anything from it


src/include/common/gate.h
```diff
@@ -33,11 +32,7 @@ class Gate {
    * Traverses the gate unless there are currently locks emplaced.  If there
    * are locks on the gate, spin until its free.
    */
-  void Traverse() {
```
mbutrovich, line 36: This is going to prevent inlining on what's ultimately a pretty short function on the txn begin critical path. We should discuss if we want this change and if there's another way to arrange the includes to make the <immintrin.h> less painful.

abalakumar083: One easy fix is to include emmintrin.h instead of immintrin.h, which is a much smaller header. I'll run it and find out the difference in time.

abalakumar083:
One issue is that I broke apart "src/include/util/time_util.h" and "test/include/execution/compiler/expression_util.h" into header and cpp files. But when I named the cpp files, src/util/time_util.cpp and test/execution/compiler/expression_util.cpp, I don't think the cpp files were actually getting linked properly. 

I think is because I was also trying to create a new sublevel directory. Is there something I need to do to the makefile in order for them to get linked?

Currently, I put the time_util.cpp file into common, and the expression_util.cpp file into test/execution.

apavlo:
> Currently, I put the time_util.cpp file into common, and the expression_util.cpp file into test/execution.

@abalakumar083 Unless @lmwnshn objects, I am okay with this.

lmwnshn:
No objection there.

PR #944, opened 2020-06-08
Switch DataTable's blocks_ from list to vector for ParallelScan.

lmwnshn:
We use TBB for parallel scan, which expects ranges of blocks. The current DataTable implementation is a `std::list<RawBlock *>` which does not allow for this sensibly.

We used to use `vector<RawBlock *>` and switched to `list<RawBlock *>` when we introduced the column API. Do you remember why this change was made? (Is it just that list iterators behave nicer than vector iterators?)

If the performance hit is significant, we can explore a couple of options:

- a custom data structure (for example something structurally like a chained hash table)
- maintaining both a list and a vector, updating the vector on-demand only on parallel scans / in a background thread / something lazy like that

But we should probably go with the simple thing if we can.

I unfortunately cannot run `perf` right now. Custom kernel doesn't support it and my hardware is meh anyway. If you're able to this would be great, otherwise I'm hoping microbenchmark will catch it.


src/storage/arrow_serializer.cpp
```diff
@@ -164,7 +164,7 @@ void ArrowSerializer::ExportTable(const std::string &file_name, std::vector<type
   const BlockLayout &layout = data_table_.accessor_.GetBlockLayout();
   auto column_ids = layout.AllColumns();
   data_table_.blocks_latch_.Lock();
-  std::list<RawBlock *> tmp_blocks = data_table_.blocks_;
+  std::vector<RawBlock *> tmp_blocks = data_table_.blocks_;
```
lmwnshn, line 167: I'm surprised we do a copy here.

src/storage/data_table.cpp
```diff
@@ -83,13 +83,13 @@ DataTable::SlotIterator DataTable::end() const {  // NOLINT for STL name compabi
...
-  if (blocks_.empty()) return {this, blocks_.end(), 0};
+  if (blocks_.empty()) return {this, blocks_.end(), blocks_.end(), 0};
   auto last_block = --blocks_.end();
   uint32_t insert_head = (*last_block)->GetInsertHead();
   // Last block is full, return the default end iterator that doesn't point to anything
-  if (insert_head == accessor_.GetBlockLayout().NumSlots()) return {this, blocks_.end(), 0};
+  if (insert_head == accessor_.GetBlockLayout().NumSlots()) return {this, blocks_.end(), blocks_.end(), 0};
   // Otherwise, insert head points to the slot that will be inserted next, which would be exactly what we want.
-  return {this, last_block, insert_head};
+  return {this, last_block, last_block, insert_head};
```
lmwnshn, line 92: Brain isn't working right now and I'm a bit suspicious of this line, could you double-check this? Thanks!

lmwnshn: heh.

src/storage/data_table.cpp
```diff
@@ -184,29 +201,34 @@ TupleSlot DataTable::Insert(const common::ManagedPointer<transaction::Transactio
...
       // insert block
       blocks_.push_back(new_block);
-      block = --blocks_.end();
+      block = new_block;
       break;
     }
 
-    if (accessor_.SetBlockBusyStatus(*block)) {
+    {
+      common::SpinLatch::ScopedSpinLatch guard(&blocks_latch_);
```
lmwnshn, line 209: Taking the latch here is concerning to me. Definitely look out for microbenchmark numbers on this one.

mbutrovich:
Fairly large regression on Insert and Scan performance at the storage layer. Need to investigate.

src/storage/data_table.cpp
```diff
@@ -68,11 +69,18 @@ DataTable::SlotIterator &DataTable::SlotIterator::operator++() {
...
+      num_advances_ = num_advances_ == SlotIterator::TILL_END ? num_advances_ : num_advances_ - 1;
+      // Cannot dereference if the next block is end(), so just use nullptr to denote
+      current_slot_ = {block_index_ >= table_->blocks_.size() ? nullptr : table_->blocks_[block_index_], 0};
+    } else {
+      // Done advancing, time to give up.
+      current_slot_ = {nullptr, 0};
+    }
   } else {
-    current_slot_ = {*block_, current_slot_.GetOffset() + 1};
+    current_slot_ = {table_->blocks_[block_index_], current_slot_.GetOffset() + 1};
```
jrolli, line 83: The "block" for this tuple slot is not changing so there is no reason to hit the vector and take a possible cache miss inside a critical section.  It would be better to use `current_slot_.GetBlock()` which is pretty much guaranteed to be a cache hit (if not already in a register).

src/storage/data_table.cpp
```diff
@@ -68,11 +69,18 @@ DataTable::SlotIterator &DataTable::SlotIterator::operator++() {
   common::SpinLatch::ScopedSpinLatch guard(&table_->blocks_latch_);
```
jrolli, line 69: We should defer acquiring this latch until absolutely necessary (i.e. referencing the vector on line 77).  Everything else is unnecessary execution in a critical section and adds potential for extra branch mispredictions while holding the latch.

src/storage/data_table.cpp
```diff
@@ -68,11 +69,18 @@ DataTable::SlotIterator &DataTable::SlotIterator::operator++() {
...
   if (current_slot_.GetOffset() == table_->accessor_.GetBlockLayout().NumSlots() - 1) {
-    ++block_;
-    // Cannot dereference if the next block is end(), so just use nullptr to denote
-    current_slot_ = {block_ == table_->blocks_.end() ? nullptr : *block_, 0};
+    if (num_advances_ > 0 || num_advances_ == SlotIterator::TILL_END) {
+      // Advance to the next block.
+      ++block_index_;
+      num_advances_ = num_advances_ == SlotIterator::TILL_END ? num_advances_ : num_advances_ - 1;
+      // Cannot dereference if the next block is end(), so just use nullptr to denote
+      current_slot_ = {block_index_ >= table_->blocks_.size() ? nullptr : table_->blocks_[block_index_], 0};
```
jrolli, line 77: This should be the only line of code that needs to be guarded by `blocks_latch_`.

src/storage/data_table.cpp
```diff
@@ -68,11 +69,18 @@ DataTable::SlotIterator &DataTable::SlotIterator::operator++() {
...
+    } else {
+      // Done advancing, time to give up.
+      current_slot_ = {nullptr, 0};
```
jrolli, line 80: The `nullptr` here is extremely opaque and makes it very hard to understand why this works given the updated constructor for `SlotIterator`.  We should use create a constant for `{nullptr, 0}` to have a clearer name for what this tuple-slot represents.

src/include/storage/data_table.h
```diff
@@ -94,18 +93,26 @@ class DataTable {
...
+
     /**
      * @warning MUST BE CALLED ONLY WHEN CALLER HOLDS LOCK TO THE LIST OF RAW BLOCKS IN THE DATA TABLE
      */
-    SlotIterator(const DataTable *table, std::list<RawBlock *>::const_iterator block, uint32_t offset_in_block)
-        : table_(table), block_(block) {
-      current_slot_ = {block == table->blocks_.end() ? nullptr : *block, offset_in_block};
+    SlotIterator(const DataTable *table, uint32_t block_index, int32_t num_advances, uint32_t offset_in_block)
+        : table_(table), block_index_(block_index), num_advances_(num_advances) {
+      current_slot_ = {block_index >= table_->blocks_.size() ? nullptr : table->blocks_[block_index], offset_in_block};
```
jrolli, line 105: Recommend adding an assertion here that effectively asserts that `offset_in_block == 0` when `nullptr` is used as the block pointer.

src/storage/data_table.cpp
```diff
@@ -137,23 +155,20 @@ bool DataTable::Update(const common::ManagedPointer<transaction::TransactionCont
   return true;
 }
 
-void DataTable::CheckMoveHead(std::list<RawBlock *>::iterator block) {
-  // Assume block is full
+void DataTable::CheckMoveHead(uint32_t block_index) {
+  // Assume block is full.
   common::SpinLatch::ScopedSpinLatch guard_head(&header_latch_);
```
jrolli, line 160: It seems like this is an unnecessary latch since the insertion head is now an atomic integer rather than a list iterator.

src/storage/data_table.cpp
jrolli: You have to take the latch if you're going to do an index lookup on the vector (otherwise it could migrate underneath your dereference because of an insert.

lmwnshn:
Subsumed into #994

PR #948, opened 2020-06-09
Extract and store the OLTPbench results for nightly build

### LINKED ISSUE #937: Convert the nightly build into a Jenkinsfile
The nightly build is currently just a script written on Jenkins. The MSE Team wants to convert it into a file stored in the repo so that we could have `Stage Step` in our nightly build that runs OLTPbench and other jobs in the future.

The way to do that is to add another `Jenkinsfile` (with a different name, maybe `Jenkinsfile-nightly`?) And configure the [terrier-nightly pipeline](http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier-nightly/activity). In the `Jenkins Configure - Build Configuration`, we change the Script Path:

Screenshot from 2020-06-03 00-46-49
 
The MSE Team is working on the nightly Jenkinsfile. We need @crd477 and @apavlo 's review after we finish it and update the Jenkins Pipeline configuration so that we could test it.

Feel free to discuss this.

skyfishjy:
- Fixes #937 
- Extract and store the OLTPbench results for nightly build

### DIFF
--- a/Jenkinsfile-nightly
+++ b/Jenkinsfile-nightly
@@ -61,12 +61,12 @@ def loop_terminals_run_oltp(list){
     // for all the nightly benchmarks, we scale the number of terminals: 1,2,4,8,16
     for(int i=0; i< list.size(); i++){
         // tatp/smallbank/noop/ycsb are running with the default config: --scale-factor=1
-        sh 'cd build && python3 ../script/testing/oltpbench/run_oltpbench.py tatp 2,35,10,35,2,14,2 --build-type=release --loader-threads=4 --terminals=${item}'
-        sh 'cd build && python3 ../script/testing/oltpbench/run_oltpbench.py smallbank 15,15,15,25,15,15 --build-type=release --loader-threads=4 --terminals=${item}'
-        sh 'cd build && python3 ../script/testing/oltpbench/run_oltpbench.py noop 100 --build-type=release --terminals=${item}'
-        sh 'cd build && python3 ../script/testing/oltpbench/run_oltpbench.py ycsb 50,5,15,10,10,10 --build-type=debug --loader-threads=4 --terminals=${item}'
+        sh "cd build && python3 ../script/testing/oltpbench/run_oltpbench.py tatp 2,35,10,35,2,14,2 --build-type=release --loader-threads=4 --terminals=${list[i]}"
+        sh "cd build && python3 ../script/testing/oltpbench/run_oltpbench.py smallbank 15,15,15,25,15,15 --build-type=release --loader-threads=4 --terminals=${list[i]}"
+        sh "cd build && python3 ../script/testing/oltpbench/run_oltpbench.py noop 100 --build-type=release --terminals=${list[i]}"
+        sh "cd build && python3 ../script/testing/oltpbench/run_oltpbench.py ycsb 50,5,15,10,10,10 --build-type=debug --loader-threads=4 --terminals=${list[i]}"
         // for TPC-C, --scale-factor=4
-        sh 'cd build && python3 ../script/testing/oltpbench/run_oltpbench.py tpcc 45,43,4,4,4 --build-type=release --loader-threads=4 --scale-factor=4 --terminals=${item}' 
+        sh "cd build && python3 ../script/testing/oltpbench/run_oltpbench.py tpcc 45,43,4,4,4 --build-type=release --loader-threads=4 --scale-factor=4 --terminals=${list[i]}"
     }
     archiveArtifacts 'build/oltpsummary/*.json'
 }
--- a/Jenkinsfile-nightly
+++ b/Jenkinsfile-nightly
@@ -1,3 +1,5 @@
+OLTP_TERMINALS = ['1','2','4','8','16']
+
 pipeline {
     agent none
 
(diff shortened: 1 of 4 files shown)


Jenkinsfile-nightly
```diff
@@ -0,0 +1,55 @@
...
+        stage('Microbenchmark') {
+            agent { label 'benchmark' }
+            when { branch 'master' }
+            steps {
+                echo 'echo y | sudo ./script/installation/packages.sh'
+                echo 'sudo apt-get install -q -y curl numactl'
+                echo 'mkdir build && cd build'
+                echo 'cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_WARNING_LEVEL=Production -DTERRIER_USE_JEMALLOC=ON ..'
+                echo 'make gflags_ep -j 4 && make googletest_ep -j 4 && make gbenchmark_ep -j 4 && make -j4 all'
+                echo 'cd ../script/micro_bench'
```
bialesdaniel, line 44: should we be calling archiveArtifact? or does this script somehow do it?

script/testing/oltpbench/constants.py
```diff
@@ -23,7 +23,10 @@
 OLTP_DEFAULT_BIN = os.path.join(OLTP_GIT_LOCAL_PATH, "oltpbenchmark")
 OLTP_DEFAULT_COMMAND_FLAGS = "--histograms  --create=true --load=true --execute=true -s 5"
 OLTP_DIR_CONFIG = os.path.join(OLTP_GIT_LOCAL_PATH, "config")
+OLTP_DIR_CUR = os.getcwd()
```
bialesdaniel, line 26: I don't think this makes sense as a constant? I'm not entirely sure how this is used. Why couldn't we use OLTP_GIT_LOCAL_PATH?

script/testing/oltpbench/test_oltpbench.py
```diff
@@ -74,10 +85,12 @@ def run_pre_test(self):
         self.install_oltp()
         self.config_xml_file()
         self.create_result_dir()
+        self.create_summary_dir()
```
bialesdaniel, line 88: Just to clarify the results are the raw stdout and the summary is the nice json?

script/testing/oltpbench/constants.py
```diff
@@ -23,7 +23,10 @@
...
 OLTP_DIR_TEST_RESULT = os.path.join(OLTP_GIT_LOCAL_PATH, "results")
+OLTP_DIR_TEST_SUMMARY_FOLDER = os.path.join(OLTP_DIR_CUR, "oltpsummary")
+OLTP_DIR_TEST_SUMMARY = os.path.join(OLTP_DIR_TEST_RESULT, "oltpbench.summary")
```
bialesdaniel, line 29: Is OLTP_DIR_TEST_SUMMARY_FOLDER the destination folder where we are going to move the OLTP_DIR_TEST_SUMMARY file?

script/testing/oltpbench/constants.py
skyfishjy: OLTP_DIR_CUR points to the file path where `run_oltpbench.py` is been called. In the `Jenkinsfile-nightly`, that is the `build` folder(which is in the subdirectory for each job, controlled by Jenkins). I didn't use `OLTP_GIT_LOCAL_PATH` because it is in the public `/tmp/` folder. There might be some concurrency problems if this script run multiple times in parallel.

Jenkinsfile-nightly
skyfishjy: you are right, I forgot it

script/testing/oltpbench/test_oltpbench.py
skyfishjy: You are correct.

script/testing/oltpbench/constants.py
skyfishjy: Yes.  Then `archiveArtifacts 'build/oltpsummary/*.json'` will archive these files

skyfishjy:
This PR is still in progress.

TODO:
Store script's parameters and arguments for each job

Discuss with @jazzplato about the `OLTP_DIR_TEST_RESULT`. It is better to only have one folder to store all results.

script/testing/oltpbench/constants.py
jazzplato: then it would be better to change the variable name from `OLTP_DIR_TEST_SUMMARY_FOLDER` to something like `OLTP_FILE_TEST_SUMMARY_FOLDER` to indicate it is a file, not a dir

Jenkinsfile-nightly
```diff
@@ -0,0 +1,78 @@
...
+            }
+        }
+    }
+    
+}
+
+@NonCPS
+def loop_terminals_run_oltp(list){
+    list.each { item ->
+        sh 'cd build && python3 ../script/testing/oltpbench/run_oltpbench.py tpcc 45,43,4,4,4 --build-type=release --loader-threads=4 --scale-factor=4 --terminals=${item}'
```
bialesdaniel, line 76: This should include all the benchmark sets

skyfishjy: It is still in progress

skyfishjy:
this PR needs some changes. I'm going to create a new one.

PR #950, opened 2020-06-10
Use regions to allocate plan generator expressions

gonzalezjo:
Title: Use regions to allocate plan generator expressions

# Proposal: Allocate plan expressions in regions

## Motivation & Design

Currently, a large number of allocations are performed as part of physical plan generation.
This presents a number of issues:
1. A large number of small allocations in a hottish path adds unnecessary overhead.
2. Managing the newly allocated memory adds unnecessary code complexity and further overhead.
3. Presenting a large number of expensive allocations encourages/normalizes the use of allocations where they might otherwise be unnecessary.

We can address these issues with a per-connection-context region allocator.

- This addresses issue 1. by using an allocator that has extremely little overhead.
- This addresses issue 2. by deferring frees to the closure of a connection context.
- This addresses issue 3. by using placement new's allocation semantics, which I consider to be more discouraging than encouraging.

It may be possible to avoid allocations entirely in certain cases going into the future.

### The path toward merge

1. Get a dev machine to the point where it can compile and run tests. Then, use it to compile and run tests (which I can't get to compile and run with Arch's gcc-10). This has only been built with `make terrier -j4`

2. Discuss this PR with others who know more about what they're doing. (Is this architecture even desirable? Are there any blatant oversights? There are a number of comments I made concerning questions--can we discuss those?)

3. Remove comments/rewrite code based on #3. This is more of an exploratory PR as it stands right now. The rewrite would involve, among other things, finding a way to break things into multiple commits. (Unfortunately pretty hard, since the changes are generally interdependent.)

4. Figure out how to track memory leaks. Then, check for memory leaks.

5. The use of `unique_ptr`s is pretty concerning. I'm not sure the expressions ever leak, and since we're trying to avoid having to free objects at a per-object level, this seems unnecessary and probably dangerous.

### Going forward

Not all allocations have been made into region allocations. Copies, specifically, remain an issue. Making it so that copies are allocated in a region would require pretty substantial code changes and refactoring, whereas eliminating the allocations entirely could be trivial. I would like to see if we can avoid deep copies and rely on references.

I'm intrigued by the large number of `unique_ptr`s associated with expressions in `expression_util.h` and `plan_generator.cpp`. I would like to see if we should/could remove them.

It seems like it will be important to document failure cases and write tests. I'm specifically curious about if region allocation stays safe as more allocations happen. 

Investigate whether or not long-lasting connections with many transactions are an issue. If it's thread safe, I *believe* that we should reset the region on transaction commit/abort. This should be doable as part of the transaction cleanup/GC process. The question is if that's safe and worthwhile.

### DIFF
--- a/src/binder/binder_context.cpp
+++ b/src/binder/binder_context.cpp
@@ -211,6 +211,7 @@ void BinderContext::GenerateAllColumnExpressions(
     auto col_cnt = schema.GetColumns().size();
     for (uint32_t i = 0; i < col_cnt; i++) {
       const auto &col_obj = schema.GetColumn(i);
+      // TODO(jordig) Should this be allocated in a region?
       auto tv_expr = new parser::ColumnValueExpression(std::string(entry.first), std::string(col_obj.Name()));
       tv_expr->SetReturnValueType(col_obj.Type());
       tv_expr->DeriveExpressionName();
--- a/src/include/network/connection_context.h
+++ b/src/include/network/connection_context.h
@@ -1,5 +1,6 @@
 #pragma once
 
+
 #include <memory>
 #include <string>
 #include <unordered_map>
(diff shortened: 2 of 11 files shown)


apavlo:
Closing for now. Need to revisit after first release.

PR #953, opened 2020-06-11
StorageInterface instantiates PRI once

mbutrovich:
`ProjectedRowInitializer` is kind of expensive to create. It's documented as such:

> More specifically, ProjectedRowInitializer will calculate an optimal layout for the given col_ids and layout, treating the vector as an unordered set of ids to include, and stores the information locally so it can be copied into new ProjectedRows. It works almost like compilation, in that the initialization process is slightly more expensive on the first invocation but cheaper on sequential ones. The correct way to use this is to get an initializer and use it multiple times. Avoid throwing these away every time you are done.

However, `StorageInterface` currently creates one every time `GetTablePR()` is called -- basically every table operation. It should be instantiated once and stashed as a member in `StorageInterface` which is exactly what this PR does.

However, I'm seeing weird behavior when I run TPC-C now where the table oids being referenced are sometimes zero. Based on my understanding on `StorageInterface` life cycle, we should be safe to make this change, so I might need @lmwnshn or @tanujnay112 help to understand what's going on.


mbutrovich:
Also, we may just end up closing this in lieu of fixing it in the new TPL2 port.

mbutrovich:
The TPC-C issue is that the delete translator doesn't need a PRI in the StorageInterface and passes a dummy array of col_oids, but because this PR moves the PRI into the StorageInterface constructor, we explode trying to request a PRI for dummy oids. Proposed solutions: need_PRI flag for StorageInterface like need_indexes, maybe pass an empty array, or special case in the storage layer. I don't like the last option.

mbutrovich:
> Proposed solutions: need_PRI flag for StorageInterface like need_indexes, **maybe pass an empty array**, or special case in the storage layer. I don't like the last option.
After talking with Prashanth, I went with this option and relaxed the semantic check on `StorageInterface`'s constructor in compilation to allow an "undefined length" array (i.e. 0).

mbutrovich:
Benchmark failure is spurious.

PR #963, opened 2020-06-15
Faster Timestamps

thepinetree:
This PR addresses the performance hit undertaken through the previous implementation of timestamps. This ports over and augments Prashanth's faster timestamp code, and rips out time_util from the system as all functionality can be replaced by types and functions in the new Timestamp and Date classes.


.gitignore
```diff
@@ -273,4 +273,7 @@ libpqxx-*.deb*
 wal.log
 
 # OLTP output
-**/oltp_output_*
\ No newline at end of file
+**/oltp_output_*
+
+# Personal github config file
```
mbutrovich, line 278: Is this a Jekyll artifact? Guessing based on the commit history.

src/execution/compiler/codegen.cpp
```diff
@@ -429,15 +428,13 @@ ast::Expr *CodeGen::PeekValue(const parser::ConstantValueExpression &const_val)
...
-      auto month = static_cast<uint32_t>(ymd.month());
-      auto day = static_cast<uint32_t>(ymd.day());
+      int32_t year = 0, month = 0, day = 0;
+      execution::sql::Date::FromNative(val).ExtractComponents(&year, &month, &day);
       return DateToSql(year, month, day);
     }
     case type::TypeId::TIMESTAMP: {
       const auto val = const_val.Peek<execution::sql::Timestamp>().ToNative();
-      auto julian_usec = terrier::util::TimeConvertor::ExtractJulianMicroseconds(static_cast<type::timestamp_t>(val));
+      auto julian_usec = execution::sql::Timestamp::FromNative(val).ToNative();
```
mbutrovich, line 437: All of the ToNative FromNative in a consecutive lines is confusing. Is there an abstraction (like, just a function) with a descriptive name and documentation that takes an execution timestamp to Julian usec? That seems to be what all of these steps accomplish.

mbutrovich: Alternatively, just some documentation of the steps being performed is good too.

.gitignore
thepinetree: Yup, thanks, updated!

src/execution/compiler/codegen.cpp
thepinetree: Right, I lost that I'm not using the timestamp_t type anymore in translation, so just using ToNative should suffice as the value represents the Julian microseconds in the first place -- updated accordingly.

src/binder/binder_util.cpp
```diff
@@ -62,22 +62,21 @@ void BinderUtil::CheckAndTryPromoteType(const common::ManagedPointer<parser::Con
         // TODO(WAN): A bit stupid to take the string view back into a string.
         switch (desired_type) {
           case type::TypeId::DATE: {
-            auto parsed_date = util::TimeConvertor::ParseDate(std::string(str_view));
+            auto parsed_date = execution::sql::Date::FromString(std::string(str_view));
```
mbutrovich, line 65: It looks like FromString just takes a string_view. Can we avoid the string instantiation now?

src/execution/sql/runtime_types.cpp
```diff
@@ -1,118 +1,636 @@
...
 
-uint64_t Timestamp::ToNative() const noexcept { return value_; }
+std::pair<bool, Timestamp> Timestamp::FromString(const char *str, std::size_t len) {
+  const char *ptr = str, *limit = ptr + len;
+
+  // Trim leading and trailing whitespace
+  while (ptr != limit && static_cast<bool>(std::isspace(*ptr))) ptr++;
+  while (ptr != limit && static_cast<bool>(std::isspace(*(limit - 1)))) limit--;
+
+  std::string suffix_str("::timestamp");
```
mbutrovich, line 390: Does it work to make this a `constexpr std::string_view` to avoid the object allocation every time?

src/include/execution/sql/runtime_types.h
```diff
@@ -228,17 +342,55 @@ class EXPORT Timestamp {
...
+   * @param hour The hour.
+   * @param min The minute.
+   * @param sec The second.
+   * @param milli The millisecond.
+   * @param micro The microsecond.
+   * @param ptr Start of timestamp string.
+   * @param limit End of timestamp string.
+   * @return The constructed timestamp if valid.
+   */
+  static std::pair<bool, Timestamp> AdjustTimezone(char c, int32_t year, int32_t month, int32_t day, int32_t hour,
```
pmenon, line 361: nit: `AdjustTimezone` -> `AdjustTimeZone`.

test/util/runtime_types_test.cpp
```diff
@@ -0,0 +1,30 @@
+#include "execution/sql/runtime_types.h"
```
pmenon, line 1: 1. Move to execution sql folder.
2. Can you bring in the tests from the TPL repo, too?
3. Add tests for the corner cases I mentioned earlier.
4. Add tests to exercise the branches in `Timestamp::AdjustTimezone()`.

src/execution/sql/runtime_types.cpp
```diff
@@ -1,118 +1,644 @@
...
+  // Trim leading and trailing whitespace
+  while (ptr != limit && static_cast<bool>(std::isspace(*ptr))) ptr++;
+  while (ptr != limit && static_cast<bool>(std::isspace(*(limit - 1)))) limit--;
 
-Timestamp Timestamp::FromNative(Timestamp::NativeType val) {
-  Timestamp ts;
-  ts.value_ = val;
-  return ts;
+  auto suffix_len = strlen(K_TIMESTAMP_SUFFIX);
+  if (EndsWith(str, len, K_TIMESTAMP_SUFFIX, suffix_len)) limit -= suffix_len;
```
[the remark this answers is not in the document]
thepinetree: Good catch, thanks

src/include/execution/sql/runtime_types.h
thepinetree: Fixed

src/execution/sql/runtime_types.cpp
```diff
@@ -1,118 +1,644 @@
...
+  *hour = time / K_MICRO_SECONDS_PER_HOUR;
+  time -= (*hour) * K_MICRO_SECONDS_PER_HOUR;
+  *min = time / K_MICRO_SECONDS_PER_MINUTE;
+  time -= (*min) * K_MICRO_SECONDS_PER_MINUTE;
+  *sec = time / K_MICRO_SECONDS_PER_SECOND;
+  *fsec = time - (*sec * K_MICRO_SECONDS_PER_SECOND);
+}
+
+// Check if a string value ends with string ending
+bool EndsWith(const char *str, std::size_t len, const char *end, std::size_t end_size) {
```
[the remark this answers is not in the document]
thepinetree: Fixed

src/execution/sql/runtime_types.cpp
```diff
@@ -1,118 +1,644 @@
...
+  // Looks good.
+  return {true, Timestamp(result)};
 }
 
-Timestamp Timestamp::FromHMSu(int32_t year, uint32_t month, uint32_t day, uint8_t hour, uint8_t minute, uint8_t sec,
-                              uint64_t usec) {
-  Timestamp ts;
-  ts.value_ = !terrier::util::TimeConvertor::TimestampFromHMSu(year, month, day, hour, minute, sec, usec);
-  return ts;
+std::pair<bool, Timestamp> Timestamp::FromYMDHMSMU(int32_t year, int32_t month, int32_t day, int32_t hour, int32_t min,
```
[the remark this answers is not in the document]
thepinetree: For an invalid date, we return {false, {}} and can handle the false case in the bytecode to raise an exception instead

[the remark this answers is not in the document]
pmenon: But why not just raise it here?

src/execution/sql/runtime_types.cpp
```diff
@@ -1,118 +1,644 @@
...
+                                             {31, 29, 31, 30, 31, 30, 31, 31, 30, 31, 30, 31}};
+constexpr int64_t K_HOURS_PER_DAY = 24;
+constexpr int64_t K_MINUTES_PER_HOUR = 60;
+constexpr int64_t K_SECONDS_PER_MINUTE = 60;
+constexpr int64_t K_MILLISECONDS_PER_SECOND = 1000;
+constexpr int64_t K_MICROSECONDS_PER_MILLISECOND = 1000;
 
-bool IsValidJulianDate(int32_t year, uint32_t month, uint32_t day) {
-  return year <= 9999 && month >= 1 && month <= 12 && day >= 1 && day <= 31;
+constexpr char K_TIMESTAMP_SUFFIX[] = "::timestamp";
```
mbutrovich, line 21: @pmenon why a `constexpr char[]` and not a `constexpr std::string_view` since we're calling `strlen` on it every time (which in theory should be `constexpr` as well but the STL doesn't have a `constexpr` `strlen`)?

pmenon: `strlen` should be optimized out. I originally suggested `sizeof` to be sure.

I'm okay with this change. In fact, you can even remove the new function with [`std::string_view::ends_with()`](https://en.cppreference.com/w/cpp/string/basic_string_view/ends_with) if we switch to C++20 ... which isn't a bad idea.

src/execution/sql/runtime_types.cpp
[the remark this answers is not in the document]
thepinetree: Mostly to match the style of FromString which I formatted based on the behavior of Parse(Date/Timestamp) in postgres_packet_writer... was also having some issues with the exception formatting before and this let me get rid of all the exceptions in this file

src/execution/sql/runtime_types.cpp
thepinetree: Should I replace K_TIMESTAMP_SUFFIX with a string_view and reface the arguments in the call to EndsWith as K_TIMESTAMP_SUFFIX.data() and K_TIMESTAMP_SUFFIX.size() instead?

src/execution/sql/runtime_types.cpp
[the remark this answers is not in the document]
pmenon: I'd say keep the interface the same and change the writer. These classes are optimized for execution.

src/execution/sql/runtime_types.cpp
pmenon: Sounds reasonable.

src/execution/sql/runtime_types.cpp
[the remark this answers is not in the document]
thepinetree: Ok, will do so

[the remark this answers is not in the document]
thepinetree: Ah right, I remember the other issue that came up... postgres_packet_util handles the INVALID type ID in TextValueToInternalValue by trying timestamps, then dates, and then falling back to varchars... I thought the reading the boolean would be faster and also less complex than the scoped try catch blocks for this section. Would that change anything?

test/util/runtime_types_test.cpp
thepinetree: Updated.

PR #967, opened 2020-06-17
Extract and store the OLTPbench results for nightly build

skyfishjy:
- Support #937 
- Extract and store the OLTPbench results for nightly build
- Unify the filename in test_oltpbench.py script

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -372,25 +372,5 @@ pipeline {
                 }
             }
         }
-
-        stage('Performance') {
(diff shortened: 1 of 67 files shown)


skyfishjy:
one testing pipeline result [here](http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/nightly-jenkins-test/detail/nightly-jenkins/35/pipeline)

apavlo:
I am thinking that we should run each of these benchmarks for longer than one minute.

PR #921, opened 2020-05-26
JUnit: add QueryMode option from command line

jazzplato:
Fix #866.

- added `--query-mode` option from the command line in `run_junit.py`
- currently only allows the following values
  - `simple`
  - `extended`
- python script pass the option to java side via `options.xml`
- if no `--query-mode` specified, the java side will not set `preferQueryMode` to any default value upon making the connection to the database
- tested ok with `--query-mode simple`
- failure occurred with `--query-mode extended`, helps need for testing the `extended` query mode

PR #969, opened 2020-06-18
fix "module 'oltpbench.constants' has no attribute 'ErrorCode'"

skyfishjy:
This error happens when: "run_oltpbench.py ycsb 50,5,15,10,10,10 --build-type=release --loader-threads=4 --terminals=4"
image


skyfishjy:
fixed in #921

PR #970, opened 2020-06-18
Dedicated test for BinderUtil::CheckAndTryPromoteType

mbutrovich:
See #977. This just has a test for BinderUtil now.


mbutrovich:
Can't get the JUnit failure to reproduce locally and it's annoying. The test that fails shouldn't even touch this codepath anyway.

mbutrovich:
We might not be able to do this. In trying to figure out the occasional failures, noticed that `strtol` requires nul-termination. https://en.cppreference.com/w/cpp/string/byte/strtol

mbutrovich:
And now clang 8 doesn't have <charconv>. Ugh.

PR #972, opened 2020-06-19
few exception messages changed

sashatankov:


src/execution/sql/runtime_types.cpp
```diff
@@ -89,7 +89,7 @@ Date Date::FromString(const std::string &str) {
 
 Date Date::FromYMD(int32_t year, uint32_t month, uint32_t day) {
   if (!IsValidJulianDate(year, month, day)) {
-    throw CONVERSION_EXCEPTION("Invalid date.");
+    throw CONVERSION_EXCEPTION("Invalid julian date.");
```
lmwnshn, line 92: It would be useful to say what year/month/day it is.

src/settings/settings_manager.cpp
```diff
@@ -46,7 +46,7 @@ void SettingsManager::ValidateSetting(Param param, const parser::ConstantValueEx
         "Value given for \"{}"
         "\" is not in its min-max bounds",
         info.name_);
-    throw SETTINGS_EXCEPTION("Invalid setting value");
+    throw SETTINGS_EXCEPTION("Setting value is out of min-max bounds");
```
lmwnshn, line 49: It would be useful to say what the value is, and what the min/max bounds are.

PR #976, opened 2020-06-19
Dynamically Generating JUnit Test Cases from SQLite Trace Files

dniu16:
1. Dynamically generated test cases for each query given input trace file. (TracefileTest.java)
2. Added a class that takes in sql statements and convert them to trace file format. (generateTrace.java)
3. Made noisepage-testfiles, a repository containing trace files, a git submodule of terrier


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "script/testing/junit/noisepage-testfiles"]
+	path = script/testing/junit/noisepage-testfiles
+	url = https://github.com/dniu16/noisepage-testfiles.git
```
apavlo, line 3: Should be https://github.com/cmu-db/noisepage-testfiles.git

dniu16:
Trying another approach

PR #984, opened 2020-06-26
Memory usage estimates, miscellaneous #include cleanups

mbutrovich:
Implements estimates for heap usage of our indexes. I've documented in their calculations where they make assumptions and in what scenarios the estimates might be wrong.

I also cleaned up some stray includes and changed some to prototypes as I found them, again.


src/storage/index/bwtree_index.cpp
```diff
@@ -17,6 +17,34 @@ void BwTreeIndex<KeyType>::PerformGarbageCollection() {
   bwtree_->PerformGarbageCollection();
 }
 
+template <typename KeyType>
+size_t BwTreeIndex<KeyType>::EstimateHeapUsage() const {
+  // This is a back-of-the-envelope calculation that could be innacurate: it does not account for deltas within the
+  // BwTree. Also the mapping table is anonymously mmap'd so getting its exact size is tricky.
+  constexpr auto max_elements_per_node = 128;  // constant from bwtree.h
```
malin1993ml, line 24: Just curious, why not directly read the constant from `bwtree.h` but hard-code it this way?

mbutrovich: I could do that, but the question then becomes which value to read. There are two (inner node max size and leaf node max size) that currently happen to have the same value. I guess I could just read the leaf node size.

Issue #982, opened 2020-06-23
SQLSmith failure connecting to terrier


jazzplato:
I am trying to use [SQLSmith](https://github.com/anse1/sqlsmith) to conduct fuzzing tests for the terrier. However, the following error occurred upon the connection.
```
# sqlsmith --target="host=localhost port=15721 dbname=terrier"
SQLsmith v1.2.1
ERROR:  Procedure not registered
```

Meanwhile, I tried using SQLSmith for Postgres, which worked normally. I have ported out the [SQLSmith query log](http://incrudibles-testing.db.pdl.cmu.edu/sqlsmith/logs/sqlsmith_queries.log) and [Postgres CSV log with the accepted queries](http://incrudibles-testing.db.pdl.cmu.edu/sqlsmith/logs/postgresql-2020-06-23_164040.csv) for troubleshooting purpose.

jazzplato:
BTW, the logs I dumped are the first 100 queries.

mbutrovich:
Same error we've seen in some other tools trying to speak Postgres: `select version()` fails.

![Screen Shot 2020-06-23 at 1 34 29 PM](https://user-images.githubusercontent.com/6855576/85435981-494fc180-b556-11ea-8880-ba1ee4846e65.png)

lmwnshn:
version() was added, closing.

PR #988, opened 2020-06-26
Add support for version() builtin

thepinetree:
Added support for `version()` builtin for Issue #982


src/execution/sql/functions/string_functions.cpp
```diff
@@ -363,4 +363,10 @@ void StringFunctions::Right(UNUSED_ATTRIBUTE exec::ExecutionContext *ctx, String
   }
 }
 
+// TODO(Deepayan): Do we need to print out OS name/version/compiler?
+void StringFunctions::Version(UNUSED_ATTRIBUTE exec::ExecutionContext *ctx, StringVal *result) {
+  const char *version = "NoisePage 1.0.0";
```
apavlo, line 368: We should `#define` these somewhere other than in this function. I think the network protocol needs it too.

apavlo:
Also, should we put whatever that file you checked in by accident (`config.yml`) in gitignore?

apavlo:
@thepinetree Also modify `postgres_defs.h` so that we return the right name+version over the network:

https://github.com/cmu-db/terrier/blob/master/src/include/network/postgres/postgres_defs.h#L24

thepinetree:
> Also, should we put whatever that file you checked in by accident (`config.yml`) in gitignore?

Emmanuel showed me this thing about how to change the view of the GitHub page so it looks more elegant (https://thepinetree.github.io/terrier/) which changes the _config.yml file. Would we want to make a site like this for our GitHub page too, otherwise is adding it to the gitignore ok?

thepinetree:
> @thepinetree Also modify `postgres_defs.h` so that we return the right name+version over the network:

@apavlo libpqxx throws an exception if the version is lower than 9 (network_test) so I can't change the version. I'm not exactly sure what server_version and session_authorization are for though, they're only used on WriteStartupResponse

apavlo:
@thepinetree Good to know! Add a warning in postgres_defs.h for that line that people should not change the string because it will break network compatibility.

src/include/common/version.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include "common/macros.h"
+
+namespace terrier::common {
+
+#define NOISEPAGE_NAME "NoisePage"
```
mbutrovich, line 7: Don't do `#define`. Prefer `constexpr std::string_view`.

thepinetree: Updated

script/testing/junit/src/FunctionsTest.java
```diff
@@ -159,4 +159,16 @@ public void testLower() throws SQLException {
...
 
+    @Test
+    public void testVersion() throws SQLException {
+        String sql = String.format("SELECT version() AS result");
+
+        Statement stmt = conn.createStatement();
+        ResultSet rs = stmt.executeQuery(sql);
+        boolean exists = rs.next();
+        assert(exists);
+        checkStringRow(rs, new String[]{"result"}, new String[]{"NoisePage 1.0.0"});
```
apavlo, line 170: Make it check that it matches with what is in version.h. Otherwise this test will break every time we update the version.

src/include/execution/sql/functions/string_functions.h
```diff
@@ -122,6 +122,11 @@ class EXPORT StringFunctions {
    * Converts the string to upper case
    */
   static void Upper(exec::ExecutionContext *ctx, StringVal *result, const StringVal &str);
+
+  /**
+   * Gets the version of the database
+   */
+  static void Version(exec::ExecutionContext *ctx, StringVal *result);
```
apavlo, line 129: This is not really a string function. Make a new file called `include/execution/sql/functions/system_functions.h`

src/include/network/postgres/postgres_defs.h
```diff
@@ -21,8 +22,9 @@ const std::unordered_map<std::string, std::string> PG_PARAMETER_STATUS_MAP = {
     {"IntervalStyle", "postgres"},
     {"is_superuser", "on"},
     {"server_encoding", "UTF8"},
+    // server_version must be at least 9 for libpqxx
```
apavlo, line 25: Change to:

```
// Warning: The DBMS must report that its 'server_version' is at least "9" for libpqxx
// Do not change this parameter unless you know what you are doing!
```

src/include/execution/sql/functions/string_functions.h
thepinetree: Updated.

src/include/network/postgres/postgres_defs.h
thepinetree: Updated.

script/testing/junit/src/FunctionsTest.java
thepinetree: I added a C++ test file (system_functions_test.cpp) that checks what's in version.h, but I couldn't find a way to get the value in the header file to the java file. Should I just take out this test then?

apavlo: You can take out the test. We are going to replace the Junit tests anyway.

PR #992, opened 2020-06-28
Terrier Pipeline Performance Stage

bialesdaniel:
I added a performance stage in the normal PR pipeline. This will allow anyone on CMU-DB team to check a PRs performance before merging, if that is a concern. Currently, to find out the information you will have to look at the stage output in Jenkins. However, in the future, we will make this easier to access. This will also be useful for when we build more automated checks.


bialesdaniel:
This has changes to the Jenkins file I'm not sure if that means that @crd477 needs to do something special when merging this.

Jenkinsfile
```diff
@@ -373,5 +373,16 @@ pipeline {
                 }
             }
         }
+        stage('Performance') {
+            agent { label: 'benchmark' }
+            steps {
+                sh 'echo $NODE_NAME'
+                sh 'echo y | sudo ./script/installation/packages.sh all'
+                sh 'mkdir build'
+                sh 'cd build && cmake -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DTERRIER_USE_ASAN=OFF -DTERRIER_USE_JEMALLOC=ON -DTERRIER_BUILD_TESTS=OFF .. && make -j$(nproc) all'
```
mbutrovich, line 382: Suggest only building the `terrier` target instead of `all`.

PR #996, opened 2020-07-01
Hook up per-component metrics settings to DBMain's Builder

mbutrovich:
We have settings to toggle per-component metrics at runtime, but I never hooked them up to DBMain's builder to enable them if they were turned on at the command line at server launch. This fixes that problem.

### DIFF
--- a/src/include/main/db_main.h
+++ b/src/include/main/db_main.h
@@ -285,23 +285,18 @@ class DBMain {
         metrics_manager = std::make_unique<metrics::MetricsManager>();
         if (use_settings_manager_) {
           if (settings_manager->GetBool(settings::Param::metrics_logging)) {
-            auto action_context = std::make_unique<common::ActionContext>(common::action_id_t(0));
             metrics_manager->EnableMetric(metrics::MetricsComponent::LOGGING, 0);
           }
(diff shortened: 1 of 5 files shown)


src/include/main/db_main.h
```diff
@@ -281,7 +281,31 @@ class DBMain {
           use_settings_manager_ ? BootstrapSettingsManager(common::ManagedPointer(db_main)) : DISABLED;
 
       std::unique_ptr<metrics::MetricsManager> metrics_manager = DISABLED;
-      if (use_metrics_) metrics_manager = std::make_unique<metrics::MetricsManager>();
+      if (use_metrics_) {
+        metrics_manager = std::make_unique<metrics::MetricsManager>();
+        if (use_settings_manager_) {
+          if (settings_manager->GetBool(settings::Param::metrics_logging)) {
+            auto action_context = std::make_unique<common::ActionContext>(common::action_id_t(0));
```
malin1993ml, line 288: What are these `action_context` for? They're not really used. Otherwise looks good!

mbutrovich: Whoops. Those are an artifact from fiddling with the `SettingsManager`. Removed.

PR #1002, opened 2020-07-01
Relax asserts in MetricsStore

mbutrovich:
There's a race condition between disabling a component's metrics and the component potentially still wanting to record one of its data points. We had asserts that would trip occasionally in this event (see [this recent Jenkins run](http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/master/593/tests/)). This relaxes the assert to just use the (new) `MetricsLogger` to print out a warning message.

### DIFF
--- a/src/storage/write_ahead_log/disk_log_consumer_task.cpp
+++ b/src/storage/write_ahead_log/disk_log_consumer_task.cpp
@@ -1,4 +1,5 @@
 #include "storage/write_ahead_log/disk_log_consumer_task.h"
+
 #include "common/resource_tracker.h"
 #include "common/scoped_timer.h"
 #include "common/thread_context.h"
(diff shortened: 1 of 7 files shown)


src/storage/write_ahead_log/disk_log_consumer_task.cpp
```diff
@@ -57,15 +58,6 @@ void DiskLogConsumerTask::DiskLogConsumerTaskLoop() {
...
 
-  bool logging_metrics_enabled =
-      common::thread_context.metrics_store_ != nullptr &&
-      common::thread_context.metrics_store_->ComponentToRecord(metrics::MetricsComponent::LOGGING);
-
-  if (logging_metrics_enabled) {
-    // start the operating unit resource tracker
-    common::thread_context.resource_tracker_.Start();
-  }
-
```
malin1993ml, line 68: I've changed the metrics recording in this file. Maybe just remove the changes here to avoid the conflicts. Then I'll just submit my new changes. Otherwise looks good!

mbutrovich: I'm not sure it works if I move it back. It's technically a bug that we currently check outside of the loop if metrics are enabled since it can be toggled off at any time.

malin1993ml: You're correct. I've changed that in my branch as well but I guess it makes sense to fix the bug first here. I can resolve the conflicts later.

PR #1004, opened 2020-07-02
Revert "Explicit instantiation of spdlog template functions (#995)"

mbutrovich:
This reverts commit 686eb69f. Not sure we want to merge this immediately, but it sounds like #995 didn't really help so this is a PR to revert it if we want.


gonzalezjo:
I didn't see this PR when I reverted that commit in #1026. For what it's worth, #1026 does this as a necessary prerequisite for statically linking spdlog.

mbutrovich:
Redundant.

PR #1006, opened 2020-07-05
Adding support for NOT expression

abalakumar083:
The NOT expression did not work in terrier, because the operator only accepted primitive booleans. These changes overload the NOT operator to also accept SQL booleans, which fixes the issue.

### DIFF
--- a/src/include/execution/sql/functions/comparison_functions.h
+++ b/src/include/execution/sql/functions/comparison_functions.h
@@ -216,6 +216,14 @@ class EXPORT ComparisonFunctions {
     return int32_t(len1) - int32_t(len2);
   }
 
+  /**
+   * Logical negation for SQL booleans
+   */
+  static void NotBoolVal(terrier::execution::sql::BoolVal *result, terrier::execution::sql::BoolVal *input) {
+    *result = terrier::execution::sql::BoolVal(!input->val_);
+    result->is_null_ = input->is_null_;
+  }
+
  private:
   /**
    * Compare two strings. Returns
--- a/src/include/execution/sql/functions/comparison_functions.h
+++ b/src/include/execution/sql/functions/comparison_functions.h
@@ -219,9 +219,9 @@ class EXPORT ComparisonFunctions {
   /**
    * Logical negation for SQL booleans
    */
-  static void NotBoolVal(terrier::execution::sql::BoolVal *result, terrier::execution::sql::BoolVal *input) {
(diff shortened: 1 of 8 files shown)


src/include/execution/vm/bytecode_handlers.h
```diff
@@ -67,6 +67,14 @@ ALL_TYPES(COMPARISONS);
 
 VM_OP_HOT void OpNot(bool *const result, const bool input) { *result = !input; }
 
+VM_OP_HOT void OpNotSql(terrier::execution::sql::BoolVal *result, terrier::execution::sql::BoolVal *input) {
```
pmenon, line 70: Move this logic into a new function in `ComparisonFunctions::NotBoolVal()`. Make sure that function is implemented in the header, too.

Try to avoid putting logic in these handlers, unless it's a "simple" bytecode (like `OpNot`). This allows them to be tested separately, and groups common SQL logic together.

src/include/execution/vm/bytecode_handlers.h
```diff
@@ -67,6 +67,10 @@ ALL_TYPES(COMPARISONS);
...
```
pmenon, line 70: Make input a `const *` to convey intention.

src/include/execution/sql/functions/comparison_functions.h
```diff
@@ -216,6 +216,14 @@ class EXPORT ComparisonFunctions {
     return int32_t(len1) - int32_t(len2);
   }
 
+  /**
+   * Logical negation for SQL booleans
+   */
+  static void NotBoolVal(terrier::execution::sql::BoolVal *result, terrier::execution::sql::BoolVal *input) {
```
pmenon, line 222: 1. Remove the namespaces.
2. Change second argument (i.e., input) to `const &` to convey intention.
3. Simplify logic:
```
result->is_null_ = input.is_null_;
result->val_ = !input.val_
```

src/include/execution/vm/bytecode_handlers.h
abalakumar083: I made result a const * as well. I think when you use * after the type, then it only requires you preserve the pointer address. And this is consistent with the other functions.

pmenon: Change it to `const terrier::execution::sql::BoolVal *input` since it isn't modified.

PR #1008, opened 2020-07-06
Modeling Augmentation

malin1993ml:
This PR contains a few augmentations of the modeling work:
1. Change the recording, mini-runners, and models for the logging and GC components.
2. Add metrics recording for the bind and execute commands.
3. Add a new direct global (interference) model.


script/model/execution_engine_runner/generate_scan.py
```diff
@@ -48,24 +64,30 @@ def generate_main_fun(fun_names):
...
-                fun_names.append(generate_scan_fun(col_num, row_num, cardinality))
+                fun_names.append(generate_scan_fun(col_num, row_num, cardinality, include_output))
 
     generate_main_fun(fun_names)
 
 
 if __name__ == '__main__':
-    generate_all()
+    aparser = argparse.ArgumentParser(description='Mini Trainer')
+    aparser.add_argument('--include_output', type=int, default=0, help='Test data split ratio')
```
17zhangw, line 91: nit: the help text should probably be about whether to write to the OutputBuffer?

script/model/mini_trainer.py
```diff
@@ -57,7 +57,7 @@ def _train_data(self, data):
             transformers.append(modeling_transformer)
 
         error_bias = 1
-        min_percentage_error = 1
+        min_percentage_error = 2
```
17zhangw, line 60: Just double-checking to make sure this is intended?

src/settings/settings_callbacks.cpp
```diff
@@ -76,7 +76,7 @@ void Callbacks::MetricsTransaction(void *const old_value, void *const new_value,
   action_context->SetState(common::ActionState::IN_PROGRESS);
   bool new_status = *static_cast<bool *>(new_value);
   if (new_status)
-    db_main->GetMetricsManager()->EnableMetric(metrics::MetricsComponent::TRANSACTION, 0);
+    db_main->GetMetricsManager()->EnableMetric(metrics::MetricsComponent::TRANSACTION, 100);
```
17zhangw, line 79: Just curious why this is set to 100 but the other are at 0? (maybe worth a comment)

src/storage/data_table.cpp
```diff
@@ -76,7 +76,7 @@ DataTable::SlotIterator &DataTable::SlotIterator::operator++() {
 DataTable::SlotIterator DataTable::end() const {  // NOLINT for STL name compability
   // TODO(Lin): We need to temporarily comment out this latch for the concurrent TPCH experiments. Should be replaced
   //  with a real solution
-  common::SpinLatch::ScopedSpinLatch guard(&blocks_latch_);
+  // common::SpinLatch::ScopedSpinLatch guard(&blocks_latch_);
```
17zhangw, line 79: I think this shouldn't be committed into master?

script/model/mini_trainer.py
malin1993ml: Yes, it's intended. Though an error higher than 1 is pretty bad...

src/settings/settings_callbacks.cpp
malin1993ml: Originally the consideration was that the time for transaction begin/commit is much faster than the tracker itself, so we apply sampling for them to reduce overhead. But since we don't execute that many transactions per second anyway, I changed this back to 0 for now.

src/storage/data_table.cpp
malin1993ml: Fixed.

PR #1009, opened 2020-07-07
Dockerfile update

bialesdaniel:
I changed the docker image to install the packages the same way that the build pipeline is installing the packages. I also added a step to copy the current directory into the docker image. That will help with local development. If this gets approved I will update the documentation here: https://github.com/cmu-db/terrier/wiki/Docker


crd477:
CI doesn't even touch this file so I expect to pass.

PR #993, opened 2020-06-28
Dynamically Generating Junit Test Cases from SQLite Trace Files

dniu16:
Dynamically generated test cases for each query given input trace file. (TracefileTest.java)
Added a class that takes in sql statements and convert them to trace file format. (generateTrace.java)
Made noisepage-testfiles, a repository containing trace files, a git submodule of terrier
Loop through noise-testfiles directory and pass each file path as environment variable in run_junit.py and call TracefileTest for each file

PR #1010, opened 2020-07-08
run_oltpbench.py: Add Debug Log Timestamps

### LINKED ISSUE #868: Python Test Harness: Add Debug Log Timestamps
The timing of debug messages is critical for determining how long certain operations took and the ordering of them. Instead of just use `print()`, the Python test harness should use the [logging](https://docs.python.org/3.7/library/logging.html) library to produce more useful messages.

This would be helpful in identifying certain problems. For example, right now the smallbank benchmark times out after 20min during loading on OSX. The debug log messages would make this more clear. Right now you have to discern this from the OLTP-Bench debug output.

benliangw:
Fix #868.


benliangw:
> Be sure to update the junit directory too.
> 
> The error messages seem to be missing timestamps too: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-1010/2/pipeline#step-283-log-3

Added LOG to the junit directory. PTAL. 
Besides, I changed the format of logging so that it will also show the full path of the source file. For example:
Snipaste_2020-07-08_14-21-42

apavlo:
We will merge this after #993 gets in.

PR #1017, opened 2020-07-09
CMake improvements

gonzalezjo:
This pull request incorporates #916 (dynamic linking) with miscellaneous fixes.

Builds should be far faster on Linux.

- This is because we do dynamic linking on Linux only. On mac OS, the linker is remarkably fast and we can get by with static linking just fine. This is good, because we can't actually do dynamic linking on mac OS without warning messages that nobody understands how to resolve in a practical fashion.

~~Debug builds now use debug optimizations.~~

- ~~This is as simple as it sounds.~~

Builds should work on the latest version of mac OS

- This was a pain to troubleshoot, but builds work flawlessly on the dev machine with tests passing.


util/CMakeLists.txt
```diff
@@ -4,9 +4,7 @@ list(REMOVE_ITEM UTIL_SRCS ${PROJECT_SOURCE_DIR}/util/execution/tpl.cpp)
 list(REMOVE_ITEM UTIL_SRCS ${PROJECT_SOURCE_DIR}/util/execution/gen_opt_bc.cpp)
 list(REMOVE_ITEM UTIL_SRCS ${PROJECT_SOURCE_DIR}/util/execution/bytecode_handlers_ir.cpp)
 add_library(util_static STATIC ${UTIL_SRCS})
```
mbutrovich, line 6: Do we want a shared version of `util` as well or punt for now? Curious how big the test binaries end up with that still being static.

CMakeLists.txt
```diff
@@ -396,13 +395,25 @@ set(TERRIER_TEST_LINK_LIBS
...
+if (APPLE) 
+    list(APPEND TERRIER_TEST_LINK_LIBS terrier_static)
+    list(APPEND TERRIER_BENCHMARK_LINK_LIBS terrier_static)
+    if (TERRIER_BUILD_SHARED) 
+        set(TERRIER_BUILD_SHARED OFF CACHE BOOL "Build the libterrier shared libraries" FORCE)
+        message("Disabling TERRIER_BUILD_SHARED due to lack of support on mac OS.")
+    endif ()
+else ()
+    list(APPEND TERRIER_TEST_LINK_LIBS terrier_shared)
+    list(APPEND TERRIER_BENCHMARK_LINK_LIBS terrier_shared)
```
mbutrovich, line 414: I think we want the benchmarks and terrier binary to be linked statically for performance reasons.

src/CMakeLists.txt
```diff
@@ -77,4 +77,4 @@ ADD_TERRIER_LIB(terrier
 ###############################################
 add_executable(terrier main/terrier.cpp)
 set_target_properties(terrier PROPERTIES ENABLE_EXPORTS 1)
-target_link_libraries(terrier terrier_static ${TERRIER_LINK_LIBS})
+target_link_libraries(terrier $<IF:$<PLATFORM_ID:Darwin>,terrier_static,terrier_shared> ${TERRIER_LINK_LIBS})
```
mbutrovich, line 80: I think we want the benchmarks and terrier binary to be linked statically for performance reasons.

util/CMakeLists.txt
gonzalezjo: Seems like tests are tiny even with util statically linked: 
```
~/w/t/b/relwithdebinfo on TerrierShared* du -h hash_index*                                                   
11M     hash_index_test
704K    hash_index_test_stripped
```

cmake_modules/SetupCxxFlags.cmake
```diff
@@ -38,7 +44,7 @@ string(TOUPPER ${CMAKE_BUILD_TYPE} CMAKE_BUILD_TYPE)
 # Set compile flags based on the build type.
 message("Configured for ${CMAKE_BUILD_TYPE} build (set with cmake -DCMAKE_BUILD_TYPE={release,debug,fastdebug,relwithdebinfo})")
 if ("${CMAKE_BUILD_TYPE}" STREQUAL "DEBUG")
-    set(CXX_OPTIMIZATION_FLAGS "-ggdb -O0 -fno-omit-frame-pointer -fno-optimize-sibling-calls")
+    set(CXX_OPTIMIZATION_FLAGS "-ggdb -Og -fno-omit-frame-pointer -fno-optimize-sibling-calls")
```
lmwnshn, line 47: Debugging experience turned to crap when trying to step around in `void PlanGenerator::Visit(const InnerNLJoin *op)` calling out to `parser::ExpressionUtil::EvaluateExpression`. I vote we don't use this.

gonzalezjo: Reverted.

gonzalezjo:
I'm unable to reproduce the `traffic_cop_test` and `network_test` crashes on the Jenkins Mac machine. Before I lose my mind, is there any reason to think that those tests aren't reliable? I checked about a dozen Jenkins runs and saw nothing to indicate that either test has a problem, but I have to ask because I don't see what the problem could be. 

Otherwise, is there any way to see more verbose logs from Jenkins, so that I can tell exactly which (if any) sub-tests failed?

util/CMakeLists.txt
gonzalezjo: I'm going to leave it alone. I don't think we want/need a shared version of util. I'll mark this as resolved, but feel free to reopen it and/or comment if you think otherwise.

gonzalezjo:
The earlier test failure seems to have been bad luck. I've merged the changes Matt wanted re: the terrier binary and benchmarks. The current Jenkins failure is misleading in that it built for both a Mac and Ubuntu machine, but docker errors led to the Jenkins run being cancelled.

gonzalezjo:
Okay, now it might _actually_ be ready for merge. The tests look fine as long as the network test failures are unrelated to the PR, which I think is the case.

util/CMakeLists.txt
```diff
@@ -18,10 +16,16 @@ configure_file(
...
 )
 
-# Find CLANG
+# Common names for Clang to search for
 set(SUPPORTED_CLANGS "clang++-8" "clang++")
 
-find_program(CLANG NAMES ${SUPPORTED_CLANGS})
+# On macOS, the clang++ we want is not the clang++ in PATH. On Linux, it is.
+if (APPLE)
+    set(CLANG "/usr/local/opt/llvm@8/bin/clang++")
```
mbutrovich, line 24: Change to `set(CLANG "${LLVM_TOOLS_BINARY_DIR}/clang++")`.

gonzalezjo:
Jenkins refuses to pull from this PR, so I'm closing it.

Issue #1013, opened 2020-07-08
ExportTableTest::CheckContent Does Not Provide Useful Feedback


apavlo:
The `export_table_test` fails randomly on Jenkins. Thanks to @crd477, we can actually see the debug output now for the failed test:

http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-938/54/tests/

This says that we are failing `ExportTableTest::CheckContent`. This function is responsible for reading the data back in from the exported file to see whether it matches what we expect:

https://github.com/cmu-db/terrier/blob/master/test/storage/export_table_test.cpp#L268

The problem, however, is that the test only checks whether `CheckContent` returns true or false. There is no useful feedback to say which of the checks inside of `CheckContent` is failing.

We need to refactor this function so that the asserts are inside of `CheckContent` rather than just returning false for the entire function invocation. This will help us begin to understand why this test case is failing randomly.

YuzeLiao:
Happened to see this issue -- sorry for not providing enough debuggability :(. If you don't mind it I'll refactor the code and try to reproduce the failure to see what went wrong.

apavlo:
@YuzeLiao That would be great! Just let us know if you don't have the time for it so that I can assign it to another student.

apavlo:
Fixed in #1019

PR #1019, opened 2020-07-13
Issue #1013: Fix Random Failure Bugs & Add More Debug Info

YuzeLiao:
Issue #1013 raised two action items of export_table_test:

- Increase debugability;
- Fix potential bugs that caused random failures.

The first one was solved by:

1. Moving assertions into `CheckContent` so that we can distinguish different failure cases; 
2. Providing the location of mismatch data, i.e., which row and column has unexpected data; 
3. Reserving the seed and exported file if there are failures so that the failure can be reproduced.

The second item was solved by identifying a rare case where when a row only contains a single null value, it was exported as `""\n`. `\n` is its only delimiter but was not properly parsed before, so when such row was generated, the test would fail.

The new version of the test in this RB has been verified by repeating 3000 times without failures.

### DIFF
--- a/test/storage/export_table_test.cpp
+++ b/test/storage/export_table_test.cpp
@@ -83,6 +83,9 @@ struct ExportTableTest : public ::terrier::TerrierTest {
           end = true;
         }
         break;
+      case '\n':
+        end = true;
+        break;
       default:
         // Normal cases, we get a general ascii char.
         break;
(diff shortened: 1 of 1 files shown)


apavlo:
@Yuze this huge! Thanks so much!

YuzeLiao:
@apavlo My pleasure! It's what I'm supposed to do :)

Issue #998, opened 2020-07-01
Unable to Join Table with Itself


apavlo:
If you try to join a table with itself, then the server crashes. It happens no matter whether you use the `JOIN...ON` syntax or not.

```
terrier=# CREATE TABLE xxx (id INT PRIMARY KEY, val INT);
CREATE TABLE
Time: 13.901 ms
terrier=# INSERT INTO xxx VALUES (1,1),(2,2),(3,3),(4,4);
INSERT 0 4
Time: 8.845 ms
terrier=# SELECT * FROM xxx AS x1 JOIN xxx AS x2 ON x1.id = x2.id;
<< DEAD >>>
terrier=# SELECT * FROM xxx AS x1, xxx AS x2 WHERE x1.id = x2.id;
<< DEAD >>>
```

Server output:
```
[2020-07-01 10:01:02.638] [execution_logger] [error] Type-checking error! 
 Line: 0, Col: 0 => undefined: 'table_pr4'
Line: 0, Col: 0 => undefined: 'table_pr4'

[2020-07-01 10:01:02.638] [execution_logger] [error] Dumping AST:
[2020-07-01 10:01:02.640] [execution_logger] [error]  File (0x62d000014e10) <line:0:0> 
|- StructDecl (0x629000064e98) <line:0:0> State
| `- FieldDecl (0x629000064e28) <line:0:0> DUMMY
|   `- IdentifierExpr (0x629000064df8) <line:0:0>  'int32' int32
|- StructDecl (0x629000064db0) <line:0:0> Output10
| |- FieldDecl (0x629000064bc0) <line:0:0> col0
| | `- IdentifierExpr (0x629000064b90) <line:0:0>  'Integer' Integer
| |- FieldDecl (0x629000064c38) <line:0:0> col1
| | `- IdentifierExpr (0x629000064c08) <line:0:0>  'Integer' Integer
| |- FieldDecl (0x629000064cb8) <line:0:0> col2
| | `- IdentifierExpr (0x629000064c88) <line:0:0>  'Integer' Integer
| `- FieldDecl (0x629000064d48) <line:0:0> col3
|   `- IdentifierExpr (0x629000064d18) <line:0:0>  'Integer' Integer
|- FunctionDecl (0x6290000650e0) <line:0:0> setupFn
| `- FunctionLitExpr (0x6290000650b0) <line:0:0>  '(*struct{int32},*ExecutionContext)->nil' 
[...]

PR #1022, opened 2020-07-17
Fix self joins which use index joins

abalakumar083:
This is a bug fix for https://github.com/cmu-db/terrier/issues/998.

The issue with self-joins occurs when deriving the index bounds for an index join. Specifically, the problem is we only use table oid to determine which column value expression is a bound for the input table. Since in a self join there are two tables with the same table oid, in some cases we end up choosing a bound which references the wrong table. 

Later in the optimizer since the parent trims the input column set based on the table aliases the child provides - the parent discards the input columns to which the incorrect bound refers to. So it's not possible to convert the column value expressions to derived value expressions which causes the observed errors. 

Anyways the fix is simple, in order to make sure the bounds refer to the correct table, we check that the column value expression chosen as the bound has both the same table oid and also the same table name as the input table.


mbutrovich:
Naive question: why is the fix in the optimizer related to table names at all? Shouldn't everything have been resolved to oids by now?

abalakumar083:
> Naive question: why is the fix in the optimizer related to table names at all? Shouldn't everything have been resolved to oids by now?

abalakumar083:
So, I think you can use table oids to do all the things you currently use table aliases for inside of optimizer groups. That's probably the better way to fix this - i'll try and see if that works.

apavlo:
@dniu16 Add the queries listed here *after* we merge this PR: https://github.com/cmu-db/terrier/issues/998

PR #1026, opened 2020-07-21
Build improvements

gonzalezjo:
This is a continuation of #1017, which Jenkins refuses to pull from. This PR does a number of things.

Builds should be far faster on Linux through dynamic linking of the terrier and test binaries.

- This is because we do dynamic linking on Linux now. On macOS, the linker is remarkably fast and we can get by with static linking just fine. This is good, because we can't actually do dynamic linking on macOS without seemingly unfixable warning spam.

Builds should "just work" work on the latest version of mac OS

- This was a pain to troubleshoot, but builds work flawlessly on the dev machine with tests passing.

Clang tooling (tidy and format) are found automatically on more Linux distributions

- At least on Arch, clang-tidy and clang-format were never found automatically. This was a huge pet-peeve of mine, so I figured I'd fix it while touching so much build stuff.

We now statically link spdlog

- This reverts the explicit template instantiation pull request and imports (in header only form) the fmt library. This should be a net-improvement in build speed--especially with ccache and/or --but more importantly, it lets us use  ccache, and, later, with the ability to just install the spdlog binaries.

We now use spdlog version 1.7.0 instead of 1.0.0

- spdlog is now faster, more stable, and more powerful. ([See, for example, the features added in 1.3.0.](https://github.com/gabime/spdlog/releases/tag/v1.3.0))

We now pull in an up-to-date version of CMake on Linux machines

- By using pip to manage CMake for us instead of apt, we get to use new CMake versions and features. As a nice bonus, we get CMake version/feature consistency on both Mac and Linux, so we don't have to ever think about about build differences due to CMake versions. (Not that that was ever much of a concern.)

We now delegate dependency management for spdlog to CMake

- Removes ~14,000 lines of code from the repository. Uses the "new" FetchContent CMake feature for idiomatic management of dependencies. Currently configured to pull spdlog 1.7.0 from the official repository.


gonzalezjo:
This is probably ready for review, JIT bug aside.

gonzalezjo:
If there's a breakage, how hard would it be to rollback?

PR #1029, opened 2020-07-22
macOS: Only use the LLVM-bundled clang for code generation

mbutrovich:
This fix is included in #1017/#1026 but the scope continues to grow and I'd like this simple fix in. This resolves the error seen on latest AppleClang:
```
Reading input 'bytecode_handlers_ir.bc' ...
Parsing into LLVM Module ...
Error parsing bytecode handler bitcode file: Unknown attribute kind (60) (Producer: 'APPLE_1_1103.0.32.62_0' Reader: 'LLVM 8.0.1')
```
We only want to use the LLVM-bundled clang for code generation.

PR #1032, opened 2020-07-24
Avoid AppleClang on ASAN debug builds for macOS

gonzalezjo:
Depending on how ccache is set up, this could be useless. This is just here for discussion and testing. Based on discussion with Matt on Slack (in #general, dated 07-24-2020) this may be able to serve as a temporary fix for the issues seen in #1029.


mbutrovich:
Note that without #1029 as well this will still use AppleClang for LLVM compilation.

gonzalezjo:
That's true. I just wanted to see if ccache would return Clang's information or AppleClang's information to CMake.

PR #1039, opened 2020-07-29
Fix update counts, run TPCC in end-to-end performance on Jenkins.

### LINKED ISSUE #1038: The number of rows affected by updates are sometimes double-counted.
Merging #994 resulted in the update counts being incorrect.

This is a bug with NoisePage that was always there, but it was exposed because we fixed #592.

In particular, updates in the presence of indexes are translated as (delete+insert). There's no additional metadata in the .tpl saying "this is an update, don't count the rows twice", so it says 2 rows are affected because it counts the same row twice (once on delete, once on insert).

This was not a problem before because indexed update was always false coming out of the optimizer and/or we weren't translating indexed updates properly anyway.

lmwnshn:
Fix #1038.

This PR adds the execCtxAddRowsAffected(exec_ctx, num_rows_affected) TPL builtin and modifies the insert/update/delete translators to invoke this builtin explicitly, instead of the StorageInterface automatically messing with the rows affected count. The previous behavior is undesirable because after translation to TPL, a (delete+insert) update becomes indistinguishable from an explicit delete followed by an explicit insert.

This PR also unhacks the TrafficCopTest's count of updated rows.

This PR also modifies the Jenkinsfile to timeout TATP and run TPCC in Jenkins.


apavlo:
@lmwnshn can you add the update to the Jenkinsfile so that we run TPC-C in CI all the time now?

Issue #1037, opened 2020-07-29
~50% increase in build time after merging TPL2


mbutrovich:
Commit ed6131b023d470556b345c2488af0117181c7f97 (TPL2, PR #994)
```
make clean; time make -j12 terrier
...
real    3m2.182s
user    29m49.936s
sys     1m28.527s
```
Commit 0043d0161e1b08ae77e2e3ec4344d9f1548f1074 (pre-TPL2, PR #1022)
```
make clean; time make -j12 terrier
...
real    2m12.421s
user    20m22.539s
sys     1m3.711s
```
I have not run any tools (time-trace, time-report, etc.) to see where it's going. This is Ubuntu 20.04 with GCC 9.3.0. Debug build with ASAN.

lmwnshn:
I hacked together a quick python script to count include-graph ancestors (only among headers)

[grr.py.txt](https://github.com/cmu-db/terrier/files/5004497/grr.py.txt). Some notable slowdowns:

- xxh3.h is included 251 times (previously, 7 times). TPL2 unified front-end and execution engine hashing. However, execution engine hashing brings in xxh3, which is a huge library. Moreover, we cannot easily break it out to a cpp file because of some enable_if template magic that is being used.
- transaction_defs.h is included 203 times (previously, 75 times).
- log_io.h is included 189 times (previously, 6 times).

Another quick hack showed that if we can remove the xxh3 dependency from hash_util, compilation time on a similar setup goes to 28min40s.

lmwnshn:
Going to declare this papered over with unity builds.
```
time ninja noisepage
ninja noisepage  554.33s user 23.73s system 880% cpu 1:05.65 total
```

PR #1051, opened 2020-07-30
Forward declare (operator and expression) translators, fix TPL AstPrettyPrint for array length.

lmwnshn:
Trying to address #1037 slightly.

Also fixed a bug where we don't print out TPL array length properly.


src/execution/ast/ast_pretty_print.cpp
```diff
@@ -43,7 +44,7 @@ class AstPrettyPrintImpl : public AstVisitor<AstPrettyPrintImpl> {
 void AstPrettyPrintImpl::VisitArrayTypeRepr(ArrayTypeRepr *node) {
   os_ << "[";
   if (node->HasLength()) {
-    os_ << node->Length();
+    os_ << node->Length()->As<ast::LitExpr>()->Int64Val();
```
lmwnshn, line 47: The only logic change in this PR.

lmwnshn:
Forgot to try making unittests. Will throw in more includes over the weekend.

PR #1050, opened 2020-07-30
Bump xxHash from 0.7.2 to 0.8.0

mbutrovich:
Part of the summer toolchain and third party updates now that TPL2 is in.

Cherry-picking some of the changes from their [releases page](https://github.com/Cyan4973/xxHash/releases):
- api : stabilize XXH3
- perf: automatic vector detection and selection at runtime
- perf: added AVX512 support
- perf: improved speed for large inputs (~+20%)
- perf: improved latency for small inputs (~10%)

LICENSE file was removed since as best as I can tell it's redundant when source code files contain it at the top.

PR #1055, opened 2020-08-03
VarlenEntry microbenchmark

mbutrovich:
As a justification for #1050 I wrote a benchmark that hashes `VarlenEntry` both inline (CRC32) and non-inline (xxh3). I figure this might as well go into master first as we debate #1050.


benchmark/storage/varlen_entry_hash_benchmark.cpp
```diff
@@ -0,0 +1,101 @@
...
+    for (auto _ : state) {                                                                        \
+      benchmark::DoNotOptimize(varlen_entry.Hash());                                              \
+    }                                                                                             \
+  }                                                                                               \
+                                                                                                  \
+  state.SetItemsProcessed(state.iterations());                                                    \
+  state.SetBytesProcessed(state.iterations() * (varlen_bytes));
+
+// NOLINTNEXTLINE
+BENCHMARK_DEFINE_F(VarlenEntryHashBenchmark, Inline4Bytes)(benchmark::State &state) { VARLEN_HASH_BENCHMARK(4); }
```
pmenon, line 42: Instead of inline and non-inline versions for different lengths, why not just have two:`BENCHMARK_DEFINE_F(VarlenEntryHashBenchmark, Inline)` and `BENCHMARK_DEFINE_F(VarlenEntryHashBenchmark, NonInline)`. Then configure the benchmarks with length ranges, 1-12 for inlined, and 12+ otherwise? This will cut the number of benchmarks and can rid of the macro.

mbutrovich: How would that data get presented? We see very different performance at different VarlenEntry sizes, so it made more sense to test the sizes in separate benchmarks.

mbutrovich:
Added a benchmark for different equality comparison scenarios to help with #1066.

PR #1056, opened 2020-08-04
Fix parallel codegen and parallel scan, force insert/update/delete to be serial.

lmwnshn:
Previously, we accidentally set codegen to always generate serial queries.

Fixing codegen exposed some bugs in parallel scan translation, which have been fixed.
Inserts, updates, and deletes are serial for now.


lmwnshn:
For CLion users, File -> Settings -> Dynamic Analysis Tools -> Sanitizers -> Leak Sanitizer, add

suppressions=" ... /CLionProjects/terrier/build-support/data/lsan_suppressions.txt"

test/execution/sql_table_vector_iterator_test.cpp
```diff
@@ -135,13 +135,16 @@ TEST_F(TableVectorIteratorTest, ParallelScanTest) {
 
   struct Counter {
     uint32_t c_;
+    exec::ExecutionContext *exec_ctx_;
```
pmenon, line 138: Do you need this?

src/include/execution/exec/execution_settings.h
```diff
@@ -27,7 +27,7 @@ class EXPORT ExecutionSettings {
   }
 
   /** @return True if parallel query execution is enabled. */
-  constexpr bool GetIsParallelQueryExecution() const { return is_parallel_query_execution_; }
+  constexpr bool GetIsParallelQueryExecutionEnabled() const { return is_parallel_query_execution_; }
```
pmenon, line 30: Nit: rename to `is_parallel_execution_enabled_` ? As written, it's a little unclear, though function name clears it up.

test/execution/sql_table_vector_iterator_test.cpp
lmwnshn: Removed

src/include/execution/exec/execution_settings.h
lmwnshn: Renamed

Issue #1049, opened 2020-07-30
[TPL2] Should we use 32 bit or 64 bit integers by default?


lmwnshn:
I think [this](https://github.com/cmu-db/terrier/blob/643b4805f1473fd3cd09c4f922e066befc5311b4/src/execution/sema/sema_expr.cpp#L245) slipped through.

We generate untyped TPL code that contains integer literals.
These integer literals need to get assigned a type.

The default in TPL1 was to have 64-bit integers.
The default in TPL2 was to have 32-bit integers.

This is a problem for the bigint test, because then it doesn't parse the integer literal correctly. This was remedied with the above "hack".

I think a cleaner solution would probably be to extend TPL's parser to have a L suffix for 64-bit integers (in which case LitExpr itself would need to be extended with a bit more metadata), or to move to 64-bit integers by default again. I don't know if the latter has performance or frame size implications, I think @pmenon at the time said "whatever makes it easier for you".

A lot of ASAN errors that me and @tanujnay112 ran into while porting TPL2 was caused by having 32-bit integers where 64-bit integers were expected and vice versa.

The most recent victim of integer sizes is #1044.

lmwnshn:
We discussed this today. We think the right way to do this is:

1. TPL parses integers to 64-bit in LitExpr (so 64-bit default).
2. Sema is where the type of the LitExpr gets resolved. But this is quite an involved change.

lmwnshn:
This issue claims another victim @17zhangw . Should probably increase priority a little..

lmwnshn:
@pmenon has a fix for this in his repo, @lmwnshn to look at it.

PR #1060, opened 2020-08-06
Mini-Runners Update for TPLv2 Engine

17zhangw:
Updates the mini-runners to work with the new TPLv2 engine.
- Adds back memory tracking during hash joins, sorts, aggregations
- Various modeling changes (index-join, etc.)
- Stabilization fixes to the mini-runners

Note that this PR depends on https://github.com/cmu-db/terrier/pull/1056/.


benchmark/runner/mini_runners.cpp
```diff
@@ -55,11 +77,21 @@ catalog::db_oid_t db_oid{0};
  */
 int64_t warmup_iterations_num{5};
 
+/**
+ * Should skip large rows iterations
+ */
+bool skip_large_rows = false;
```
malin1993ml, line 83: I believe this is to skip the rerun iteration instead of the warmup iteration? Please make the document more specific.

benchmark/runner/mini_runners.cpp
```diff
@@ -26,13 +27,34 @@
...
 
+/**
+ * Should start small-row only scans
+ */
+bool rerun_start = false;
+
+/**
+ * Number of rerun iterations
+ */
+int64_t rerun_iterations = 5;
```
malin1993ml, line 46: nit: I think we're using 10 as default now?

benchmark/runner/mini_runners.cpp
```diff
@@ -640,6 +769,23 @@ class MiniRunners : public benchmark::Fixture {
     return plan;
   }
 
+  std::unique_ptr<planner::AbstractPlanNode> IndexNLJoinCorrector(
```
malin1993ml, line 772: nit: Maybe rename this to `IndexNLJoinChecker`? When you say corrector I thought this is actually going to fix something in the plan node. But this is just correctness checking, right? Similar with the function below.

script/model/data_class/opunit_data.py
```diff
@@ -91,10 +93,19 @@ def _interval_get_mini_runner_data(filename, model_results_path):
...
+    :param model_map: the map from OpUnit to the mini model
+    :param predict_cache: cache for the mini model prediction
+    :param trim: % of too high/too low anomalies to prune
+    :return: the list of Data for execution operating units
+    """
 
-def _execution_get_mini_runner_data(filename, model_map, predict_cache):
     # Get the mini runner data for the execution engine
     data_map = {}
+    anomaly_map = {}
```
malin1993ml, line 108: I wouldn't call this `anomaly_map`... maybe call this `raw_data_map`.

script/model/mini_trainer.py
```diff
@@ -145,15 +162,16 @@ def train(self):
                          help='Prediction results of the mini models')
     aparser.add_argument('--save_path', default='trained_nogen_model', help='Path to save the mini models')
     aparser.add_argument('--ml_models', nargs='*', type=str,
-                         default=["lr", "rf", "nn", 'huber', 'svr', 'kr', 'gbm'],
+                         default=["lr", "rf", "gbm"],
                          help='ML models for the mini trainer to evaluate')
     aparser.add_argument('--test_ratio', type=float, default=0.2, help='Test data split ratio')
+    aparser.add_argument('--trim', default=0.2, type=float, help='% of values to remove from both top and bottom')
```
malin1993ml, line 168: I don't think we need this trim thing anymore, right?

script/model/training_util/global_data_constructing_util.py
```diff
@@ -244,21 +268,61 @@ def _predict_grouped_opunit_data(data_list, mini_model_map, model_results_path):
...
                 # Don't modify prediction cache
                 y_pred = copy.deepcopy(y_pred)
                 y_pred[0][data_info.TARGET_CSV_INDEX[Target.MEMORY_B]] = adj_mem
 
             pipeline_y_pred += y_pred[0]
 
+        pipeline_y = copy.deepcopy(pipeline_y_pred)
+        if should_mult and simulate_cache:
+            # Scale elapsed time by 40% (this is a hack)
+            pipeline_y[-1] = pipeline_y[-1] * 0.4
```
malin1993ml, line 282: Though it's a hack, we should also apply to cycles, CPU times, and cache misses.

src/brain/operating_unit_recorder.cpp
```diff
@@ -160,6 +160,7 @@ void OperatingUnitRecorder::AggregateFeatures(brain::ExecutionOperatingUnitType
   // TODO(wz2): Populate actual num_rows/cardinality after #759
   size_t num_rows = 1;
   size_t cardinality = 1;
+  size_t num_iterations = 0;
```
malin1993ml, line 163: Call it `num_loops`. This would be similar to Postgres terminology and that's also the term I used in the paper.

src/include/brain/operating_unit.h
```diff
@@ -69,15 +69,17 @@ class ExecutionOperatingUnitFeature {
    * @param num_keys Number of keys
    * @param cardinality Estimated cardinality
    * @param mem_factor Memory adjustment factor
+   * @param num_iterations Number of iterations
```
malin1993ml, line 72: Oh and similarly, please rename the `num_iterations` here and in other places to `num_loops`.

benchmark/runner/mini_runners.cpp
17zhangw: Updated the documentation. This flag is actually used for only running up to the warmup limit (i.e TPCC).

script/model/mini_trainer.py
17zhangw: I kept the trim since the script doesn't default to medians for reducing to a single data point. The trim is supposed to be the x% trim.

malin1993ml: Got it!

src/execution/vm/vm.cpp
```diff
@@ -459,7 +459,7 @@ void VM::Interpret(const uint8_t *ip, Frame *frame) {  // NOLINT
 
   OP(ExecutionContextAddRowsAffected) : {
     auto *exec_ctx = frame->LocalAt<exec::ExecutionContext *>(READ_LOCAL_ID());
-    auto rows_affected = frame->LocalAt<int64_t>(READ_LOCAL_ID());
+    auto rows_affected = frame->LocalAt<int32_t>(READ_LOCAL_ID());
```
17zhangw, line 462: Patch suggested by @lmwnshn! Might be worth reverting this part once #1049 gets resolved.

PR #1067, opened 2020-08-07
Build improvements

gonzalezjo:
This is a continuation of #1017 and #1026. The PR does a number of things.

Builds should be far faster on Linux through dynamic linking of the terrier and test binaries.

- This is because we do dynamic linking on Linux now. On macOS, the linker is remarkably fast and we can get by with static linking just fine. This is good, because we can't actually do dynamic linking on macOS without seemingly unfixable warning spam.

Builds should "just work" work on the latest version of mac OS

- This was a pain to troubleshoot, but builds work flawlessly on a dev machine with all tests passing.

- Whatever issues existed with address sanitizer not doing anything on macOS appear to be gone.

- (07-29-2020) Jenkins uses clang-8 instead of AppleClang (thanks, Matt!)

- (07-29-2020) Jenkins correctly refers to the new compiler as clang-8 instead of AppleClang in the macOS stage names.

- (07-29-2020) Jenkins resolves LLVM_DIR using brew.

- (07-29-2020) Note: LLVM_DIR still has to be set via an environment variable, but I believe that that's desired behavior. I can make it automatic, but that would involve assuming that brew is installed and/or making it only automatic.

Clang tooling (tidy and format) are found automatically on more Linux distributions

- At least on Arch, clang-tidy and clang-format were never found automatically. This was a huge pet-peeve of mine, so I figured I'd fix it while touching so much build stuff.

We now statically link spdlog

- Required reverting the explicit logging template instantiation PR.

- Imports (in header only form) the fmt library.

- While those two points might normally be expected to cause a regression in build times, it actually doesn't. Because they facilitate the static linking of spdlog, they actually help to speed up builds.

We now use spdlog version 1.7.0 instead of 1.0.0

- Introduces a vast number of utility, performance, and stability improvements. ([See, for example, the features added in 1.3.0.](https://github.com/gabime/spdlog/releases/tag/v1.3.0))

We now pull in an up-to-date version of CMake on Linux machines

- By using pip to manage CMake for us instead of apt, we get to use new CMake versions and features. As a nice bonus, we get CMake version/feature consistency on both Mac and Linux, so we don't have to ever think about about build differences due to CMake versions. (Not that that was ever much of a concern.)

- It's worth discussing how important this is. I strongly prefer this approach since it guarantees consistency across platforms, but as we migrate to Ubuntu 20.04, maybe this will become less important.

We now delegate dependency management for spdlog to CMake

- Removes ~14,000 lines of code from the repository. Uses the "new" FetchContent CMake feature for idiomatic management of dependencies. Currently configured to pull spdlog 1.7.0 from the official repository.

We now compile gtest with address sanitizer when TERRIER_USE_ASAN=ON

- (07-29-2020) Fixes test failure on macOS

We now compile util_static and test_util differently for benchmarks

- This preserves old performance characteristics while having minimal impact on build times and avoiding benchmark crashes.

- Making benchmarks more dynamically linked was probably the better (and certainly the easier) solution, but I'm under the impression that there is some preference to keep benchmarks statically linked.

    - Or maybe that's just what I'm telling myself to rationalize the amount of time I spent on this one issue 🤔


gonzalezjo:
Passes on Jenkins: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/gonzalezjo-terrier-fork/detail/TerrierShared2/15/pipeline/114

The benchmark failure is from a supposed performance regression, but I would ignore that based on how noisy some of the benchmarks appear to be. Benchmark code should be essentially unaffected by this PR.

gonzalezjo:
I'm pretty alarmed that this passed through the normal Jenkins pipeline without any Jenkinsfile tweaking on master. I'd expect it to fail. @mbutrovich do you know what's up with that?

gonzalezjo:
I've been attempting to out why it's slower with ccache. I can't reproduce a slowdown for non-ccached builds with gcc-7. I'll try setting up gcc-9 and seeing what happens.

The problem is that a lot of the changes _are_ incompatible with master without bringing in a bunch of other changes. The spdlog bump, for example, requires the CMake upgrade and the ASAN linking changes.

gonzalezjo:
This is currently being split into a bunch of parts. Closing.

PR #1052, opened 2020-07-31
Added FilterTrace and filtered select1-5.test

dniu16:
Added FilterTrace, a class that filter out desired traces from a trace file, and its target filter-trace.

Instruction included in junit/README.md. Takes in a list of skip keywords. Queries that contain any keyword will be tagged as "skip" while the others will have the hash updated with postgresql. When TracefileTest is called, queries with the "skip" tag will be ignored.

Added filtered select1-5.test (queries without CASE, BETWEEN, WHERE, abs(, and WHEN in them).

select1.test: 14/97 failed
Failed query: Line 860, 1019, 1783, 3563, 4094, 4608, 5199, 5608, 6081, 6603, 7658, 9635, 10963, 11317

select2.test: 82 tests successful

select3.test: 157 tests successful

select4.test: 1043 tests successful, most of the tests are insert

select5.test: 704 tests successful, no query all insert

PR #1072, opened 2020-08-10
Update testing README to reflect current structure

jkosh44:


jkosh44:
Just a note, some other parts of this read me might be out of date as well. For example running `python3 junit/run_junit.py` gives me an error, but running `python ./junit/run_junit.py --build-type=debug --query-mode=simple` doesn't give me an error. Also `junit/README.md` also seems to reference non-existent files.

As I get more familiar with the project I can try and make those updates if someone else isn't already planning on it.

apavlo:
Let's merge these changes into #1052

PR #1073, opened 2020-08-11
Support for SET statements.

lmwnshn:
SET talks directly to the settings manager, poking the values.

With parser SET support yoinked from @thepinetree, thanks!

## Known issues

- SettingsManager assumes that 1 setting = 1 constant value expression. This is not necessarily the case in postgres, which can have `SET foo TO bar,baz`.
- Legacy? `ActionContext` and `SettingsCallback` stuff is not being used by `DBMain` or throughout the system.

## Example usage

Support for invalid values (type mismatch):
```
terrier=# set record_buffer_segment_size to a;
ERROR:  invalid value for parameter "record_buffer_segment_size": "a"
terrier=# \errverbose
ERROR:  22023: invalid value for parameter "record_buffer_segment_size": "a"
LOCATION:  /home/cmudb/CLionProjects/terrier/src/settings/settings_manager.cpp:272
```

Support for invalid values (out of range):
```
terrier=# set record_buffer_segment_size to -1;
ERROR:  -1 is outside the valid range for parameter "record_buffer_segment_size" (1 .. 1000000)
terrier=# \errverbose
ERROR:  22023: -1 is outside the valid range for parameter "record_buffer_segment_size" (1 .. 1000000)
LOCATION:  /home/cmudb/CLionProjects/terrier/src/settings/settings_manager.cpp:140
```

Support for valid value but runtime error:
```
terrier=# set record_buffer_segment_size to 1;
ERROR:  Callback failed after setting "record_buffer_segment_size" from "100000" to "1", revert succeeded.
terrier=# \errverbose
ERROR:  XX000: Callback failed after setting "record_buffer_segment_size" from "100000" to "1", revert succeeded.
LOCATION:  /home/cmudb/CLionProjects/terrier/src/settings/settings_manager.cpp:140
```

Support for valid value being set (I don't know how to show a setting, you can see it changed in the error message):
```
terrier=# set record_buffer_segment_size to 100001;
SET
terrier=# set record_buffer_segment_size to 1;
ERROR:  Callback failed after setting "record_buffer_segment_size" from "100001" to "1", revert succeeded.

terrier=# set record_buffer_segment_size to 5;
SET
terrier=# set record_buffer_segment_size to 1;
ERROR:  Callback failed after setting "record_buffer_segment_size" from "5" to "1", revert succeeded.
```

---

Also added support for DEFAULT. There was a bug with the settings manager that wouldn't let you define more than one string setting, that was fixed.


src/network/postgres/postgres_network_commands.cpp
```diff
@@ -123,6 +123,17 @@ Transition SimpleQueryCommand::Exec(const common::ManagedPointer<ProtocolInterpr
     t_cop->BeginTransaction(connection);
   }
 
+  // Set statements are manually handled here.
```
mbutrovich, line 126: What if one comes across in Extended Query?

lmwnshn: Added some logic to Bind and Execute.

mbutrovich:
Seems there are real failures in CI.

PR #1078, opened 2020-08-13
Hackathon hyperbolic trig functions

jkosh44:
This PR was created by cherry picking the following commit: https://github.com/apavlo/terrier/commit/e13a2c3098c8a76eef15a6c649d3b39a9a964c73 and then manually resolving conflicts, updating certain sections, and converting the JUnit tests to SQL trace tests.

The original PR can be found here: https://github.com/apavlo/terrier/pull/4

This works towards resolving #1001

Issue #1001, opened 2020-07-01
Merge SQL Functions from May 2020 Hack-a-Thon


apavlo:
There are several PRs in my personal fork from our May2020 hack-a-thon. These are new SQL functions written by the students last semester:

https://github.com/apavlo/terrier/pulls

We need to merge them together into a clean branch and then submit them as a single PR to the master repo. I have reviewed some of the students code but not all of it.

We will also need to port their modifications to [FunctionsTest.java](https://github.com/cmu-db/terrier/blob/master/script/testing/junit/src/FunctionsTest.java) to use @dniu16's new SQLite trace file format when he gets that merged.

I will try to go through and approve the PRs. This would be a good beginner project for somebody to help us in dealing with the conflicts.

wgy8283335:
@apavlo. I'd like to take this task. As we need to merge them together into a clean branch, maybe we should create a new branch based on master branch in the cmu-db/terrier project. I do not have the rights to create it, could you help me?
![image](https://user-images.githubusercontent.com/22066046/88353085-c6a56680-cd8e-11ea-8eab-5cf0fdd7c9aa.png)

wgy8283335:
@apavlo Dear Professor Pavlo. Do I need to create a new branch based on master branch in the "cmu-db/terrier" repository? Or I just review those commits in the "apavlo/terrier" repository?

apavlo:
@wgy8283335 Here is a rough outline of the steps on how to proceed with this task:

1. Fork apavlo/terrier:master into your own branch. It already contains some merged functions:
    a. https://github.com/apavlo/terrier/pull/7
    b. https://github.com/apavlo/terrier/pull/5
    c. https://github.com/apavlo/terrier/pull/4
2. Update your new branch to pull in the changes from cmu-db/terrier:master
3. Manually merge in the remaining pull requests from apavlo/terrier into your new branch: https://github.com/apavlo/terrier/pulls

For this last step, there will be many conflicts in the files because the everyone was updating the same ones. But these should be easy to manually resolve.

You'll also want to move the test queries out of `FunctionsTest.java` and add them to [functions.sql](https://github.com/cmu-db/terrier/blob/master/script/testing/junit/traces/functions.sql). We removed most of the old Java test files.

apavlo:
@wgy8283335 After discussing this internally, we think that it would be better to do this in stages with separate PRs, rather than one giant PR. I would start with reviving my current master branch and submit a PR after updating that.

[...]

PR #1080, opened 2020-08-13
Hackathon log2

jkosh44:
This PR was created by cherry picking this commit: https://github.com/apavlo/terrier/commit/5fdc3d8fcb4a2607df2fa0bcd52fa5c00fccdfd0 and applying some manual updates.

The original PR can be found here: https://github.com/apavlo/terrier/pull/5

Works towards resolving #1001 

A couple of notes about this PR:
- This will likely have some minor merge conflicts with https://github.com/cmu-db/terrier/pull/1078, I will resolve as we merge
- As far as I can tell postgres doesn't have a `log2` function, the closest I could find was `log(base, x)`
- Because of some precision issues the results of the test trace from terrier is off by one in the least significant digit from the results from postgres

### DIFF
--- a/script/testing/junit/traces/functions.test
+++ b/script/testing/junit/traces/functions.test
@@ -41,12 +41,12 @@ SELECT tan(double_val) AS result FROM functions1 WHERE is_null = 1
 1 values hashing to 68b329da9893e34099c7d8ad5cb9c940
 
 query I nosort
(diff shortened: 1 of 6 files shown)


jkosh44:
This will be combined with https://github.com/cmu-db/terrier/pull/1084 and https://github.com/apavlo/terrier/pull/16

PR #1081, opened 2020-08-13
Increase microbenchmark lax_tolerance from 30 to 50.

lmwnshn:
Suggested by @apavlo. The microbenchmarks are noisy and we end up merging a lot of stuff that fails that stage. Does anyone have objections?


apavlo:
To be clear, this fix **only** affects new benchmarks until we have at least 20 days of data from the nightly build.

PR #1085, opened 2020-08-14
Fix AdaptiveCheckTest in sql_filter_manager_test.cpp

gonzalezjo:
Noise is apparently massive in some situations, which can cause spurious test failures. As a result, we disable this test on Apple machines and get false positives on others. However, steps can be taken to decrease the noise and increase the signal for this test, such that false positives are exceedingly unlikely.

To do so, this commit increases the average number of samples taken and dramatically increasing the overhead of the "slow" filters, respectively. For added effect, we exaggerate selectivity differences as well. This is mostly apparent in execution traces.

This _should_ (we have to wait for Jenkins results) fix the test failures noted by Wan, which would allow this test to be ran on Apple machines. It would also fix some spurious test failures that I've encountered.


gonzalezjo:
Thanks for the quick review and merge @lmwnshn!

PR #1090, opened 2020-08-16
Hackathon math funcs

jkosh44:
This PR implements the following built-in math functions
- log2
- log10
- ceil
- floor
- truncate

I accomplished this by cherry picking the following commits and applying some manual updates
- log2
  - apavlo@5fdc3d8
- log10, ceil, floor
  - apavlo@3ad961a
  - apavlo@5bdd36e
- truncate
  - https://github.com/apavlo/terrier/pull/18/commits/35700fd3887c1fb16a8e3b68e5f64a5c88765b59

The original PRs can be found here:
- log2: apavlo#5
- log10, ceil, floor: apavlo#6
- truncate: https://github.com/apavlo/terrier/pull/18

Works towards resolving #1001

A couple of notes about this PR:
- As far as I can tell postgres doesn't have a `log2` function, the closest I could find was `log(base, x)`
- The postgresql function `truncate` is actually to remove all the rows of a table (https://www.postgresql.org/docs/9.1/sql-truncate.html). The equivalent function in postgres is `trunc` (https://www.postgresqltutorial.com/postgresql-trunc/)
- Because of some precision limitations the results of the test traces for many of these functions from terrier is different from the results from postgres


src/execution/vm/bytecode_generator.cpp
```diff
@@ -2375,7 +2395,15 @@ void BytecodeGenerator::VisitBuiltinCallExpr(ast::CallExpr *call) {
...
     case ast::Builtin::Sin:
-    case ast::Builtin::Tan: {
+    case ast::Builtin::Floor:
+    case ast::Builtin::Ceil:
+    case ast::Builtin::Log10:
+    case ast::Builtin::Tan:
+    case ast::Builtin::Truncate: {
+      VisitBuiltinTrigCall(call, builtin);
+      break;
+    }
```
apavlo, line 2405: Why have this when the next case (`ast::Builtin::Log2`) also calls `VisitBuiltinTrigCall`?

src/include/catalog/postgres/pg_proc.h
```diff
@@ -73,10 +73,15 @@ constexpr proc_oid_t TAN_PRO_OID = proc_oid_t(90);
...
 constexpr proc_oid_t UPPER_PRO_OID = proc_oid_t(93);
+constexpr proc_oid_t CEIL_PRO_OID = proc_oid_t(107);
+constexpr proc_oid_t FLOOR_PRO_OID = proc_oid_t(132);
+constexpr proc_oid_t LOG10_PRO_OID = proc_oid_t(134);
 constexpr proc_oid_t COSH_PRO_OID = proc_oid_t(127);
 constexpr proc_oid_t SINH_PRO_OID = proc_oid_t(128);
 constexpr proc_oid_t TANH_PRO_OID = proc_oid_t(129);
+constexpr proc_oid_t LOG2_PRO_OID = proc_oid_t(133);
 constexpr proc_oid_t VERSION_PRO_OID = proc_oid_t(150);
+constexpr proc_oid_t TRUNCATE_PRO_OID = proc_oid_t(125);
```
apavlo, line 84: AFAIK, the oid values don't mean anything. We should just make them increase by one.

src/execution/vm/bytecode_generator.cpp
jkosh44: Fixed

src/include/catalog/postgres/pg_proc.h
jkosh44: Fixed

src/include/catalog/postgres/pg_proc.h
```diff
@@ -71,19 +71,24 @@ constexpr proc_oid_t SIN_PRO_OID = proc_oid_t(89);
...
+constexpr proc_oid_t LOG10_PRO_OID = proc_oid_t(98);
+constexpr proc_oid_t LOG2_PRO_OID = proc_oid_t(99);
+constexpr proc_oid_t LOWER_PRO_OID = proc_oid_t(100);
+constexpr proc_oid_t UPPER_PRO_OID = proc_oid_t(101);
+constexpr proc_oid_t VERSION_PRO_OID = proc_oid_t(102);
+
+constexpr proc_oid_t NP_RUNNERS_EMIT_INT_PRO_OID = proc_oid_t(103);
+constexpr proc_oid_t NP_RUNNERS_EMIT_REAL_PRO_OID = proc_oid_t(104);
+constexpr proc_oid_t NP_RUNNERS_DUMMY_INT_PRO_OID = proc_oid_t(105);
+constexpr proc_oid_t NP_RUNNERS_DUMMY_REAL_PRO_OID = proc_oid_t(106);
```
apavlo, line 89: We should have the runner oids start at a higher # (say 1000) instead of mixing them in with regular SQL functions.

src/include/catalog/postgres/pg_proc.h
```diff
@@ -71,19 +71,24 @@ constexpr proc_oid_t SIN_PRO_OID = proc_oid_t(89);
...
 
 // SQL Function
-constexpr proc_oid_t DATE_PART_PRO_OID = proc_oid_t(101);
+constexpr proc_oid_t DATE_PART_PRO_OID = proc_oid_t(107);
```
apavlo, line 92: Move with this up with the other SQL functions.

jkosh44: Fixed

src/include/catalog/postgres/pg_proc.h
jkosh44: Fixed

PR #1096, opened 2020-08-19
Support for LIKE and NOT LIKE

wuwenw:
We previously have LIKE and NOT LIKE implemented as Select function and SQL function but not handled in the codegen, this PR mainly add support for LIKE and NOT LIKE during vector filtering.

All testcases in [PR](https://github.com/apavlo/terrier/blob/d00ddd28d9b92f5b23715f62ac7f7846452d3ad2/script/testing/junit/src/FunctionsTest.java#L262) from Hackathon has passed, but I haven't included them in a trace file.


script/testing/junit/traces/select.test
```diff
@@ -150,3 +150,336 @@ SELECT * FROM temp3 WHERE id=1 OR time=300 AND id=3 OR id=4 AND time=200 OR id=2
 
 statement ok
 DROP TABLE temp3
+
+# LIKE AND NOT LIKE
+statement ok
+CREATE TABLE like_tbl (int_val INT, double_val DECIMAL,str_i_val VARCHAR(32),str_a_val VARCHAR(32),is_null INT);
+
+statement ok
+INSERT INTO like_tbl (int_val, double_val, str_i_val, str_a_val, is_null) VALUES (1, 2.0, '123456', 'AbCdEf', 0);
```
lmwnshn, line 159: Sorry, but can you update the test to include more than one string? I am concerned about this case happening in the future:

- we break LIKE and it matches any string
- but it is not caught because there's only one value with is_null = 0 anyway

Suggestions:

```
INSERT INTO like_tbl (int_val, double_val, str_i_val, str_a_val, is_null) VALUES (1, 2.0, '123456', 'AbCdEF', 0);
INSERT INTO like_tbl (int_val, double_val, str_i_val, str_a_val, is_null) VALUES (1, 2.0, '123456', 'AbcdEf', 0);
INSERT INTO like_tbl (int_val, double_val, str_i_val, str_a_val, is_null) VALUES (1, 2.0, '123456', 'abcdEf', 0);
```

lmwnshn: Also, I'm not sure why there is a is_null column to begin with. Is it possible to do WHERE str_a_val IS NOT NULL?

src/execution/sql/operators/like_operators.cpp
```diff
@@ -75,7 +75,9 @@ bool Like::Impl(const char *str, size_t str_len, const char *pattern, size_t pat
       return false;
     }
   }
-
+  while (plen > 0 && *p == '%') {
```
lmwnshn, line 78: What does this do? The loop condition above was `plen > 0 && slen > 0`, so for us to have reached here, it must be the case that `plen <= 0 || slen <= 0`. Given your condition here, this executes when `slen <= 0`. Your code advances through the rest of the pattern, so if the string runs out of characters right as the pattern reaches a %, you will always say the string fits the pattern.

I am not convinced that this is correct, consider pattern `terrier%groundhog` and string `terrier`. What does your code currently do for that case?

src/execution/sql/vector_operations/like.cpp
```diff
@@ -47,9 +47,6 @@ void TemplatedLikeOperation(const Vector &a, const Vector &b, TupleIdList *tid_l
     throw EXECUTION_EXCEPTION(fmt::format("Inputs to (NOT) LIKE must be VARCHAR, left {} right {}.",
                                           TypeIdToString(a.GetTypeId()), TypeIdToString(b.GetTypeId())));
   }
-  if (a.IsConstant()) {
```
lmwnshn, line 50: As far as I can tell, this is fine...

script/testing/junit/traces/select.test
lmwnshn:
```
terrier=# create table foo (a text);
CREATE TABLE
terrier=# insert into foo values (null);
INSERT 0 1
terrier=# insert into foo values ('a');
INSERT 0 1
terrier=# select * from foo where a is not null;
 a 
---
 a
(1 row)
```

Seems to work for me, and is less hacky looking.

src/execution/sql/operators/like_operators.cpp
lmwnshn: I am mistaken, LGTM.

script/testing/junit/traces/select.test
apavlo: I think we should put some of the `LIKE` in the `WHERE` clause just to be safe.

wuwenw: added

wuwenw:
Added some more testcases to cover the following cases:
1. Only one string in the table
2. Several strings in the table
3. LIKE in the WHERE and SELECT str_val LIKE

Issue #1040, opened 2020-07-30
Column order output is not consistent when using joins


abalakumar083:
So there's an issue with joins right now where the left table's columns and the right table's columns will be flipped in the output occasionally. For example the following queries,
```
CREATE TABLE xxx (id INT PRIMARY KEY, val INT);  
CREATE TABLE yyy (id INT PRIMARY KEY, val INT);
INSERT INTO xxx VALUES (1,1),(2,2),(3,3),(4,4);
INSERT INTO yyy VALUES (1,10),(2,9),(4,8);
select * from xxx INNER JOIN yyy ON xxx.id = yyy.id;
```
result in the following output:
```
 id | val | id | val 
----+-----+----+-----
  1 |  10 |  1 |   1
  2 |   9 |  2 |   2
  4 |   8 |  4 |   4
(3 rows)
```
The reason for this (well at least in the above example) is because the output tables are ordered in whatever order `regular_alias_table_map_` is iterated through (relevant function is `BinderContext::GenerateAllColumnExpressions` ), but since `regular_alias_table_map_` is a unordered map there's no set order for the iteration.

apavlo:
@soumilm after we merge in the upcoming CTE PR.

PR #1100, opened 2020-08-19
Left outer join 

abalakumar083:
Adds support for left outer join queries. In addition it includes minor changes in the binder in order to fix https://github.com/cmu-db/terrier/issues/1040, since a fix is needed for left outer join to work properly.


script/testing/junit/traces/select.test
```diff
@@ -150,3 +150,28 @@ SELECT * FROM temp3 WHERE id=1 OR time=300 AND id=3 OR id=4 AND time=200 OR id=2
...
+CREATE TABLE yyy (id INT PRIMARY KEY, val INT);
+
+statement ok
+INSERT INTO xxx VALUES (2,2), (3,3), (4,4), (5,5);
+
+statement ok
+INSERT INTO yyy VALUES (2,2), (3,3), (4,4);
+
+query I nosort
+SELECT * FROM xxx LEFT OUTER JOIN yyy ON xxx.id = yyy.id;
```
apavlo, line 169: Are there more complex queries than this? For example, trying joining on the non-pkey columns.

And what about joining with an inequality:

`SELECT * FROM xxx LEFT OUTER JOIN yyy ON xxx.id > yyy.id + 1;`

src/include/execution/compiler/work_context.h
```diff
@@ -76,6 +76,11 @@ class WorkContext {
    */
   void SetExpressionCacheEnable(bool val) { cache_enabled_ = val; }
 
+  /**
+   * Advances the WorkContext to the next step in the pipeline
+   */
+  void AdvancePipelineIter() { pipeline_iter_++; }
```
apavlo, line 82: Maybe rename to `AdvancePipelineIterator`?

src/include/execution/sql/join_hash_table.h
```diff
@@ -272,4 +274,79 @@ inline HashTableEntryIterator JoinHashTable::Lookup<true>(const hash_t hash) con
...
+ *
+ * Users use the OOP-ish iteration API:
+ * for (JoinHashTableIterator iter(table); iter.HasNext(); iter.Next()) {
+ *   auto row = iter.GetCurrentRowAs<MyFancyAssType>();
+ *   ...
+ * }
+ *
+ * // TODO(pmenon): Vectorized version, too.
+ */
+class HashTableNaiveIterator {
```
apavlo, line 295: Why is this the "naive" iterator? The comment above says that it is the "Join Hash Table Iterator"?

abalakumar083: I agree naming is an issue. There's two different iterators for the join hash table. One iterator takes a hash and lets you iterate over all the matching hash values called HashTableEntryIterator. This iterator simply iterates over all the entries in the hash table. I didn't know what to call it, because it is also a entry iterator.

src/execution/compiler/pipeline.cpp
```diff
@@ -238,28 +238,28 @@ ast::FunctionDecl *Pipeline::GenerateRunPipelineFunction(query_id_t query_id) co
       op->BeginPipelineWork(*this, &builder);
     }
 
+    // var pipelineState = @tlsGetCurrentThreadState(...)
+    auto exec_ctx = compilation_context_->GetExecutionContextPtrFromQueryState();
+    auto tls = codegen_->ExecCtxGetTLS(exec_ctx);
+    auto state = codegen_->TLSAccessCurrentThreadState(tls, state_.GetTypeName());
```
tanujnay112, line 244: Why was this moved up?

src/execution/compiler/operator/hash_join_translator.cpp
```diff
@@ -252,6 +273,48 @@ void HashJoinTranslator::CheckRightMark(WorkContext *ctx, FunctionBuilder *funct
...
+    auto left_mark = codegen->AccessStructMember(codegen->MakeExpr(build_row_var_), build_mark_);
+
+    // If mark is true, then row was not matched
+    If check_condition(function, left_mark);
+    {
+      // // outerJoinConsumer(queryState, buildRow);
+      std::initializer_list<ast::Expr *> args{GetQueryStatePtr(),
+                                              codegen->MakeExpr(GetPipeline()->GetPipelineStateVar()),
+                                              codegen->MakeExpr(build_row_var_)};
+      function->Append(codegen->Call(outer_join_consumer_, args));
```
tanujnay112, line 309: Could it not be possible that while you created the pipeline for outer join consumer you would also want to make sure you called the initializepipeline and the finishpipeline for the outer join consumer's pipeline?

src/execution/compiler/pipeline.cpp
abalakumar083: Actually looking at this again, I think it's a mistake. I needed the `pipelineState` inside of `runPipeline` even in the parallel case, because I needed to pass it to leftOuterConsumer. But I think this is rare. I'll revert the changes and move this code into hash_join_translator.cpp.

src/execution/compiler/operator/hash_join_translator.cpp
abalakumar083: I think this is taken care of. For the examples that I'm running the last operator in the pipeline is the `outputTranslator` and inside the `outputTranslator` finishpipeline is where the result is finalized and I can see the corresponding tpl (`@resultBufferFinalize(queryState.execCtx)`) being generated.

apavlo (dismissed):
LGTM, but I am not the best person to review this code.

src/execution/compiler/pipeline.cpp
```diff
@@ -243,23 +243,23 @@ ast::FunctionDecl *Pipeline::GenerateRunPipelineFunction(query_id_t query_id) co
...
 
+      // SerialWork(queryState, pipelineState)
       builder.Append(
           codegen_->Call(GetWorkFunctionName(), {builder.GetParameterByPosition(0), codegen_->MakeExpr(state_var_)}));
     }
 
     // Let the operators perform some completion work in this pipeline.
-    for (auto op : steps_) {
-      op->FinishPipelineWork(*this, &builder);
+    for (auto op_iter = steps_.rbegin(); op_iter < steps_.rend(); op_iter++) {
```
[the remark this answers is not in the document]
abalakumar083: Correct me if this wrong, I think you need to finish the pipeline work in the opposite order that they're started. 

For example, the simplest type of left outer join query generates a pipeline with operators as [OutputTranslator, HashJoinTranslator, SeqScanTranslator]. If in this case I have OutputTranslator finish finalizing the output before HashJoinTranslator adds the extra rows, then the extra rows don't appear properly.

src/execution/compiler/operator/hash_join_translator.cpp
```diff
@@ -216,7 +238,7 @@ void HashJoinTranslator::CheckJoinPredicate(WorkContext *ctx, FunctionBuilder *f
   auto *codegen = GetCodeGen();
 
   auto cond = ctx->DeriveValue(*join_plan.GetJoinPredicate(), this);
-  if (join_plan.RequiresLeftMark()) {
+  if (join_plan.GetLogicalJoinType() == planner::LogicalJoinType::LEFT_SEMI) {
```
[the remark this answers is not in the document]
abalakumar083: I thought this exists to prevent rows from the left table from appearing multiple times in the output. I didn't think this is appropriate for left outer joins, because the same left row can match with different right rows.

src/execution/compiler/operator/hash_join_translator.cpp
```diff
@@ -61,6 +63,26 @@ void HashJoinTranslator::DefineHelperStructs(util::RegionVector<ast::StructDecl
   decls->push_back(struct_decl);
 }
 
+void HashJoinTranslator::DefineHelperFunctions(util::RegionVector<ast::FunctionDecl *> *decls) {
```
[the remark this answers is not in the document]
abalakumar083: So I'm not sure how to get the table vector which corresponds to the right table row from within `HashJoinTranslator`, if I can get access to that then I can refactor `outerJoinConsumer` to take both the buildRow and the rightRow as arguements. This isn't an issue right now, because since I'm only calling `outerJoinConsumer` for the left outer join specfic portion - the right values are always NULL. 

(Also I would prefer refactoring this as well, not only for the code duplication, but would be easier if someone wanted to implement right outer join or outer join later)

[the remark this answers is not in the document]
pmenon: In this case, materialize the right row and call `outerJoinConsumer`. Reduce duplication in generated code.

src/execution/compiler/operator/hash_join_translator.cpp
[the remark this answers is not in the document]
pmenon: You're right. I misread where this change was happening.

src/execution/compiler/pipeline.cpp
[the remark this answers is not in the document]
pmenon: Technically, these translator finalizers should be independent of each other. Adding the `@resultBufferFinalize()` was a mistake.

The correct solution is to:
1. Remove the call in `OutputTranslator` to finalize the buffer [here](https://github.com/cmu-db/terrier/blob/master/src/execution/compiler/operator/output_translator.cpp#L46).
2. Finalize the output buffer *after* the query executes [here](https://github.com/cmu-db/terrier/blob/master/src/execution/compiler/executable_query.cpp#L161).

For now:
1. Change to:
```
for (auto iter = Begin(), end = End(); iter != end; ++iter) { ... }
```
2. Ditto for the calls to `BeginPipelineWork()`
3. Leave TODO here indicating the order of invocation shouldn't matter.
4. Create an issue to remove the finalize call in output translator, remove associated builtins and bytecodes, too.

tanujnay112 (approved):
LGTM mostly, one test in compiler failing looks like because the two join consumers are declared in reverse order.

src/execution/compiler/operator/hash_join_translator.cpp
```diff
@@ -51,14 +55,44 @@ HashJoinTranslator::HashJoinTranslator(const planner::HashJoinPlanNode &plan, Co
...
   if (GetPlanAs<planner::HashJoinPlanNode>().RequiresLeftMark()) {
     fields.push_back(codegen->MakeField(build_mark_, codegen->BoolType()));
   }
   ast::StructDecl *struct_decl = codegen->DeclareStruct(build_row_type_, std::move(fields));
   struct_decl_ = struct_decl;
   decls->push_back(struct_decl);
+
+  /* Probe row declaration */
+  // TODO(abalakum): support mini-runners for this struct as well
+  fields = codegen->MakeEmptyFieldList();
```
pmenon, line 71: You should only need the probe row if you have a left outer join.

src/execution/compiler/operator/hash_join_translator.cpp
```diff
@@ -231,8 +275,17 @@ void HashJoinTranslator::CheckJoinPredicate(WorkContext *ctx, FunctionBuilder *f
...
+    // var probeRow : ProbeRow
+    auto probe_row_type = codegen->MakeExpr(probe_row_type_);
+    auto probe_row = codegen->MakeExpr(probe_row_var_);
+    function->Append(codegen->DeclareVarNoInit(probe_row_var_, probe_row_type));
+    // Fill row.
+    FillProbeRow(ctx, function, codegen->MakeExpr(probe_row_var_));
+    // joinConsumer(queryState, pipelineState, buildRow, probeRow);
+    std::initializer_list<ast::Expr *> args{GetQueryStatePtr(), codegen->MakeExpr(GetPipeline()->GetPipelineStateVar()),
+                                            codegen->MakeExpr(build_row_var_), codegen->AddressOf(probe_row)};
+    function->Append(codegen->Call(join_consumer_, args));
```
pmenon, line 288: Only call the outer join if you have a left outer-join. Otherwise, inline the consumption code.

src/include/execution/compiler/operator/hash_join_translator.h
```diff
@@ -37,12 +37,20 @@ class HashJoinTranslator : public OperatorTranslator {
                      Pipeline *pipeline);
 
   /**
-   * Declare the build-row struct used to materialized tuples from the build side of the join.
-   * @param decls The top-level declarations for the query. The build-row struct will be
-   *                        registered here after it's been constructed.
+   * Declare the build-row struct and probe-row struct used to materialized tuples from the build side
```
pmenon, line 40: Fix these comments to reflect under what conditions structs are actually generated.

src/execution/compiler/operator/hash_join_translator.cpp
```diff
@@ -252,6 +305,60 @@ void HashJoinTranslator::CheckRightMark(WorkContext *ctx, FunctionBuilder *funct
...
+    // var buildRow = @joinHTIterGetRow()
+    function->Append(
+        codegen->DeclareVarWithInit(build_row_var_, codegen->JoinHTIteratorGetRow(jht_iter_expr, build_row_type_)));
+
+    auto left_mark = codegen->AccessStructMember(codegen->MakeExpr(build_row_var_), build_mark_);
+
+    // If mark is true, then row was not matched
+    If check_condition(function, left_mark);
+    {
+      // var probeRow : ProbeRow
```
pmenon, line 337: These are for build rows, whose probe-side attributes are null?

pmenon (changes requested):
You should only be generating and *using* the outer join consumer function if you have left outer join. Otherwise, the probe-side consumption should be inlined using `context->Push()`.

src/execution/compiler/operator/hash_join_translator.cpp
abalakumar083: I'm not sure how to parse the question exactly. 

These probeRows are for the buildRows that weren't matched, so for left outer join need to output the buildRow + NULLs. Since, I'm passing a probe row to the `joinConsumer` function which it uses to read the right-side values, I set the probe row values to NULL before passing it.

rohanaggarwal7997 (approved):
Giving thumbs up to tanuj's review to give it an approving review

pmenon (approved):
LGTM

PR #1109, opened 2020-08-22
Unix domain sockets & findings from offcputime analysis

gonzalezjo:
# Unix Domain Sockets 
## Background

While investigating our offcputime, I was struck by how much time was spent strictly on network tasks. EC2 numbers pointed to about 8% being spent on just flushing write buffers, while my own profiles from VMware Workstation suggested that about 80% of offcputime was spent on flushing write buffers. (Small update: Restarting the VM brought that number down drastically. I went from ~120 t/s -> ~220+t/s on 1 terminal from faster writes!) 

I wanted to see if Postgres was similarly affected on a VM. As it turns out, it absolutely is, and it is affected to the exact same extent. However, Postgres can cheat and use Unix domain sockets (think named pipes, but accessed via the filesystem) on supported tests. I sought to evaluate the extent to which Unix domain sockets could impact our performance by reducing offcputime.

## Procedure

Disclaimer: I can't run oltpbench with more than about 10 terminals without the DB dying. This is the case for master *and* my code. I get these errors on the server:

```
[2020-08-22 17:29:34.801] [network_logger] [error] Error writing: Connection reset by peer
[2020-08-22 17:29:34.801] [network_logger] [error] Error writing: Connection reset by peer
[2020-08-22 17:29:34.801] [network_logger] [error] Error writing: Connection reset by peer
[2020-08-22 17:29:34.801] [network_logger] [error] Error when filling read buffer
[2020-08-22 17:29:34.801] [network_logger] [error] Error when filling read buffer
[2020-08-22 17:29:34.801] [network_logger] [error] Error when filling read buffer
[2020-08-22 17:29:34.802] [network_logger] [error] Error when filling read buffer
```

All benchmarks were performed with 200 connection threads on a machine with twenty-four free cores. The benchmark used was tpcc from oltpbench. tpcc is a nice workload for this PR, since it really exaggerates protocol overhead. The terrier server was restarted prior to each run of the benchmark. See the section titled "The PR" for my configuration file. The virtual machine software used is VMware Workstation 15. 

The virtual machine is running Ubuntu 18.04 LTS with kernel version 5.4.0-42-generic.

The bare metal machine is running Arch Linux with kernel version 5.4.57-1-lts.

## Results

| Machine Type 	| Terminals 	| Scale Factor 	| Socket Type 	| Transactions / Second 	|
|--------------	|-----------	|--------------	|-------------	|-----------------------	|
| Bare metal   	| 10        	| 10           	| UNIX        	| 8346 (+34%)           	|
| Bare metal   	| 10        	| 10           	| INET        	| 6214                  	|
| Bare metal   	| 1         	| 1            	| UNIX        	| 914 (+37%)            	|
| Bare metal   	| 1         	| 1            	| INET        	| 668                   	|
| VMware       	| 10        	| 10           	| UNIX        	| 3728 (+24%)           	|
| VMware       	| 10        	| 10           	| INET        	| 3005                  	|
| VMware       	| 1         	| 1            	| UNIX        	| 289 (+28%)            	|
| VMware       	| 1         	| 1            	| INET        	| 226                   	|

## Discussion

To be honest, I'm pretty confused. The results go entirely against my expectations. I expected the gains from Unix domain sockets on bare metal to be a small fraction of the gains from a virtual machine. My interpretation of bare metal profiling data suggested that I had at most ~10% to gain from this PR. I was wrong on both counts. I can consistently reproduce the speedups that I have measured here. I'll revisit this with VTune again, profiling the code with and without Unix domain sockets, and see what's up.

## The PR and its future

This PR creates a new network test, two new settings, and a "new" (it extracts out existing code) method in terrier_server. It supports simultaneous usage of both Unix and networked/INET sockets, with logs displaying relevant socket information, and with graceful handling and recovery from errors.

Since this adds a potentially useful feature from two big DBs (Postgres and MySQL) with next to no overhead, it's nice from a features perspective. That said, I **don't think it makes a ton of sense to merge this PR**. It (1) doesn't bring us much closer to project goals, and (2) isn't actually fixing any IO-protocol-induced-gap between us and Postgres/MySQL in benchmarks, since as it turns out, oltpbench can't actually use their Unix domain sockets. 

In light of the last comment, I should note that to benchmark this PR, you would want to use [my fork of Matt's fork of oltpbench]( https://github.com/gonzalezjo/oltpbench/tree/unix-domain-sockets). The real change is just two lines, though for others' convenience and for the sake of reproducibility, I've also changed the default settings of the benchmark to mirror what I've used. Instructions for running the benchmark are available [here](https://github.com/cmu-db/terrier/wiki/Profiling-on-EC2).

### Investigating the source of the speedup

At this point, this is my main goal. I'd like to try to reconcile my interpretation of previous profiling data with the results that I've collected. This will involve more profiling, but I expect it to be straightforward.

### Investigating libevent/epoll

This seems like a dead end.

After running oltpbench with Wireshark and being stunned by the amount of packets being sent by tpcc, I wondered if libevent and epoll could be potential sources of overhead. I've never been a fan of epoll--the only decent API that libevent supports on Linux--so I tried [porting things over](https://github.com/gonzalezjo/terrier/tree/unix-domain-sockets) to libev. The idea was that I'd be able to use io_uring, a much more performant (and, if implemented properly, probably much cleaner to use) kernel API. After our meeting, this didn't seem worth it. So for now, work on libev and io_uring support is on pause.

In either case, I would hope that libevent's epoll backend would be more than fast enough for a mere ten terminals. epoll is plenty fast for our cases, but that's assuming that it's implemented well, and it's very hard to tell if that's the case from libevent's codebase. Attempts to update the libevent version and to use edge-triggered epoll had no impact on performance, so I think we're still far away from any theoretical epoll limits.

### Investigating loopback and TCP overhead

This is probably also a dead end, but less of a dead end than the libevent/epoll stuff.

I'd like to work on improving our INET socket overhead, but at this point, that might not be doable without being a kernel engineer. Nothing I tried measurably reduced loopback and/or INET overhead, and I tried a lot. Still, I think it's worth digging some more. 

One question I have is how much of the speedup comes from avoiding TCP, and how much comes from using a glorified pipe instead of loopback. This would be interesting to know, but I have no idea how I'd measure, and I'm not convinced that I'd be able to make anything useful from the answer. 

## If I don't think this should be merged, then why did I make this giant PR?

Good question. Four reasons:

0. I think that there is potential for good discussion from the findings here.

1. I've spent way too much time investigating off CPU time and don't want it to go to waste. I want the information I've collected to be accessible to others. I needed to write this all down anyway. It only made sense to clean up my notes, put them in markdown, and share them.

2. While Unix domain sockets aren't particularly practical, they do demonstrate a serious source of overhead in a way that anyone can verify--we no longer need to speculate, for example, as to whether or not (and/or, to what extent) loopback overhead might be costing us. Hence, as an academic exercise, I found writing this PR to be extremely useful as a way to gain a deeper insight into the performance characteristics of the system.

3. By sharing the findings from my investigation in this PR, I'm hoping that someone smart might realize a way to claw back some of the overhead losses.


gonzalezjo:
Further updates: epoll overhead is real and fairly significant--I think it's responsible for an avoidable ~5-10% of loss on localhost throughput--but designing around it is extremely painful. And of course, alternatives have overhead of their own. 

From investigations of macOS vs. Linux (thanks Matt), Postgres in a VM vs. Postgres on bare metal, and Noisepage in a VM vs. Noisepage on bare metal, it seems like our TPC-C throughput is very heavily constrained by networking and other kernel/OS related overhead. While efforts like this PR can reduce that overhead substantially, a huge amount of overhead remains.


I believe that the extended query protocol, and to a lesser extent, its implementation (through epoll) present a _major_ system performance bottleneck on benchmarks like TPC-C. I have also been plagued with issues while profiling from variance--both in the statistical and literal sense--in network stack performance across platforms, kernel versions, and even CPU microcode versions(!)


This makes me question whether such network-IO intensive workloads, such as TPC-C, are a good fit for modeling and sanity-checking our performance data.

src/network/terrier_server.cpp
```diff
@@ -38,44 +43,114 @@ TerrierServer::TerrierServer(common::ManagedPointer<ProtocolInterpreter::Provide
...
+      sin.sin_addr.s_addr = INADDR_ANY;
+      sin.sin_port = htons(port_);
+
+      return sin;
+    } else {
+      // Builds the socket path name
+      // Naming conventions come from https://www.postgresql.org/docs/9.3/runtime-config-connection.html
+      const std::string socket_path = fmt::format("{0}/.s.PGSQL.{1}", socket_directory_, port_);
+      struct sockaddr_un sun = {0};
+      memset(&sun, 0, sizeof(struct sockaddr_un));
```
thepinetree, line 71: What is the purpose of the memset here?

test/network/network_test.cpp
```diff
@@ -85,7 +86,8 @@ class NetworkTests : public TerrierTest {
       server_ =
           std::make_unique<TerrierServer>(common::ManagedPointer<ProtocolInterpreter::Provider>(&protocol_provider_),
                                           common::ManagedPointer(handle_factory_.get()),
-                                          common::ManagedPointer(&thread_registry_), port_, connection_thread_count_);
+                                          common::ManagedPointer(&thread_registry_), port_, connection_thread_count_,
+                                          true /* Use unix domain socket */, socket_directory_);
```
thepinetree, line 90: Does using the unix domain socket here force it to be used on all the tests? If yes, I would recommend making a new file to test the domain socket behavior individually and rename this file to correspond to internet socket tests.

gonzalezjo: No. While the Unix domain socket is created and left open for all network tests, it only actually gets tested by the one test I've created.

src/network/terrier_server.cpp
gonzalezjo: Totally unneeded. Was based on what we used to do for network sockets, and I left that in by accident. Thanks for catching that!

test/network/network_test.cpp
thepinetree: Gotcha, could you clarify the difference in setup in network_test.cpp:196 that causes the Unix domain socket to be used here in a comment.

thepinetree (approved):
LGTM

gonzalezjo:
So, I can confirm that the default Postgres configuration

Postgres definitely gains a lot from using Unix domain sockets. Benched with postgres running over Optane (NVMe, not pmem):

```
jordig@dev5:~/oltpbench/oltpbench$ pgbench -c 10 -T 60 -j 10 -M extended -P 1 -h 127.0.0.1 -p 5433
Password: 
starting vacuum...end.
progress: 1.0 s, 12806.3 tps, lat 0.775 ms stddev 0.139
progress: 2.0 s, 13542.9 tps, lat 0.738 ms stddev 0.115
progress: 3.0 s, 13529.2 tps, lat 0.739 ms stddev 0.110
progress: 4.0 s, 13707.6 tps, lat 0.730 ms stddev 0.127
progress: 5.0 s, 13835.9 tps, lat 0.723 ms stddev 0.119
progress: 6.0 s, 13776.8 tps, lat 0.726 ms stddev 0.112
progress: 7.0 s, 13784.6 tps, lat 0.725 ms stddev 0.116
progress: 8.0 s, 13637.0 tps, lat 0.733 ms stddev 0.117
progress: 9.0 s, 13154.1 tps, lat 0.760 ms stddev 0.189


^C


jordig@dev5:~/oltpbench/oltpbench$ pgbench -c 10 -T 60 -j 10 -M extended -P 1
starting vacuum...end.
progress: 1.0 s, 17512.7 tps, lat 0.567 ms stddev 0.099
progress: 2.0 s, 17673.7 tps, lat 0.566 ms stddev 0.090
progress: 3.0 s, 17570.8 tps, lat 0.569 ms stddev 0.095
progress: 4.0 s, 17414.5 tps, lat 0.574 ms stddev 0.097
progress: 5.0 s, 17401.7 tps, lat 0.575 ms stddev 0.098
progress: 6.0 s, 17128.6 tps, lat 0.584 ms stddev 0.098
progress: 7.0 s, 17420.6 tps, lat 0.574 ms stddev 0.095
progress: 8.0 s, 17277.9 tps, lat 0.579 ms stddev 0.097
progress: 9.0 s, 17086.5 tps, lat 0.585 ms stddev 0.096

```


Results are reproducible: Postgres regularly gains around 30% from using domain sockets.

src/network/terrier_server.cpp
```diff
@@ -38,44 +43,113 @@ TerrierServer::TerrierServer(common::ManagedPointer<ProtocolInterpreter::Provide
...
+      const std::string socket_path = fmt::format("{0}/.s.PGSQL.{1}", socket_directory_, port_);
+      struct sockaddr_un sun = {0};
 
-  struct sockaddr_in sin;
-  std::memset(&sin, 0, sizeof(sin));
-  sin.sin_family = AF_INET;
-  sin.sin_addr.s_addr = INADDR_ANY;
-  sin.sin_port = htons(port_);
+      // Validate pathname
+      if (socket_path.length() >= 108 /* Max Unix socket path length */ ) {
```
gengkev, line 73: Allegedly this limit is 104 on Mac OS (which I don't use so I can't confirm): https://serverfault.com/questions/641347/check-if-a-path-exceeds-maximum-for-unix-domain-socket

Would it be appropriate to use `sizeof(sun.sun_path)` instead?

gonzalezjo: Good idea. I think so. I'll do that now.

thepinetree (approved):
LGTM

src/include/main/db_main.h
```diff
@@ -229,17 +229,20 @@ class DBMain {
      * @param traffic_cop argument to the ConnectionHandleFactor
      * @param port argument to TerrierServer
      * @param connection_thread_count argument to TerrierServer
+     * @param use_unix_domain_sockets argument to TerrierServer
+     * @param socket_directory argument to TerrierServer
      */
     NetworkLayer(const common::ManagedPointer<common::DedicatedThreadRegistry> thread_registry,
                  const common::ManagedPointer<trafficcop::TrafficCop> traffic_cop, const uint16_t port,
-                 const uint16_t connection_thread_count) {
+                 const uint16_t connection_thread_count, const bool use_unix_domain_sockets,
```
mbutrovich, line 237: As a general style nit: We should prefer enums over a boolean arg here. It gives the option of easily adding another value to an enum like `SocketType` without refactoring the signature of the function.

mbutrovich:
Requested a review from @lmwnshn so he can take a quick pass to make sure it doesn't conflict with anything he's poking at in the network layer.

lmwnshn (approved):
I think error handling would be detectable if it were made an assert. Additionally, some changes to commenting would be nice. I am in the process of refactoring the network layer and will probably change stuff in terrier_server in a couple of days. Overall, LGTM though! Would be fine with merging as-is.

src/network/terrier_server.cpp
```diff
@@ -39,44 +42,109 @@ TerrierServer::TerrierServer(common::ManagedPointer<ProtocolInterpreter::Provide
...
+      if (socket_path.length() > sizeof(sun.sun_path) /* Max Unix socket path length */) {
+        NETWORK_LOG_ERROR(fmt::format("Domain socket name too long (must be <= {} characters)", sizeof(sun.sun_path)));
+        throw NETWORK_PROCESS_EXCEPTION(fmt::format("Failed to name {} socket.", socket_description));
+      }
 
-  if (listen_fd_ < 0) {
-    NETWORK_LOG_ERROR("Failed to open socket: {}", strerror(errno));
-    throw NETWORK_PROCESS_EXCEPTION("Failed to open socket.");
+      sun.sun_family = AF_UNIX;
+      std::memcpy(sun.sun_path, socket_path.c_str(), sizeof(sun.sun_path));
```
lmwnshn, line 77: I think you want:
```
      const char *socket_path_cstr = socket_path.c_str();
      auto min_len = std::min(sizeof(sun.sun_path), strlen(socket_path.c_str()));
      std::memcpy(sun.sun_path, socket_path_cstr, min_len);
```

lmwnshn (changes requested):
Found a bug while merging it into my own branch.

src/network/terrier_server.cpp
gonzalezjo: I'm not sure why this would be a bug.

src/include/network/terrier_server.h
```diff
@@ -25,6 +26,9 @@
 
 namespace terrier::network {
 
+// The name is based on https://www.postgresql.org/docs/9.3/runtime-config-connection.html
+constexpr std::string_view UNIX_DOMAIN_SOCKET_FORMAT_STRING = "{0}/.s.PGSQL.{1}";
```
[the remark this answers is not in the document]
gonzalezjo: Will do.

src/network/connection_dispatcher_task.cpp
```diff
@@ -10,19 +10,25 @@
...
       num_handlers_(num_handlers),
       dedicated_thread_owner_(dedicated_thread_owner),
       connection_handle_factory_(connection_handle_factory),
       thread_registry_(thread_registry),
       interpreter_provider_(interpreter_provider),
       next_handler_(0) {
-  RegisterEvent(listen_fd, EV_READ | EV_PERSIST, METHOD_AS_CALLBACK(ConnectionDispatcherTask, DispatchConnection),
-                this);
+  for (int listen_fd : file_descriptors) {
+    if (listen_fd >= 0) {
```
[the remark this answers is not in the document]
gonzalezjo: That's intentional. Do you want me to change this? If I do, then it'd make sense to change how we handle invalid file descriptors elsewhere. 

We already warn (and provide some nice information) whenever we fail to get a file descriptor. Whenever that happens, though, we handle it gracefully: it's possible that we failed to get a Unix domain socket but still get a network socket, or vice versa. In those cases, we keep the server up. This is vaguely similar to what Postgres does, where it may keep listening to a network socket even if it fails to get a Unix domain socket.

src/network/terrier_server.cpp
```diff
@@ -39,44 +42,109 @@ TerrierServer::TerrierServer(common::ManagedPointer<ProtocolInterpreter::Provide
...
+  constexpr auto conn_backlog = common::Settings::CONNECTION_BACKLOG;
+  constexpr auto is_networked_socket = type == NETWORKED_SOCKET;
+  constexpr auto socket_description = std::string_view(is_networked_socket ? "networked" : "Unix domain");
+
+  auto &socket_fd = is_networked_socket ? network_socket_fd_ : unix_domain_socket_fd_;
+
+  // Gets the appropriate sockaddr for the given SocketType. Abuse a lambda and auto to specialize the type.
+  auto socket_addr = ([&] {
+    if constexpr (is_networked_socket) {  // NOLINT
+      struct sockaddr_in sin = {0};
```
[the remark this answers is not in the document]
gonzalezjo: They're the exact same. Do we have a preference for style?

src/network/terrier_server.cpp
```diff
@@ -39,44 +42,109 @@ TerrierServer::TerrierServer(common::ManagedPointer<ProtocolInterpreter::Provide
...
+  int status = bind(socket_fd, reinterpret_cast<struct sockaddr *>(&socket_addr), sizeof(socket_addr));
+  if (status < 0) {
+    NETWORK_LOG_ERROR("Failed to bind {} socket: {}", socket_description, strerror(errno), errno);
+
+    // We can recover from exactly one type of error here, contingent on this being a Unix domain socket.
+    if constexpr (!is_networked_socket) {  // NOLINT
+      auto recovered = false;
+
+      if (errno == EADDRINUSE) {
+        // I find this disgusting, but it's the approach favored by a bunch of software that uses Unix domain sockets.
```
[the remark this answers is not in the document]
gonzalezjo: Yep. Good idea.

src/network/terrier_server.cpp
```diff
@@ -39,44 +42,109 @@ TerrierServer::TerrierServer(common::ManagedPointer<ProtocolInterpreter::Provide
...
+        // BSD syslogd, for example, does this in *every* case--error handling or not--and I'm not one to question it.
+        recovered = !std::remove(fmt::format(UNIX_DOMAIN_SOCKET_FORMAT_STRING, socket_directory_, port_).c_str()) &&
+                    bind(socket_fd, reinterpret_cast<struct sockaddr *>(&socket_addr), sizeof(socket_addr)) >= 0;
+      }
+
+      if (recovered) {
+        NETWORK_LOG_INFO("Recovered! Managed to bind {} socket by purging a pre-existing bind.", socket_description)
```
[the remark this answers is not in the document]
gonzalezjo: Good catch.

src/network/terrier_server.cpp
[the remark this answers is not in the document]
lmwnshn: Resolved as "yup, exactly the same".

src/network/connection_dispatcher_task.cpp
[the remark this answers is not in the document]
lmwnshn: Hm. I guess I am OK with this for now (to limit the PR's scope, and because errors are unlikely to happen in practice). I am in favor of the caller swallowing the exception and doing nothing, since there are many fd's getting passed in and I am uneasy about the semantics of "some items in the list will work and some won't".

lmwnshn (approved):
LGTM.

PR #1111, opened 2020-08-24
Common Table Expressions (Simple, recursive and iterative)

tanujnay112:
This PR brings in Common Table Expressions (simple, recursive and iterative) along with a few other minor fixes to other parts of the system:

1) Adds functionality for nested pipelines
2) Adds the capability of having duplicate aliases by making a new AliasType that is a class with essentially a string and a serial number.
3) Made sure we order output columns from select * queries like Postgres does.

44k of the diff lines come from editing the generated pg_parse code to add support for the ITERATIVE keyword.


src/execution/compiler/compilation_context.cpp
```diff
@@ -1,5 +1,7 @@
 #include "execution/compiler/compilation_context.h"
 
+#include <execution/compiler/operator/ind_cte_scan_leader_translator.h>
```
lmwnshn, line 3: `""` not `<>`

sample_tpl/cte_scan_temp_table_insert.tpl
```diff
@@ -0,0 +1,56 @@
+// INSERT INTO temp_table SELECT colA FROM test_1 WHERE colA BETWEEN 495 AND 505
```
lmwnshn, line 1: Nit: follow the formatting of the other .tpl files?

sample_tpl/ind_cte_scan_temp_table_insert.tpl
```diff
@@ -0,0 +1,62 @@
+// Test that Accumulate() returns false on empty
```
lmwnshn, line 1: Nit: follow the formatting of the other .tpl files?

sample_tpl/cte_scan_temp_table_insert.tpl
```diff
@@ -0,0 +1,56 @@
...
+// Select * from temp_table
+// Returns the number of tuples inserted (11)
+fun main(execCtx: *ExecutionContext) -> int32 {
+  // Init inserter
+  var col_oids: [1]uint32
+  col_oids[0] = 1 // colA
```
lmwnshn, line 7: Can you use the testCatalogLookup builtin? The intent is to prevent strange .tpl file breakage when test tables are edited.

sample_tpl/cte_scan_temp_table_insert.tpl
```diff
@@ -0,0 +1,56 @@
...
+
+  var col_types: [1]uint32
+  col_types[0] = 4 // colA
```
lmwnshn, line 10: Can you use the testCatalogLookup builtin? The intent is to prevent strange .tpl file breakage when test tables are edited.

sample_tpl/cte_scan_temp_table_insert.tpl
```diff
@@ -0,0 +1,56 @@
...
+  var temp_col_oids: [1]uint32
+  temp_col_oids[0] = 2147483649 // colA
```
lmwnshn, line 12: Wait what's this?

sample_tpl/ind_cte_scan_temp_table_insert.tpl
```diff
@@ -0,0 +1,62 @@
...
+// Returns the number of tuples returned (should be 0)
+
+fun main(exec_ctx: *ExecutionContext) -> int32 {
+  // Initialize CTE Scan Iterator
+
+  var col_oids: [1]uint32
+  col_oids[0] = 1
```
lmwnshn, line 8: Can you use the testCatalogLookup builtin? The intent is to prevent strange .tpl file breakage when test tables are edited.

sample_tpl/ind_cte_scan_temp_table_insert.tpl
```diff
@@ -0,0 +1,62 @@
...
+  var col_types: [1]uint32
+  col_types[0] = 4
```
lmwnshn, line 10: Can you use the testCatalogLookup builtin? The intent is to prevent strange .tpl file breakage when test tables are edited.

sample_tpl/ind_cte_scan_temp_table_insert.tpl
```diff
@@ -0,0 +1,62 @@
...
+
+  var temp_col_oids: [1]uint32
+  temp_col_oids[0] = -2147483647 // colA
```
lmwnshn, line 13: What's this?

sample_tpl/ind_cte_scan_temp_table_insert.tpl
```diff
@@ -0,0 +1,62 @@
...
+
+  var cte_scan: IndCteScanIterator
+  @indCteScanInit(&cte_scan, exec_ctx, -2147483648, temp_col_oids, col_types, false)
```
lmwnshn, line 16: Magic number?

sample_tpl/cte_scan_temp_table_insert.tpl
```diff
@@ -0,0 +1,56 @@
...
+  var cte_scan_iterator: CteScanIterator
+  @cteScanInit(&cte_scan_iterator, execCtx, -2147483648, temp_col_oids, col_types)
```
lmwnshn, line 14: Magic number?

script/testing/junit/traces/cte.sql
```diff
@@ -0,0 +1,36 @@
...
+WITH cte(y) AS (SELECT 1 AS "x") SELECT y FROM cte;
+WITH cte(x) AS (WITH cte2(y) AS (SELECT 1) SELECT y FROM cte2) SELECT x FROM cte;
+WITH cte(x) AS (SELECT 1), cte2(y) AS (SELECT 2) SELECT * FROM cte INNER JOIN cte2 ON cte.x+1 = cte2.y;
+WITH cte(x) AS (SELECT 1), cte2(y) AS (SELECT 2) SELECT cte.x, cte2.y FROM cte INNER JOIN cte2 ON cte.x+1 = cte2.y;
+
+CREATE TABLE tree (node INTEGER, parent INTEGER);
+INSERT INTO tree VALUES (1,NULL), (10, 1), (11, 1), (100, 10), (101, 10), (110, 11), (111, 11);
+WITH RECURSIVE cte(x) AS (SELECT 1 UNION ALL SELECT tree.node FROM tree INNER JOIN cte ON tree.parent=cte.x) SELECT * FROM cte;
+
+with cte(x,x) as (select 1, 2) select * from cte;
```
lmwnshn, line 13: Nit: would be nice to have consistent capitalization in this file (eg all SQL keywords uppercase)

src/binder/bind_node_visitor.cpp
```diff
@@ -454,6 +455,75 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::SelectStatement> node
   BinderContext context(context_);
   context_ = common::ManagedPointer(&context);
 
+  for (auto &ref : node->GetSelectWith()) {
+    // Store CTE table name
+    cte_table_name_.push_back(ref->GetAlias());
+
+    bool inductive =
```
lmwnshn, line 462: Nit: short comment on what inductive means? Or a general high-level explanation of what this block of code does here.

lmwnshn: In general it is not obvious to me what the new code is doing and how to extend it if necessary, I think an overview comment would be very helpful here.

src/binder/bind_node_visitor.cpp
```diff
@@ -753,10 +843,13 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::TableRef> node) {
     // Save the previous context
     auto pre_context = context_;
     node->GetSelect()->Accept(common::ManagedPointer(this).CastManagedPointerTo<SqlNodeVisitor>());
+
+    //    TERRIER_ASSERT(num_aliases == num_columns, "Not enough aliases for all columns");
```
lmwnshn, line 847: Dead code.

src/binder/binder_context.cpp
```diff
@@ -77,40 +80,98 @@ void BinderContext::AddNewTable(const std::string &new_table_name,
...
   nested_table_alias_map_[table_alias] = column_alias_map;
 }
 
+void BinderContext::AddCTETable(const std::string &table_name,
+                                const std::vector<common::ManagedPointer<parser::AbstractExpression>> &select_list,
+                                const std::vector<parser::AliasType> &col_aliases) {
+  if (nested_table_alias_map_.find(table_name) != nested_table_alias_map_.end()) {
+    throw BINDER_EXCEPTION("Duplicate cte table definition", common::ErrorCode::ERRCODE_DUPLICATE_TABLE);
+  }
+  //  std::vector<catalog::Schema::Column> schema_columns;
```
lmwnshn, line 129: Dead code.

src/binder/binder_context.cpp
```diff
@@ -77,40 +80,98 @@ void BinderContext::AddNewTable(const std::string &new_table_name,
...
+  std::unordered_map<parser::AliasType, type::TypeId, parser::AliasType::HashKey> nested_column_mappings;
+  for (size_t i = 0; i < col_aliases.size(); i++) {
+    //    catalog::Schema::Column col(col_aliases[i].GetName(), select_list[i]->GetReturnValueType(), false,
```
lmwnshn, line 132: Dead code.

src/binder/binder_context.cpp
```diff
@@ -77,40 +80,98 @@ void BinderContext::AddNewTable(const std::string &new_table_name,
...
+    //                                parser::ConstantValueExpression(select_list[i]->GetReturnValueType()),
+    //                                TEMP_OID(catalog::col_oid_t, i));
+    //    schema_columns.push_back(col);
+    TERRIER_ASSERT(select_list[i]->GetReturnValueType() != type::TypeId::INVALID, "CTE column type not resolved");
+    nested_column_mappings[col_aliases[i]] = select_list[i]->GetReturnValueType();
+  }
+
+  //  catalog::Schema cte_schema(schema_columns);
```
lmwnshn, line 140: Dead code.

src/brain/operating_unit_recorder.cpp
```diff
@@ -131,7 +132,8 @@ size_t OperatingUnitRecorder::ComputeKeySize(catalog::table_oid_t tbl_oid, const
     AdjustKeyWithType(col.Type(), &key_size, num_key);
   }
 
-  TERRIER_ASSERT(key_size > 0, "KeySize must be greater than 0");
+  // TODO(preetang, rohanaff, gautamj, wz2): The following assert is not true
+  // TERRIER_ASSERT(key_size > 0, "KeySize must be greater than 0");
```
lmwnshn, line 136: What happened here?

src/catalog/schema.cpp
```diff
@@ -43,6 +44,25 @@ std::unique_ptr<Schema> Schema::DeserializeSchema(const nlohmann::json &j) {
   return std::make_unique<Schema>(columns);
 }
 
+// Schema::Schema(std::vector<std::string> column_aliases, std::vector<type::TypeId> column_types) {
```
lmwnshn, line 47: Dead code.

src/execution/compiler/compilation_context.cpp
```diff
@@ -290,7 +310,6 @@ void CompilationContext::Prepare(const planner::AbstractPlanNode &plan, Pipeline
                                                   planner::PlanNodeTypeToString(plan.GetPlanNodeType())));
     }
   }
-
```
lmwnshn, line 293: Unnecessary change.

src/execution/compiler/compiler.cpp
```diff
@@ -79,6 +79,7 @@ void Compiler::Run(Compiler::Callbacks *callbacks) {
   }
 
   callbacks->EndPhase(Phase::SemanticAnalysis, this);
+  //  ast::AstPrettyPrint::Dump(std::cout, GetAST());
```
lmwnshn, line 82: Dead code.

src/execution/compiler/operator/cte_scan_leader_translator.cpp
```diff
@@ -0,0 +1,180 @@
+#include "execution/compiler/operator/cte_scan_leader_translator.h"
+#include "execution/compiler/codegen.h"
+#include "execution/compiler/compilation_context.h"
+#include "execution/compiler/function_builder.h"
+#include "execution/compiler/work_context.h"
+#include "parser/expression/constant_value_expression.h"
+//
```
lmwnshn, line 7: Dead code.

src/execution/compiler/operator/cte_scan_leader_translator.cpp
```diff
@@ -0,0 +1,180 @@
...
+namespace terrier::execution::compiler {
+
+//
```
lmwnshn, line 10: Dead code.

src/execution/compiler/operator/cte_scan_leader_translator.cpp
```diff
@@ -0,0 +1,180 @@
...
+
+  std::vector<uint16_t> attr_sizes;
+  for (const auto &column : plan.GetTableSchema()->GetColumns()) {
+    attr_sizes.push_back(column.AttrSize());
+  }
+
+  auto offsets = storage::StorageUtil::ComputeBaseAttributeOffsets(attr_sizes, 0);
+
+  // Build the map from Schema columns to underlying columns
+  for (const auto &column : plan.GetTableSchema()->GetColumns()) {
```
lmwnshn, line 36: This code looks quite familiar and I feel it exists somewhere else, but maybe it doesn't.

src/execution/compiler/operator/cte_scan_translator.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+parser::ConstantValueExpression DummyCVE() {
+  return terrier::parser::ConstantValueExpression(type::TypeId::INTEGER, execution::sql::Integer(0));
+}
+
+CteScanTranslator::CteScanTranslator(const planner::CteScanPlanNode &plan, CompilationContext *compilation_context,
+                                     Pipeline *pipeline)
+    : SeqScanTranslator(plan, compilation_context, pipeline) {
+  if (plan.GetChildrenSize() > 0) {
+    compilation_context->Prepare(*(plan.GetChild(0)), pipeline);
```
lmwnshn, line 22: Assert that there is either (1) zero children or (2) only one child?

src/execution/compiler/operator/seq_scan_translator.cpp
```diff
@@ -1,12 +1,16 @@
 #include "execution/compiler/operator/seq_scan_translator.h"
 
+#include <execution/compiler/operator/cte_scan_leader_translator.h>
```
lmwnshn, line 3: `""` not `<>`, next line too

src/include/catalog/catalog_defs.h
```diff
@@ -39,6 +39,11 @@ constexpr trigger_oid_t INVALID_TRIGGER_OID = trigger_oid_t(NULL_OID);
 constexpr type_oid_t INVALID_TYPE_OID = type_oid_t(NULL_OID);
 constexpr view_oid_t INVALID_VIEW_OID = view_oid_t(NULL_OID);
 
+constexpr uint32_t TEMP_OID_MASK = 1 << 31;
+#define TEMP_OID(strongtype, x) (strongtype(x | catalog::TEMP_OID_MASK))
```
lmwnshn, line 43: I think we strongly prefer functions over macros, unless there is a good reason for doing this.

src/include/catalog/schema.h
```diff
@@ -242,6 +263,8 @@ class Schema {
     }
   }
 
+  //  explicit Schema(std::vector<std::string> column_aliases, std::vector<type::TypeId> column_types);
```
lmwnshn, line 266: Dead code.

src/include/catalog/catalog_defs.h
soumilm: As I think was discussed with Tanuj, this will be staying as a macro for the time being

lmwnshn:
@soumilm What's the status of this PR? Is it ready for me to take a pass on it?

src/CMakeLists.txt
```diff
@@ -75,6 +75,6 @@ ADD_TERRIER_LIB(terrier
 ###############################################
 # Terrier executable
 ###############################################
-add_executable(terrier main/terrier.cpp)
+add_executable(terrier main/terrier.cpp execution/ast/ast_pretty_print.cpp include/execution/ast/ast_pretty_print.h include/execution/compiler/operator/cte_scan_translator.h include/planner/plannodes/cte_scan_plan_node.h execution/compiler/operator/cte_scan_translator.cpp planner/plannodes/cte_scan_plan_node.cpp include/execution/sql/cte_scan_iterator.h execution/sql/cte_scan_iterator.cpp)
```
mbutrovich, line 78: Guessing this is a CLion goof.

src/catalog/schema.cpp
```diff
@@ -1,6 +1,7 @@
 #include "catalog/schema.h"
 
 #include "common/json.h"
+#include "parser/expression/constant_value_expression.h"
```
mbutrovich, line 4: Stray include.

src/include/catalog/schema.h
```diff
@@ -79,6 +79,27 @@ class Schema {
...
+     */
+    Column(std::string name, const type::TypeId type, const bool nullable,
+           const parser::AbstractExpression &default_value, const col_oid_t oid)
+        : name_(std::move(name)),
+          type_(type),
+          attr_size_(type::TypeUtil::GetTypeSize(type_)),
+          nullable_(nullable),
+          oid_(oid),
+          default_value_(default_value.Copy()) {
+      // TODO(Rohan,Gautam,Preetansh): We need to manually set this as we cannot get this from Output
```
mbutrovich, line 98: What is "this" in this comment?

src/include/catalog/catalog_defs.h
mbutrovich: Can you document this conversation? Because my immediate response was scrolling by this was "Get rid of that macro."

mbutrovich:
Just skimming and a couple questions re: blast radius outside of CTEs that should be documented for other people.
- There's a map of temp tables in the `CatalogAccessor` now. Who owns these? What's their lifecycle? Why isn't ownership given to the catalog, even if it's taken away at the end by dropping the table at the end of the query?

src/parser/postgresparser.cpp
```diff
@@ -2026,4 +2034,71 @@ std::unique_ptr<VariableSetStatement> PostgresParser::VariableSetTransform(Parse
...
+            colnames.emplace_back(parser::AliasType(column, i));
+            i++;
+          }
+        }
+        //        if (colnames.empty()) {
+        //          for (auto &expr : select->GetSelectColumns()) {
+        //            // TODO(tanujnay112) route the "?column?" codepath through traffic cop
+        //            colnames.push_back(expr->GetAlias().empty() ? "?column?" : expr->GetAlias());
+        //          }
+        //        }
```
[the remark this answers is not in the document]
soumilm: CTEs are a little complicated in how column names get picked -- if you consider the query

`WITH xxx(a) AS (SELECT 1, b as c, d, 2) ...`

The table `xxx` will have four columns: 

1. The first column name is `a`, since this was specified with the CTE name in `WITH xxx(a)`
2. The second column name is `c`, since no name was specified for the second column, and the alias for the second column is `c`
3. The third column name is `d`, since neither of the above two items were specified, so the column name itself is used
4. The fourth column name is `?column?`, since there is no column name

We do all of this casing in the binder, once column names/`*`'s have been bound to specific columns, so we handle the fourth case (where none of the other three options provide a column name to use) in the binder as well.

src/binder/bind_node_visitor.cpp
```diff
@@ -753,10 +855,12 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::TableRef> node) {
     // Save the previous context
     auto pre_context = context_;
     node->GetSelect()->Accept(common::ManagedPointer(this).CastManagedPointerTo<SqlNodeVisitor>());
+
     // TODO(WAN): who exactly should save and restore contexts? Restore the previous level context
     context_ = pre_context;
-    // Add the table to the current context at the end
-    context_->AddNestedTable(node->GetAlias(), node->GetSelect()->GetSelectColumns());
+    if (node->GetCteType() == parser::CTEType::INVALID) {
```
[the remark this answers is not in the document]
soumilm: This means "not a CTE" -- so this loop is casing on whether the TableRef argument is a CTE table or some other table,

preetansh (approved):

src/catalog/catalog_accessor.cpp
```diff
@@ -217,4 +220,8 @@ common::ManagedPointer<storage::BlockStore> CatalogAccessor::GetBlockStore() con
   return catalog_->GetBlockStore();
 }
 
+void CatalogAccessor::RegisterTempTable(table_oid_t table_oid, const common::ManagedPointer<storage::SqlTable> table) {
+  temp_tables_[table_oid] = table;
```
mush-zhang, line 224: Will this temp_table grow arbitrarily large?

tanujnay112: No its scope is limited to the scope of the catalogaccessor

mbutrovich:
Just a reminder that I have unaddressed questions above, since this PR was just described as "more or less ready to merge" in the meeting.

lmwnshn:
It would also be helpful to provide a small patch for what exactly changed in libpg_query as we may want to upgrade in the future (if merged as is, we probably either do not upgrade libpg_query or we give up on CTEs when we upgrade).

tanujnay112:
> Just a reminder that I have unaddressed questions above, since this PR was just described as "more or less ready to merge" in the meeting.



> Just skimming and a couple questions re: blast radius outside of CTEs that should be documented for other people.
> 
> * There's a map of temp tables in the `CatalogAccessor` now. Who owns these temp tables? What's their lifecycle? How and when are they freed? Why isn't ownership given to the catalog, even if it's taken away at the end by dropping the table at the end of the query?
@mbutrovich 
Sorry I thought you were asking to add documentation to answer such questions. But to reiterate here, this temp table is mainly used for interaction with temporary tables in a query, more specifically in CTE's. The CTE execution operators will handle the full lifecycle of the SQLTable objects. More specifically, the CTEScanIterator family deals with this lifecycle. The point of making these temporary tables not owned by the catalog was to avoid the catalog processing overhead (putting things in indexes and such) on these temp tables that are really only relevant for the lifetime of one query.

src/binder/binder_context.cpp
```diff
@@ -156,14 +225,18 @@ bool BinderContext::SetColumnPosTuple(common::ManagedPointer<parser::ColumnValue
     }
     // Check nested table
     for (auto &entry : current_context->nested_table_alias_map_) {
-      bool get_match = entry.second.find(col_name) != entry.second.end();
+      auto iter = entry.second.find(alias_name);
+      bool get_match = iter != entry.second.end();
+      auto matches = std::count_if(entry.second.begin(), entry.second.end(),
+                                   [=](auto it) { return entry.second.key_eq()(it.first, alias_name); });
```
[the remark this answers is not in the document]
tanujnay112: The point of matches is to make sure there's a unique match. If there are two or more matches, then the requested name is deemed ambiguous. I'll document this.

mush-zhang (approved):
LGTM

script/testing/junit/run_junit.py
```diff
@@ -30,6 +30,7 @@ def section_header(title):
     # Step 1: Run the regular JUnit tests.
     LOG.info(section_header("JUNIT TESTS"))
     test_command_regular = constants.JUNIT_TEST_CMD_JUNIT
+    print(test_command_regular)
```
mbutrovich, line 33: Do we want to keep this change or was it just for debugging?

script/testing/junit/traces/cte.sql
```diff
@@ -0,0 +1,36 @@
+WITH cte(x) AS (SELECT 1) SELECT x FROM cte;
```
mbutrovich, line 1: What's the source on these? Are some sqlite test cases and some our own?

src/binder/binder_context.cpp
```diff
@@ -5,13 +5,15 @@
 #include <string>
 #include <tuple>
 #include <unordered_map>
+#include <unordered_set>
```
mbutrovich, line 8: Looks like an unnecessary include.

src/binder/binder_context.cpp
```diff
@@ -134,8 +185,26 @@ void BinderContext::SetColumnPosTuple(const std::string &col_name,
   expr->SetReturnValueType(column_object.Type());
 }
 
+void BinderContext::SetTableName(common::ManagedPointer<parser::ColumnValueExpression> expr,
+                                 common::ManagedPointer<parser::SelectStatement> node) {
+  if (node->GetSelectTable() != nullptr) {
```
mbutrovich, line 190: What if it is `nullptr`? Can that happen? Is it a failure scenario or a no-op? If `nullptr` means binding failure, we should throw. If `nullptr` can't happen, we should have an assertion. If the null check is just forcing a no-op, we should document the scenario when that happens.

src/catalog/catalog_accessor.cpp
```diff
@@ -79,6 +79,9 @@ bool CatalogAccessor::SetTablePointer(table_oid_t table, storage::SqlTable *tabl
 }
 
 common::ManagedPointer<storage::SqlTable> CatalogAccessor::GetTable(table_oid_t table) const {
+  if (UNLIKELY(catalog::IsTempOid(table))) {
+    return temp_tables_.find(table)->second;
```
mbutrovich, line 83: What happens if this lookup fails? If it can't we should store the result in an `UNUSED_ATTRIBUTE` local variable and then do an assert that it's not `end()`. All assumptions about code behavior should be asserted.

src/catalog/catalog_accessor.cpp
mbutrovich, line 224: Similarly, there should be an assertion here that the oid is not already in the map, if that assumption is being made. Assert any assumptions about behavior so that when the next person goes to violate that assumption, they get an error.

src/include/binder/bind_node_visitor.h
```diff
@@ -102,6 +102,7 @@ class BindNodeVisitor final : public SqlNodeVisitor {
   /** Catalog accessor */
   const common::ManagedPointer<catalog::CatalogAccessor> catalog_accessor_;
   const catalog::db_oid_t db_oid_;
+  std::vector<std::string> cte_table_name_;
```
mbutrovich, line 105: Is this state that should be in the `BinderSherpa` instead?

lmwnshn:
In the middle of reviewing when I saw something in the code that I wanted to test, should this be supported?

```
noisepage=# create table foo (a int);
CREATE TABLE
noisepage=# insert into foo values (999);
INSERT 0 1
noisepage=# with foo as (select 1) select * from foo;
ERROR:  Duplicate alias "foo"

(postgres returns 1)
```

src/include/catalog/catalog_accessor.h
```diff
@@ -380,6 +381,19 @@ class EXPORT CatalogAccessor {
...
 
+  /**
+   * Registers a temporary table in this accessor
+   * @param table_oid The temporary oid of this table
+   * @param table The temp table being registered
+   */
+  void RegisterTempTable(table_oid_t table_oid, common::ManagedPointer<storage::SqlTable> table);
+
+  /**
+   * Allocates and returns a new temporary oid. These oids are only valid for the lifetime of this accessor
```
mbutrovich, line 392: How does this work with caching? What happens if another query comes along and tries to use the cached plans but for some reason the oids don't match? Is that possible? Are there other implications for query caching with this design?

src/include/catalog/schema.h
```diff
@@ -79,6 +79,27 @@ class Schema {
...
+     */
+    Column(std::string name, const type::TypeId type, const bool nullable,
+           const parser::AbstractExpression &default_value, const col_oid_t oid)
+        : name_(std::move(name)),
+          type_(type),
+          attr_size_(type::TypeUtil::GetTypeSize(type_)),
+          nullable_(nullable),
+          oid_(oid),
+          default_value_(default_value.Copy()) {
+      // TODO(Rohan,Gautam,Preetansh): We need to manually set the max varlen size here
```
mbutrovich, line 98: comment says it's for non-VARLEN, assert that.

src/include/parser/table_ref.h
```diff
@@ -2,6 +2,7 @@
 
 #include <memory>
 #include <string>
+#include <unordered_set>
```
mbutrovich, line 5: Is this a stray include?

src/include/storage/sql_table.h
```diff
@@ -231,6 +231,17 @@ class SqlTable {
    */
   ProjectionMap ProjectionMapForOids(const std::vector<catalog::col_oid_t> &col_oids);
 
+  /**
+   * Clears the contents of this table and reinitializes it
+   */
+  void Reset();
```
mbutrovich, line 237: Who is it used by, what purpose does it serve, and when is it not safe to use it? This seems like a blunt instrument ripe for abuse in the future that'll cause races between the garbage collector and the table.

src/include/storage/sql_table.h
```diff
@@ -231,6 +231,17 @@ class SqlTable {
...
+
+  /**
+   * Copies all data from src to this table
+   * @param txn The current transaction context
+   * @param src The source table to copy from
+   */
+  void CopyTable(common::ManagedPointer<transaction::TransactionContext> txn, common::ManagedPointer<SqlTable> src);
```
mbutrovich, line 244: Same question.

src/storage/sql_table.cpp
```diff
@@ -86,6 +66,57 @@ ProjectionMap SqlTable::ProjectionMapForOids(const std::vector<catalog::col_oid_
...
+  while (it != src->end()) {
+    const TupleSlot slot = *it;
+    // Only fill the buffer with valid, visible tuples
+    if (!src->Select(txn, slot, projected_row)) {
+      it++;
+      continue;
+    }
+
+    // TODO(tanujnay112) I don't like how I have to hardcode this
+    auto *redo = txn->StageWrite(catalog::db_oid_t(999), catalog::table_oid_t(999), pr_init);
```
mbutrovich, line 91: What are these magic values?

src/storage/sql_table.cpp
```diff
@@ -86,6 +66,57 @@ ProjectionMap SqlTable::ProjectionMapForOids(const std::vector<catalog::col_oid_
...
+void SqlTable::CopyTable(const common::ManagedPointer<transaction::TransactionContext> txn,
+                         const common::ManagedPointer<SqlTable> src) {
+  //  uint32_t filled = 0;
+  auto it = src->begin();
+  std::vector<catalog::col_oid_t> col_oids;
+  for (auto &cols : table_.column_map_) {
+    col_oids.push_back(cols.first);
+  }
+  auto pr_init = InitializerForProjectedRow(col_oids);
+  void *buffer = alloca(pr_init.ProjectedRowSize());
```
mbutrovich, line 80: Do we use this style of allocation anywhere else in the system?

mbutrovich (changes requested):
First pass.

src/execution/sql/cte_scan_iterator.cpp
```diff
@@ -0,0 +1,71 @@
...
+    all_columns.push_back(col);
+    col_oids_.emplace_back(catalog::col_oid_t(schema_cols_ids[i]));
+  }
+
+  // Create the table in the catalog.
+  catalog::Schema cte_table_schema(all_columns);
+  cte_table_ = new storage::SqlTable(exec_ctx->GetAccessor()->GetBlockStore(), cte_table_schema);
+  exec_ctx->GetAccessor()->RegisterTempTable(table_oid, common::ManagedPointer(cte_table_));
+
+  auto cte_table_local = cte_table_;
```
lmwnshn, line 29: auto * so that it is clearer no crazy copy is being done

lmwnshn: or better, storage::SqlTable *

src/execution/sql/cte_scan_iterator.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+  // We are deferring it in both commit and abort because we need to delete the temp table regardless of transaction
+  // outcome. We use deferred actions to guarantee memory safety with the garbage collector.
+  // TODO(Rohan): explore API change to deferred actions for unconditional actions to avoid commit and abort actions
```
lmwnshn, line 33: This seems like a wider discussion, can you open an issue?

src/execution/sql/cte_scan_iterator.cpp
```diff
@@ -0,0 +1,71 @@
...
+storage::ProjectedRow *CteScanIterator::GetInsertTempTablePR() {
+  // We need all the columns
+  storage::ProjectedRowInitializer pri = cte_table_->InitializerForProjectedRow(col_oids_);
+  auto txn = exec_ctx_->GetTxn();
+  table_redo_ = txn->StageWrite(exec_ctx_->DBOid(), cte_table_oid_, pri);
+  return table_redo_->Delta();
+}
+
+storage::TupleSlot CteScanIterator::TableInsert() {
+  exec_ctx_->RowsAffected()++;  // believe this should only happen in root plan nodes, so should reflect count of query
```
lmwnshn, line 64: We don't do this any more. See the execCtxAddRowsAffected or whatever that was

src/execution/sql/ind_cte_scan_iterator.cpp
```diff
@@ -0,0 +1,83 @@
+
```
lmwnshn, line 1: delete newline

src/execution/sql/ind_cte_scan_iterator.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "transaction/deferred_action_manager.h"
+#include "transaction/transaction_context.h"
+
+namespace noisepage::execution::sql {
+
+IndCteScanIterator::IndCteScanIterator(exec::ExecutionContext *exec_ctx, catalog::table_oid_t table_oid,
+                                       uint32_t *schema_cols_ids, uint32_t *schema_cols_type, uint32_t num_schema_cols,
+                                       bool is_recursive)
+    : exec_ctx_{exec_ctx},
+      cte_scan_1_{
```
lmwnshn, line 16: see comment in header about cte_scan_1, _2, _3, purpose unclear

src/include/execution/compiler/pipeline.h
```diff
@@ -175,6 +181,16 @@ class Pipeline {
    */
   std::string CreatePipelineFunctionName(const std::string &func_name) const;
 
+  /**
+   * @return A vector of expressions that initialize, run and teardown a nested pipeline
+   */
+  std::vector<ast::Expr *> CallSingleRunPipelineFunction() const;
```
lmwnshn, line 187: Suggestion: Generate instead of Call, or in front of Call

src/include/execution/compiler/pipeline.h
```diff
@@ -175,6 +181,16 @@ class Pipeline {
...
+
+  /**
+   * @return A vector of expressions that do the work of running a pipeline function and its dependencies
+   */
+  std::vector<ast::Expr *> CallRunPipelineFunction() const;
```
lmwnshn, line 192: Suggestion: Generate instead of Call, or in front of Call

src/include/execution/compiler/pipeline.h
```diff
@@ -241,31 +259,51 @@ class Pipeline {
...
+  StateDescriptor &GetPipelineStateDescriptor() { return nested_ ? parent_->state_ : state_; }
+
  private:
   // A unique pipeline ID.
   uint32_t id_;
   // The compilation context this pipeline is part of.
   CompilationContext *compilation_context_;
   // The code generation instance.
   CodeGen *codegen_;
+  // Cache of common identifiers.
```
lmwnshn, line 283: Maybe move these back to their original location?

src/include/execution/compiler/pipeline.h
```diff
@@ -241,31 +259,51 @@ class Pipeline {
...
+  ast::Identifier state_var_;
+  // The pipeline state.
+  StateDescriptor state_;
+  // The pipeline operating unit feature vector state.
+  StateDescriptor::Entry oufeatures_;
   // Operators making up the pipeline.
   std::vector<OperatorTranslator *> steps_;
   // The driver.
   PipelineDriver *driver_;
+  // pointer to parent pipeline (only applicable if this is a nested pipeline)
```
lmwnshn, line 293: See comment on nested_, I think this is quite potentially very misleading

src/include/execution/compiler/pipeline.h
```diff
@@ -241,31 +259,51 @@ class Pipeline {
   /** @return The vector of pipeline operators that make up the pipeline. */
   const std::vector<OperatorTranslator *> &GetTranslators() const { return steps_; }
 
+  void InjectStartPipelineTracker(FunctionBuilder *builder) const;
```
lmwnshn, line 262: Why are these functions now in pipeline.h?

src/include/execution/compiler/pipeline.h
```diff
@@ -241,31 +259,51 @@ class Pipeline {
...
+
+  void InjectEndResourceTracker(FunctionBuilder *builder, query_id_t query_id) const;
```
lmwnshn, line 264: Why are these functions now in pipeline.h?

src/include/execution/compiler/pipeline.h
```diff
@@ -241,31 +259,51 @@ class Pipeline {
...
+
+  ast::Identifier GetRunPipelineFunctionName() const;
```
lmwnshn, line 266: Move this with the other GetRun... functions?

src/include/execution/compiler/operator/cte_scan_leader_translator.h
```diff
@@ -0,0 +1,97 @@
+#pragma once
+
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+#include "execution/compiler/operator/cte_scan_provider.h"
+#include "execution/compiler/operator/operator_translator.h"
+#include "execution/compiler/pipeline.h"
+#include "planner/plannodes/cte_scan_plan_node.h"
```
lmwnshn, line 10: Can this be forward declared?

src/include/execution/compiler/operator/cte_scan_leader_translator.h
```diff
@@ -0,0 +1,97 @@
+#pragma once
+
+#include <string>
```
lmwnshn, line 3: unused?

src/include/execution/compiler/operator/cte_scan_leader_translator.h
```diff
@@ -0,0 +1,97 @@
...
+#include <unordered_map>
+#include <utility>
```
lmwnshn, line 5: unused?

src/include/execution/compiler/operator/cte_scan_leader_translator.h
```diff
@@ -0,0 +1,97 @@
...
+namespace noisepage::execution::compiler {
+
+/**
+ * CteScanLeader Translator
+ */
+class CteScanLeaderTranslator : public OperatorTranslator, CteScanProvider {
+ public:
+  /**
+   * Constructor
+   * @param plan The ctescanplanndoe this translator is generating code for
```
lmwnshn, line 21: typo

src/include/execution/compiler/operator/ind_cte_scan_leader_translator.h
```diff
@@ -0,0 +1,92 @@
+#pragma once
+
+#include <unordered_map>
```
lmwnshn, line 3: unused?

src/include/execution/compiler/operator/ind_cte_scan_leader_translator.h
```diff
@@ -0,0 +1,92 @@
...
+#include <utility>
```
lmwnshn, line 4: unused?

src/include/execution/compiler/operator/ind_cte_scan_leader_translator.h
```diff
@@ -0,0 +1,92 @@
...
+#include <vector>
+#include "execution/compiler/operator/cte_scan_provider.h"
+#include "execution/compiler/operator/operator_translator.h"
+#include "execution/compiler/pipeline.h"
+#include "planner/plannodes/cte_scan_plan_node.h"
```
lmwnshn, line 9: can forward declare?

src/include/execution/compiler/operator/ind_cte_scan_leader_translator.h
```diff
@@ -0,0 +1,92 @@
...
+namespace noisepage::execution::compiler {
+
+/**
+ * IndCteScanLeader Translator
+ */
+class IndCteScanLeaderTranslator : public OperatorTranslator, CteScanProvider {
+ public:
+  /**
+   * Constructor
+   * @param plan The ctescanplanndoe this translator is generating code for
```
lmwnshn, line 20: typo

src/optimizer/optimizer.cpp
```diff
@@ -113,6 +184,18 @@ std::unique_ptr<planner::AbstractPlanNode> Optimizer::ChooseBestPlan(
                                       std::move(children_plans), std::move(children_expr_map));
   OPTIMIZER_LOG_TRACE("Finish Choosing best plan for group " + std::to_string(id.UnderlyingValue()));
 
+  if (op->Contents()->GetOpType() == OpType::CTESCAN && !child_groups.empty()) {
+    NOISEPAGE_ASSERT(child_groups.size() <= 2, "CTE should not have more than 2 children.");
```
lmwnshn, line 188: what are valid numbers of children to have? 0? 1? 2? could you please document each case happens?

src/optimizer/optimizer.cpp
```diff
@@ -63,6 +72,68 @@ std::unique_ptr<planner::AbstractPlanNode> Optimizer::BuildPlanTree(transaction:
...
+      if (*leader != nullptr) {
+        std::vector<std::unique_ptr<planner::AbstractPlanNode>> adopted_children;
+        plan->MoveChildren(&adopted_children);
+        NOISEPAGE_ASSERT(adopted_children.size() == 1, "CTE leader should have 1 child");
+        (*leader)->AddChild(std::move(adopted_children[0]));
+        auto cte_scan = dynamic_cast<planner::CteScanPlanNode *>(plan.Get());
+        cte_scan->SetLeader(leader->CastManagedPointerTo<const planner::CteScanPlanNode>());
+      }
+    }
+
```
lmwnshn, line 96: a bit tough to follow the logic, can you add some comments?

src/optimizer/plan_generator.cpp
```diff
@@ -299,7 +300,8 @@ void PlanGenerator::Visit(const QueryDerivedScan *op) {
     auto colve = output.CastManagedPointerTo<parser::ColumnValueExpression>();
 
     // Get offset into child_expr_ma
```
lmwnshn, line 302: not your typo, but can you add the p back?

src/optimizer/plan_generator.cpp
```diff
@@ -1063,4 +1065,90 @@ void PlanGenerator::Visit(const Analyze *analyze) {
                      .Build();
 }
 
+///////////////////////////////////////////////////////////////////////////////
+// CTE
+///////////////////////////////////////////////////////////////////////////////
+
+void PlanGenerator::Visit(const CteScan *cte_scan) {
+  // CteScan has the same output schema as the child plan!
+  NOISEPAGE_ASSERT(children_plans_.size() <= 2, "CteScan needs at most 2 child plans");
```
lmwnshn, line 1074: similar to previous comment, when are 0/1/2 child plans OK?

src/optimizer/plan_generator.cpp
```diff
@@ -1063,4 +1065,90 @@ void PlanGenerator::Visit(const Analyze *analyze) {
...
+      auto tve = output_expr.CastManagedPointerTo<parser::ColumnValueExpression>();
+
+      // output schema of a plan node is literally the columns of the table.
+      // there is no such thing as an intermediate column here!
+      columns.emplace_back(tve->GetColumnName(), tve->GetReturnValueType(), tve->Copy());
+    }
+
+    auto cte_scan_out = std::make_unique<planner::OutputSchema>(std::move(child_columns));
+    (void)cte_scan_out;
+    if (children_plans_.size() == 2) {
```
lmwnshn, line 1110: Any asserts we can make here? What about 1 child?

src/optimizer/plan_generator.cpp
```diff
@@ -1063,4 +1065,90 @@ void PlanGenerator::Visit(const Analyze *analyze) {
...
+                         .SetScanPredicate(common::ManagedPointer<parser::AbstractExpression>(predicate))
+                         .Build();
+    }
+  } else {
+    // make schema from output columns
+    std::vector<planner::OutputSchema::Column> columns;
+    for (auto &output_expr : output_cols_) {
+      auto tve = output_expr.CastManagedPointerTo<parser::ColumnValueExpression>();
+
+      // output schema of a plan node is literally the columns of the table.
```
lmwnshn, line 1138: this seems slightly repetitive

src/optimizer/query_to_operator_transformer.cpp
```diff
@@ -50,6 +71,78 @@ void QueryToOperatorTransformer::Visit(common::ManagedPointer<parser::SelectStat
   predicates_ = {};
   transaction::TransactionContext *txn_context = accessor_->GetTxn().Get();
 
+  if (!op->GetSelectWith().empty()) {
```
lmwnshn, line 74: this is a pretty painful function to read, can you add some guide comments in front of each block of code?

src/optimizer/rules/rewrite_rules.cpp
```diff
@@ -449,6 +521,71 @@ void RewritePullFilterThroughMarkJoin::Transform(common::ManagedPointer<Abstract
...
+
+  auto derived_node = derived_query_get->Contents()->GetContentsAs<LogicalQueryDerivedGet>();
+  std::vector<std::unique_ptr<AbstractOptimizerNode>> derived_children;
+  derived_children.push_back(left_node->Copy());
+  auto new_derived =
+      std::make_unique<OperatorNode>(Operator(common::ManagedPointer<BaseOperatorNodeContents>(derived_node->Copy()))
+                                         .RegisterWithTxnContext(context->GetOptimizerContext()->GetTxn()),
+                                     std::move(derived_children), context->GetOptimizerContext()->GetTxn());
+
+  //  auto new_derived_query_get_contents =
```
lmwnshn, line 567: Dead code?

test/execution/ind_cte_scan_test.cpp
```diff
@@ -0,0 +1,381 @@
...
+};
+
+TEST_F(IndCteScanTest, IndCTEEmptyAccumulateTest) {
+  // Test that Accumulate() returns false on empty
+
+  // Create cte_table
+  uint32_t col_oids[1] = {exec_ctx_->GetAccessor()->GetNewTempOid()};
+  uint32_t cte_table_col_type[1] = {4};  // {INTEGER}
+
+  // auto cte_scan = new noisepage::execution::sql::IndCteScanIterator(exec_ctx_.get(), cte_table_col_type, 1);
```
lmwnshn, line 37: Dead code?

test/execution/ind_cte_scan_test.cpp
```diff
@@ -0,0 +1,381 @@
...
+                           index_oid.UnderlyingValue(),
+                           col_oids.data(),
+                           static_cast<uint32_t>(col_oids.size())};
+  index_iter.Init();
+
+  // Create cte_table
+  uint32_t cte_table_col_type[1] = {4};  // {INTEGER}
+  uint32_t cte_table_col_ids[1] = {exec_ctx_->GetAccessor()->GetNewTempOid()};
+
+  // auto cte_scan = new noisepage::execution::sql::IndCteScanIterator(exec_ctx_.get(), cte_table_col_type, 1);
```
lmwnshn, line 159: dead code

test/execution/ind_cte_scan_test.cpp
```diff
@@ -0,0 +1,381 @@
...
+                           table_oid.UnderlyingValue(),
+                           index_oid.UnderlyingValue(),
+                           col_oids.data(),
+                           static_cast<uint32_t>(col_oids.size())};
+  index_iter.Init();
+
+  // Create cte_table
+  uint32_t cte_table_col_type[1] = {4};  // {INTEGER}
+
+  // auto cte_scan = new noisepage::execution::sql::IndCteScanIterator(exec_ctx_.get(), cte_table_col_type, 1);
```
lmwnshn, line 231: dead code

test/execution/ind_cte_scan_test.cpp
```diff
@@ -0,0 +1,381 @@
...
```
lmwnshn, line 310: dead code

lmwnshn (changes requested):
Nice! A lot cleaner and less dead code than when I did a first pass some weeks back.

A couple of comments.

Also, are any of the tests better expressed as tracefile tests instead of C++ tests? I didn't look too closely at those.

src/include/brain/brain_defs.h
```diff
@@ -103,6 +103,7 @@ enum class ExecutionOperatingUnitType : uint32_t {
    * This gets dropped right now... :(
    */
   LIMIT,
+  CTE_SCAN,
```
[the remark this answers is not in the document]
lmwnshn: Maybe @17zhangw or @malin1993ml can weigh in here -- same as SEQ_SCAN?

src/brain/operating_unit_recorder.cpp
```diff
@@ -614,6 +615,19 @@ void OperatingUnitRecorder::Visit(const planner::LimitPlanNode *plan) {
   AggregateFeatures(plan_feature_type_, key_size, num_keys, plan, 1, 1);
 }
 
+void OperatingUnitRecorder::Visit(const planner::CteScanPlanNode *plan) {
+  VisitAbstractPlanNode(plan);
```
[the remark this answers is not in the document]
17zhangw: Looks good to me.

src/optimizer/input_column_deriver.cpp
```diff
@@ -96,6 +101,61 @@ void InputColumnDeriver::Visit(const Limit *op) {
...
+  for (size_t i = 0; i < gexpr_->GetChildrenGroupsSize(); i++) {
+    auto child_exprs = op->GetChildExpressions()[i];
+    std::vector<common::ManagedPointer<parser::AbstractExpression>> new_child_exprs;
+    new_child_exprs.reserve(child_exprs.size());
+    for (auto &elem : child_exprs) {
+      new_child_exprs.push_back(elem);
+    }
+    child_exprs.clear();
+    child_exprs = new_child_exprs;
+    child_cols.push_back(std::move(child_exprs));
```
17zhangw, line 133: Can there be a comment included here explaining why this logic is necessary/correct?

src/storage/sql_table.cpp
soumilm: We don't really use `alloca` in other places, but this is a pointer with a very short lifecycle that we would free almost immediately if we were to put it on the heap so we're just `alloca`ing it instead

src/parser/expression/abstract_expression.cpp
```diff
@@ -48,7 +48,7 @@ common::hash_t AbstractExpression::Hash() const {
 
 bool AbstractExpression::operator==(const AbstractExpression &rhs) const {
   if (expression_type_ != rhs.expression_type_) return false;
-  if (alias_ != rhs.alias_) return false;
+  if (!(alias_ == rhs.alias_)) return false;
```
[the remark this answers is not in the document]
soumilm: We refactored aliases at some point so they now store an id as well (to handle queries such as `SELECT 1 as x, 2 as x` and `WITH cte(x,x) AS (SELECT 1,2) SELECT * FROM cte`), so aliases have an `==` function but not a `!=` function.

src/optimizer/plan_generator.cpp
soumilm: We have an assert at the top of this function that `children_plans_` has size 0,1 or 2, and we are inside the `if` statement for the case where it is non-empty, so we do have a guarantee that the the `else` case is for when there is only one child. So I don't think we need an additional assert here.

src/optimizer/plan_generator.cpp
soumilm: I added a comment in the code for this, but:
0 children: the node is not a leader (i.e. it reads from the temp table but does not write)
1 child: the node is the first time we refer to a _simple_ CTE table, in which case we populate the table as well (this is a leader); the child is the plan for the query used to populate the table
2 children: the node is a leader for an _inductive_ CTE table, in which case the two children are the plans for the base and inductive queries for the table

src/include/planner/plannodes/cte_scan_plan_node.h
```diff
@@ -0,0 +1,246 @@
...
+    /**
+     * Build the limit plan node
+     * @return plan node
+     */
+    std::unique_ptr<CteScanPlanNode> Build() {
+      std::vector<catalog::col_oid_t> col_oids;
+      for (auto &col : table_schema_.GetColumns()) {
+        col_oids.push_back(col.Oid());
+      }
+      return std::unique_ptr<CteScanPlanNode>(
```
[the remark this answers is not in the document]
soumilm: The constructor we're using here is private, so I think `make_unique` won't work in this case

script/testing/junit/traces/cte.sql
soumilm: These are all our own tests, tested against postgres

src/include/execution/sql/ind_cte_scan_iterator.h
```diff
@@ -0,0 +1,83 @@
...
+  ~IndCteScanIterator() = default;
+
+  /**
+   * This class cannot be copied or moved
+   */
+  DISALLOW_COPY_AND_MOVE(IndCteScanIterator);
+
+ private:
+  exec::ExecutionContext *exec_ctx_;
+  CteScanIterator cte_scan_1_;
```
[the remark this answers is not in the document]
soumilm: I added explanations as comments

src/include/common/strong_typedef.h
```diff
@@ -234,6 +234,12 @@ class StrongTypeAlias {
  private:
   IntType val_;
 };
+
+/**
+ * gets the type of the underlying value of a given strongtype at compile time
+ */
+template <typename T>
+using get_strongtype_underlying_t = std::decay_t<decltype(std::declval<T>().UnderlyingValue())>;
```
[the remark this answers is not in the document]
tanujnay112: how would I get the underlying type and templatize the MakeTempOid function as opposed to making it a macro then

src/execution/compiler/pipeline.cpp
```diff
@@ -340,33 +416,52 @@ ast::FunctionDecl *Pipeline::GenerateTearDownPipelineFunction() const {
...
+      auto pipeline_state = codegen_->TLSAccessCurrentThreadState(codegen_->MakeExpr(tls), state.GetTypeName());
+      auto call = codegen_->Call(GetTearDownPipelineStateFunctionName(), {state_ptr, pipeline_state});
+      builder.Append(codegen_->MakeStmt(call));
+    }
   }
   return builder.Finish();
 }
 
 void Pipeline::GeneratePipeline(ExecutableQueryFragmentBuilder *builder) const {
   // Declare the pipeline state.
```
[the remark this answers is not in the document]
tanujnay112: Added individual comments where nested functionality differs

src/execution/compiler/pipeline.cpp
```diff
@@ -340,33 +416,52 @@ ast::FunctionDecl *Pipeline::GenerateTearDownPipelineFunction() const {
...
   // Generate pipeline state initialization and tear-down functions.
   builder->DeclareFunction(GenerateSetupPipelineStateFunction());
   builder->DeclareFunction(GenerateTearDownPipelineStateFunction());
 
   // Generate main pipeline logic.
   builder->DeclareFunction(GeneratePipelineWorkFunction());
+  builder->DeclareFunction(GenerateRunPipelineFunction());
+  builder->DeclareFunction(GenerateInitPipelineFunction());
+  auto teardown = GenerateTearDownPipelineFunction();
+  builder->DeclareFunction(teardown);
```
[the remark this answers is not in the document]
tanujnay112: RegisterStep did the equivalent before

src/include/catalog/catalog_accessor.h
tanujnay112: It should still work. All uses of this are restricted to the backend of the execution engine so cached plans should behave the same way. I'll make note of this.

src/include/execution/compiler/pipeline.h
tanujnay112: Clang-tidy complained about too much padding with the original arrangement

src/include/execution/compiler/pipeline.h
```diff
@@ -241,31 +259,51 @@ class Pipeline {
...
   bool check_parallelism_;
-  // All pipelines this one depends on completion of.
-  std::vector<Pipeline *> dependencies_;
-  // Cache of common identifiers.
-  ast::Identifier state_var_;
-  // The pipeline state.
-  StateDescriptor state_;
-  // The pipeline operating unit feature vector state.
-  StateDescriptor::Entry oufeatures_;
+  // Whether or not this is a nested pipeline
```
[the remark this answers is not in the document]
tanujnay112: Refactored nested pipelines to not be misleading and do what it suggests

src/include/planner/plannodes/cte_scan_plan_node.h
```diff
@@ -0,0 +1,246 @@
...
+  // Update schema
+  //===--------------------------------------------------------------------===//
+
+  /**
+   * Output schema for the node. The output schema contains information on columns of the output of the plan
+   * node operator
+   * @param schema output schema for plan node
+   */
+  void SetTableSchema(catalog::Schema schema) {
+    // TODO(preetang): Test for memory leak
```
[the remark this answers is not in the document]
tanujnay112: It doesn't lol, made sure and removed

src/optimizer/optimizer.cpp
```diff
@@ -56,6 +58,14 @@ std::unique_ptr<OptimizeResult> Optimizer::BuildPlanTree(transaction::Transactio
     PlanGenerator generator(optimize_result->GetPlanMetaData());
     auto best_plan = ChooseBestPlan(txn, accessor, root_id, phys_properties, output_exprs, &generator);
     optimize_result->SetPlanNode(std::move(best_plan));
+
+    // Assign CTE Schema to each CTE Node
+    for (auto &table : context_->GetCTETables()) {
+      planner::CteScanPlanNode *leader = nullptr;
+      common::ManagedPointer<planner::CteScanPlanNode> ldr = common::ManagedPointer(leader);
+      ElectCTELeader(common::ManagedPointer(best_plan), table, &ldr);
```
jkosh44, line 66: This could be due to my lack of C++ knowledge, but how can we use `best_plan` here after having moved it into `optimize_result` on line 60?

lmwnshn: @jkosh44 I think you're right.

jkosh44: You should be fine to move `optimize_result->SetPlanNode(std::move(best_plan));` from line 60 to after this loop.

src/parser/postgresparser.cpp
```diff
@@ -2061,6 +2064,69 @@ std::unique_ptr<VariableSetStatement> PostgresParser::VariableSetTransform(Parse
   return result;
 }
 
+<<<<<<< HEAD
```
jkosh44, line 2067: Think you may have left this in by accident

soumilm: Yup, just fixed that

src/optimizer/optimizer.cpp
soumilm: Fixed

apavlo:
Replaced by #1590

PR #1123, opened 2020-08-26
Templated perf monitor and benchmark

mbutrovich:
Perf events can be opened to be enabled, disabled, and read out as a single sys call respectively. However, you can only do that if you don't want to count children threads in the parent thread that instantiates the perf counters. Since it looks like we're only using the `false` inherit flag right now, I templated the class for a performance speedup when we want to use perf counters without inheriting counters.


mbutrovich:
For reference, here are numbers on my NUC for the new benchmark:
```
--------------------------------------------------------------------
Benchmark                             Time           CPU Iterations
--------------------------------------------------------------------
PerfMonitorBenchmark/Basic        11731 ns      11731 ns      64216    83.249k items/s
PerfMonitorBenchmark/Inherit      15895 ns      15894 ns      45351    61.441k items/s
```

mbutrovich:
Still need to revise documentation a bit then it'll be ready for review.

test/common/perf_monitor_test.cpp
```diff
@@ -2,92 +2,71 @@
...
+  perf_counters parent_counters, catalog_counters, sleep_counters;
   parent_monitor.Start();
-  std::thread thread1(CreateAndDestroyCatalog, &catalog_counters);
-  std::thread thread2(JustSleep, &sleep_counters);
+  std::thread thread1(CreateAndDestroyCatalog<inherit, perf_counters>, &catalog_counters);
+  std::thread thread2(JustSleep<inherit, perf_counters>, &sleep_counters);
   thread1.join();
   thread2.join();
   parent_monitor.Stop();
   parent_counters = parent_monitor.Counters();
```
lmwnshn, line 51: Does this test now essentially just start and stop the monitor?

Happy to LGTM the rest otherwise.

mbutrovich: The tests are hard to evaluate deterministically since depending on your kernel flags, perf counters might be 0. On a Mac they'll definitely be 0. I have tests there as a sanity check for compilation and to also have a quick test to toss a breakpoint in on a system to verify the counters are working correctly.

src/include/common/perf_monitor.h
```diff
@@ -93,7 +102,7 @@ class PerfMonitor {
    * accumulate all.
```
lmwnshn, line 102: You need to wipe out this comment for doxygen to pass, but LGTM otherwise.

PR #1127, opened 2020-08-26
Improved performance of AccessCurrentThreadState in ThreadStatecontainer.

arvindsaik:
Removed tbb::enumerable::thread_specific and added std::map with a mutex. Flame graphs for TPCC benchmark with different configurations can be found here : https://drive.google.com/drive/folders/1QQcSaYiGVYMi8gPsb8kXZxEehKR0ucTH?usp=sharing.

TPCC CPU Utilisation 
- Single threaded: 1.09% with TBB → 0.54% w/o TBB
- Multi threaded (8): 1.09% with TBB → 0.74% w/o TBB


apavlo:
If the overhead of your solution increases with higher core counts, what happens when you go above eight threads?

abhijithanilkumar:
We found that `unordered_map` has poor performance compared to `map`. And it is more or less constant, 0.50-0.60 as we scale. Once we switch to map, the numbers are:

1 terminal, SF 1 : 1.09% with TBB → 0.54% w/o TBB
8 terminals, SF 8 : 1.09% with TBB → 0.48% w/o TBB
16 terminals, SF 16 : 1.09% with TBB → 0.55% w/o TBB

arvindsaik:
Performance Report
2 threads -> 0.58%
4 threads -> 0.55%
6 threads -> 0.57%
8 threads -> 0.48%
10 threads -> 0.66%
12 threads -> 0.60%
14 threads -> 0.66%
16 threads -> 0.55%
18 threads -> 0.69%

eppingere (changes requested):
Looks pretty good just some tweaking to take a look at

src/execution/sql/thread_state_container.cpp
```diff
@@ -47,7 +47,8 @@ ThreadStateContainer::TLSHandle::~TLSHandle() {
 
 // The actual container for all thread-local state for participating threads
 struct ThreadStateContainer::Impl {
-  tbb::enumerable_thread_specific<TLSHandle> states_;
+  tbb::mutex states_mutex_;
+  std::map<std::thread::id, std::shared_ptr<TLSHandle>> states_;  // NOLINT
```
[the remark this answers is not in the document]
arvindsaik: I think since we are trying to move away from TBB it is probably better to stick with the std::map implementation?

arvindsaik:
- Added spin latch, Removed tbb::mutex
- Converted shared_ptr to pointer.
- Optimised AccessCurrentThreadState() function

New TPCC performance numbers:
1 thread : 0.23%
36 threads : 0.40%

src/execution/sql/thread_state_container.cpp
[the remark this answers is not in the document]
eppingere: Sounds good

src/execution/sql/thread_state_container.cpp
```diff
@@ -1,13 +1,13 @@
 #include "execution/sql/thread_state_container.h"
 
-#include <tbb/enumerable_thread_specific.h>
+#include <tbb/mutex.h>
```
pmenon, line 3: I don't think you need this header if you're using `common::SpinLatch`. Thinking again, why isn't the header for `common::SpinLatch` included here?

src/execution/sql/thread_state_container.cpp
```diff
@@ -47,7 +47,8 @@ ThreadStateContainer::TLSHandle::~TLSHandle() {
 
 // The actual container for all thread-local state for participating threads
 struct ThreadStateContainer::Impl {
-  tbb::enumerable_thread_specific<TLSHandle> states_;
+  common::SpinLatch states_latch_;
+  std::map<std::thread::id, TLSHandle *> states_;
```
pmenon, line 51: Why `std::map` and not `std::unordered_map` ?

pmenon: Also, why not use `std::unique_ptr<>` for the values? That should simplify your cleanup logic.

src/execution/sql/thread_state_container.cpp
```diff
@@ -83,35 +87,52 @@ void ThreadStateContainer::Reset(const std::size_t state_size, const ThreadState
 }
 
 byte *ThreadStateContainer::AccessCurrentThreadState() {
-  auto &tls_handle = impl_->states_.local();
-  return tls_handle.State();
+  common::SpinLatch::ScopedSpinLatch guard(&impl_->states_latch_);
+  auto iter = impl_->states_.find(std::this_thread::get_id());
+  byte *state = nullptr;
```
pmenon, line 92: This pattern will perform two lookups into the hash table, which might explain the performance you're seeing. What you want to do is:
```
auto &tls_handle = impl_->states_[std::this_thread::get_id()];
if (!tls_handle) {
  tls_handle = std::make_unique<TLSHandle>(this);
}
return tls_handle->State();
```

Check the syntax, but the above should be faster. You should also consider `std::unordered_map::reserve()` at `ThreadStateContainer` construction time to pre-allocate room to avoid resizing while holding the latch here.

src/execution/sql/thread_state_container.cpp
```diff
@@ -83,35 +87,52 @@ void ThreadStateContainer::Reset(const std::size_t state_size, const ThreadState
...
+    state = iter->second->State();
+  }
+  return state;
 }
 
 void ThreadStateContainer::CollectThreadLocalStates(std::vector<byte *> *container) const {
   container->clear();
+  common::SpinLatch::ScopedSpinLatch guard(&impl_->states_latch_);
   container->reserve(impl_->states_.size());
   for (auto &tls_handle : impl_->states_) {
```
pmenon, line 107: The loop variable isn't a `TLSHandle` any more. Suggest changing the name to `entry`, or using structured binding:
```
for (auto &[tid, tls_handle] : imple_->states_) {
  ...
}
```

src/execution/sql/thread_state_container.cpp
```diff
@@ -83,35 +87,52 @@ void ThreadStateContainer::Reset(const std::size_t state_size, const ThreadState
...
 }
 
 void ThreadStateContainer::CollectThreadLocalStateElements(std::vector<byte *> *container,
                                                            const std::size_t element_offset) const {
   container->clear();
+  common::SpinLatch::ScopedSpinLatch guard(&impl_->states_latch_);
   container->reserve(impl_->states_.size());
   for (auto &tls_handle : impl_->states_) {
-    container->push_back(tls_handle.State() + element_offset);
+    container->push_back(tls_handle.second->State() + element_offset);
```
pmenon, line 118: Ditto for loop variable renaming.

src/execution/sql/thread_state_container.cpp
```diff
@@ -83,35 +87,52 @@ void ThreadStateContainer::Reset(const std::size_t state_size, const ThreadState
...
   }
 }
 
 void ThreadStateContainer::IterateStates(void *const ctx, ThreadStateContainer::IterateFn iterate_fn) const {
+  common::SpinLatch::ScopedSpinLatch guard(&impl_->states_latch_);
   for (auto &tls_handle : impl_->states_) {
-    iterate_fn(ctx, tls_handle.State());
+    iterate_fn(ctx, tls_handle.second->State());
```
pmenon, line 125: Ditto for loop variable renaming.

src/execution/sql/thread_state_container.cpp
```diff
@@ -83,35 +87,52 @@ void ThreadStateContainer::Reset(const std::size_t state_size, const ThreadState
...
   }
 }
 
 void ThreadStateContainer::IterateStatesParallel(void *const ctx, ThreadStateContainer::IterateFn iterate_fn) const {
-  tbb::parallel_for_each(impl_->states_, [&](auto &tls_handle) { iterate_fn(ctx, tls_handle.State()); });
+  common::SpinLatch::ScopedSpinLatch guard(&impl_->states_latch_);
+  tbb::parallel_for_each(impl_->states_.begin(), impl_->states_.end(), [&](auto &tls_handle) {
```
pmenon, line 131: 1. Warning: this can lead to dead-lock if the iteration function, for any reason, calls back into `ThreadStateContainer`. In practice, query code only invokes this function after modifications to the states have completed. So, grabbing a lock here isn't strictly necessary.
2. You can remove the `if` clause below since NULL elements should never exist in the map.

pmenon (changes requested):
Some minor changes. Check them out and re-run perf. 

It isn't clear to my why `cland-tidy` was complaining about the iteration, but hopefully the cleanup might make the errors more clear, if not remove them altogether.

You can also remove `ExecutionSettings` and the `strong_typedef.h` includes from the header.

src/execution/sql/thread_state_container.cpp
arvindsaik: 1) std::map is giving better performance in the benchmarks as compared to std::unordered_map.
2) We did use unique_ptr but that gave us other clang-tidy issues.

arvindsaik: 1) std::map consistently performed better in the benchmarks. I am not able to reason this out though.
2) We did try this but it gave us other clang-tidy issues and so we moved away from it.

src/execution/sql/thread_state_container.cpp
arvindsaik: In our code, find is happening just once. If the entry is not found, it is being inserted into the map for future look ups.

pmenon: Not quite: the version above probes the map **exactly** once per lookup, regardless of whether it's a hit or miss. Your version accesses it twice on miss. This means double the hashing and double the memory references, all under a latch.

This function is called only once-per-pipeline 90% of the time (for TPCC it's probably 100%). This means calling threads incur the full double-map-access since the states are reset in each pipeline.

src/execution/sql/thread_state_container.cpp
pmenon: Do you recall what `clang-tidy` issues came up with `std::unique_ptr`? Apply the changes here and let's see what happens.

src/execution/sql/thread_state_container.cpp
arvindsaik: The initial state of the map is empty. With the code above, every call to AccessCurrentThread would result in thread id not being found and just returning state for a new TLSHandle object. I am not sure the map will even be used in that case. Are you thinking of populating the map in the constructor of the ThreadStateContainer class instead of in the AccessCurrentThreadState function?

arvindsaik:
Regarding the clang-tidy issue, the previous version of our code was exactly as you suggested - without the nullptr checks in the parallel_for_each. It still gave a clang-tidy error on some specific machines. I introduced the checks to see if it will fix the clang-tidy issues.

src/execution/sql/thread_state_container.cpp
pmenon: Take a look at the docs for [`std::unordered_map::operator[]`](https://en.cppreference.com/w/cpp/container/unordered_map/operator_at). It inserts the key and default initializes the value if the key is new. In both cases, it returns a reference to the value. The version above checks if the returned value is `nullptr` (indicating a miss) and move-assigns the value to a new state. This achieves the single map access guarantee.

arvindsaik: For `std::unordered_map::reserve()` shall I pass double the number of cores? We did run 36 threads on an 18 core machine.

pmenon: Try three versions: no reserve, reserve 1x core count, and reserve 2x core count, all with `std::unordered_map<std::thread::id, std::unique_ptr<>>` to see performance.

Ideally, the `ThreadStateContainer` constructor should accept an expected thread count. But, the scope of that change is much larger.

apavlo:
Blocked on #1206

arvindsaik:
Order based on performance:
1. map + normal pointer
1 thread : 0.23%
 36 threads : 0.40%

2. map + unique_ptr :
1 thread : 0.42%
36 threads : 0.45%

3. unordered_map + unique_ptr
1 thread: 0.43
36 threads: 0.45

@pmenon thinks that it is better to go with 3 as it is safer, modern and intuitive and the performance difference with high thread counts is 0.05% (not much).

src/include/execution/sql/thread_state_container.h
```diff
@@ -4,15 +4,10 @@
...
-#include "common/strong_typedef.h"
 #include "execution/sql/memory_pool.h"
 
-namespace terrier::execution::exec {
-class ExecutionSettings;
-}  // namespace terrier::execution::exec
-
 namespace terrier::execution::sql {
-
+using byte = std::byte;
```
pmenon, line 10: Ugh, I didn't realize `byte` is defined in `strong_typedef.h`. Re-include that header file.

pmenon (changes requested):
Small include change, but LGTM otherwise.

pmenon (approved):
LGTM.

lmwnshn (approved):
Rubber stamp.

PR #1129, opened 2020-08-27
Fix FindClangTools.cmake

gonzalezjo:
A long-standing pet peeve of mine is that FindClangTools.cmake isn't particularly portable. This can lead to clang tools not being found on certain configurations. This PR aims to fix that.

From #1067.

### DIFF
--- a/cmake_modules/FindClangTools.cmake
+++ b/cmake_modules/FindClangTools.cmake
@@ -41,6 +41,7 @@ endif()
 if (CLANG_TOOLS_VERSION)
   find_program(CLANG_TIDY_BIN
       NAMES clang-tidy-${CLANG_TOOLS_VERSION}
+      clang-tidy
       PATHS
       ${ClangTools_PATH}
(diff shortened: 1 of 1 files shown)


mbutrovich:
Will this potentially result in it finding a tool binary from a different version of LLVM if someone has multiple versions installed?

gonzalezjo:
Yes. Since we've all been using CLion with its built-in clang-tidy and clang-format binaries without issue, I don't think that that should pose a problem. It is certainly an improvement over not working at all.

mbutrovich:
> Yes. That's mostly the point of this PR: currently, we don't let people run clang-format or clang-tidy at all unless they have LLVM 8. This seems overly burdensome

That's what we want. It's impossible for people to pass CI if their tools don't match the same version. We've seen differences from one version to the next, even with the same checks. If someone wants to use a newer version on CLion then that's up to them, but the wiki has instructions on how to override that as well.

gonzalezjo:
Understood. I'll close the PR, then.

PR #1130, opened 2020-08-27
Revert "Explicit instantiation of spdlog template functions (#995)"

gonzalezjo:
This reverts commit 686eb69f from #995. From #1067. See also: #1004.

The PR doesn't do seem to do a ton for build speeds and is a blocker for continued work on spdlog.


mbutrovich:
Can we still optimize the includes without doing explicit instantiation? Is that still compatible with where we want to go with spdlog?

gonzalezjo:
I don't think so. In fact, the includes get _marginally_ more annoying with the follow-up PR to this. That said, the follow-up is statically linked spdlog. We'd be including spdlog headers that essentially preprocess-out 90% of the stuff we're including soon. That said, I can take a look.

PR #1131, opened 2020-08-27
Apple fixes

gonzalezjo:
A collection of fixes for macOS builds. 

1. Brings the toolchain closer toward what we have on Ubuntu by migrating to clang-8. 

2. Uses asan for gtest. While this applies to builds on all platforms, the bug it addresses (asan error) only ever appears on macOS. Needed for (1) to function. 

3. Fixes a build issue where the wrong LLVM is used on macOS.

Might take a bit to test this.


cmake_modules/san-config.cmake
```diff
@@ -37,7 +37,7 @@ if (${TERRIER_USE_ASAN})
       message(SEND_ERROR "Cannot use dynamic linking when ASAN and UBSAN are both enabled")
     endif ()
   endif ()
-  set(CXX_COMMON_FLAGS "${CXX_COMMON_FLAGS} -fsanitize=address")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address")
```
mbutrovich, line 40: Isn't `CXX_COMMON_FLAGS` already appended to `CMAKE_CXX_FLAGS`? This has the side effect of removing the flag from other compilation steps that use `CXX_COMMON_FLAGS`.

gonzalezjo: That is a bug. Good catch. Here was my reasoning: its appended, but later in the main CMakeLists.txt. By the time it's appended, other steps have already passed, and many dependencies have already been configured _not_ to use a sanitizer.


However, this breaks from expected behavior for util's CMakeList. I'll fix that right now.

gonzalezjo:
All checks passed, and Jenkins ran using my Jenkins file. I think this should be an easy review.

mbutrovich:
This LGTM!

mbutrovich (approved):

PR #1134, opened 2020-08-29
Add support for recording feature values directly via a new TPL builtin, execCtxRecordFeature.

lmwnshn:
This PR adds support for directly recording to the metrics feature vector via a new TPL builtin `@execCtxRecordFeature`.

Summary:

- Counters are implemented to follow the definition in [brain_defs.h](https://github.com/cmu-db/terrier/pull/1134/files#diff-4c774f0c39f07eb16ef19af8143d421aR8), courtesy of @17zhangw 
- All recording is gated by [CompilationContext::IsCountersEnabled()](https://github.com/cmu-db/terrier/pull/1134/files#diff-a8dce8335eae87fbb935cabb7632ef0cR121), which is currently not exposed but in theory could be hooked up to the settings manager. All the code in this PR should be easily ripped out in the future once we have real table statistics.
- Counters are stashed into a TPL program's `queryState` as the counters may need to be maintained across TPL functions within the same query.
- Convenience functions are defined in `OperatorTranslator`: `CounterDeclare`, `CounterSet`, `CounterAdd`. These may be useful for adding new counters. They automatically use the `if (IsCountersEnabled()) ...` guard.
- Modified the [BwTree](https://github.com/cmu-db/terrier/pull/1134/files#diff-f0c28d9553fd33b2315722bd96f3c6cfR6794) to expose a `GetSize()` method, as well as our [hash index](https://github.com/cmu-db/terrier/pull/1134/files#diff-15f0a2eea833a260da9002625361dff7R74). This is for `IDX_SCAN::NUM_ROWS`.

Known issues:

- All counters force serial pipelines right now. This will be addressed as me and @17zhangw sync up...


src/execution/compiler/operator/seq_scan_translator.cpp
```diff
@@ -215,6 +217,8 @@ void SeqScanTranslator::ScanVPI(WorkContext *ctx, FunctionBuilder *function, ast
       function->Append(assign);
       // Push to parent.
       ctx->Push(function);
+
+      CounterAdd(function, num_scans_, 1);
```
17zhangw, line 221: This is a small concern. `num_scans_` records the number of tuples that match the filter and are passed on to the next translator. This means that as far as the model is concerned, the cost of scanning 1 million tuples to emit 1 tuple is the same as the cost of scanning 1 tuple to emit 1 tuple.

Is there a way to get the raw number of tuples in the table (or at least a close approximation)?

src/execution/compiler/operator/sort_translator.cpp
```diff
@@ -233,6 +249,20 @@ void SortTranslator::FinishPipelineWork(const Pipeline &pipeline, FunctionBuilde
     } else {
       function->Append(codegen->SorterSort(sorter_ptr));
     }
+
+    FeatureRecord(function, brain::ExecutionOperatingUnitType::SORT_BUILD,
+                  brain::ExecutionOperatingUnitFeatureAttribute::NUM_ROWS, pipeline,
+                  num_sort_build_rows_.Get(GetCodeGen()));
+    FeatureRecord(function, brain::ExecutionOperatingUnitType::SORT_BUILD,
+                  brain::ExecutionOperatingUnitFeatureAttribute::CARDINALITY, pipeline,
+                  codegen->CallBuiltin(ast::Builtin::SorterGetTupleCount, {sorter_ptr}));
```
17zhangw, line 258: Since we don't necessarily want to track data uniqueness, this might be the best we can do? (although the prediction results will be slightly affected since `sort_build` does depend on the input data cardinality).

@malin1993ml

src/include/brain/brain_defs.h
```diff
@@ -5,36 +5,136 @@ namespace terrier::brain {
...
+   * PROJECTION
+   * num_rows:
+   * cardinality:
+   * This shouldn't be emitted?
+   */
   PROJECTION,
+  /**
+   * OUTPUT
+   * num_rows: # rows being output
+   * cardinality: 1 for network output TODO(WAN): will when is this not 1?
```
17zhangw, line 98: In response to the TODO, it is set to 0 when using the NoOpResultConsumer

src/include/brain/brain_defs.h
```diff
@@ -5,36 +5,136 @@ namespace terrier::brain {
...
 
+  /**
+   * AGGREGATE_BUILD
+   * num_rows: # input rows to aggregation
+   * cardinality: # unique values (although probably more correct is # of unique values based on group by clause)
+   */
   AGGREGATE_BUILD,
+  /**
+   * AGGREGATE_ITERATE
+   * num_rows: # rows output by aggregation  // TODO(WAN): when is this not 1?
```
17zhangw, line 16: Is this TODO still current?

src/include/brain/brain_defs.h
```diff
@@ -5,36 +5,136 @@ namespace terrier::brain {
...
+   * SORT_ITERATE
+   * num_rows: # tuples output
+   * cardinality: # unique values
+   */
   SORT_ITERATE,
 
+  /**
+   * SEQ_SCAN
+   * num_rows: # tuples output (uncertain whether it's # accessed vs # after applying the filters)
+   * cardinality: # unique values // TODO(WAN): this should be the same right?
```
17zhangw, line 59: In response to TODO, cardinality can probably be set to num_rows without anything going noticeably wrong (since a seq scan doesn’t really depend on table cardinality).

On a separate note, should probably amend the num_rows definition once a previous comment gets resolved?

17zhangw (changes requested):
Generally LGTM. Just a few questions.

src/execution/compiler/operator/seq_scan_translator.cpp
malin1993ml: I think the concern is valid. Maybe directly utilize the `SqlTable::GetNumTuple` function?

src/execution/compiler/operator/sort_translator.cpp
malin1993ml: Yeah I think this is the best we can do. I would add a TODO here to document that in theory we should use the number of unique tuples as cardinality, but given the potential overhead and engineering complexity we just use the size of the sorter.

src/include/brain/brain_defs.h
```diff
@@ -5,36 +5,136 @@ namespace terrier::brain {
...
   HASHJOIN_PROBE,
 
+  /**
+   * IDXJOIN
+   * num_rows: N/A
+   * cardinality: N/A
+   * This feature doesn't actually exist.
+   * The recorder emits a IDX_SCAN (with num_loops = # rows output by the outer loop)
+   */
   IDXJOIN,
```
malin1993ml, line 41: Do we still need this if the feature doesn't actually exist? Also, it seems that we haven't modified the `index_join_translator` to add the counters, especially for the `num_loops` feature?

src/include/brain/brain_defs.h
```diff
@@ -5,36 +5,136 @@ namespace terrier::brain {
...
+   * IDX_SCAN
+   * num_rows: size of index
+   * cardinality: size of scan
+   */
   IDX_SCAN,
 
+  /**
+   * INSERT
+   * num_rows: # input tuples
+   * cardinality: # unique values
```
malin1993ml, line 72: This is just the same as `num_rows`.

src/include/brain/brain_defs.h
```diff
@@ -5,36 +5,136 @@ namespace terrier::brain {
...
+   */
   INSERT,
+  /**
+   * UPDATE
+   * num_rows: # input tuples
+   * cardinality: # unique values
```
malin1993ml, line 78: Same here.

src/include/brain/brain_defs.h
```diff
@@ -5,36 +5,136 @@ namespace terrier::brain {
...
+   */
   UPDATE,
+  /**
+   * DELETE
+   * num_rows: # input tuples
+   * cardinality: # unique values
```
malin1993ml, line 84: Same here.

src/include/brain/brain_defs.h
```diff
@@ -5,36 +5,136 @@ namespace terrier::brain {
...
+   * NL_JOIN
+   */
   NL_JOIN,
+  /**
+   * SORT
+   */
   SORT,
+  /**
+   * STATIC_AGGREGATE
+   */
```
malin1993ml, line 131: Since none of these features actually exist, shouldn't we just remove them?

src/include/execution/compiler/compilation_context.h
```diff
@@ -117,6 +117,9 @@ class CompilationContext {
    */
   CompilationMode GetCompilationMode() const { return mode_; }
 
+  /** @return True if we should collect counters in TPL, used for Lin's models. */
+  bool IsCountersEnabled() const { return true; }
```
malin1993ml, line 121: Yeah I think we should probably set this to `false` by default.

src/execution/compiler/operator/sort_translator.cpp
malin1993ml: I was also running TPCC just now with this branch. I noticed that this value as well as the cardinality value for `SORT_ITERATE` are recorded as a very high corrupted value. This probably needs double-check.

malin1993ml:
While I was testing against the TPCH queries, I noticed that there's another high-level thing: there're certain translators that have [arithmetic features](https://github.com/cmu-db/terrier/blob/master/src/include/brain/brain_defs.h#L40), such as the predicate for seq scan, index scan, or joins. Those features are already recorded in the pipeline features, but we need to update their counters as well. Thus, we need another API in the `ExecutionContext`:
`RecordArithmeticFeature(pipeline_id_t pipeline_id, translator_id_t translator_id, uint32_t value);`
This function will **multiply** the `num_rows`  and `cardinality` for all the arithmetic features by `value`. `num_rows` and `cardinality` are equal because arithmetics don't really care the uniqueness. We need to add this to almost all the translators you have. It's multiplication because we may have merged multiple arithmetics under the same type together (the original values would just be the number of occurrences). For most translators, this value would just be equal to the `num_rows` value of the primary feature of the translator (such as `num_rows` for the `SEQ_SCAN` OU of the `seq_scan_translator`). I think the only two exceptions would be `HASHJOIN_PROBE` and `IDX_SCAN`, where we're going to use their `cardinality` to set the values for the arithmetics in the same translator. Let me know whether this makes sense, and we can also chat offline. The rest of the PR is awesome.

src/execution/compiler/operator/seq_scan_translator.cpp
lmwnshn: Changed it to add the tuple count from the VPI.

For the record, looks like `SqlTable::GetNumTuple` is borked for getting approximate counts.

src/execution/compiler/operator/sort_translator.cpp
lmwnshn: It works for small test cases for me.. Does it go above int32_max? I wonder if we are hitting #1049.

src/execution/compiler/operator/seq_scan_translator.cpp
malin1993ml: I think your change works.

src/execution/compiler/operator/sort_translator.cpp
malin1993ml: I don't think it goes above int32_max. I just directly compiled your branch and executed TPCC against it with `--metrics_pipeline` turned on. Maybe try TPCC locally to see whether you can reproduce?

lmwnshn: Fixed, it was a bug with bytecode generation.

src/include/brain/brain_defs.h
lmwnshn: I think in practice I can always just record the feature as 1, then, right? I'm not sure I know which consumer is consuming it.

src/include/brain/brain_defs.h
lmwnshn: Yes; an aggregation typically smashes a whole bunch of values into a single value right? When will there be more than 1 value in an aggregation's output? I cannot think of any right now.

src/include/brain/brain_defs.h
lmwnshn: Done.

src/include/brain/brain_defs.h
malin1993ml: I think you can leave this particular field as it is. In practice it will be set outside, and you're right that you won't know which consumer it is.

src/include/brain/brain_defs.h
malin1993ml: Well in the case of a group by there will be more values (same number of rows as the distinct values in the group by column).

src/include/brain/brain_defs.h
lmwnshn: Removed.

src/include/brain/brain_defs.h
lmwnshn: Removed.

src/include/brain/brain_defs.h
```diff
@@ -70,6 +61,10 @@ enum class ExecutionOperatingUnitType : uint32_t {
    * IDX_SCAN
    * num_rows: size of index
    * cardinality: size of scan
+   *
+   * Special case: in the case of an index join, it is also recorded as an IDX_SCAN where:
+   * num_rows: number of rows output by the outer loop
+   * cardinality: N/A
```
lmwnshn, line 67: @malin1993ml @17zhangw Can you confirm this N/A part? Or should cardinality be the same as NUM_ROWS?

malin1993ml: `num_rows` and `cardinality` would actually be the same as a regular index scan. That is, the size of the index and the size of the scan per `idx_scan` invocation. If calculating a per-invocation value is not convenient, you can also just use the value of the last invocation as an approximation.

lmwnshn: I do not understand what num_loops would be recorded as then (the number of times the outer loop is accessed).

lmwnshn: That said, I have pushed my changes.

src/include/brain/brain_defs.h
lmwnshn: Got it.

lmwnshn:
I think I have added the arithmetic features and addressed review comments, could you take a look to see if I understood what is needed? Thanks! @malin1993ml @17zhangw

src/include/brain/brain_defs.h
17zhangw: I took a look at the changes to the `IndexJoinTranslator`. So the existing `num_scans_index_` essentially tracks how many tuples are retrieved from the index. So this would be recorded as the `cardinality` (which is what is done now).

For `num_loops`, it would essentially be tracking the number of times `@indexIteratorScanKey(&index_iter)` is invoked (which equals # tuples emit by the index join plan's child and equals # times the index is looked up on).

src/execution/compiler/operator/operator_translator.cpp
```diff
@@ -142,4 +161,52 @@ void OperatorTranslator::FeatureRecord(FunctionBuilder *function, brain::Executi
...
+          ast::Expr *record =
+              codegen->ExecCtxRecordFeature(GetExecutionContext(), pipeline_id, feature.GetFeatureId(), attrib, val);
+          function->Append(record);
+        }
+      }
+    }
+  }
+}
+
+void OperatorTranslator::FeatureArithmeticRecordMul(FunctionBuilder *function, const Pipeline &pipeline,
```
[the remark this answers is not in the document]
malin1993ml: Thanks for following up on this, William! I think you're correct that we need to differentiate the translator ids when doing the arithmetic multiplication. And we do need to modify the `operating_unit_recorder` to only merge arithmetics of the same translator.

src/brain/operating_unit_recorder.cpp
```diff
@@ -383,9 +384,8 @@ void OperatingUnitRecorder::Visit(const planner::IndexScanPlanNode *plan) {
 
 void OperatingUnitRecorder::VisitAbstractJoinPlanNode(const planner::AbstractJoinPlanNode *plan) {
   if (plan_feature_type_ == ExecutionOperatingUnitType::HASHJOIN_PROBE ||
-      plan_feature_type_ == ExecutionOperatingUnitType::NL_JOIN ||
-      plan_feature_type_ == ExecutionOperatingUnitType::IDXJOIN) {
-    // Right side stiches together outputs
+      plan_feature_type_ == ExecutionOperatingUnitType::DUMMY) {
+    // Right side stitches together outputs
```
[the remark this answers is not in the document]
lmwnshn: Fixed with your suggestion of moving things [around](https://github.com/cmu-db/terrier/pull/1134/commits/073ab53e009633fee373f1fe185dc8043f2366ff#diff-74b764e1da242750e292732b021c45e2R432).

src/brain/operating_unit_recorder.cpp
```diff
@@ -222,7 +222,13 @@ void OperatingUnitRecorder::AggregateFeatures(brain::ExecutionOperatingUnitType
     }
   }
 
-  auto feature = ExecutionOperatingUnitFeature(current_translator_->translator_id_, type, num_rows, key_size, num_keys,
+  // This is a hack.
```
lmwnshn, line 225: @17zhangw

src/execution/compiler/operator/operator_translator.cpp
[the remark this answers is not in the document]
lmwnshn: Modified by having each translator denote whether it should own its features or pass it through to some other translator. Right now, projection and nested_loop_join passes it through.

src/execution/compiler/operator/index_join_translator.cpp
```diff
@@ -87,13 +98,27 @@ void IndexJoinTranslator::PerformPipelineWork(WorkContext *context, FunctionBuil
...
+                 GetCodeGen()->CallBuiltin(ast::Builtin::IndexIteratorGetSize, {GetCodeGen()->AddressOf(index_iter_)}));
   // @indexIteratorFree(&index_iter_)
   FreeIterator(function);
 }
 
+void IndexJoinTranslator::FinishPipelineWork(const Pipeline &pipeline, FunctionBuilder *function) const {
+  FeatureRecord(function, brain::ExecutionOperatingUnitType::IDX_SCAN,
+                brain::ExecutionOperatingUnitFeatureAttribute::NUM_ROWS, pipeline, CounterVal(index_size_));
+  FeatureRecord(function, brain::ExecutionOperatingUnitType::IDX_SCAN,
+                brain::ExecutionOperatingUnitFeatureAttribute::CARDINALITY, pipeline, CounterVal(num_scans_index_));
```
malin1993ml, line 116: To match the models, can we change this to per-loop num scans, if num loops > 0 (i.e., int(num_scans_index_ / num_loops_))?

src/brain/operating_unit_recorder.cpp
```diff
@@ -222,7 +222,14 @@ void OperatingUnitRecorder::AggregateFeatures(brain::ExecutionOperatingUnitType
     }
```
malin1993ml, line 222: We also need to check that the pipeline ids are the same before we merge features here, since now we differentiate the multiplication by the pipeline ids.

malin1993ml (changes requested):
I think overall this is great! I've checked the TPCC recordings, except some minor comments I left below, all the other features look correct!

I also ran the `tpch_runner` to check the counters as well, but I observed something weird. Q1-p3 has `num_rows` > 0 for `AGG_BUILD`, but `cardinality` (size of the hash table) = 0. And subsequently, Q1-p2 and Q1-p3 have `num_rows` = 0 for all their OUs. This is wrong because an aggregation with `num_rows` > 0 should always have return values. I don't know whether this is related to the counters we added, but could you double-check on this?

Q18-p6 also has all the OUs with `num_rows` = 0, which also seems suspicious to me.

And I also noticed some arithmetic features for TPCH queries still have `num_rows` = 0 despite all the translators in the pipeline have `num_rows` > 0. The last `OP_INTEGER_PLUS_OR_MINUS` feature for Q4-p3 is an example of this. This is less important, but could you double-check as well if you have time?

lmwnshn:
@thepinetree This is the branch, you can probably just build on top of it. We might want to call briefly.

src/execution/compiler/operator/output_translator.cpp
```diff
@@ -39,11 +45,19 @@ void OutputTranslator::PerformPipelineWork(terrier::execution::compiler::WorkCon
...
+  CounterAdd(function, num_output_, 1);
 }
 
 void OutputTranslator::FinishPipelineWork(const Pipeline &pipeline, FunctionBuilder *function) const {
   auto exec_ctx = GetExecutionContext();
   function->Append(GetCodeGen()->CallBuiltin(ast::Builtin::ResultBufferFinalize, {exec_ctx}));
+
+  FeatureRecord(function, brain::ExecutionOperatingUnitType::OUTPUT,
+                brain::ExecutionOperatingUnitFeatureAttribute::NUM_ROWS, pipeline, CounterVal(num_output_));
+  FeatureRecord(function, brain::ExecutionOperatingUnitType::OUTPUT,
```
malin1993ml, line 58: A minor thing: we should remove setting the cardinality for `OUTPUT` since you don't really know which consumer it is and it would be set outside.

src/execution/compiler/operator/hash_aggregation_translator.cpp
```diff
@@ -387,6 +387,7 @@ void HashAggregationTranslator::UpdateAggregates(WorkContext *context, FunctionB
 
   // Advance aggregate.
   AdvanceAggregate(function, agg_payload, agg_values);
+  FeatureArithmeticRecordAdd(function, context->GetPipeline(), GetTranslatorId(), codegen->Const32(1));
```
lmwnshn, line 390: I am trying to convince myself that this is correct.

src/brain/operating_unit_recorder.cpp
lmwnshn: Added a check on the set of translators.

src/execution/compiler/operator/static_aggregation_translator.cpp
```diff
@@ -185,6 +185,7 @@ void StaticAggregationTranslator::PerformPipelineWork(WorkContext *context, Func
     CounterAdd(function, num_agg_outputs_, 1);
   } else {
     UpdateGlobalAggregate(context, function);
+    FeatureArithmeticRecordAdd(function, context->GetPipeline(), GetTranslatorId(), codegen->Const32(1));
```
lmwnshn, line 188: And here too.

src/execution/compiler/operator/hash_aggregation_translator.cpp
malin1993ml: Emm I'm surprised that we even need to do this. I think in theory we just need to multiply the arithmetic features by `num_agg_inputs_` (which you already did). By default each OU has a [`num_rows` = 1](num_agg_inputs_), and the arithmetic OUs will be invoked by `num_agg_inputs_` times. Adding another counter does not seem to be the best direction for me. Unless the arithmetics have a `num_rows` = 0 to begin with? Then that might be a bug in the operating unit recorder. Also, the multiplication of `num_agg_outputs_` in the `AGG_ITERATE` phase seems redundant to me...

lmwnshn: Removed, it was a different bug.

Could you elaborate on the redundancy?

lmwnshn:
@malin1993ml I had a bug where I was multiplying the arithmetic OU every iteration instead of at the end. Due to #1049, this bug overwrites random memory and may explain weird things that you've seen before.

SF1, I am not seeing anything too weird for e.g. Q1 now, I think:

```
0, 3, 0, 6, OP_INTEGER_COMPARE;AGG_BUILD;OP_INTEGER_PLUS_OR_MINUS;OP_DECIMAL_MULTIPLY;SEQ_SCAN;OP_DECIMAL_PLUS_OR_MINUS, 6001215;5914748;5914748;17744244;6001215;59147480, 4;128;4;8;4;8, 1;32;1;1;1;1, 6001215;4;5914748;17744244;6001215;59147480, 1;0.343284;1;1;1;1, 0;0;0;0;0;0, 1599493930515711, 0, 0, 0, 0, 0, 0, 0, 0, 47512, 2215103
0, 2, 0, 2, AGG_ITERATE;SORT_BUILD, 4;4, 188;128, 40;32, 4;4, 1;0.34375, 0;0, 1599493932730821, 0, 0, 0, 0, 0, 0, 0, 0, 45248, 6
0, 1, 0, 2, SORT_ITERATE;OUTPUT, 4;4, 188;188, 40;40, 4;1, 1;1, 0;0, 1599493932730828, 0, 0, 0, 0, 0, 0, 0, 0, 1, 2
```

src/execution/compiler/operator/hash_aggregation_translator.cpp
malin1993ml: I see! In that case, wouldn't the `num_rows` and `cardinality` features for `AGG_BUILD` and `AGG_ITERATE` also be recorded many times instead of just one time in the end? (Though the eventual value is correct.) Maybe worth double-checking other counters.

For the redundancy - Basically I'm wondering would the `AGG_BUILD` and `AGG_ITERATE` have the same translator id or not? If they have the same translator id, then all the arithmetics would be multiplied twice in both of the two pipelines, right?

src/execution/compiler/operator/delete_translator.cpp
```diff
@@ -41,6 +45,15 @@ void DeleteTranslator::PerformPipelineWork(WorkContext *context, FunctionBuilder
   for (const auto &index_oid : indexes) {
     GenIndexDelete(function, context, index_oid);
   }
+
+  FeatureRecord(function, brain::ExecutionOperatingUnitType::DELETE,
```
lmwnshn, line 49: Move this to FinishPipelineWork.

src/execution/compiler/operator/output_translator.cpp
```diff
@@ -20,6 +20,12 @@ OutputTranslator::OutputTranslator(const planner::AbstractPlanNode &plan, Compil
       output_struct_(GetCodeGen()->MakeFreshIdentifier("OutputStruct")) {
   // Prepare the child.
   compilation_context->Prepare(plan, pipeline);
+
+  num_output_ = CounterDeclare("num_output");
+}
+
+void OutputTranslator::InitializePipelineState(const Pipeline &pipeline, FunctionBuilder *function) const {
```
lmwnshn, line 27: Move to InitializeQueryState.

src/execution/compiler/operator/seq_scan_translator.cpp
```diff
@@ -288,6 +298,14 @@ void SeqScanTranslator::PerformPipelineWork(WorkContext *context, FunctionBuilde
   if (declare_local_tvi) {
     function->Append(codegen->TableIterClose(codegen->MakeExpr(tvi_var_)));
   }
+
+  FeatureRecord(function, brain::ExecutionOperatingUnitType::SEQ_SCAN,
```
lmwnshn, line 302: Move this to FinishPipelineWork.

src/execution/compiler/operator/update_translator.cpp
```diff
@@ -77,6 +81,16 @@ void UpdateTranslator::PerformPipelineWork(WorkContext *context, FunctionBuilder
   }
   function->Append(GetCodeGen()->ExecCtxAddRowsAffected(GetExecutionContext(), 1));
 
+  CounterAdd(function, num_updates_, 1);
+
+  FeatureRecord(function, brain::ExecutionOperatingUnitType::UPDATE,
```
lmwnshn, line 86: Move this to FinishPipelineWork.

lmwnshn (commented):
Fix me.

malin1993ml (changes requested):
I think this is good! A few minor things and then I think we're done!

src/execution/compiler/operator/hash_aggregation_translator.cpp
```diff
@@ -426,28 +436,47 @@ void HashAggregationTranslator::PerformPipelineWork(WorkContext *context, Functi
...
+    FeatureRecord(function, brain::ExecutionOperatingUnitType::AGGREGATE_BUILD,
+                  brain::ExecutionOperatingUnitFeatureAttribute::CARDINALITY, pipeline,
+                  codegen->CallBuiltin(ast::Builtin::AggHashTableGetTupleCount, {agg_ht.GetPtr(codegen)}));
+    FeatureArithmeticRecordMul(function, pipeline, GetTranslatorId(), CounterVal(num_agg_inputs_));
+  } else {
+    FeatureRecord(function, brain::ExecutionOperatingUnitType::AGGREGATE_ITERATE,
+                  brain::ExecutionOperatingUnitFeatureAttribute::NUM_ROWS, pipeline, CounterVal(num_agg_outputs_));
+    FeatureRecord(function, brain::ExecutionOperatingUnitType::AGGREGATE_ITERATE,
+                  brain::ExecutionOperatingUnitFeatureAttribute::CARDINALITY, pipeline,
+                  codegen->CallBuiltin(ast::Builtin::AggHashTableGetTupleCount, {agg_ht.GetPtr(codegen)}));
```
malin1993ml, line 479: Let's add the multiplication back here...

malin1993ml (changes requested):
I actually took my last review back... I just realized that the arithmetics on two sides of an operator would end up in two different pipelines. And since we record the features at a per-pipeline basis, the two multiplication actually should not affect each other at all, unless you put the two sides in a single pipeline (which I think won't happen in our engine). So we don't need to remove those multiplications and instead should just add the multiplication back to the `AGG_ITERATE` side. Please think about whether this makes sense.

src/execution/compiler/operator/hash_aggregation_translator.cpp
lmwnshn: Added it to FinishPipelineWork (the convention I'm running with now is for all recording to happen there by default).

Your reasoning makes sense, the translator ID is the same but the pipeline ID is not the same.

malin1993ml (approved):

malin1993ml:
Awesome! I think there're still tests failing? After the build passes I think we're ready to merge!

lmwnshn:
Interesting, I don't think I touched the index create test and I can't reproduce this locally. Going to merge master and see what happens.

src/execution/sema/sema_checking.cpp
```diff
@@ -226,6 +226,12 @@ Sema::CheckResult Sema::CheckComparisonOperands(parsing::Token::Type op, const S
     return {result_type, left, right};
   }
 
+  // If both input types are integers, we don't need to do any work.
```
lmwnshn, line 229: Hey @pmenon, could you quickly weigh in on this change?

left: MemberExpr, it is queryState.some_uint32
right: LitExpr, integer literal in particular, treated as int32

I'm not sure if:
- there is a cast for narrowing/widening integers
- there is a more fine-grained way of comparing types here than IsIntegerType which should be used

Thanks!

pmenon: There isn't a means to down- or up-cast primitive integers at the moment. For now, I think what you have is okay. But, put a `TODO` putting the blame on me and revisiting with a fix. Also, create an issue for it.

lmwnshn: #1174. Thanks!

malin1993ml (approved):
This is great! I've tested that this works on both TPCC and TPCH. Thanks for the awesome work, Wan!

Only thing is that there still seems to be a unit test failing. I think we're good to go once the tests pass.

src/execution/sema/sema_checking.cpp
```diff
@@ -226,6 +226,12 @@ Sema::CheckResult Sema::CheckComparisonOperands(parsing::Token::Type op, const S
...
+  if (left->GetType()->IsIntegerType() == right->GetType()->IsIntegerType()) {
```
lmwnshn: whoops

Issue #1110, opened 2020-08-23
Test Aggregation on All Possible Types 


dniu16:
Make a new trace file aggregates.test and have all possible aggregations (with and without GROUP BY) on all possible types.

Issue #1125, opened 2020-08-26
AVG(columnA) when columnA's type is INT is not working. 


githubxxcc:
The average aggregation is failing when working on Integer type columns.  

### Error output:
```
[2020-08-26 16:39:47.759] [execution_logger] [error] ERROR: Line: 13, Col: 0 => cannot assign type 'Integer' to type 'Real'
```

### Reproce error
```
DROP TABLE IF EXISTS accounts;
CREATE TABLE accounts (user_id INT PRIMARY KEY, email INT NOT NULL, phone INT NOT NULL);
INSERT INTO accounts (user_id, email, phone) VALUES (0, 99, 10);

# THIS WILL BREAK!!!
SELECT AVG(email) FROM accounts;
```

PR #1136, opened 2020-08-30
Aggregation test (#1110) and statement error code optional flag

dniu16:
The aggregation test here are all hand-written by myself. I tried to add the sqlite aggregation traces but failed. I modified FilterTrace and MogSqlite to handle the sqlite aggregation trace files, but it seems that the syntax is a lot different from Postgresql.

For example from sqlite aggregation:
SELECT AVG ( + col1 ) FROM tab2 AS cor0 WHERE NULL > NULL
SELECT DISTINCT - ( AVG ( 68 ) ) FROM tab1 WHERE ( + 98 + + col0 ) IS NULL

which causes I/O error for Postgresql.

Regarding the aggregation test I wrote myself there has been I/O errors with certain statements. 

1. AVG aggregation error on INTEGER. (SELECT AVG(col0) FROM region)

2. SELECT COUNT(col2) FROM region when col2 is of type varchar, float or timestamp

3. For column with INTEGER type, insert statement failed for first insert 1 then insert 1.0 whereas postgresql succeeded

4. SELECT MAX(col2)/MIN(col2) FROM region failed for col2 type timestamp, succeeded for postgresql

5. With table column types col0 bigint, col1 float, col2 timestamp, I/O error on "SELECT MAX(col0) FROM region GROUP BY col2", but succeeded for "SELECT MAX(col0) FROM region GROUP BY col0". Both statements pass for postgresql. Same condition applies to sum/avg/min.

There are also some hash value mismatch which I think are rounding errors

Regarding the statement error code optional flag, when a sql statement is executed, if a SQLException is thrown and statement error code is specified (e.g. Statement error 10000), the actual error code will be compared to 10000 and if the result don't match an Exception would be thrown.


turingcompl33t:
Tracking 1. from your list of bugs with issue [#1125](https://github.com/cmu-db/terrier/issues/1125) and I'm working on it currently.

Issue #804, opened 2020-03-03
Concat Operator (||) Does Not Work


apavlo:
We are not able to use the concat operator `||` in queries.

## Client
```
terrier=# CREATE TABLE xxx (id INT);
CREATE TABLE
terrier=# INSERT INTO xxx VALUES (1),(2),(3);
INSERT 0 3
terrier=# SELECT 'A' || 'P' FROM xxx;
server closed the connection unexpectedly
        This probably means the server terminated abnormally
        before or while processing the request.
The connection to the server was lost. Attempting reset: Failed.
```

## Server
```
[2020-03-03 13:09:04.528] [execution_logger] [error] Type-checking error! 
 Line: 0, Col: 0 => cannot assign type 'Integer' to type 'StringVal'

[2020-03-03 13:09:04.528] [execution_logger] [error] Dumping AST:
[2020-03-03 13:09:04.530] [execution_logger] [error]  File (0x62900006bbb0) <line:0:0> 
|- StructDecl (0x629000069ad8) <line:0:0> State
| `- FieldDecl (0x629000069a68) <line:0:0> DUMMY
|   `- IdentifierExpr (0x629000069a38) <line:0:0>  'int32' int32
|- StructDecl (0x6290000699f0) <line:0:0> Output4
| `- FieldDecl (0x629000069980) <line:0:0> col0
|   `- IdentifierExpr (0x629000069950) <line:0:0>  'StringVal' StringVal
|- FunctionDecl (0x629000069d20) <line:0:0> setupFn
| `- FunctionLitExpr (0x629000069cf0) <line:0:0>  '(*struct{int32},*ExecutionContext)->nil' 
|   `- BlockStmt (0x629000069c60) <line:0:0> 
|     `- ReturnStmt (0x62900006bd78) <line:0:0> 
|- FunctionDecl (0x629000069f78) <line:0:0> teardownFn
| `- FunctionLitExpr (0x629000069f48) <line:0:0>  '(*struct{int32},*ExecutionContext)->nil' 
|   `- BlockStmt (0x629000069eb8) <line:0:0> 
|     `- ReturnStmt (0x62900006bdd0) <line:0:0> 
|- FunctionDecl (0x62900006b490) <line:0:0> pipeline0
| `- FunctionLitExpr (0x62900006b460) <line:0:0>  '(*struct{int32},*ExecutionContext)->nil' 
|   `- BlockStmt (0x62900006a110) <line:0:0> 
[...]

PR #839, opened 2020-03-22
Fixed Concat operator handling.

VenkatDatta:
**Added handling for Concat (||) operator.**

Bug: #804 

Test:
```
terrier=# CREATE TABLE xxx (id VARCHAR);
CREATE TABLE
terrier=# INSERT INTO xxx VALUES ('One'), ('Two');
INSERT 0 2
terrier=# SELECT 'Id: ' || id  from xxx;
 OPERATOR_CONCAT VARCHAR OPERATOR_CONCAT id 
--------------------------------------------
 Id: One
 Id: Two
(2 rows)

terrier=# SELECT 'A' || 'P'  as test from xxx;
 test 
------
 AP
 AP
(2 rows)

```

PR #1137, opened 2020-08-30
Fix concat operator

jkosh44:
The sql parser was correctly parsing "||" as CONCAT but there was
no tpl parsing, tpl ast generation, semantic checking, or
bytecode generation for concat. This commit adds tpl ast
generation, semantic checking, and bytecode generation. tpl
parsing still isn't working properly. We reuse the concat
built-in bytecode handling as the underlying implementation.

Originally CONCAT was being treated as a normal BinaryOpExpr,
however CONCAT needs access to the Execution Context in order
to allocate space for a new string. Normal BinaryOpExpr don't
have a reference to the execution context. Therefore we needed
to create a new type of expression, StringBinaryOpExpr, which
has a reference to the execution context.

StringBinaryOpExpr was made to be generalized for use with  any
string binary operation. Some other options would have been:
- Add a reference to the Execution Context to BinaryOpExpr.
However this is probably wasteful since the majority of binary
operations do not need a reference to the Execution Context.
- Create a specific operation for CONCAT.
- Generalize further and make some StringOpExpr, that instead
of having a left and right expression, has a vector of
expressions. This would allow it to be reused on string
operations that need a reference to the Execution Context and
have any amount of arguments.

Many elements were inspired by/taken out of the following PR:
https://github.com/cmu-db/terrier/pull/839/files

Fixes: #804

Co-authored-by: VenkatDatta <nhvenkatdatta@gmail.com>

### DIFF
--- a/src/execution/ast/ast_dump.cpp
+++ b/src/execution/ast/ast_dump.cpp
@@ -275,6 +275,13 @@ void AstDumperImpl::VisitBinaryOpExpr(BinaryOpExpr *node) {
   DumpExpr(node->Right());
 }
 
+void AstDumperImpl::VisitStringBinaryOpExpr(StringBinaryOpExpr *node) {
+  DumpExpressionCommon(node);
+  DumpToken(node->Op());
+  DumpExpr(node->Left());
+  DumpExpr(node->Right());
+}
+
 void AstDumperImpl::VisitComparisonOpExpr(ComparisonOpExpr *node) {
   DumpExpressionCommon(node);
   DumpToken(node->Op());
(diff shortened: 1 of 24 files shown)


jkosh44:
This ended up being a lot more code than I thought it'd be. I started off by trying to have StringBinaryOpExpr extend BinaryOpExpr, but that ended up being a little messy and required a bunch of casting. 

I wasn't able to figure out how to get a reference to the Execution Context during tpl parsing so if someone can help me with that it would be greatly appreciated.

I listed a couple of alternate approaches in the commit message. I made StringBinaryOpExpr to try and generalize all binary string operators, but I can't actually find any others besides concat. After finishing I think the general StringOpExpr would probably be a better approach, but before switching it all out I wanted a second opinion.

pmenon:
Remove support for `||` and just handle `CONCAT(...)`. Then it becomes like any other builtin you've added before.

To support `||` we need `pg_catalog.pg_operator` so we can map `||` to the function `concat()`.

jkosh44:
Yeah, that's a much better approach actually. I'm just gonna close this PR then and open a new one for the built in Concat.

PR #1139, opened 2020-08-30
Add support for variadic functions in the catalog and builtin CONCAT function

jkosh44:


src/catalog/database_catalog.cpp
```diff
@@ -2193,6 +2198,12 @@ void DatabaseCatalog::BootstrapProcContexts(const common::ManagedPointer<transac
   SetProcCtxPtr(txn, postgres::TRIM2_PRO_OID, func_context);
   txn->RegisterAbortAction([=]() { delete func_context; });
 
+  func_context = new execution::functions::FunctionContext("concat", type::TypeId::VARCHAR,
+                                                           {type::TypeId::VARCHAR, type::TypeId::VARCHAR},
+                                                           execution::ast::Builtin::Concat, true);
+  SetProcCtxPtr(txn, postgres::CONCAT_PRO_OID, func_context);
+  txn->RegisterAbortAction([=]() { delete func_context; });
```
mbutrovich, line 2205: I'm surprised we do so much registering of abort actions to cleanup pointers when other parts of catalog bootstrapping asserts that we can't fail. Do we envision this bootstrap txn ever failing, @tanujnay112?

pmenon (changes requested):
So, CONCAT is actually a variable argument function. Is this something the catalog supports? If so, can you implement that version? The TPL API becomes: `@concat(str1, str2, str2, ...)` and the C++ API is something like `StringFunctions::Concat(StringVal *result, ExecutionContext *ctx, StringVal **inputs, uint32_t num_inputs)`. It should be pretty simple to implement.

jkosh44:
> So, CONCAT is actually a variable argument function. Is this something the catalog supports? If so, can you implement that version? The TPL API becomes: `@concat(str1, str2, str2, ...)` and the C++ API is something like `StringFunctions::Concat(StringVal *result, ExecutionContext *ctx, StringVal **inputs, uint32_t num_inputs)`. It should be pretty simple to implement.

I'll take a look at this and see

src/catalog/database_catalog.cpp
jkosh44: > I'm surprised we do so much registering of abort actions to cleanup pointers when other parts of catalog bootstrapping asserts that we can't fail. Do we envision this bootstrap txn ever failing, @tanujnay112?

I'm not really sure why all the abort actions are there. I tested it and it does work fine without them. Would the DB still even be able to start up if this txn aborted?

jkosh44:
> > So, CONCAT is actually a variable argument function. Is this something the catalog supports? If so, can you implement that version? The TPL API becomes: `@concat(str1, str2, str2, ...)` and the C++ API is something like `StringFunctions::Concat(StringVal *result, ExecutionContext *ctx, StringVal **inputs, uint32_t num_inputs)`. It should be pretty simple to implement.
> 
> I'll take a look at this and see

@pmenon So I went to look in the catalog and it doesn't look like we support functions with variable argument lenght
https://github.com/cmu-db/terrier/blob/dabbbef03678288cd4a6b4a34284055dbaa3f0c1/src/include/catalog/database_catalog.h#L309-L311
https://github.com/cmu-db/terrier/blob/dabbbef03678288cd4a6b4a34284055dbaa3f0c1/src/include/catalog/database_catalog.h#L333-L335

I think this is probably blocked until that functionality gets added. It doesn't seem like something trivial to add either.

jkosh44:
I'm closing this for now to avoid Jenkins builds

src/catalog/postgres/builder.cpp
```diff
@@ -765,12 +765,8 @@ IndexSchema Builder::GetProcNameIndexSchema(db_oid_t db) {
...
   columns.back().SetOid(indexkeycol_oid_t(2));
 
-  columns.emplace_back("proallargs", type::TypeId::VARBINARY, MAX_NAME_LENGTH, false,
-                       parser::ColumnValueExpression(db, PRO_TABLE_OID, PROALLARGTYPES_COL_OID));
-  columns.back().SetOid(indexkeycol_oid_t(3));
-
-  // Unique, not primary
-  IndexSchema schema(columns, storage::index::IndexType::BWTREE, true, false, false, true);
+  // Non-Unique, not primary
+  IndexSchema schema(columns, storage::index::IndexType::BWTREE, false, false, false, false);
```
jkosh44, line 769: I'm honestly not super confident on these boolean parameters, if someone could double check them.

src/catalog/database_catalog.cpp
```diff
@@ -2980,22 +2986,33 @@ proc_oid_t DatabaseCatalog::GetProcOid(common::ManagedPointer<transaction::Trans
...
+    const std::vector<type_oid_t> variadic = {GetTypeOidForType(type::TypeId::VARIADIC)};
+    auto variadic_varlen = storage::StorageUtil::CreateVarlen(variadic);
+
+    // Search through results and see if any match the parsed function by argument types
+    for (auto &tuple : results) {
+      auto table_pr = pg_proc_all_cols_pri_.InitializeRow(buffer);
+      bool UNUSED_ATTRIBUTE visible = procs_->Select(txn, tuple, table_pr);
+      storage::VarlenEntry index_all_arg_types = *reinterpret_cast<storage::VarlenEntry *>(
+          table_pr->AccessForceNotNull(pg_proc_all_cols_prm_[postgres::PROALLARGTYPES_COL_OID]));
+      // variadic functions will match any argument types
```
jkosh44, line 3005: Is this accurate, or do we need to also check that all arguments have the same type?

src/catalog/database_catalog.cpp
tanujnay112: At that moment, I just put this in because every other Bootstrapping function was doing so as well. But yeah now that I think about it we shouldn't have the bootstrapping transaction fail. Otherwise, the whole system pretty much does not proceed from there right.

src/include/execution/vm/bytecode_handlers.h
```diff
@@ -1490,9 +1490,8 @@ VM_OP_WARM void OpCharLength(terrier::execution::sql::Integer *result, terrier::
 }
 
 VM_OP_WARM void OpConcat(terrier::execution::sql::StringVal *result, terrier::execution::exec::ExecutionContext *ctx,
-                         const terrier::execution::sql::StringVal *left,
-                         const terrier::execution::sql::StringVal *right) {
-  terrier::execution::sql::StringFunctions::Concat(result, ctx, *left, *right);
+                         const int64_t *num_inputs, const terrier::execution::sql::StringVal **inputs) {
```
tanujnay112, line 1493: Other functions in this file that pass in arrays and such don't pass in the length of the array or in this case number of inputs as a pointer but the actual value. Is there a specific reason this is a pointer by any chance?

src/include/execution/sql/functions/string_functions.h
```diff
@@ -29,8 +29,9 @@ class StringFunctions {
     Length(result, ctx, str);
   }
 
-  /** Compute CONCAT(left, right). */
-  static void Concat(StringVal *result, exec::ExecutionContext *ctx, const StringVal &left, const StringVal &right);
+  /** Compute CONCAT(str1, str2, str3, ...). */
+  static void Concat(StringVal *result, exec::ExecutionContext *ctx, const int64_t &num_inputs,
```
tanujnay112, line 33: Don't use references on primitive types.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -3484,6 +3496,20 @@ LocalVar BytecodeGenerator::NewStaticString(ast::Context *ctx, const ast::Identi
...
 
+LocalVar BytecodeGenerator::CreateArgumentArray(const util::RegionVector<execution::ast::Expr *> &arguments,
+                                                const uint32_t start, const uint32_t end, execution::ast::Type *type) {
+  auto num_args = end - start;
+  const auto array_type = ast::ArrayType::Get(num_args, type->PointerTo());
+  LocalVar args = GetCurrentFunction()->NewLocal(array_type);
+  auto arr_elem_ptr = GetCurrentFunction()->NewLocal(type->PointerTo()->PointerTo());
+  for (uint32_t i = 0; i < num_args; i++) {
+    GetEmitter()->EmitLea(arr_elem_ptr, args, i * 8);
+    LocalVar input_string = VisitExpressionForLValue(arguments[i + start]);
```
tanujnay112, line 3507: It seems like you're using this for the right-hand side of an assignment. Use VisitExpressionForRValue instead as it looks like this being treated like an R-value.

tanujnay112 (changes requested):

src/catalog/database_catalog.cpp
jkosh44: If that's the case I can remove this line from the PR, and I can open a separate PR to remove it from the other functions

src/include/execution/vm/bytecode_handlers.h
jkosh44: No good reason, I'll de-reference it in vm.cpp so it's not a pointer here.

src/execution/vm/bytecode_generator.cpp
jkosh44: So for this I'm a little confused tbh. @pmenon can you weigh in here? I thought that these were supposed to be LValues and we were mistakenly using RValues. @pmenon recently opened this issue: https://github.com/cmu-db/noisepage/issues/1167

src/execution/sema/sema_builtin.cpp
```diff
@@ -2586,23 +2586,20 @@ void Sema::CheckBuiltinStringCall(ast::CallExpr *call, ast::Builtin builtin) {
...
 
-      // checking to see if the third argument is a string
-      if (!call->Arguments()[2]->GetType()->IsSpecificBuiltin(ast::BuiltinType::StringVal)) {
-        ReportIncorrectCallArg(call, 2, ast::StringType::Get(GetContext()));
-        return;
+      // checking to see if the arguments are strings
+      for (uint32_t i = 1; i < call->NumArgs(); i++) {
+        auto *resolved_type = call->Arguments()[i]->GetType();
+        if (!resolved_type->IsSpecificBuiltin(ast::BuiltinType::StringVal)) {
+          ReportIncorrectCallArg(call, 1, ast::StringType::Get(GetContext()));
```
pmenon, line 2600: Fix index for error message.

src/execution/sql/functions/string_functions.cpp
```diff
@@ -9,20 +9,27 @@
 
 namespace terrier::execution::sql {
 
-void StringFunctions::Concat(StringVal *result, exec::ExecutionContext *ctx, const StringVal &left,
-                             const StringVal &right) {
-  if (left.is_null_ || right.is_null_) {
+void StringFunctions::Concat(StringVal *result, exec::ExecutionContext *ctx, const int64_t num_inputs,
+                             const StringVal *inputs) {
```
pmenon, line 13: The `inputs` array needs to store *pointers* to string values to avoid copying each string value into the array. So:

```
Concat(StringVal *result, ExecutionContext *ctx, const StringVal *inputs[], uint32_t num_inputs)
```

src/execution/sql/functions/string_functions.cpp
```diff
@@ -9,20 +9,27 @@
...
+  if (num_inputs == 0) {
```
pmenon, line 14: Technically, this shouldn't be possible. It will be caught earlier during planning or even in code-gen. I'd change this to an assert ensuring at least one input.

src/execution/sql/functions/string_functions.cpp
```diff
@@ -9,20 +9,27 @@
...
     *result = StringVal::Null();
     return;
   }
 
-  const std::size_t length = left.GetLength() + right.GetLength();
+  std::size_t length = 0;
+  for (int64_t i = 0; i < num_inputs; i++) {
+    length += inputs[i].GetLength();
```
pmenon, line 21: You need to only count (and copy) non-NULL inputs.

src/execution/sql/functions/string_functions.cpp
```diff
@@ -9,20 +9,27 @@
...
+      std::memcpy(ptr + offset, inputs[i].GetContent(), inputs[i].GetLength());
+      offset += inputs[i].GetLength();
+    }
+  }
 
-  // Copy contents into result.
-  std::memcpy(ptr, left.GetContent(), left.GetLength());
-  std::memcpy(ptr + left.GetLength(), right.GetContent(), right.GetLength());
-  *result = StringVal(ptr, length);
+  *result = length == 0 ? StringVal::Null() : StringVal(ptr, length);
```
pmenon, line 32: I don't think `CONCAT` ever returns NULL.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -2350,6 +2350,17 @@ void BytecodeGenerator::VisitBuiltinStringCall(ast::CallExpr *call, ast::Builtin
       }
       break;
     }
+    case ast::Builtin::Concat: {
+      const auto num_inputs = call->NumArgs() - 1;
+      LocalVar num_inputs_var =
+          GetCurrentFunction()->NewLocal(ast::BuiltinType::Get(call->GetType()->GetContext(), ast::BuiltinType::Int64));
+      GetEmitter()->EmitAssignImm8(num_inputs_var, static_cast<int64_t>(num_inputs));
```
pmenon, line 2357: The length is a compile-time constant, directly emit the value in the bytecode rather than assigning to a temporary. Also, using an 8-byte count may be overkill?

src/execution/vm/bytecode_generator.cpp
pmenon: SQL values should use L-values. But, this function shouldn't make assumptions on how to derive child expressions. Either (1) add a boolean indicating whether to use L- or R-values, or (2) pull this logic into `CONCAT` generation. I'd prefer the latter until we have enough use-cases for this to make refactor into a common function.

src/execution/vm/vm.cpp
```diff
@@ -2112,9 +2112,9 @@ void VM::Interpret(const uint8_t *ip, Frame *frame) {  // NOLINT
   OP(Concat) : {
     auto *result = frame->LocalAt<sql::StringVal *>(READ_LOCAL_ID());
     auto *exec_ctx = frame->LocalAt<exec::ExecutionContext *>(READ_LOCAL_ID());
-    auto *left = frame->LocalAt<const sql::StringVal *>(READ_LOCAL_ID());
-    auto *right = frame->LocalAt<const sql::StringVal *>(READ_LOCAL_ID());
-    OpConcat(result, exec_ctx, left, right);
+    auto *num_inputs = frame->LocalAt<const int64_t *>(READ_LOCAL_ID());
+    auto **inputs = frame->LocalAt<const sql::StringVal **>(READ_LOCAL_ID());
+    OpConcat(result, exec_ctx, *num_inputs, *inputs);
```
pmenon, line 2117: Change to (1) use an immediate for the input count, and (2) use a proper array of pointers for the inputs.

src/include/execution/sql/functions/string_functions.h
```diff
@@ -29,8 +29,8 @@ class StringFunctions {
     Length(result, ctx, str);
   }
 
-  /** Compute CONCAT(left, right). */
-  static void Concat(StringVal *result, exec::ExecutionContext *ctx, const StringVal &left, const StringVal &right);
+  /** Compute CONCAT(str1, str2, str3, ...). */
+  static void Concat(StringVal *result, exec::ExecutionContext *ctx, const int64_t num_inputs, const StringVal *inputs);
```
pmenon, line 33: Ditto, change signature.

src/include/type/type_id.h
```diff
@@ -30,7 +30,8 @@ enum class TypeId : uint8_t {
   DATE,
   VARCHAR,
   VARBINARY,
-  PARAMETER_OFFSET
+  PARAMETER_OFFSET,
+  VARIADIC,
```
pmenon, line 34: This isn't a real `Type`, but neither is `PARAMETER_OFFSET`. Will need to refactor later.

pmenon (changes requested):
I think you'll need to add additional type-checking on these variadics to ensure someone doesn't pass non-strings into `CONCAT`.

If it's a significant amount of work, I'd prefer creating an issue and punt for later.

src/include/type/type_id.h
jkosh44: So the actual type from postgres is "any" (https://github.com/cmu-db/noisepage/issues/1176). We can just rename this to that?

src/execution/vm/bytecode_generator.cpp
jkosh44: When you say it's a compile-time constant you're talking about when compiling the tpl, not when compiling the C++ right? How would I emit the number directly in the bytecode?

jkosh44:
> I think you'll need to add additional type-checking on these variadics to ensure someone doesn't pass non-strings into `CONCAT`.
> 
> If it's a significant amount of work, I'd prefer creating an issue and punt for later.

I'm not really sure what you mean by this to be honest. I thought that `sema_builtin` was doing exactly this?

src/include/execution/sql/functions/string_functions.h
jkosh44: Updated

src/include/execution/vm/bytecode_handlers.h
jkosh44: Updated

src/execution/sema/sema_builtin.cpp
jkosh44: Updated

src/execution/sql/functions/string_functions.cpp
jkosh44: updated

src/execution/sql/functions/string_functions.cpp
jkosh44: Updated

src/execution/sql/functions/string_functions.cpp
jkosh44: Updated

src/execution/sql/functions/string_functions.cpp
jkosh44: Updated

src/execution/vm/bytecode_generator.cpp
jkosh44: I pulled it into CONCAT generation, and kept it as L-Value (which I think is what you were saying).

src/execution/vm/vm.cpp
jkosh44: Updated (2), waiting on input for (1) from other question.

src/include/execution/sql/functions/string_functions.h
jkosh44: updated

cmake_modules/FindLibev.cmake
```diff
@@ -0,0 +1,45 @@
+# - Try to find Libevent
```
pmenon, line 1: Accidental include?

src/execution/sql/functions/string_functions.cpp
```diff
@@ -9,19 +9,25 @@
...
+  std::size_t length = 0;
+  for (int64_t i = 0; i < num_inputs; i++) {
+    if (!inputs[i]->is_null_) {
+      length += inputs[i]->GetLength();
+    }
   }
 
-  const std::size_t length = left.GetLength() + right.GetLength();
   char *const ptr = ctx->GetStringAllocator()->PreAllocate(length);
+  for (int64_t i = 0, offset = 0; i < num_inputs; i++) {
```
pmenon, line 24: Technically, `offset` should be a `std::size_t`. I'd also go ahead and change the type on line 17 for cosistency.

src/execution/vm/bytecode_generator.cpp
pmenon: > When you say it's a compile-time constant you're talking about when compiling the tpl, not when compiling the C++ right?

Correct.

> How would I emit the number directly in the bytecode?

Take a look at how `EmitDerefN()` works. You should create a new function `BytecodeEmitter::EmitConcat()` that takes in the argument count as a `uint32_t`.

src/include/execution/sql/functions/string_functions.h
```diff
@@ -29,8 +29,9 @@ class StringFunctions {
     Length(result, ctx, str);
   }
 
-  /** Compute CONCAT(left, right). */
-  static void Concat(StringVal *result, exec::ExecutionContext *ctx, const StringVal &left, const StringVal &right);
+  /** Compute CONCAT(str1, str2, str3, ...). */
+  static void Concat(StringVal *result, exec::ExecutionContext *ctx, const int64_t num_inputs,
```
pmenon, line 33: Should make the input count unsigned, probably a `uint32_t`.

src/include/type/type_id.h
pmenon: I'd say leave as it. This'll need to be refactored later outside this PR.

pmenon (changes requested):
Small changes, then it's probably good to go.

cmake_modules/FindLibev.cmake
jkosh44: oops yeah

jkosh44: removed

src/execution/sql/functions/string_functions.cpp
jkosh44: Fixed

src/execution/vm/bytecode_generator.cpp
jkosh44: Updated

src/include/execution/sql/functions/string_functions.h
jkosh44: Fixed

jkosh44:
@pmenon I believe I've addressed all your comments except for 

> I think you'll need to add additional type-checking on these variadics to ensure someone doesn't pass non-strings into `CONCAT`.
> 
> If it's a significant amount of work, I'd prefer creating an issue and punt for later.

I'm still not really sure what this entails.

src/execution/vm/vm.cpp
```diff
@@ -2112,9 +2112,9 @@ void VM::Interpret(const uint8_t *ip, Frame *frame) {  // NOLINT
   OP(Concat) : {
     auto *result = frame->LocalAt<sql::StringVal *>(READ_LOCAL_ID());
     auto *exec_ctx = frame->LocalAt<exec::ExecutionContext *>(READ_LOCAL_ID());
-    auto *left = frame->LocalAt<const sql::StringVal *>(READ_LOCAL_ID());
-    auto *right = frame->LocalAt<const sql::StringVal *>(READ_LOCAL_ID());
-    OpConcat(result, exec_ctx, left, right);
+    auto **inputs = frame->LocalAt<const sql::StringVal **>(READ_LOCAL_ID());
+    auto num_inputs = READ_UIMM4();
+    OpConcat(result, exec_ctx, num_inputs, inputs);
```
pmenon, line 2117: Nit: try to be consistent with parameter ordering between bytecode and C++ functions.

pmenon (changes requested):
You also need to modify the `Concat` bytecode encoding in `bytecodes.h` [here](https://github.com/cmu-db/noisepage/blob/master/src/include/execution/vm/bytecodes.h#L595) so that `num_inputs` is an immediate unsigned 4-byte integer (e.g., like the last operand to `CSVReaderInit` [here](https://github.com/cmu-db/noisepage/blob/master/src/include/execution/vm/bytecodes.h#L493))

pmenon:
Don't worry about variadic function type-check for now. Ideally, this checking should be done during function lookup in binding. Having a check during `TPL` semantic analysis is good for non-SQL use-cases (i.e., UDFs), but moving as much checking earlier in the pipeline is generally the correct approach.

src/execution/vm/vm.cpp
jkosh44: Updated

jkosh44:
> You also need to modify the `CONCAT` bytecode encoding in `bytecodes.h` [here](https://github.com/cmu-db/noisepage/blob/master/src/include/execution/vm/bytecodes.h#L595) so that `num_inputs` is an immediate unsigned 4-byte integer (e.g., like the last operand to `CSVReaderInit [here](https://github.com/cmu-db/noisepage/blob/master/src/include/execution/vm/bytecodes.h#L493))

Updated

pmenon (approved):
LGTM

src/catalog/postgres/builder.cpp
mbutrovich: Those look reasonable.

mbutrovich:
> Don't worry about variadic function type-check for now. Ideally, this checking should be done during function lookup in binding. Having a check during `TPL` semantic analysis is good for non-SQL use-cases (i.e., UDFs), but moving as much checking earlier in the pipeline is generally the correct approach.

Let's open an issue for this once this gets merged.

lmwnshn (approved):
Just echoing the LGTM.

PR #1140, opened 2020-08-31
Add Asin/Acos out of range error handling

### LINKED ISSUE #1098: Input validation on acos and asin will crash the DB
If the input of acos or asin is not in between -1 and 1, then a runtime exception is thrown. This will cause the DB to crash.

https://github.com/cmu-db/terrier/blob/cdb728710c4c26a7a593a0c9da36c4cef2e823ca/src/include/execution/sql/operators/numeric_operators.h#L42

https://github.com/cmu-db/terrier/blob/cdb728710c4c26a7a593a0c9da36c4cef2e823ca/src/include/execution/sql/operators/numeric_operators.h#L54

jkosh44:
This commit adds an exception type for arithmetic exceptions. The
traffic cop now catches this exception and returns an appropriate
error message.

Asin and Acos will now raise this exception when given an invalid
input.

Fixes #1098


jkosh44:
This is the same error code used by postgress for invalid input on acos and asin

https://github.com/postgres/postgres/blob/53095b5fe650270118bc2ab77416d08e19472cd3/src/backend/utils/adt/float.c#L1768-L1776

mbutrovich:
Do we want a different exception type for classes of errors? I envisioned exception types maybe per-component. What does a separate exception type here buy us over just an ExecutionException with the properly defined code and message?

jkosh44:
> Do we want a different exception type for classes of errors? I envisioned exception types maybe per-component. What does a separate exception type here buy us over just an ExecutionException with the properly defined code and message?
> 
> It doesn't seem like we need the traffic cop to know how to handle every possible exception in the system.

I think using ExecutionException is probably fine, and I can test it out. I was just nervous because it was already thrown from a bunch of places I didn't know what implications it would have to catch it and return an error in the traffic cop.

jkosh44:
Update: after discussing with @mbutrovich the new approach is to reuse `ExecutionException`. Before we can do that, we first need to add an error code to `ExecutionException`. Currently `ExecutionException` seems to only be thrown in the case of software bugs and it crashes the DB. A consequence of that change would be that `ExecutionException` will no longer crash the DB and instead just abort the current transaction and send an error to the user.

jkosh44:
Pending #1146

jkosh44:
Who ever merges this, can they please change the commit message from
"This commit adds an exception type for arithmetic exceptions. The
traffic cop now catches this exception and returns an appropriate
error message.

Asin and Acos will now raise this exception when given an invalid
input.

Fixes #1098"

to 
"Asin and Acos now raise an ExecutionException when given an 
invalid input, so that it is handled properly.

Fixes #1098"

mbutrovich (approved):
LGTM

PR #1143, opened 2020-09-01
Substitute GC with multi-threaded Deferred Action Framework

mush-zhang:
This PR modified the existing deferred action framework (DAF) and garbage collector, handing the functionalities of garbage collection to DAF. Now DAF takes care of memory deallocation, unlinking version chains, transaction context deletion, and index GC invocation.

Changes:
- DAF: Process deferred actions and index GC. Use concurrent queue for the centralized deferred action queue to alleviate contention.
- GC: remove all functionalities of GC. The PerformGarbageCollection() remains as an interface to invoke the actual deferred action processing and memory management in DAF.
- DataTable: handles truncate version chain
- TxnContext: handles unlinking
- TxnManager: Upon commit/abort, instead of holding latch and moving commit/abort actions to a centralized queue commit/abort actions are invoked with DAF as a parameter. Also, register deferred action to clean up the transaction context upon commit/abort. As current DAF does not have any information on the contents of the actions processed, we keep two atomic counters in the transaction manager to fake the number of unlinked txns and the number of deleted txns processed in each DAF invocation.
- GC Metric: Record the number of deferred actions and txns processed over time.
- TPCC Benchmark: Add benchmark case for running with DAF metrics with and without logging enabled respectively.
- Benchmark Config, Benchmark util, mini runner: Configure number of DAF threads used in env variables.
- Various signature changes.

Note:
Currently faking the API (return type) for GC so that other parts of the system do not need to be changed too much. GC can be removed completely in the future.


apavlo:
Closing. @mush-zhang will submit a new PR

PR #1144, opened 2020-09-01
Fix AVG Aggregations for INT Column Types.

### LINKED ISSUE #1125: AVG(columnA) when columnA's type is INT is not working.

turingcompl33t:
Closes #1125. 

The TPL code generated for aggregations always resembles the following:

```
// helper struct
struct AggValues {
    agg_term_attr0: Real
}

// within pipeline serial work
for (@tableIterAdvance(tvi)) {
    var vpi = @tableIterGetVPI(tvi)
    for (; @vpiHasNext(vpi); @vpiAdvance(vpi)) {
        slot = @vpiGetSlot(vpi)
        var aggValues: AggValues
        aggValues.agg_term_attr0 = @vpiGetInt(vpi, 0)
        @aggAdvance(&queryState.aggs.agg_term_attr0, &aggValues.agg_term_attr0)
    }
}
```

Previously, we always failed semantic analysis on the assignment `aggValues.agg_term_attr0 = @vpiGetInt(vpi, 0)` because the conversion `Integer` to `Real` is not one of the ones we have implemented.

This PR addresses this issue not by defining this conversion, but rather by updating the logic in `static_aggregation_translator.cpp` to select the correct type for temporary aggregation terms in the `AggValues` helper struct. For AVG, the type of the column over which the aggregation is computed is used. For all other aggregations, the return type of the aggregation itself is used (the same logic that existed previously).


src/execution/compiler/operator/static_aggregation_translator.cpp
```diff
@@ -161,6 +160,19 @@ void StaticAggregationTranslator::UpdateGlobalAggregate(WorkContext *ctx, Functi
...
 }
 
+ast::Expr *StaticAggregationTranslator::GetValuesStructTypeForTerm(
+    const common::ManagedPointer<parser::AggregateExpression> &term) {
+  auto codegen = GetCodeGen();
+  if (term->GetExpressionType() == parser::ExpressionType::AGGREGATE_AVG) {
+    // No guarantee that we have derived the return type of the child
+    // expression at this point in query processing, so we need to do it here.
+    TERRIER_ASSERT(term->GetChildrenSize() >= 1, "No column name given.");
+    const_cast<parser::AbstractExpression *>(term->GetChild(0).Get())->DeriveReturnValueType();
```
lmwnshn, line 170: I would actually be quite concerned if we need to derive return value type all the way here, this should be the binder and/or optimizer's job. I would prefer an assert saying that the type isn't INVALID.

src/include/execution/compiler/operator/static_aggregation_translator.h
```diff
@@ -5,6 +5,7 @@
 #include "execution/compiler/operator/operator_translator.h"
 #include "execution/compiler/pipeline.h"
 #include "execution/compiler/pipeline_driver.h"
+#include "planner/plannodes/aggregate_plan_node.h"
```
lmwnshn, line 8: Could you forward declare AggregateExpression instead? This helps a lot with compile times, because including the plan nodes tends to bring in a ton of code. (If you're interested, I have a script for counting the effects of an include.)

src/execution/compiler/operator/static_aggregation_translator.cpp
turingcompl33t: Good point. I neglected to consider that we necessarily `Visit()` all of the children of the aggregate expression in the plan before we get to this step in processing. I wonder if that means we can remove a similar invocation in the implementation of `DeriveReturnValueType()` for `AggregateExpression` because we can make a similar assumption? I'll make the update.

src/include/execution/compiler/operator/static_aggregation_translator.h
turingcompl33t: Ah, rookie mistake. Thanks for the catch!

src/execution/compiler/operator/static_aggregation_translator.cpp
```diff
@@ -161,6 +160,19 @@ void StaticAggregationTranslator::UpdateGlobalAggregate(WorkContext *ctx, Functi
   }
 }
 
+ast::Expr *StaticAggregationTranslator::GetValuesStructTypeForTerm(
+    const common::ManagedPointer<parser::AggregateExpression> &term) {
+  auto codegen = GetCodeGen();
+  if (term->GetExpressionType() == parser::ExpressionType::AGGREGATE_AVG) {
```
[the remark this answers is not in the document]
githubxxcc: Like the correct way of doint this should be: https://github.com/cmu-db/terrier/blob/efcb7e6311c784500cba3b9baa8d2c9fdf11f503/src/execution/compiler/operator/hash_aggregation_translator.cpp#L110-L113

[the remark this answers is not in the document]
turingcompl33t: Great point! Should have looked at how the other aggregation was dealing with this issue from the start... I'll update the PR.

turingcompl33t:
@lmwnshn Adding the in-progress label for now because I want to look into writing some SQL tests for the aggregations.

lmwnshn:
@turingcompl33t I was under the impression that we would use the .test trace stuff instead of .java files

turingcompl33t:
@lmwnshn You're right, definitely the more sustainable solution. I'll generate a trace for these queries instead.

githubxxcc:
But I believe the current sql file traces won't allow you to specify the exact aggregation results, only the number of rows, and query status.

lmwnshn:
No, you should be able to obtain the actual value. I think we've asked @dniu16 to disable the hashing thing, but if he hasn't, you can specify a hash of the resutls.

turingcompl33t:
Yes and the hash will work, but it is going to take some additional updates because while our output for various aggregations is correct, it doesn't match the format of the Postgres output exactly e.g. the column names will need to be labeled properly (instead of just `?column?`) for the hash to match. But I can look at that today.

turingcompl33t:
So tracefile testing for aggregations is proving to be an issue, but not for the reason I cited above (it looks like the column headers are not included as input to the hash, so we're safe there). The problem is the precision with which we report resulkts for various output column types. For instance, in Postgres an AVG over an INT column produces:

```
pg# SELECT AVG(count) FROM test;
        avg
--------------------
 1.5000000000000000
```

Whereas we produce:

```
terrier# SELECT AVG(count) FROM test;
 ?column?
----------
 1.500000
```

and it's easy to see why: in `postgres_packet_writer.cpp` we just extract the underlying `double` from the output value and call `std::to_string()` on it, then write this to the packet. So originally I just thought we'd need to modify the logic for writing `DECIMAL` outputs, easy enough. But while testing this I came on another issue where Postgres is doing something more intelligent than we are when determining the output column type. For instance in a MAX over `DECIMAL` columns, Postgres is happy to truncate all of those unnecessary 0s from the output:

```
pg# SELECT MAX(count) FROM test;
 max
-----
   2
```

Whereas we, because the column type over which the aggregate is computed is `DECIMAL`, end up writing all the 0s the same way we did for AVG:

```
terrier=# SELECT MAX(count) FROM test;
 ?column?
----------
 2.000000
```

This makes sense because for MIN, MAX, and SUM we derive the output type directly from the column over which the aggregation is computed. I'm not sure how Postgres is getting away with this, or even if it is preferred; it seems like an inconsistent way to handle results.

So my thinking is that we push the tracefile test for aggregations off for now and revisit this issue in the future because it is not a major correctness or performance concern, whereas we have other outstanding bugs that are. Curious as to what your thoughts on this are @lmwnshn and @githubxxcc. Let me know if I'm overlooking something obvious.

turingcompl33t:
Per the short discussion at the standup on Friday, we're not concerned about the output format differences between our implementation and Postgres at the moment. Comprehensive testing for aggregations is already being addressed by [#1136](https://github.com/cmu-db/terrier/pull/1136/files) so I believe this PR should just be merged.

lmwnshn (approved):
LGTM.

PR #1146, opened 2020-09-02
Handle Execution Exceptions

jkosh44:
Currently there are various places where ExecutionException can be
thrown from, mostly only reachable through software bugs. Before
this commit they will be fatal and cause the DB to crash.

This commit now handles ExecutionExceptions byt causing the
transaction to abort, and returning an error message to the user.

### DIFF
--- a/src/traffic_cop/traffic_cop.cpp
+++ b/src/traffic_cop/traffic_cop.cpp
@@ -391,7 +391,17 @@ TrafficCopResult TrafficCop::RunExecutableQuery(const common::ManagedPointer<net
 
   const auto exec_query = portal->GetStatement()->GetExecutableQuery();
 
-  exec_query->Run(common::ManagedPointer(exec_ctx), execution_mode_);
+  try {
+    exec_query->Run(common::ManagedPointer(exec_ctx), execution_mode_);
+  } catch (ExecutionException &e) {
+    /*
+     * An ExecutionException is thrown in the case of some failure caused by a software bug or caused by some data
+     * exception. In either case we abort the current transaction and return an error to the client.
+     */
+    connection_ctx->Transaction()->SetMustAbort();
+    auto error = common::ErrorData(common::ErrorSeverity::ERROR, e.what(), e.code_);
+    return {ResultType::ERROR, error};
+  }
 
   if (connection_ctx->TransactionState() == network::NetworkTransactionStateType::BLOCK) {
     // Execution didn't set us to FAIL state, go ahead and return command complete
--- a/src/traffic_cop/traffic_cop.cpp
+++ b/src/traffic_cop/traffic_cop.cpp
@@ -398,8 +398,11 @@ TrafficCopResult TrafficCop::RunExecutableQuery(const common::ManagedPointer<net
      * An ExecutionException is thrown in the case of some failure caused by a software bug or caused by some data
      * exception. In either case we abort the current transaction and return an error to the client.
      */
+    // TODO(Matt): evict this plan from the statement cache when the functionality is available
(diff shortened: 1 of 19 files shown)


jkosh44:
For the error code I chose `ERRCODE_INTERNAL_ERROR` for the majority of errors. Postgres describes this error as "this is for "can't-happen"(https://github.com/postgres/postgres/blob/eb892102e01a2073df9250d65e33ec1ed21798df/src/backend/utils/errcodes.txt#L492) conditions and software bugs (PostgreSQL-specific error class)". The reason being that the majority of these errors should have been caught in the semantic checking phase (If I'm incorrect about any of those please let me know).

The two exceptions that were given a different error code are 
- When a value is out of range of the target type if a cast. I used `ERRCODE_NUMERIC_VALUE_OUT_OF_RANGE`. I wasn't able to find this in semantic checking and postgres uses this error code for similar errors
  - https://github.com/postgres/postgres/blob/a57d312a7706321d850faa048a562a0c0c01b835/src/backend/utils/adt/numeric.c#L1773
  - https://github.com/postgres/postgres/blob/dad75eb4a8d5835ecc795d7a7978e7702e4d5912/src/backend/utils/adt/int8.c#L116
- When a CSV column is out of range, I used `ERRCODE_DATA_EXCEPTION`. I wasn't able to find similar postgres errors.

src/traffic_cop/traffic_cop.cpp
```diff
@@ -391,7 +391,17 @@ TrafficCopResult TrafficCop::RunExecutableQuery(const common::ManagedPointer<net
...
-  exec_query->Run(common::ManagedPointer(exec_ctx), execution_mode_);
+  try {
+    exec_query->Run(common::ManagedPointer(exec_ctx), execution_mode_);
+  } catch (ExecutionException &e) {
+    /*
+     * An ExecutionException is thrown in the case of some failure caused by a software bug or caused by some data
+     * exception. In either case we abort the current transaction and return an error to the client.
+     */
+    connection_ctx->Transaction()->SetMustAbort();
+    auto error = common::ErrorData(common::ErrorSeverity::ERROR, e.what(), e.code_);
```
mbutrovich, line 402: Let's toss this in just because we can:
```
error.AddField(common::ErrorField::LINE, std::to_string(e.GetLine()));
error.AddField(common::ErrorField::FILE, e.GetFile());
```

src/traffic_cop/traffic_cop.cpp
```diff
@@ -391,7 +391,17 @@ TrafficCopResult TrafficCop::RunExecutableQuery(const common::ManagedPointer<net
...
 
-  exec_query->Run(common::ManagedPointer(exec_ctx), execution_mode_);
+  try {
+    exec_query->Run(common::ManagedPointer(exec_ctx), execution_mode_);
+  } catch (ExecutionException &e) {
+    /*
+     * An ExecutionException is thrown in the case of some failure caused by a software bug or caused by some data
+     * exception. In either case we abort the current transaction and return an error to the client.
+     */
+    connection_ctx->Transaction()->SetMustAbort();
```
mbutrovich, line 401: Let's add a TODO that in the future we might want to evict this plan from the statement cache (no functionality for that at the moment). No sense in keeping a busted plan around.

src/traffic_cop/traffic_cop.cpp
jkosh44: Done

src/traffic_cop/traffic_cop.cpp
jkosh44: Done...hope you don't mind I assigned it to you.

PR #1149, opened 2020-09-02
Performance Fix - Use col_ids during codegen of Index scan

rohanaggarwal7997:
Currently, the codegen of index scans in terrier/noisepage call the function col_oid_to_ids every time they have to fetch a tuple or operate on it. This PR, modifies that and generates the col_ids in TPL and directly uses them in the scan.


apavlo:
@rohanaggarwal7997 Do you have benchmark numbers for this?

rohanaggarwal7997:
No

rohanaggarwal7997:
I changed the description

mbutrovich:
After seeing the changes, I'm not really comfortable with breaking a clean abstraction of leaking the notion of col_ids up above the storage layer. We're basically caching the first part of the computation that the storage layer does by stashing the col_id vector in the layer above, and then still doing expensive computations at each IndexIterator::Init when we do the ProjectedRowInitializer creation.

It's a huge breaking of abstractions for what's likely less than 1% of on-cpu time.

PR #1074, opened 2020-08-12
Compilation time, breaking out storage_defs and hash_util

lmwnshn:
Another quick attempt at #1037. At a high level, hash_util (and transitively storage_defs) became expensive includes. So breaking those out of header files into cpp files is good.

- Measured with the Python script linked in #1037, [diff.csv.txt](https://github.com/cmu-db/terrier/files/5065020/diff.csv.txt)
  - hash_util.h : -222 includes
  - xxh3.h : -217 includes

- New build time, 23.13 minutes user time,
  - ```
    [100%] Built target terrier
    make -j12 terrier  1387.90s user 79.82s system 867% cpu 2:49.18 total
    ```
- We had hash_t defined in hash_util and strong_typedef. Shot the former.

PR #1152, opened 2020-09-04
Shrink xxHash include

gonzalezjo:
It's much smaller. This helps a lot with build times. (~10-15% faster for me.) There ought to be some low-hanging fruit that would make it possible to shrink xxhash even more, but I'll revisit this PR tomorrow.

This was inspired by #1074. Wan gave me the go-ahead to make this a separate PR. Hoping that this could be useful with the issue that #1074 faces.

### DIFF
--- a/third_party/xxHash/xxh3.h
+++ b/third_party/xxHash/xxh3.h
@@ -32,11 +32,6 @@
    - xxHash source repository : https://github.com/Cyan4973/xxHash
 */
 
-/* Note :
-   This file is separated for development purposes.
-   It will be integrated into `xxhash.c` when development phase is complete.
-*/
-
 #ifndef XXH3_H
 #define XXH3_H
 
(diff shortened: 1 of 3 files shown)


gonzalezjo:
> LGTM.
> 
> We may want to upgrade xxhash in the future. What notes would you leave for future you? @gonzalezjo

Good question! I'd note that: (1) we don't plan on supporting ARM, PowerPC, 32-bit x86, or (and this is important, since there's a bunch of big-endian specific code) any ISA that uses big-endian integers. 

Almost all (just not the typedefs) of the 32-bit and 128-bit xxHash API code is deletable. All of the streaming APIs are deletable. 


With what's noted above, you cover just about everything here. For the rest, you'd want to just sit down and evaluate macros with the help of your IDE, and try to remove stuff like Visual Studio specific code.

PR #1153, opened 2020-09-04
Modeling Support for CREATE INDEX

17zhangw:
Changes:
- Adds a mini-runner model for `CREATE INDEX`
- Adds logic to generate numeric filenames (execution_SEQ#.csv)
- Strips out `simulate_cache`, reverts a prior change removing mini_trainer models

Other comments:
- OOM at small cardinalities once we hit a certain index size / key size combination
- Verification produces an ~20% elapsed error


src/brain/operating_unit_recorder.cpp
```diff
@@ -203,6 +211,15 @@ void OperatingUnitRecorder::AggregateFeatures(brain::ExecutionOperatingUnitType
...
     cardinality = 1;  // extract from plan num_rows (this is the scan size)
+  } else if (type == ExecutionOperatingUnitType::CREATE_INDEX) {
+    // We extract the num_rows and cardinality from the table name if possible
+    // This is a special case for mini-runners
+    std::string idx_name = reinterpret_cast<const planner::CreateIndexPlanNode *>(plan)->GetIndexName();
+    auto mrpos = idx_name.find("minirunners__");
+    if (mrpos != std::string::npos) {
+      num_rows = atoi(idx_name.c_str() + mrpos + sizeof("minirunners__") - 1);
+      cardinality = num_rows;
+    }
```
17zhangw, line 222: We can probably replace this once counters are merged.

script/model/training_util/data_transforming_util.py
```diff
@@ -105,6 +105,7 @@ def _tuple_num_log_predict_transform(x, y):
     OpUnit.OP_DECIMAL_DIVIDE: _tuple_num_linear_transformer,
     OpUnit.OP_DECIMAL_COMPARE: _tuple_num_linear_transformer,
     OpUnit.OUTPUT: _tuple_num_linear_transformer,
+    OpUnit.CREATE_INDEX: _tuple_num_linear_transformer,
```
malin1993ml, line 108: I think this should be `linear_log_transformer`?

17zhangw: Yes. This should be `linear_log_transformer`.

malin1993ml (changes requested):
I think the code itself is good! But a high-level thing is that we need to track the memory spent to create the actual index. I think we can expose a `SetMemory` method in the `ExecutionContext`, and add a built-in similar to Wan's counters to set the memory usage at the end of the index creation pipeline (before the `EndPipelineTracker` call). The [API](https://github.com/cmu-db/terrier/blob/master/src/include/storage/index/index.h#L73) for getting the estimated index size already exists. And we can add an if statement in the `EndPipelineTracker` so that it does not [set the memory](https://github.com/cmu-db/terrier/blob/master/src/execution/exec/execution_context.cpp#L48) again if it's already set. Let me know if the plan makes sense.

17zhangw:
So a recent change I pushed now allows us to record the estimated index heap size `Index::EstimateHeapUsage` as the recorded memory for a `create index` pipeline. Current changes do not adjust for concurrent threads but will depend on the merge order of this PR relative to #1156.

malin1993ml (approved):

Issue #1113, opened 2020-08-24
Record query traces


malin1993ml:
One necessary functionality for the pilot is to record the traces of past queries and forecast future workloads. For now, we'll skip the forecasting steps, but we need to record the query traces.

We'll piggyback the metrics framework for this. Here's a sketch for the steps:
1. First, please set up [Clion](https://github.com/cmu-db/terrier/wiki/Develop-and-Debug-with-CLion) for your development and read through the [development guideline](https://github.com/cmu-db/terrier/wiki/Coding-Guideline).
1. Add a `metrics_query_trace` command line flag in `settings_defs.h`. This is similar to other metrics, like `metrics_pipeline` or `metrics_bind_command`. 
2. Hook up `metrics_query_trace` to the system by modifying `db_main.h`. Again, this would be similar to other metrics in the system.
3. Add a new `query_trace_metric.h` file in `src/include/metrics`. This metric component will have two metric files: `query_text.csv` and `query_trace.csv`. A good example to look at is `logging_metric.h` where the metric component is also responsible for two files.
4. You will add two hooks in `metrics_store.h` to connect the related places (described below) in the codebase to `MetricsStore` to put data into `QueryTraceMetric`.
5. In `traffic_cop.cpp`, add a function call in `CodegenPhysicalPlan` to record two things for `query_text.csv`: the query id (using `exec_query->GetQueryId()`) and the query text (using `portal->GetStatement()->GetQueryText()`).
6. In `traffic_cop.cpp`, add a function call in `RunExecutableQuery` to record two things for `query_trace.csv`: the query id (using `exec_query->GetQueryId()`) and the current timestamp (using `MetricsUtil::Now()`). 
7. Enable the query trace metrics by passing in the `-metrics_query_trace` flag when executing the terrier binary. Test whether the query texts and traces are recorded correctly using OLTPBench.

malin1993ml:
Addressed by #1161.

PR #1161, opened 2020-09-07
Query Trace Recording

rabbit721:
This PR adds new infrastructure to record SQL queries #1113.


src/traffic_cop/traffic_cop.cpp
```diff
@@ -364,6 +366,16 @@ TrafficCopResult TrafficCop::CodegenPhysicalPlan(
...
 
   // TODO(Matt): handle code generation failing
+  auto &resource_metrics = common::thread_context.resource_tracker_.GetMetrics();
+
+  const bool query_trace_metrics_enabled =
+      common::thread_context.metrics_store_ != nullptr &&
+      common::thread_context.metrics_store_->ComponentToRecord(metrics::MetricsComponent::QUERY_TRACE);
+  if (query_trace_metrics_enabled) {
+    common::thread_context.metrics_store_->RecordQueryText(exec_query->GetQueryId(),
+                                                           portal->GetStatement()->GetQueryText(), resource_metrics);
```
malin1993ml, line 376: We don't need the `resource_metrics` for query texts. We can just remove that one. Instead, we can append the timestamp at the end.

src/traffic_cop/traffic_cop.cpp
```diff
@@ -406,6 +418,17 @@ TrafficCopResult TrafficCop::RunExecutableQuery(const common::ManagedPointer<net
...
 
+  auto &resource_metrics = common::thread_context.resource_tracker_.GetMetrics();
+
+  const bool query_trace_metrics_enabled =
+      common::thread_context.metrics_store_ != nullptr &&
+      common::thread_context.metrics_store_->ComponentToRecord(metrics::MetricsComponent::QUERY_TRACE);
+
+  if (query_trace_metrics_enabled) {
+    common::thread_context.metrics_store_->RecordQueryTrace(exec_query->GetQueryId(), metrics::MetricsUtil::Now(),
+                                                            resource_metrics);
```
malin1993ml, line 429: Similarly, we don't need the `resource_metrics` here as well.

malin1993ml (changes requested):
I think overall this is good! Just two minor changes.

A high-level thing is that I think we also need to add a test case for the actual data recording like [here](https://github.com/cmu-db/terrier/blob/master/test/metrics/metrics_test.cpp#L74). For that, you can reference the [traffic_cop_test](https://github.com/cmu-db/terrier/blob/master/test/traffic_cop/traffic_cop_test.cpp). You pretty much need to [enable all the layers](https://github.com/cmu-db/terrier/blob/master/test/traffic_cop/traffic_cop_test.cpp#L41) of the system. Then create a table and issue some simple queries to the system. The query id and the timestamp might be difficult to check, but at least you can check that the query string is correct and the right number of records is recorded. The `create table` query may not be logged but that's okay.

src/traffic_cop/traffic_cop.cpp
rabbit721: Removed `resource_metrics`.

src/traffic_cop/traffic_cop.cpp
rabbit721: Removed `resource_metrics`.

src/traffic_cop/traffic_cop.cpp
```diff
@@ -364,6 +366,15 @@ TrafficCopResult TrafficCop::CodegenPhysicalPlan(
       common::ManagedPointer<const std::string>(&portal->GetStatement()->GetQueryText()));
 
   // TODO(Matt): handle code generation failing
+
+  const bool query_trace_metrics_enabled =
+      common::thread_context.metrics_store_ != nullptr &&
+      common::thread_context.metrics_store_->ComponentToRecord(metrics::MetricsComponent::QUERY_TRACE);
+  if (query_trace_metrics_enabled) {
+    common::thread_context.metrics_store_->RecordQueryText(exec_query->GetQueryId(),
+                                                           portal->GetStatement()->GetQueryText());
```
malin1993ml, line 375: Let's add the timestamp as the third parameter here. May also be useful.

test/metrics/metrics_test.cpp
```diff
@@ -199,6 +213,56 @@ TEST_F(MetricsTests, TransactionCSVTest) {
...
+    EXPECT_TRUE(false);
+  }
+
+  std::this_thread::sleep_for(std::chrono::seconds(1));
+
+  metrics_manager_->Aggregate();
+  const auto aggregated_data = reinterpret_cast<QueryTraceMetricRawData *>(
+      metrics_manager_->AggregatedMetrics().at(static_cast<uint8_t>(MetricsComponent::QUERY_TRACE)).get());
+  EXPECT_NE(aggregated_data, nullptr);
+  EXPECT_GE(aggregated_data->query_trace_.size(), 2);  // 2 data point recorded
```
malin1993ml, line 248: Should this be `EXPECT_GE` or `EXPECT_EQ`?

test/metrics/metrics_test.cpp
```diff
@@ -199,6 +213,56 @@ TEST_F(MetricsTests, TransactionCSVTest) {
...
+  if (!(aggregated_data->query_trace_.empty())) {
+    EXPECT_GE(aggregated_data->query_trace_.begin()->timestamp_, 2);  // 1 records: create table
```
malin1993ml, line 250: Same here. Also, is the "create table" comment valid here?

test/metrics/metrics_test.cpp
```diff
@@ -199,6 +213,56 @@ TEST_F(MetricsTests, TransactionCSVTest) {
...
+  }
+  EXPECT_GE(aggregated_data->query_text_.size(), 2);  // 2 data point recorded
```
malin1993ml, line 252: Same here.

malin1993ml (changes requested):
Thanks for the updates! Only a few nits left. Also, you need to merge the latest master and resolve the conflicts.

test/metrics/metrics_test.cpp
rabbit721: Fixed it to be `EXPECT_EQ`.

malin1993ml (approved):
LGTM! Just need to pass the build.

PR #1162, opened 2020-09-09
Run clang-format-10 on codebase

mbutrovich:
This is in preparation of bumping our toolchain to LLVM 10. It seems the only major behavior change is how it groups included files. We've seen this behavior already from people (like myself) submitting PRs formatted with the CLion-bundled clang-format (which is newer than the 8 our repo uses).

There are 3 lines that ended up incompatible with clang-format-8 after running 10, so I ran 8 over it again to undo those. This will at least bring the diff down dramatically for the PR that switches over to 10.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -137,7 +137,7 @@ endif ()
 # COMPILER TOOLS
 ###################################################
 
-set(CLANG_TOOLS_VERSION "10")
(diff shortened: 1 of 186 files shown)


mbutrovich:
So, not compatible since it fails check-format with clang-format-8. We'll have to coordinate how to orchestrate this switch.

PR #1164, opened 2020-09-09
Add EXPORT to bytecode handler to fix TPCH runner compiled mode.

lmwnshn:
Add missing EXPORT to bytecode handler.
Fixes TPCH Q4, Q16, Q19 in compiled mode - it was segfaulting because nothing got compiled, and that was because symbols were missing.

Also, don't forget to build the `tpl` target so that `bytecode_handlers_ir.bc` is generated. You may need to regenerate that file every time you switch branches that mess with bytecodes.

Let me know if this works for you? @malin1993ml Thanks!


malin1993ml (approved):
It works! This is super helpful. Thanks!

PR #1170, opened 2020-09-10
ITP Protocol and Basic Replication Integration

tpan496:
Installs basic replication code from Gus to the terrier master. 
- Basic ITP protocol that involves `StartReplication` and `StopReplication` commands.
- Changes to the log manager and networking layer to incorporate replication logs.
- A basic `CreateDatabase` test where the database oid of the master and replica should match.


tpan496:
@lmwnshn finished code cleanup, ready for review. thanks!

src/CMakeLists.txt
```diff
@@ -12,7 +12,7 @@ list(REMOVE_ITEM TERRIER_SRCS ${PROJECT_SOURCE_DIR}/src/main/terrier.cpp)
 # spdlog
 file(GLOB_RECURSE SPDLOG_SOURCES
         ${PROJECT_SOURCE_DIR}/third_party/spdlog/*.h
-)
+        )
```
lmwnshn, line 15: Nit: unnecessary change.

src/CMakeLists.txt
```diff
@@ -22,7 +22,7 @@ add_subdirectory(${PROJECT_SOURCE_DIR}/third_party/libpg_query/ libpg_query.a)
 file(GLOB_RECURSE BWTREE_SOURCES
         ${PROJECT_SOURCE_DIR}/third_party/bwtree/*.cpp
         ${PROJECT_SOURCE_DIR}/third_party/bwtree/*.h
-)
+        )
```
lmwnshn, line 25: Nit: unnecessary change.

src/CMakeLists.txt
```diff
@@ -32,14 +32,14 @@ list(APPEND TERRIER_SRCS ${PROJECT_SOURCE_DIR}/third_party/nlohmann/json.hpp)
 file(GLOB_RECURSE MADOKA_SOURCES
         ${PROJECT_SOURCE_DIR}/third_party/madoka/*.cc
         ${PROJECT_SOURCE_DIR}/third_party/madoka/*.h
-)
+        )
```
lmwnshn, line 35: Nit: unnecessary change.

src/CMakeLists.txt
```diff
@@ -32,14 +32,14 @@ list(APPEND TERRIER_SRCS ${PROJECT_SOURCE_DIR}/third_party/nlohmann/json.hpp)
...
 list(APPEND TERRIER_SRCS ${MADOKA_SOURCES})
 
 # libcount
 file(GLOB_RECURSE LIBCOUNT_SOURCES
         ${PROJECT_SOURCE_DIR}/third_party/libcount/*.cc
         ${PROJECT_SOURCE_DIR}/third_party/libcount/*.h
-)
+        )
```
lmwnshn, line 42: Nit: unnecessary change.

src/CMakeLists.txt
```diff
@@ -70,11 +70,11 @@ ADD_TERRIER_LIB(terrier
         STATIC_LINK_LIBS ${TERRIER_STATIC_LINK_LIBS}
         STATIC_PRIVATE_LINK_LIBS ${TERRIER_STATIC_PRIVATE_LINK_LIBS}
         DEPENDENCIES ${TERRIER_DEPENDENCIES}
-)
+        )
```
lmwnshn, line 73: Nit: unnecessary change.

src/CMakeLists.txt
```diff
@@ -70,11 +70,11 @@ ADD_TERRIER_LIB(terrier
...
 
 ###############################################
 # Terrier executable
 ###############################################
-add_executable(terrier main/terrier.cpp)
+add_executable(terrier main/terrier.cpp include/storage/write_ahead_log/replication_log_consumer_task.h storage/write_ahead_log/replication_log_consumer_task.cpp include/storage/write_ahead_log/log_consumer_task.h)
```
lmwnshn, line 78: Nit: this should not be necessary (and might actually be incorrect).

src/include/main/db_main.h
```diff
@@ -240,6 +241,8 @@ class DBMain {
       server_ = std::make_unique<network::TerrierServer>(common::ManagedPointer(provider_),
                                                          common::ManagedPointer(connection_handle_factory_),
                                                          thread_registry, port, connection_thread_count);
+      server_->RegisterProtocol(port, common::ManagedPointer(provider_), CONNECTION_THREAD_COUNT,
```
lmwnshn, line 244: Does this logic belong in the network layer? Or does it belong in TerrierServer? IMO it belongs in TerrierServer. Also, if you move it to TerrierServer you don't need to include settings.h.

src/include/network/itp/itp_packet_writer.h
```diff
@@ -27,22 +27,24 @@ class ITPPacketWriter : public PacketWriter {
...
+    BeginPacket(NetworkMessageType::ITP_REPLICATION_COMMAND).AppendValue(message_id).AppendValue(data_size);
+  }
 
   /**
    * End the Replication command
    */
   void EndReplicationCommand() { EndPacket(); }
 
   /**
-   * Writes a Stop Replication packet
```
lmwnshn, line 38: Nit: it seems like this change does not do anything except reorder the code.

src/include/network/itp/itp_protocol_interpreter.h
```diff
@@ -85,7 +85,12 @@ class ITPProtocolInterpreter : public ProtocolInterpreter {
   void SetPacketMessageType(common::ManagedPointer<ReadBuffer> in) override;
 
  private:
+  InputPacket curr_input_packet_{};
```
lmwnshn, line 88: Please add a comment describing this.

src/include/network/itp/itp_protocol_interpreter.h
```diff
@@ -85,7 +85,12 @@ class ITPProtocolInterpreter : public ProtocolInterpreter {
...
+  std::unordered_map<std::string, std::string> cmdline_options_;
```
lmwnshn, line 89: Please add a comment describing this + I haven't seen how this is used yet, but I'm suspicious of keeping cmdline options here. It should probably be a setting.

src/include/network/network_defs.h
```diff
@@ -16,6 +16,9 @@ class ReadBuffer;
 // want anyone using it to directly access the socket downstream
 STRONG_TYPEDEF_HEADER(connection_id_t, uint16_t);
 
+// For threads
+#define CONNECTION_THREAD_COUNT 4
```
lmwnshn, line 20: It is not clear to me what this is doing, can you elaborate on the comment?

lmwnshn: We already have connection_thread_count being passed in to TerrierServer, why does RegisterProtocol need its own count? Are they related?

src/include/network/network_io_utils.h
```diff
@@ -221,6 +221,13 @@ class ReadBufferView {
     return result;
   }
 
+  /**
+   * @param bytes The amount of bytes between the current cursor and rest of the buffer view
+   * @return Whether there is any more bytes between the cursor and
+   *         the end of the buffer view
+   */
+  bool HasMore(size_t bytes = 1) { return offset_ + bytes <= size_; }
```
lmwnshn, line 229: This is identical to Buffer::HasMore, but I understand that you might need similar functionality in the ReadBufferView.

That said, we're trying to move away from having default arguments. Consider rewriting to `HasMore(size_t bytes)` and maybe if you really hate calling `HasMore(1)` then you can have an `IsFinished` or something like that which calls `HasMore(1)` for you.

src/include/network/network_io_wrapper.h
```diff
@@ -37,6 +39,25 @@ class NetworkIoWrapper {
     RestartState();
   }
 
+  explicit NetworkIoWrapper(const std::string &ip_address, uint16_t port)
+      : sock_fd_(socket(AF_INET, SOCK_STREAM, 0)),
```
lmwnshn, line 43: This does not check the error code of socket.

lmwnshn: In general, this function seems to be a hack. Is there a reason to not create the socket from whatever is calling this?

src/include/network/network_io_wrapper.h
```diff
@@ -37,6 +39,25 @@ class NetworkIoWrapper {
...
+        in_(std::make_unique<ReadBuffer>()),
+        out_(std::make_unique<WriteQueue>()) {
+    struct sockaddr_in serv_addr;
+    memset(&serv_addr, 0, sizeof(serv_addr));
+    serv_addr.sin_family = AF_INET;
+    serv_addr.sin_addr.s_addr = inet_addr(ip_address.c_str());
+    serv_addr.sin_port = htons(port);
+
+    int64_t ret UNUSED_ATTRIBUTE = connect(sock_fd_, reinterpret_cast<sockaddr *>(&serv_addr), sizeof(serv_addr));
+    // TODO(Gus): we need better exception handling here
```
lmwnshn, line 53: Gus is graduated, assign TODOs to me (WAN) or yourself (whatever you use).

src/include/network/protocol_interpreter.h
```diff
@@ -90,7 +90,7 @@ class ProtocolInterpreter {
    * @param in The ReadBuffer to read input from
    * @return whether the packet header is valid or not
    */
-  bool TryReadPacketHeader(const common::ManagedPointer<ReadBuffer> in) {
+  virtual bool TryReadPacketHeader(const common::ManagedPointer<ReadBuffer> in) {
```
lmwnshn, line 93: Is there a reason to make this virtual?

src/network/itp/itp_protocol_interpreter.cpp
```diff
@@ -40,4 +42,54 @@ void ITPProtocolInterpreter::SetPacketMessageType(const common::ManagedPointer<R
...
+  curr_input_packet_.msg_type_ = in->ReadValue<NetworkMessageType>();
+  curr_input_packet_.len_ = in->ReadValue<uint32_t>() - sizeof(uint32_t);
+  if (curr_input_packet_.len_ > PACKET_LEN_LIMIT) {
+    NETWORK_LOG_ERROR("Packet size {} > limit {}", curr_input_packet_.len_, PACKET_LEN_LIMIT);
+    throw NETWORK_PROCESS_EXCEPTION("Packet too large");
+  }
+
+  // Extend the buffer as needed
+  if (curr_input_packet_.len_ > in->Capacity()) {
+    // Allocate a larger buffer to copy bytes off from the I/O layer's buffer
```
lmwnshn, line 82: As far as I can tell, this is the only change from the original function.

src/network/itp/itp_protocol_interpreter.cpp
```diff
@@ -40,4 +42,54 @@ void ITPProtocolInterpreter::SetPacketMessageType(const common::ManagedPointer<R
   curr_input_packet_.msg_type_ = in->ReadValue<NetworkMessageType>();
 }
 
+bool ITPProtocolInterpreter::TryBuildPacket(common::ManagedPointer<ReadBuffer> in) {
+  if (!TryReadPacketHeader(in)) return false;
```
lmwnshn, line 46: As far as I can tell, this is identical to the existing function.

src/include/network/protocol_interpreter.h
lmwnshn: After looking at the new implementations, I do not think this should be made virtual. I think `TryBuildPacket`'s override is identical to the original, and `TryReadPacketHeader`'s override is almost identical except for a copy of the buffer (which could happen elsewhere).

src/include/storage/recovery/replication_log_provider.h
```diff
@@ -1,19 +1,110 @@
 #pragma once
 
+#include <queue>
 #include <memory>
+#include <utility>
+
+#include <string>
```
lmwnshn, line 7: Surprised format isn't grouping these together, or maybe you need to run that.

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -57,15 +58,18 @@ class LogManager : public common::DedicatedThreadOwner {
   LogManager(std::string log_file_path, uint64_t num_buffers, std::chrono::microseconds serialization_interval,
              std::chrono::microseconds persist_interval, uint64_t persist_threshold,
              common::ManagedPointer<RecordBufferSegmentPool> buffer_pool,
-             common::ManagedPointer<terrier::common::DedicatedThreadRegistry> thread_registry)
+             common::ManagedPointer<terrier::common::DedicatedThreadRegistry> thread_registry,
+             const std::string& ip_address = "", uint16_t replication_port = 0)
```
lmwnshn, line 62: Avoid default args, especially if they can be buggy.

src/network/terrier_server.cpp
```diff
@@ -42,41 +42,43 @@ void TerrierServer::RunServer() {
...
-  int retval = bind(listen_fd_, reinterpret_cast<struct sockaddr *>(&sin), sizeof(sin));
-  if (retval < 0) {
-    NETWORK_LOG_ERROR("Failed to bind socket: {}", strerror(errno));
-    throw NETWORK_PROCESS_EXCEPTION("Failed to bind socket.");
-  }
-  retval = listen(listen_fd_, conn_backlog);
-  if (retval < 0) {
-    NETWORK_LOG_ERROR("Failed to create listen socket: {}", strerror(errno));
-    throw NETWORK_PROCESS_EXCEPTION("Failed to create listen socket.");
+  // int conn_backlog = common::Settings::CONNECTION_BACKLOG;
```
lmwnshn, line 45: Dead code.

test/storage/replication_test.cpp
```diff
@@ -0,0 +1,239 @@
...
+  storage::ReplicationLogProvider *replica_log_provider_;
+  storage::RecoveryManager *replica_recovery_manager_;
+  network::ITPCommandFactory *replica_itp_command_factory_;
+  network::ITPProtocolInterpreter::Provider *replica_itp_protocol_provider_;
+  network::ConnectionHandleFactory *replica_connection_handle_factory_;
+  trafficcop::TrafficCop *replica_tcop_;
+  network::TerrierServer *replica_server_;
+
+  void SetUp() override {
+    TerrierTest::SetUp();
```
lmwnshn, line 97: This is an "old style" test prior to DBMain existing, take a look at how tests are brought up now. We don't need to manually create all the components, I think.

test/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ file(GLOB_RECURSE TEST_UTIL_SRCS ${PROJECT_SOURCE_DIR}/test/test_util/*.cpp ${PR
 # Test library
 ###############################################
 if (TERRIER_BUILD_TESTS OR TERRIER_BUILD_BENCHMARKS)
-    add_library(test_util STATIC ${TEST_UTIL_SRCS})
+    add_library(test_util STATIC ${TEST_UTIL_SRCS} storage/replication_test.cpp)
```
lmwnshn, line 7: Nit: should not be necessary.

src/include/storage/write_ahead_log/replication_log_consumer_task.h
```diff
@@ -0,0 +1,72 @@
+#pragma once
+
+#include <utility>
+#include <vector>
+#include <string>
+#include <memory>
+
+#include "network/network_io_utils.h"
```
lmwnshn, line 8: Nit: Some of these includes can probably be forward declarations.

src/include/storage/write_ahead_log/log_consumer_task.h
```diff
@@ -0,0 +1,40 @@
+#pragma once
+
+#include <condition_variable>  // NOLINT
+#include <utility>
+#include <vector>
+
+#include "common/container/concurrent_blocking_queue.h"
```
lmwnshn, line 7: Nit: most of these includes should be forward declarations.

src/network/itp/itp_protocol_interpreter.cpp
```diff
@@ -20,7 +20,9 @@ Transition ITPProtocolInterpreter::Process(common::ManagedPointer<ReadBuffer> in
     NETWORK_LOG_ERROR("Encountered exception {0} when parsing packet", e.what());
     return Transition::TERMINATE;
   }
-  auto command = command_factory_->PacketToCommand(common::ManagedPointer(&curr_input_packet_));
+
+  std::shared_ptr<ITPNetworkCommand> command =
```
lmwnshn, line 24: Nit: shared pointers tend to confuse ownership. Who owns this command?

src/network/terrier_server.cpp
```diff
@@ -42,41 +42,43 @@ void TerrierServer::RunServer() {
...
+
+  for (auto &protocol : protocols_) {
```
lmwnshn, line 47: Nit: this logic probably should be broken out into a function. Additionally, this seems to duplicate existing code.

lmwnshn (changes requested):
Per our meeting 2020-09-10, we probably will not merge this code, so you don't need to address the comments. I did a quick review though so that you can see what some issues were.

This gives us have a sense of what we had. Later, when we meet with Ricky we should get a sense of where we want to go.

PR #1171, opened 2020-09-10
Remove varlen_entry_benchmark from CI

mbutrovich:
It's just too noisy to be reliable. Added documentation that performance should be looked at critically using the benchmark locally if exercised code is changed.


lmwnshn (approved):
LGTM.

PR #1172, opened 2020-09-10
Prefer fdatasync over fsync of WAL on supported platforms

mbutrovich:
[fdatasync](https://linux.die.net/man/2/fdatasync) is used as an optimization on Linux since we don't care about all of the file's metadata being persisted, just the contents. macOS does not provide this syscall.

### DIFF
--- a/src/include/storage/write_ahead_log/log_io.h
+++ b/src/include/storage/write_ahead_log/log_io.h
@@ -5,6 +5,7 @@
 #include <sys/types.h>
 #include <sys/uio.h>
 #include <unistd.h>
+
 #include <cerrno>
 #include <cstring>
 #include <string>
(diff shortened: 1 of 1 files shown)


src/include/storage/write_ahead_log/log_io.h
```diff
@@ -120,10 +119,15 @@ class BufferedLogWriter {
   }
 
   /**
-   * Call fsync to make sure that all writes are consistent.
+   * Call fsync to make sure that all writes are consistent. fdatasync is used as an optimization on Linux since we
+   * don't care able all of the file's metadata being persisted, just the contents.
```
lmwnshn, line 123: Typo: about

lmwnshn (approved):
LGTM, typo though.

PR #1179, opened 2020-09-13
Parallel Query Usage Tracking

17zhangw:
PR adds support for the followign:
- Tracking resource usage for parallel query execution
- Tracking counters with parallel pipelines
- `CREATE INDEX` instrumentation


lmwnshn:
Per our meeting earlier, we won't merge this. We will introduce hooks into the execution context instead.

PR #1181, opened 2020-09-15
Register abort action - trying functor

preetansh:


preetansh:
@gonzalezjo Can you please have a look at this?

mbutrovich:
Do we have any performance measurements for this? Glancing at the changes, I'm not sure where we're expecting the win. It looks like allocating a `std::function` object was replaced with allocating a class that wraps a `std::function` object.

gonzalezjo:
The goal of a functor is to avoid the cost of std::functions by encapsulating their behavior in stuff that the compiler understands at compile-time. This can even be done with dynamic dispatch and still be way more efficient than the plain std::function. Matt is correct here. This abstracts away (but doesn't eliminate) the std::function. The goal should be to abstract abort actions into their own functors.

PR #1187, opened 2020-09-15
Refactor FunctionContext bootstrapping in DatabaseCatalog

### LINKED ISSUE #1184: Remove abort actions from bootstrap logic in DatabaseCatalog
Throughout the database catalog we register abort actions when adding the function context of builtin functions. An example is below. Doing so shouldn't be necessary because we assert that the transaction doesn't fail and if it did fail it would leave the DB in a non-functioning state.

However removing these lines causes a bunch of tests failures, the cause seems to be memory leaks. Some preliminary investigation seems to indicate that some tests rely on these abort actions to clean up memory.

We should remove these lines from the database catalog and fix the tests to not rely on them.

https://github.com/cmu-db/noisepage/blob/master/src/catalog/database_catalog.cpp#L2089-L2092

Brought up in #1139

mbutrovich:
The `TearDown` logic in `CreateDatabase` wasn't doing anything because it's an AbortAction that queues a CommitAction. I moved the abort action registration for setting `FunctionContext` pointers to the correct spot, and refactored a lot of copy-paste code. I also removed some comments that seemed redundant in the face of self-documenting code. 

Fix #1184.

### DIFF
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -2136,6 +2136,12 @@ bool DatabaseCatalog::SetProcCtxPtr(common::ManagedPointer<transaction::Transact
   TERRIER_ASSERT(write_lock_.load() == txn->FinishTime(),
                  "Setting the object's pointer should only be done after successful DDL change request. i.e. this txn "
                  "should already have the lock.");
+
+  // The catalog owns this pointer now, so if the txn ends up aborting, we need to make sure it gets freed.
+  txn->RegisterAbortAction([=](transaction::DeferredActionManager *deferred_action_manager) {
+    deferred_action_manager->RegisterDeferredAction([=]() { delete func_context; });
+  });
+
   // Do not need to store the projection map because it is only a single column
   auto oid_pri = procs_oid_index_->GetProjectedRowInitializer();
 
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -2071,14 +2071,14 @@ void DatabaseCatalog::BootstrapProcContexts(const common::ManagedPointer<transac
   BootstrapProcContext(txn, postgres::REPEAT_PRO_OID, "repeat", type::TypeId::VARCHAR,
                        {type::TypeId::VARCHAR, type::TypeId::INTEGER}, execution::ast::Builtin::Repeat, true);
 
-  BootstrapProcContext(txn, postgres::TRIM_PRO_OID, "btrim", type::TypeId::VARCHAR,
-                       {type::TypeId::VARCHAR}, execution::ast::Builtin::Trim, true);
+  BootstrapProcContext(txn, postgres::TRIM_PRO_OID, "btrim", type::TypeId::VARCHAR, {type::TypeId::VARCHAR},
+                       execution::ast::Builtin::Trim, true);
 
   BootstrapProcContext(txn, postgres::TRIM2_PRO_OID, "btrim", type::TypeId::VARCHAR,
                        {type::TypeId::VARCHAR, type::TypeId::VARCHAR}, execution::ast::Builtin::Trim2, true);
 
-  BootstrapProcContext(txn, postgres::CONCAT_PRO_OID, "concat", type::TypeId::VARCHAR,
-                       {type::TypeId::VARIADIC}, execution::ast::Builtin::Concat, true);
+  BootstrapProcContext(txn, postgres::CONCAT_PRO_OID, "concat", type::TypeId::VARCHAR, {type::TypeId::VARIADIC},
+                       execution::ast::Builtin::Concat, true);
 
   BootstrapProcContext(txn, postgres::LPAD_PRO_OID, "lpad", type::TypeId::VARCHAR,
                        {type::TypeId::VARCHAR, type::TypeId::INTEGER, type::TypeId::VARCHAR},
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -1616,7 +1616,9 @@ bool DatabaseCatalog::CreateIndexEntry(const common::ManagedPointer<transaction:
   return true;
 }
 
-type_oid_t DatabaseCatalog::GetTypeOidForType(const type::TypeId type) { return type_oid_t(static_cast<uint8_t>(type)); }
+type_oid_t DatabaseCatalog::GetTypeOidForType(const type::TypeId type) {
+  return type_oid_t(static_cast<uint8_t>(type));
+}
 
 void DatabaseCatalog::InsertType(const common::ManagedPointer<transaction::TransactionContext> txn, type_oid_t type_oid,
                                  const std::string &name, const namespace_oid_t namespace_oid, const int16_t len,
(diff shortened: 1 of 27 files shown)


src/include/catalog/database_catalog.h
```diff
@@ -615,6 +616,21 @@ class DatabaseCatalog {
...
+   * @param proc_oid oid to associate with this proc's and its context
+   * @param func_name Name of function
+   * @param func_ret_type Return type of function
+   * @param args_type Vector of argument types
+   * @param builtin Which builtin this context refers to
+   * @param is_exec_ctx_required true if this function requires an execution context var as its first argument
+   */
+  void BootstrapProcContext(common::ManagedPointer<transaction::TransactionContext> txn, proc_oid_t proc_oid,
+                            std::string &&func_name, type::TypeId func_ret_type, std::vector<type::TypeId> &&args_type,
+                            execution::ast::Builtin builtin, bool is_exec_ctx_required);
```
mbutrovich, line 632: When `is_exec_ctx_required` true? The constructor we rely on for `FunctionContext` has a default arg (separate debate) so it's assigned inconsistently. My suspicion was that string functions would need it to allocate, but that wasn't always the case. Could use a sanity check.

mbutrovich: Documenting the answer from Slack:

> Joseph Koshakow From what I saw, all string functions were getting an ExecutionContext. Some probably don't need it, but it made Semantic checking slightlier easier in that we group all of the string functions together and for all of them check that their first argument is an execution context.
> Matt Butrovich hm, okay. so worst case scenario we're passing it in and we dont actually need it
> Joseph Koshakow Yeah I think for a lot of them we just pass it through and never use it. It can definitely be cleaned up IMO.  Like exp for example takes the execution context, but then in the bytecode_generator just ignores it.

src/catalog/database_catalog.cpp
```diff
@@ -1869,432 +1848,294 @@ void DatabaseCatalog::BootstrapProcs(const common::ManagedPointer<transaction::T
...
+                       {type::TypeId::VARCHAR, type::TypeId::INTEGER}, execution::ast::Builtin::Left, true);
+
+  BootstrapProcContext(txn, postgres::RIGHT_PRO_OID, "right", type::TypeId::VARCHAR,
+                       {type::TypeId::VARCHAR, type::TypeId::INTEGER}, execution::ast::Builtin::Right, true);
+
+  BootstrapProcContext(txn, postgres::REPEAT_PRO_OID, "repeat", type::TypeId::VARCHAR,
+                       {type::TypeId::VARCHAR, type::TypeId::INTEGER}, execution::ast::Builtin::Repeat, true);
+
+  BootstrapProcContext(txn, postgres::TRIM_PRO_OID, "btrim", type::TypeId::VARCHAR,
+                       {type::TypeId::VARCHAR, type::TypeId::VARCHAR}, execution::ast::Builtin::Trim, true);
```
jkosh44, line 2075: I'm just noticing this now, but I feel like this is a mistake (not introduced by this PR). This has the same signature as TRIM2 below, and looking at semantic checking/bytecode generation it looks like this should only have 1 argument. From testing though it seems to work with one arg, not really sure how.

mbutrovich: I thought that looked off. If you don't mind, could you push a fix to the branch when you get a chance? You should have access since you're a reviewer.

src/catalog/database_catalog.cpp
```diff
@@ -1869,432 +1848,294 @@ void DatabaseCatalog::BootstrapProcs(const common::ManagedPointer<transaction::T
...
+
+  BootstrapProcContext(txn, postgres::TRIM2_PRO_OID, "btrim", type::TypeId::VARCHAR,
+                       {type::TypeId::VARCHAR, type::TypeId::VARCHAR}, execution::ast::Builtin::Trim2, true);
+
+  BootstrapProcContext(txn, postgres::CONCAT_PRO_OID, "concat", type::TypeId::VARCHAR,
+                       {type::TypeId::VARIADIC, type::TypeId::VARIADIC}, execution::ast::Builtin::Concat, true);
```
jkosh44, line 2081: Similarly CONCAT only needs one VARIADIC arg...that's my bad.

src/include/catalog/database_catalog.h
jkosh44: I'm replying here again for recording purposes. Some string functions need it to allocate memory for new strings, and the NpRunners need it as well to access the output buffer. We end up passing it to all string functions for simplicity (which is not necessarily a good approach). Some other functions, like `exp` get it passed by mistake and just ignore it.

src/catalog/database_catalog.cpp
jkosh44: yep, I'll update this and Concat. The rest looks good to me.

jkosh44 (approved):
LGTM

jkosh44:
Hey @mbutrovich Looks like there's some test failures. It's the same one's that I described in the issue. I think some of those tests were relying on the abort functionality to clean up some memory. Now the tests have memory leaks.

mbutrovich:
I'll take a look. The tests likely predate or at least didn't look at our Postgres protocol stuff. The protocol doesn't allow `CREATE DATABASE` to run inside of a transaction block, so exercising abort logic doesn't make sense.

jkosh44:
Those are the exact same test failures that I was encountering when I removed the abort actions. I was just quickly looking at `catalog_test` and I think one of the problems is probably here
https://github.com/cmu-db/noisepage/blob/70f7e9ee2708ec2ca40c2e5affb58f437abfe5cd/test/catalog/catalog_test.cpp#L193-L198

We try and call `CreateDatabase` twice, which actually goes through and bootstrapping process twice. The second time all those `FunctionContext`s are never cleaned up.

I think it actually might be a bug here on line 108: https://github.com/cmu-db/noisepage/blob/70f7e9ee2708ec2ca40c2e5affb58f437abfe5cd/src/catalog/catalog.cpp#L99-L110
Why would we go the the bootstrapping processing if creating the database entry failed. Maybe it should be `if(bootstrap && success)`.

mbutrovich:
Yeah maybe we can chat about it in the lab sometime. There was some discussion about that in a deferred events meeting. I think @jrolli had mentioned that you have to finish bootstrapping regardless of whether you're gonna keep the database, maybe to stash pointers in the tables to teardown? We'd have to look at the logic again.

jrolli:
That’s correct.  You’ve already created all of the catalog tables and indexes by the time you try to create an entry for the database in DatabaseTable.  Without bootstrapping, the information is not seen during the call to teardown on abort and you will leak all of that memory.

mbutrovich:
I'm confused by why those `FunctionContext` objects need dedicated abort actions currently. Teardown should catch them even in an abort scenario, otherwise how is the catalog getting torn down successfully in the general case?

jrolli:
After a quick look, that looks like the issue.  We currently have a double-free on aborts from bootstrap because of those additional abort actions registered.  Teardown should be the only abort action created by bootstrap.

jrolli:
After a quick look, that looks like the issue.  We currently have a double-free on aborts from bootstrap because of those additional abort actions registered.  Teardown should be the only abort action created by bootstrap.

mbutrovich:
> After a quick look, that looks like the issue. We currently have a double-free on aborts from bootstrap because of those additional abort actions registered. Teardown should be the only abort action created by bootstrap.
That's what I thought, but instead of getting double-free ASAN warnings on master, we get leaks when I strip out the abort actions. I don't get it. I can also see them getting read out of the table and deleted on `Teardown`. It's baffling.

Issue #1188, opened 2020-09-16
Implement support for SHOW.


lmwnshn:
Right now, we can SET settings to new values, but we can't easily check the current value of the setting without using a debugger. It would be nice to have SHOW. For only showing setting values, this should be similar to the existing code for supporting SET -- instead of calling `Set` you call `Get`.

lmwnshn:
Whoever implements this, could you add tests for #1189? Thanks!

apavlo:
We need this to support `SHOW TRANSACTION ISOLATION LEVEL` for the new version of OLTP-Bench

PR #1189, opened 2020-09-16
Allow SET to accept values that come in as VARCHAR despite having a different type.

lmwnshn:
This relies on existing binder functionality, so the support is the same as the binder.
Some examples are booleans coming in as strings and timestamps coming in as strings.
This is a quirk of the Postgres parser.

Before:
```
terrier=# set parallel_execution to TRUE;
ERROR:  invalid value for parameter "parallel_execution": "true"
```

After:
```
terrier=# set parallel_execution to TRUE;
SET
```


malin1993ml (approved):
LGTM.

apavlo:
Can we add SQL tests for this?

mbutrovich:
> Can we add SQL tests for this?

That's probably a punt until #1188.

apavlo:
At the very least we could test valid/invalid inputs

lmwnshn:
Per the earlier discussion, will open an issue and assign it to whoever implements SHOW. If this passes, it merges.

PR #1192, opened 2020-09-18
Toolchain related fixes

gonzalezjo:
This collection of fixes allows us to build using clang-10 and gcc-10.

Note that this does not allow us to build using llvm-10's libraries. This is purely addressing the new warnings. In the process, though, it does make some functional changes: 

1. RowsAffected and NumRows now return a uint32_t. Technically speaking, this limits the number of rows we can return to ~4 billion. Practically speaking, we were always limited to ~4 billion because the network protocol is uses uint32_ts to tell the client how many rows it's returning. 

This just propagates the fundamental network limits through the type system. 

(While I'm pretty sure a whole lot of other things will break before TrafficCop if we ever try to return 4 billion rows, this change had to be made in order for us to build under gcc-10)

2. The cast-safety functions in binder_util and cast_operators have been adjusted to deal with a bizarre edge-case that clang-10 now checks for. It boils down to floats and doubles quantizing in an unfortunate and counter-intuitive manner. Example here: https://godbolt.org/z/M14jdb

These two functional changes are in need of review.

### DIFF
--- a/src/binder/binder_util.cpp
+++ b/src/binder/binder_util.cpp
@@ -182,7 +182,13 @@ void BinderUtil::CheckAndTryPromoteType(const common::ManagedPointer<parser::Con
 
 template <typename Output, typename Input>
 bool BinderUtil::IsRepresentable(const Input int_val) {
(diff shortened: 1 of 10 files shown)


mbutrovich:
LGTM

PR #1193, opened 2020-09-20
Latency int bug

bialesdaniel:
There is currently a bug in the oltpbench scripts where the latency numbers are converted into ints before being sent to the performance storage service. This causes problems with the Grafana visualizations because a number of the latency metrics end up being <1. This causes /0 errors as you can see here https://stats.noise.page/grafana/d/latency_trends/latency-trends?orgId=1

This PR fixes this issue and refactors the code by implementing reuse to give a single place to control the parsing of latency metric labels. This will be especially helpful if https://github.com/oltpbenchmark/oltpbench/pull/338 gets merged (though it also works without that other PR).

I tested this code on both pipelines:
Terrier: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/latency-int-bug/3/pipeline/375
Nightly: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier-nightly/detail/latency-int-bug/5/pipeline/

### DIFF
--- a/Jenkinsfile-nightly
+++ b/Jenkinsfile-nightly
@@ -4,7 +4,7 @@ pipeline {
     environment{
         //Do not change. 
         //Performance Storage Service(Django) authentication information. The credentials can only be changed on Jenkins webpage
-        PSS_CREATOR= credentials('pss-creator')
+        PSS_CREATOR= credentials('pss-creator-testing')
     }
     options {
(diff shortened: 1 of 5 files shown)


skyfishjy (commented):
LGTM

rohanaggarwal7997 (approved):

PR #1195, opened 2020-09-22
Build external projects with same compiler as system

mbutrovich:
This solves an issue that macOS students are seeing after #1131 merged.
```
Undefined symbols for architecture x86_64:
 "___asan_version_mismatch_check_apple_clang_1103", referenced from:
   _asan.module_ctor in libgtest.a(gtest-all.cc.o)
   _asan.module_ctor in libgmock_main.a(gmock_main.cc.o)
   _asan.module_ctor in libgflags.a(gflags.cc.o)
   _asan.module_ctor in libgmock_main.a(gmock-all.cc.o)
   _asan.module_ctor in libgflags.a(gflags_reporting.cc.o)
   _asan.module_ctor in libgflags.a(gflags_completions.cc.o)
ld: symbol(s) not found for architecture x86_64`
```


mbutrovich:
And of course that breaks on a Linux build.

gonzalezjo:
You can always add `if (APPLE)` but generally setting CMAKE_C(XX)_COMPILER is a source of nightmares. The idiomatic solution--as unfortunate as this is--is to have users source something like this:


```
LLVM_DIR=sh(script: "brew --prefix llvm@10", label: "Fetching LLVM path", returnStdout: true).trim()
CC="${LLVM_DIR}/bin/clang"
CXX="${LLVM_DIR}/bin/clang++"
```

Before actually building the project. After doing so, things *should* work.

mbutrovich:
Other CMake and toolchain changes are coming...

Issue #1196, opened 2020-09-22
Missing {bwtree,hash}_index.h documentation


gonzalezjo:
Missing bwtree_index.h and hash_index.h are both missing documentation. Doxygen 1.8.14+ flag this. This is therefore a blocker to the bump to Ubuntu 20.04 (which uses Doxygen 1.8.17)

Tagging #879.

mbutrovich:
Fixed by #1197.

PR #1197, opened 2020-09-22
Fix complaints from new Doxygen versions

gonzalezjo:
Copy-pastes docs from index.h into hash_index.h and bwtree_index.h. No functional change.

This is necessary to fix issue #1196.


src/include/storage/index/bwtree_index.h
```diff
@@ -44,34 +44,98 @@ class BwTreeIndex final : public Index {
...
 
  public:
+  /**
+   * @return type of the index. Note that this is the physical type, not extracted from the underlying schema or other
+   * catalog metadata. This is mostly used for debugging purposes.
+   */
   IndexType Type() const final { return IndexType::BWTREE; }
 
+  /**
+   * Invoke garbage collection on the index. For some underlying index types this may be a no-op.
```
lmwnshn, line 54: Could you remove the "no-op" part? Or better, "This is NOT a no-op", or something like that.

lmwnshn (approved):
LGTM, though I have one minor nitpick.

PR #1199, opened 2020-09-23
SQLite Test Traces: Support for Error Flag + Value List

dniu16:
1. I talked with Ricky yesterday and he suggested that instead of comparing the hash, we should compare the whole result list instead. I modified the code so that if a flag "Include Outputs" is specified in the comments before a query in the sql input file, the whole result list will be checked, comparing values one by one with a 0.000001 error precision. 

2. Error flag for non-select queries as before. If a non-select query fails and an error code is specified in the trace, the codes will be compared.

3. This PR include modification for aggregation test. I'll submit another PR later for aggregation test files


apavlo:
@dniu16 There is a failing SQL in `insert.test` at line 68. Is this expected?

script/testing/junit/src/TracefileTest.java
```diff
@@ -104,12 +109,17 @@ public static DynamicTest executeSelectQuery(int num, String cur_sql) {
             ResultSet rs = statement.getResultSet();
             res = mog.processResults(rs);
             // create an executable for the query
-            String hash2 = TestUtility.getHashFromDb(res);
-            String message = "Failure at Line " + num + ": " + "\n" + cur_sql + "\n" +
-                    "Query expected " + parsed_num_result + " results, got " + res.size() + " results"
-                    + "\n" + res;
-            boolean len_match = getCheckLength(check_expected, parsed_num_result, res.size());
-            exec = () -> checkResultMatch(parsed_hash, hash2, message, len_match, res);
+            if(onlyResult){
```
apavlo, line 112: Add comment that explains what this means.

script/testing/junit/src/TracefileTest.java
```diff
@@ -118,6 +128,27 @@ public static DynamicTest executeSelectQuery(int num, String cur_sql) {
...
+                if(!queryResult.get(i).equals("")){
+                    throw new Exception("Value '' Mismatch");
+                }
+            }else{
+                double one = Double.parseDouble(res.get(i));
+                double two = Double.parseDouble(queryResult.get(i));
+                if(Math.abs(one-two)>precision){
+                    throw new Exception("Expected " + queryResult + " but have " + res);
+                }
+            }
```
apavlo, line 148: Is this trying to parse all values to a double and then check whether they are equal? What if the type isn't a double?

script/testing/junit/src/TracefileTest.java
```diff
@@ -184,14 +229,18 @@ public static void checkResultMatch(String hash1, String hash2, String message,
                 List<String> new_res = new ArrayList<>();
                 for(String i:res){
                     try{
-                        int cur = (int)Double.parseDouble(i);
+                        int cur = (int)Math.round(Double.parseDouble(i));
```
apavlo, line 232: This seems like a hack. What are you trying to do here?

script/testing/junit/src/moglib/Constants.java
```diff
@@ -10,4 +10,10 @@
     public static final String SEPARATION = "----";
     public static final String DEST_DIR = "traces";
     public static final String VALUES = "value";
+    public static final String SKIPIF = "skipif";
+    public static final String ONLYIF = "onlyif";
+    public static final String NUM_OUTPUT_FLAG = "No of outputs";
```
apavlo, line 15: Should this be `"Num of outputs"`?

script/testing/junit/src/FilterTrace.java
```diff
@@ -33,70 +31,80 @@ public static void main(String[] args) throws Throwable {
...
+            if(cur_sql.contains("SELECT")){
+                boolean contain_aggregation = false;
+                for(String i: aggregations){
+                    if(cur_sql.contains(i)){
+                        contain_aggregation = true;
+                    }
+                }
+                if(!contain_aggregation){
+                    skip_flag = true;
+                }
```
apavlo, line 57: What is this doing? Why are aggregations skipped?

apavlo (changes requested):
The code seems unwieldy. I am not sure what is going on in some cases. Please address comments.

script/testing/junit/src/TracefileTest.java
dniu16: I was trying to deal with the float precision. If the hashes don't match but the result length are the same, I converted doubles from the result list to int and compared them again.

script/testing/junit/src/FilterTrace.java
dniu16: For this part we skip queries that don't contain aggregation keywords. I was using this to filter out traces from the sqlite files, but then realized that the sqlite syntax is a lot different

script/testing/junit/src/TracefileTest.java
githubxxcc: I think specifying a delta to account for this precision drifting might be better than merely rounding.

script/testing/junit/src/FilterTrace.java
```diff
@@ -33,70 +31,80 @@ public static void main(String[] args) throws Throwable {
         Connection conn = db.getDbTest().newConn();
         List<String> tab = getAllExistingTableName(mog,conn);
         removeExistingTable(tab,conn);
+        List<String> aggregations = new ArrayList<>();
+        aggregations.add("COUNT");
+        aggregations.add("MAX");
+        aggregations.add("MIN");
+        aggregations.add("AVG");
+        aggregations.add("SUM");
```
apavlo, line 39: What is going on here? Why do we need this?

script/testing/junit/src/GenerateTrace.java
```diff
@@ -71,25 +74,44 @@ public static void main(String[] args) throws Throwable {
...
                 String hash = TestUtility.getHashFromDb(res);
                 String queryResult = "";
-                if(expected_result_num>=0){
-                    queryResult = "Expected " + expected_result_num + " values hashing to " + hash;
+                if(include_result){
+                    for(String i:res){
+                        queryResult += i;
+                        queryResult += "\n";
+                    }
+                    queryResult = queryResult.trim();
```
apavlo, line 82: Please explain what this means. What does it mean to include result?

script/testing/junit/src/GenerateTrace.java
```diff
@@ -71,25 +74,44 @@ public static void main(String[] args) throws Throwable {
...
+                            queryResult = res.size() + " values hashing to " + hash;
+                        }
+                        if(res.size() < Constants.DISPLAY_RESULT_SIZE){
+                            queryResult = "";
+                            for(String i:res){
+                                queryResult += i;
+                                queryResult += "\n";
+                            }
+                            queryResult = queryResult.trim();
+                        }
```
apavlo, line 97: There is a lot going on here but no documentation / comments.

script/testing/junit/src/TracefileTest.java
```diff
@@ -118,6 +133,43 @@ public static DynamicTest executeSelectQuery(int num, String cur_sql) {
...
+                if(!queryResult.get(i).equals("")){
+                    throw new Exception("Value '' Mismatch");
+                }
+            }else{
+                try{
+                    double one = Double.parseDouble(res.get(i));
+                    double two = Double.parseDouble(queryResult.get(i));
+                    if(Math.abs(one-two)>precision){
+                        throw new Exception("Expected " + queryResult + " but have " + res);
+                    }
```
apavlo, line 156: As we talked about, we don't want to do this rounding. We want to match our own output.

script/testing/junit/src/TracefileTest.java
apavlo: Remove the rounding.

apavlo (changes requested):
Please add code comments to explain what is happening.

script/testing/junit/src/TracefileTest.java
dniu16: So this is for the case that hash don't match but result list size match. I did a work around here to cast doubles to int and then compare the hash again. How should I deal with floating difference without the rounding? We might only get the hash from the trace instead of exact results.

script/testing/junit/src/FilterTrace.java
dniu16: We don't need this file for the PR. I was using FilterTrace to filter out traces from sqlite files.

script/testing/junit/src/TracefileTest.java
dniu16: Yeah there's no rounding here. Within this function I compare the query result list to the list from trace.

apavlo: What if the value is a string and not a double?

script/testing/junit/src/TracefileTest.java
apavlo: For now people will have to modify the tracefile by hand to add the value list.

apavlo (changes requested):

script/testing/junit/src/TracefileTest.java
dniu16: Then it will enter the catch clause below comparing the values.

script/testing/junit/src/TracefileTest.java
dniu16: I see. If a value list is used (instead of hash), we won't enter this piece of code. Instead, the checkEqual method will be called (the other unresolved comment)

githubxxcc: For clarification here, I asked Jeff to support this precision aware comparison (but yeah, rounding might not be the best way). 
Quick question here, If we need to enter the value by ourselves, is the trace test for regression test only? 
As we are simply comparing our noisepage result (ran in the JUnit test) with the noisepage results(manually generated and inputed from manual runs of same trace SQLs).

PR #1200, opened 2020-09-23
Network layer refactor.

lmwnshn:
This is an opinionated nitpicky refactor of the network layer. There are no real logic changes.

I was digging around in the network layer last week because I thought I would be building on it for the `Messenger` component, but that is turning out to be independent of the existing network layer. So merging this PR is not necessary if votes are not in favor.

- connection_handle.cpp : replace macro-based state machine with templates, less lines of code overall, `HandleDispatch()` and other libevent callback type of functions do not need to have dummy arguments any more (libevent didn't need them, our macros did).
- Add guide-style comments, arguably redundant, perhaps useful.
- Forward declare more stuff. I need to watch CI to see if the Mac builds complain without some system includes.


gonzalezjo:
Please don't take my reply as a review (I've only skimmed files changed list on GitHub) but I think this is a huge improvement. The comment changes are awesome. Thanks for fixing mine!

src/include/common/event_util.h
```diff
@@ -35,6 +35,14 @@ class EventUtil {
   }
 
  public:
+  /**
+   * EVENT_ACTIVATE_OR_TIMEOUT_ONLY can be supplied to event_new() to indicate that the event can only be activated
+   * by an explicit event_active or by a timeout firing.
+   */
+  static constexpr const int EVENT_ACTIVATE_OR_TIMEOUT_ONLY = -1;
```
mbutrovich, line 42: I think this actually forces static storage duration (i.e. prevents the compiler from completely eliding allocating it and just inlining the constant expression). I think you want to drop the `static` and then the `constexpr` is free to be elided as an allocated variable.

lmwnshn: I get `non-static data error cannot be constexpr: did you intend to make it static?` if I drop the static.

I wrote a quick test (attached) where the compiler at `O0` seems to be able to constant fold it away across multiple translation units, maybe it is fine?

image

mbutrovich: That's a class member though. `static` has different meanings at the namespace level, block level, and class member level.

src/network/connection_handle.cpp
```diff
@@ -1,188 +1,257 @@
...
+                                   std::unique_ptr<ProtocolInterpreter> interpreter)
+    : io_wrapper_(std::make_unique<NetworkIoWrapper>(sock_fd)),
+      conn_handler_task_(task),
+      traffic_cop_(tcop),
+      protocol_interpreter_(std::move(interpreter)) {
+  context_.SetCallback(Callback, this);
+  context_.SetConnectionID(static_cast<connection_id_t>(sock_fd));
+}
+
+ConnectionHandle::~ConnectionHandle() { context_.Reset(); }
```
mbutrovich, line 149: Is reseting necessary if the `ConnectionContext` is about be to destroyed? I took a peek at `Reset` and it doesn't seem necessary to be called here.

src/network/connection_handle.cpp
```diff
@@ -1,188 +1,257 @@
...
+  workpool_event_ = conn_handler_task_->RegisterManualEvent(
+      [](int fd, int16_t flags, void *arg) { static_cast<ConnectionHandle *>(arg)->HandleEvent(fd, flags); }, this);
+
+  network_event_ = conn_handler_task_->RegisterEvent(
+      io_wrapper_->GetSocketFd(), EV_READ | EV_PERSIST,
+      [](int fd, int16_t flags, void *arg) { static_cast<ConnectionHandle *>(arg)->HandleEvent(fd, flags); }, this);
+}
+
+void ConnectionHandle::HandleEvent(int fd, int16_t flags) {
+  // TODO(WAN): convince myself that this is correct. Why should it be TERMINATE?
```
mbutrovich, line 161: I think that's the idle disconnect.

src/network/connection_handle.cpp
```diff
@@ -1,188 +1,257 @@
...
+  /** Define a function (ManagedPointer<ConnectionHandle> -> Transition) that calls the underlying handle's function. */
+#define DEF_HANDLE_WRAPPER_FN(function_name)                                               \
+  static Transition function_name(const common::ManagedPointer<ConnectionHandle> handle) { \
+    return handle->function_name();                                                        \
+  }
+
+  DEF_HANDLE_WRAPPER_FN(GetResult);
+  DEF_HANDLE_WRAPPER_FN(Process);
+  DEF_HANDLE_WRAPPER_FN(TryRead);
+  DEF_HANDLE_WRAPPER_FN(TryWrite);
```
mbutrovich, line 78: Is `TryWrite` ever even used?

mbutrovich (commented):
Little stuff.

src/network/connection_handle.cpp
lmwnshn: In Process::PROCESS and Write::WAKEUP, the function is called.
I don't know off the top of my head if those two states occur, but it doesn't seem implausible.
A breakpoint on TryWrite does get hit.

src/network/connection_handle.cpp
lmwnshn: thanks!

src/network/connection_handle.cpp
lmwnshn: you're right

mbutrovich (approved):
LGTM

PR #1201, opened 2020-09-24
Add a Messenger subsystem.

lmwnshn:
## Blast radius

- Adds a new dependency on ZeroMQ via CMake.
- Adds a default port 9022 and allows code to open more ports.
- Existing code should be unaffected.

## What

This PR adds support for a Messenger subsystem by bringing in [ZeroMQ](https://zeromq.org/). To quote the ZeroMQ docs,
>     In the ZeroMQ universe, sockets are doorways to fast little background communications engines
>     that manage a whole set of connections automagically for you.

We open a single ZeroMQ context `zmq::context_t` on every NoisePage process. A ZeroMQ context abstracts away the many TCP, IPC, and in-process transports that ZeroMQ manages on the program's behalf.

The following ZeroMQ transports are opened by default on every terrier instance. In increasing order of speed (slow to fast),
- `tcp://*:9022` : regular TCP socket on port 9022.
- `ipc:///tmp/noisepage-ipc0`: interprocess communication file, which ZeroMQ implements with Unix domain sockets.
- `inproc://noisepage-inproc`: fancy ZeroMQ in-process communication, uses the ZeroMQ context.

All I/O is written by a background I/O thread. Old documentation claims that each I/O thread can comfortably handle about 1 GBps of data. By default, ZeroMQ only has one background I/O thread, this can be easily increased if necessary.

The nice thing about ZeroMQ is that you just swap your connection string from TCP to IPC and everything just works*. In practice, the important part is that you now have a high speed channel to communicate complex messages in-process between threads (inproc), inter-process (ipc), or over the network (tcp).

\* Assuming that you wrote the server and client protocols properly. ZeroMQ is prone to misbehaving during the transport setup and teardown phases if you're not careful, which usually manifests as lost messages.

## Why

We envision using the Messenger for replication and for the pilot. Having ZeroMQ abstract away all the socket and IPC stuff is nice. I have a separate branch that tries to pick between raw TCP sockets and IPC automagically and that code is much less readable.

## How

If familiar with ZeroMQ, we are implementing a ROUTER(server)-DEALER(client) pattern with the usual REQ-REP message envelope protocol but without the typical ROUTER(server)-DEALER(server) inproc workerpool. If not familiar, you shouldn't need to know the details, but I'm trying to document `messenger.cpp` with all relevant ZeroMQ background.


src/include/messenger/messenger.h
```diff
@@ -0,0 +1,84 @@
...
+  ConnectionId MakeConnection(const ConnectionDestination &target, std::optional<std::string> identity);
+
+  void SendMessage(ConnectionId *connection_id, std::string message);
+
+ private:
+  static constexpr int MESSENGER_PORT = 9022;
+  static constexpr const char *MESSENGER_DEFAULT_TCP = "tcp://*:9022";
+  static constexpr const char *MESSENGER_DEFAULT_IPC = "ipc:///tmp/noisepage-ipc0";
+  static constexpr const char *MESSENGER_DEFAULT_INPROC = "inproc://noisepage-inproc";
+  static constexpr const char *MESSENGER_INTERNAL_INPROC = "inproc:://messenger-internal-inproc";
```
githubxxcc, line 64: Maybe could drop the protocol from: 
`inproc://noisepage-inproc` 
to 
`noisepage-inproc` ?

Since you already have different functions for ConnectionDestination?

src/messenger/messenger.cpp
```diff
@@ -0,0 +1,242 @@
...
+    }
+  }
+};
+
+}  // namespace terrier
+
+namespace terrier::messenger {
+
+/** An abstraction around successful connections made through ZeroMQ. */
+class ConnectionId {
```
githubxxcc, line 133: This also needs to be put into header right? I am getting incomplete type error whenI tried to obtain a conn id.

lmwnshn: You are right. I have pushed an update.

src/include/messenger/messenger.h
lmwnshn: If I'm understanding correctly, inproc:// is necessary. The way connections in ZeroMQ work is that you must always specify a protocol (tcp://, ipc://, inproc://) and a destination. `noisepage-inproc` by itself has no protocol.

If you mean, what's MESSENGER_INTERNAL_INPROC -- yeah, I'm getting rid of that at some point. That was in case we want a messenger worker pool.

src/messenger/messenger.cpp
```diff
@@ -211,16 +203,16 @@ void Messenger::Terminate() {
 }
 
 ConnectionId Messenger::MakeConnection(const ConnectionDestination &target, std::optional<std::string> identity) {
```
githubxxcc, line 205: Is it better to return a unique_ptr here? It makes easier for caller of this to own a `ConnectionId` I guess

githubxxcc: I am having this use case in mind:
My pilotmanager will want to create a connection with the messenger, but not necessary at the construction (or it will need to somehow reset a connection during the lifetime of the pilotmanager). 
This means the pilotmanager preferably needs to have a member field to a connectionid, but returning a ConnectionId here kind of enforces the lifetime to be the lifetime of the enclosing function call. 

Or is there a better way for me to have a connection that lives longer than the enclosing function's lifetime?

githubxxcc: Actually, I think for my use case, this is fine. I think I could, and should, have a connectionId initialized at the constructor of my pilot manager.

githubxxcc:
How should the user(e.g., pilot manager/replica manager) exposes its message processing logic to the messenger? Let's say my pilot manager expects something like: 
```
SEND Connect
RECV ConnectAck
SEND Quit 
RECV QuitAck
```

How should I supply this logic to the Messenger layer? (I am currently seeing this coded in the MessengerLogic class, and what's your plan going forward for customized logic from different users?)

tpan496:
I am getting a CMake error when I'm trying to compile with this.
```
CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
Please set them or make sure they are set and tested correctly in the CMake files:
ZMQ_INCLUDE_DIR
```
Am I missing something here?

githubxxcc:
> I am getting a CMake error when I'm trying to compile with this.
> 
> ```
> CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
> Please set them or make sure they are set and tested correctly in the CMake files:
> ZMQ_INCLUDE_DIR
> ```
> 
> Am I missing something here?

You on mac?

tpan496:
> > I am getting a CMake error when I'm trying to compile with this.
> > ```
> > CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
> > Please set them or make sure they are set and tested correctly in the CMake files:
> > ZMQ_INCLUDE_DIR
> > ```
> > 
> > 
> > Am I missing something here?
> 
> You on mac?

No, I’m using CLion on PC. Is this supposed to be an issue with Mac?

githubxxcc:
> > > I am getting a CMake error when I'm trying to compile with this.
> > > ```
> > > CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
> > > Please set them or make sure they are set and tested correctly in the CMake files:
> > > ZMQ_INCLUDE_DIR
> > > ```
> > > 
> > > 
> > > Am I missing something here?
> > 
> > 
> > You on mac?
> 
> No, I’m using CLion on PC. Is this supposed to be an issue with Mac?

Have you installed zmq, and cppzmq? Not sure how it will work out on windows though.

tpan496:
> > > > I am getting a CMake error when I'm trying to compile with this.
> > > > ```
> > > > CMake Error: The following variables are used in this project, but they are set to NOTFOUND.
> > > > Please set them or make sure they are set and tested correctly in the CMake files:
> > > > ZMQ_INCLUDE_DIR
> > > > ```
> > > > 
> > > > 
> > > > Am I missing something here?
> > > 
> > > 
> > > You on mac?
> > 
> > 
> > No, I’m using CLion on PC. Is this supposed to be an issue with Mac?
> 
> Have you installed zmq, and cppzmq? Not sure how it will work out on windows though.

Oh, I was not aware that I need to install this as an extra dependency. It's working after I installed this. Thanks!

src/messenger/messenger.cpp
```diff
@@ -0,0 +1,256 @@
...
+  return ConnectionId(common::ManagedPointer(zmq_ctx_), target, identifier);
+}
+
+void Messenger::SendMessage(common::ManagedPointer<ConnectionId> connection_id, std::string message) {
+  ZmqMessage msg{connection_id->routing_id_, message};
+  ZmqUtil::SendMsg(common::ManagedPointer(connection_id->socket_), msg);
+}
+
+void Messenger::ServerLoop() {
+  common::ManagedPointer<zmq::socket_t> socket{zmq_default_socket_};
```
githubxxcc, line 237: If it's the `zmq_default_socket_` that the server loop will always be recving from. How about those sockets opened with `MakeConnection`?

src/messenger/messenger_logic.cpp
```diff
@@ -0,0 +1,34 @@
+#include "messenger/messenger_logic.h"
+
+#include "common/json.h"
+#include "loggers/messenger_logger.h"
+#include "messenger/messenger.h"
+
+namespace terrier::messenger {
+
+void MessengerLogic::ProcessMessage(std::string_view sender, std::string_view message) {
```
githubxxcc, line 9: How are we handling different connection's logic (Like the model's vs the replica's)? 
Will the `MessengerLogic` have some sort of pointer to the logic of them, and invoke based on sender here? 

Will it be better if the MessengerLogic expose a function to register user-specific logic (model's, replica's) during the runtime?

src/messenger/messenger.cpp
```diff
@@ -0,0 +1,256 @@
...
+    return messenger::ZmqMessage{identity, payload};
+  }
+
+  /** @return Send the specified ZmqMessage (identity and payload) over the socket. */
+  static void SendMsg(common::ManagedPointer<zmq::socket_t> socket, const messenger::ZmqMessage &msg) {
+    zmq::message_t identity_msg(msg.identity_.data(), msg.identity_.size());
+    zmq::message_t delimiter_msg("", 0);
+    zmq::message_t payload_msg(msg.payload_.data(), msg.payload_.size());
+    bool ok = true;
+
```
githubxxcc, line 152: I actually managed to make it work without deleting the send `identity_msg`

src/messenger/messenger.cpp
```diff
@@ -0,0 +1,256 @@
...
+    char buf[MAX_ROUTING_ID_LEN];
+    size_t routing_id_len = MAX_ROUTING_ID_LEN;
+    socket->getsockopt(ZMQ_ROUTING_ID, &buf, &routing_id_len);
+    return std::string(buf, routing_id_len);
+  }
+
+  /** @return The next string to be read off the socket. */
+  static std::string Recv(common::ManagedPointer<zmq::socket_t> socket, zmq::recv_flags flags) {
+    zmq::message_t message;
+    auto received = socket->recv(message, flags);
```
githubxxcc, line 130: This call may raise errors when the serverloop terminates (because I believe this `recv` is a blocking call), and it will raise errors 'host not reachable'. 
What is the best way to handler someone calls `messenger->Terminate()` when it is blocked in this `recv` call?

src/messenger/messenger.cpp
lmwnshn: oh, interesting. That's weird. I wonder if that means the python API is adding it automatically but the C++ api does not.

src/messenger/messenger.cpp
lmwnshn: Is `messenger->Terminate()` able to be called? If so, I can add an explicit call to `zmq_ctx->shutdown()` which causes all blocking socket operations to end with `ETERM`.

src/messenger/messenger.cpp
lmwnshn: ahhh. My bad. I should be doing a poll. Addressed in latest push.

src/messenger/messenger.cpp
lmwnshn: Resolved in latest push.

src/messenger/messenger.cpp
lmwnshn: It turns out that ROUTER and DEALER sockets behave differently. I will update this information inside the slides.

src/messenger/messenger_logic.cpp
lmwnshn: Replaced with callback-per-message functionality.

lmwnshn:
@githubxxcc Ready for review. Please feel free to comment on absolutely anything that doesn't make sense - if you have to spend more than 5 minutes on googling any specific thing, I should provide more information. Thanks!

src/include/messenger/messenger.h
```diff
@@ -119,31 +123,25 @@ class Messenger : public common::DedicatedThreadTask {
    * @param connection_id   The connection to send the message over.
    * @param message         The message to be sent.
    * @param callback        The callback function to be invoked on the response.
+   * @param recv_msg_id     The receiver's message ID, e.g., sent in response or to invoke preregistered functions.
+   *                        To invoke preregistered functions, use static_cast<uint8_t>(Messenger::BuiltinCallback).
```
githubxxcc, line 127: Why is the callback enum number being used as a recv_msg_id?

src/messenger/messenger.cpp
```diff
@@ -179,9 +182,7 @@ class MessengerPolledSockets {
     if (!writer_.items_.empty()) {
       std::scoped_lock lock(mutex_);
```
githubxxcc, line 183: There could potentially be a race on the check of empty.

src/messenger/messenger.cpp
```diff
@@ -364,32 +374,29 @@ ConnectionId Messenger::MakeConnection(const ConnectionDestination &target) {
   return ConnectionId(common::ManagedPointer(this), target, identity_);
 }
 
-void Messenger::SendMessage(common::ManagedPointer<ConnectionId> connection_id, std::string message, CallbackFn fn) {
-  uint64_t message_id = message_id_++;
+void Messenger::SendMessage(common::ManagedPointer<ConnectionId> connection_id, const std::string &message,
```
githubxxcc, line 377: This function needs to be made thread-safe as well?

src/include/messenger/messenger.h
githubxxcc: I guess a more direct naming for this id from its usage is: `callback_id`, or `recv_cb_id`

githubxxcc (approved):
LGTM other a few questions on concurrency. 

Would be building model server manager based on this, but it looks rather possible the model server manager would work as expected since replica test had been proven to work.

test/messenger/messenger_test.cpp
```diff
@@ -0,0 +1,224 @@
...
+    spin_until_done();
+  };
+
+  VoidFn replica1_fn = [=]() {
+    auto replica1 = BuildDBMain(port_replica1, port_messenger_replica1, "replica1");
+    replica1->GetNetworkLayer()->GetServer()->RunServer();
+
+    // Set up a connection to the primary.
+    auto messenger = replica1->GetMessengerLayer()->GetMessenger();
+    ConnectionDestination dest_primary = Messenger::GetEndpointIPC("primary", port_messenger_primary);
```
tpan496, line 134: These tests should still work with replicas on TCP right? Or do we prefer IPC?

src/include/messenger/messenger.h
```diff
@@ -0,0 +1,183 @@
...
+  /**
+   * Connect to the specified target destination.
+   *
+   * @param target      The destination to be connected to.
+   * @return            A new ConnectionId. See warning!
+   *
+   * @warning           DO NOT USE THIS ConnectionId FROM A DIFFERENT THREAD THAN THE CALLER OF THIS FUNCTION!
+   *                    Make a new connection instead, connections are cheap.
+   */
+  ConnectionId MakeConnection(const ConnectionDestination &target);
```
tpan496, line 116: I think it would be more helpful if we have some examples of these? e.g. Correct ordering to use MakeConnection and ListenForConnection; if two messengers are communicating with each other, should both be calling `ListenForConnection` and `MakeConnection`?

tpan496 (approved):
LGTM, have some small questions.

githubxxcc:
> Adds a new dependency on ZeroMQ via CMake. Can someone tell me if the equivalent of libzmq5-dev or libzmq3-dev exists on Mac?

I managed to make it work by compiling and installing the cppzmq from source + `brew install zmq`

githubxxcc:
It will be nice to have a `SendMessage` that expects the callback to be a builtincallback.

src/messenger/messenger.cpp
lmwnshn: Yeah, stale reads are OK, it will get caught by the next poll. I will add a comment.

test/messenger/messenger_test.cpp
lmwnshn: It should work with TCP. Doesn't matter.

src/include/messenger/messenger.h
lmwnshn: You shouldn't need `ListenForConnection` at all. Referring to noisepage design slides, that is if you want it to be available on a new endpoint. You can just `MakeConnection` to the default endpoints as shown in this example.

Do you have concrete examples of what you want to see as examples?

src/include/main/db_main.h
```diff
@@ -381,6 +412,7 @@ class DBMain {
       db_main->execution_layer_ = std::move(execution_layer);
       db_main->traffic_cop_ = std::move(traffic_cop);
       db_main->network_layer_ = std::move(network_layer);
+      db_main->messenger_layer_ = std::move(messenger_layer);
```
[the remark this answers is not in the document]
lmwnshn: Sure

src/messenger/messenger.cpp
lmwnshn: Comment added.

src/include/messenger/messenger.h
lmwnshn: Renamed. Perhaps we should just call everything callback ID since that is what it really is..

src/messenger/messenger.cpp
lmwnshn: Added latch to unordered_map, made int atomic, but I think we don't need to do it for the socket -- see comment.

lmwnshn:
@githubxxcc Do you have an idea of what API you would like? Is it basically the same as SendMessage, but it does the cast for you?

src/include/messenger/messenger.h
tpan496: Oh, I see, I was a little bit confused by the function name. I thought it was listening for connections from some remote destination.

githubxxcc:
> @githubxxcc Do you have an idea of what API you would like? Is it basically the same as SendMessage, but it does the cast for you?

This is what I have for now for your ref:

```
  /**
   *  Sends a message and expects a default builtin callbackj
   * @param connection_id   The connection to send the message over.
   * @param message         The message to be sent.
   * @param callback_id        The builtin callback to be invoked.
   * @param recv_msg_id     The receiver's message ID, e.g., sent in response or to invoke preregistered functions.
   */
  void SendMessage(common::ManagedPointer<ConnectionId> connection_id, const std::string &message, BuiltinCallback callback_id,
                   uint64_t recv_msg_id);
```

src/messenger/messenger.cpp
```diff
@@ -0,0 +1,480 @@
...
+
+namespace terrier::messenger {
+
+ConnectionId::ConnectionId(common::ManagedPointer<Messenger> messenger, const ConnectionDestination &target,
+                           const std::string &identity)
+    : target_name_(target.GetTargetName()) {
+  // Create a new DEALER socket and connect to the server.
+  socket_ = std::make_unique<zmq::socket_t>(*messenger->zmq_ctx_, ZMQ_DEALER);
+  socket_->set(zmq::sockopt::routing_id, identity);
+  socket_->connect(target.GetDestination());
```
githubxxcc, line 301: Should this one be `bind`?

lmwnshn: No, it should be connect.

You call connect() on sockets that are bind()'d.

jrolli:
@lmwnshn, removed `ready-for-ci` tag because it appears there is a missing include statement or installed dependency based off of repeated missing zmq-related declarations in `src/messenger/messenger.cpp`.

PR #1202, opened 2020-09-24
Fix visiting SQL values in bytecode generation.

### LINKED ISSUE #1167: Builtin SQL functions should use L-values during bytecode generation
All builtin SQL function require pointers to SQL values (i.e., L-values) to work. This is to avoid copying fat objects through the stack. However, during bytecode generation, they all seem to use R-values (see [here](https://github.com/cmu-db/noisepage/blob/ab16754a777929329fe2461e6516b81694c1808e/src/execution/vm/bytecode_generator.cpp#L1637) (2020-09-04: updated your link to be a permalink rather than current main branch)). Not sure how this got borked (maybe porting TPL2?). As is, queries like the following fail:

```
psql (12.4 (Ubuntu 12.4-0ubuntu0.20.04.1), server 9.5devel)
Type "help" for help.

[terrier] # create table xxx (a int, b int, c double);
CREATE TABLE
Time: 7.839 ms
[terrier] # insert into xxx values (1,2,0.3),(4,5,-0.6),(7,8,9.0);
INSERT 0 3
Time: 9.986 ms
[terrier] # select a,a+b,sin(c) from xxx order by a;
Time: 261.789 ms
[] # 
```

We need to revisit `BytecodeGenerator` and modify each function to use L-values instead of R-values if the TPL function wants SQL types. It's probably clearer to have an explicit `BytecodeGenerator::VisitExpressionForSQLValue()` which dispatches to `BytecodeGenerator::VisitExpressionForLValue()` to avoid repetition.

### LINKED ISSUE #1237: Returning SQL values doesn't work in VM but works in compiled.
# Bug Report

## Summary

Returning SQL values doesn't work in VM but works in compiled.

## Environment

**OS**: Ubuntu 20.04 (LTS)
**Compiler**: GCC 9.3.0

## Steps to Reproduce

```
fun foo() -> Integer {
  return @intToSql(12345)
}

fun main() -> int {
  var x = foo()
  if (x == @intToSql(12345)) {
    return 1
  }

  return 0
}
```

### Expected Behavior

```
[2020-10-16 05:52:42.490] [execution_logger] [info] VM main() returned: 1
[2020-10-16 05:52:42.490] [execution_logger] [info] ADAPTIVE main() returned: 1
[2020-10-16 05:52:42.606] [execution_logger] [info] JIT main() returned: 1
```

### Actual Behavior

```
[2020-10-16 05:52:42.490] [execution_logger] [info] VM main() returned: 0
[2020-10-16 05:52:42.490] [execution_logger] [info] ADAPTIVE main() returned: 0
[2020-10-16 05:52:42.606] [execution_logger] [info] JIT main() returned: 1
```

### More Data

```
[2020-10-16 05:52:41.721] [execution_logger] [info] TPL Bytecode Count: 645
[2020-10-16 05:52:41.721] [execution_logger] [info] TPL initialized ...
[2020-10-16 05:52:41.722] [execution_logger] [info] 
CPU Info: 
  Processors: 1
  Model:      Intel(R) Xeon(R) W-1250 CPU @ 3.30GHz
  Cores:      6 physical, 12 logical
  Mhz:        4400.66
  Caches: 
    L1: 32.00 KB (64 byte line)
    L2: 256.00 KB (64 byte line)
    L3: 12288.00 KB (64 byte line)
Features: sse4_2 avx avx2 

[2020-10-16 05:52:41.722] [execution_logger] [info] Welcome to TPL (ver. 0.1)
[2020-10-16 05:52:41.722] [execution_logger] [info] Compiling and running file: /home/kapi/CLionProjects/noisepage/sample_tpl/return-sql.tpl

Data: 
  Data section size 0 bytes (0 locals)

Function 0 <foo>:
  Frame size 12 bytes (1 parameter, 2 locals)
    param  hiddenRv:  offset=0       size=8       align=8       type=*Integer
    local      tmp1:  offset=8       size=4       align=4       type=int32  

  0x00000000    AssignImm4                                      local=&tmp1  i32=12345
  0x0000000c    InitInteger                                     local=&hiddenRv  local=tmp1
  0x00000018    Return                                          

Function 1 <main>:
  Frame size 60 bytes (1 parameter, 8 locals)
    param  hiddenRv:  offset=0       size=8       align=8       type=*int32 
    local         x:  offset=8       size=16      align=8       type=Integer
    local      tmp1:  offset=24      size=1       align=1       type=bool   
    local      tmp2:  offset=25      size=2       align=1       type=Boolean
    local      tmp3:  offset=32      size=16      align=8       type=Integer
    local      tmp4:  offset=48      size=4       align=4       type=int32  
    local      tmp5:  offset=52      size=4       align=4       type=int32  
    local      tmp6:  offset=56      size=4       align=4       type=int32  

  0x00000000    Call                                            func=<foo>  local=&x  
  0x0000000c    AssignImm4                                      local=&tmp4  i32=12345
  0x00000018    InitInteger                                     local=&tmp3  local=tmp4
  0x00000024    EqualInteger                                    local=&tmp2  local=&x  local=&tmp3
  0x00000034    ForceBoolTruth                                  local=&tmp1  local=&tmp2
  0x00000040    JumpIfFalse                                     local=tmp1  target=0x000068
  0x0000004c    AssignImm4                                      local=&tmp5  i32=1
  0x00000058    Assign4                                         local=hiddenRv  local=tmp5
  0x00000064    Return                                          
  0x00000068    AssignImm4                                      local=&tmp6  i32=0
  0x00000074    Assign4                                         local=hiddenRv  local=tmp6
  0x00000080    Return                                          

[2020-10-16 05:52:42.490] [execution_logger] [info] VM main() returned: 0
[2020-10-16 05:52:42.490] [execution_logger] [info] ADAPTIVE main() returned: 0
[2020-10-16 05:52:42.606] [execution_logger] [info] JIT main() returned: 1
[2020-10-16 05:52:42.606] [execution_logger] [info] Parse: 0.018549 ms, Type-check: 0.028111 ms, Code-gen: 0.041711 ms, Interp. Exec.: 0.066872 ms, Adaptive Exec.: 0.093155 ms, Jit+Exec.: 116.083 ms
[2020-10-16 05:52:42.807] [execution_logger] [info] TPL cleanly shutdown ...

Process finished with exit code 0
```

lmwnshn:
Fix #1167 and fix #1237.

This restores support for trig functions on SQL values, among other things.

On a separate note, @pmenon do you have a rule of thumb that people can use for LValue/RValue visits?
This RValue code has existed since TPL1. I think the general approach that we took in TPL1 (and that others are taking now) is, if it doesn't work as a LValue, try a RValue.
IIRC the most common way LValue blows up is people using `@builtin(@anotherBuiltin())`, so people tend towards RValue by default.

Pending test cases, blocked on #1199.


src/execution/vm/bytecode_generator.cpp
```diff
@@ -1833,13 +1842,13 @@ void BytecodeGenerator::VisitBuiltinArithmeticCall(ast::CallExpr *call, ast::Bui
 
   switch (builtin) {
     case ast::Builtin::Abs: {
-      LocalVar src = VisitExpressionForRValue(call->Arguments()[0]);
+      LocalVar src = VisitExpressionForSQLValue(call->Arguments()[0]);
```
pmenon, line 1845: Change the trailing call to `GetExecutionResult()->SetDestination(dest);` (i.e., remove `dest.ValueOf()`. That should fix your nesting problem.

pmenon (changes requested):
Looks good. All functions _producing_ SQL values should also set the destination explicitly like `GetExecutionResult()->SetDestination(dest);` and not `dest.ValueOf()` since we want callers to have a _pointer_ to the SQL values, not the value itself.

src/execution/vm/bytecode_generator.cpp
lmwnshn: That worked, thanks!

src/execution/vm/bytecode_generator.cpp
```diff
@@ -503,10 +503,11 @@ void BytecodeGenerator::VisitReturnStmt(ast::ReturnStmt *node) {
     LocalVar result;
     if (node->Ret()->GetType()->IsSqlValueType()) {
       result = VisitExpressionForSQLValue(node->Ret());
+      BuildAssign(rv, result, node->Ret()->GetType());
```
lmwnshn, line 506: @pmenon Could you check this one? Thanks!

pmenon: Since you already have a destination address to store the result, you can get by with:
```
if (node->Ret() != nullptr) {
  LocalVar rv = GetCurrentFunction()->GetReturnValueLocal();
  TERRIER_ASSERT(rv.GetAddressMode() == LocalVar::AddressMode::Address, "RV expected to be address.");
  // rv is a double pointer to the result. So we want the recursive call to store the
  // result of the computation into *rv. Hence, we set the destination to rv.ValueOf().
  VisitExpressionForRValue(node->Ret(), rv.ValueOf());
}
```

pmenon (changes requested):
Nice! Can you also add a sample TPL file to check if you can return SQL values from functions. Try single returns and complex/nested returns too.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -500,12 +500,10 @@ void BytecodeGenerator::VisitUnaryOpExpr(ast::UnaryOpExpr *node) {
 void BytecodeGenerator::VisitReturnStmt(ast::ReturnStmt *node) {
   if (node->Ret() != nullptr) {
     LocalVar rv = GetCurrentFunction()->GetReturnValueLocal();
-    LocalVar result;
     if (node->Ret()->GetType()->IsSqlValueType()) {
-      result = VisitExpressionForSQLValue(node->Ret());
-      BuildAssign(rv, result, node->Ret()->GetType());
+      VisitExpressionForSQLValue(node->Ret(), rv);
```
lmwnshn, line 504: @pmenon Could you please check this?

Writing tests is blocked on #1237.

pmenon: Ack, I was wrong. Let's make a copy of the returned value:

```
if (node->Ret()->GetType()->IsSqlValueType()) {
  LocalVar result = VisitExpressionForSQLValue(node->Ret());
  BuildDeref(rv.ValueOf(), result, node->Ret()->GetType());
}
```

pmenon (changes requested):
Small fix for returns (which should fix returning #1237). Also, you missed a few changes:
1. SQL casts [here](https://github.com/cmu-db/noisepage/blob/6083450b33d8bdf59b05b55016cada3e20e1c88f/src/execution/vm/bytecode_generator.cpp#L584).
2. Setting values in `VPI`s [here](https://github.com/cmu-db/noisepage/blob/6083450b33d8bdf59b05b55016cada3e20e1c88f/src/execution/vm/bytecode_generator.cpp#L837) and [here](https://github.com/cmu-db/noisepage/blob/6083450b33d8bdf59b05b55016cada3e20e1c88f/src/execution/vm/bytecode_generator.cpp#L843).
3. SQL math [here](https://github.com/cmu-db/noisepage/blob/6083450b33d8bdf59b05b55016cada3e20e1c88f/src/execution/vm/bytecode_generator.cpp#L3223).
5. SQL comparisons [here](https://github.com/cmu-db/noisepage/blob/6083450b33d8bdf59b05b55016cada3e20e1c88f/src/execution/vm/bytecode_generator.cpp#L3310).

pmenon (approved):
LGTM.

PR #1206, opened 2020-09-25
Ubuntu 20.04 version bump PR

gonzalezjo:
This PR does more than I want it to do, hasn't gone entirely through Jenkins yet, etc. It couldn't be further from being merge-ready. Just creating this to track progress and get this onto "normal" CI. Certain commits are just broken / don't do what I want them to do, etc.

Think of this more as an exploratory PR.


src/include/settings/settings_defs.h
```diff
@@ -1,270 +1,125 @@
...
+
+                        SETTING_bool(metrics_gc,
+                                     "Metrics collection for the GarbageCollector component (default: false).", false,
+                                     true, terrier::settings::Callbacks::MetricsGC)
+
+                            SETTING_bool(metrics_execution,
+                                         "Metrics collection for the Execution component (default: false).", false,
+                                         true, terrier::settings::Callbacks::MetricsExecution)
+
+                                SETTING_bool(metrics_pipeline,
```
lmwnshn, line 95: Formatting here is wack.

(found by @weihang7)

PR #1208, opened 2020-09-28
Parallel Query Instrumentation through "hooks"

### LINKED ISSUE #1116: Race conditions in multi-threaded use of output buffer.
As found by @17zhangw, the current implementation of output buffers (specifically [AllocOutputSlot](https://github.com/cmu-db/terrier/blob/master/src/include/execution/exec/output.h#L58)) has the following problems in a multi-threaded setting:

1. It can hand out the same slots. This is because `num_tuples_` is not guarded and two threads can be calling this function at the same time.
2. It can invoke the output callback while a slot is still being written to. This is because we assumed in the original design that if we're asking for a new slot, we're done with the old one.

17zhangw:
Fix #1116, #1225, #1209.

At a high level, this PR introduces support for parallel query instrumentation. Parallel queries are instrumented such that each discrete chunk of work is recorded separately (regardless of whether the task is executed by the same thread or not). Furthermore, each pipeline state now stashes an `ExecOUFeatureVector` in which pipeline features can be stashed and mutated before being logged with the MetricsManager.

Certain parallel operations have a special post-parallel processing step (i.e building the global hash table) which is written in C++. To handle recording features and metrics for the post-processing step, the PR introduces the concept of hooks: a hook is a TPL function handle that is stored in the ExecutionContext which can then be invoked by both TPL/C++ code.

To control code generation, the PR introduces a few extra settings flags and enables support for initializing `ExecutionSettings` based on the SettingsManager for network queries.
- `counters_enable`: whether counters are enabled or not
- `pipeline_metrics_enable`: whether pipeline metrics should be recorded
- `num_parallel_execution_threads`: how many threads to use for parallel query execution

In addition, to support parallel query execution over psql, the PR also introduces an auxiliary fix for #1116 by adding an `OutputBuffer` to the pipeline state corresponding with the pipeline containing an `OutputTranslator`. This thread-local `OutputBuffer` allows individual threads to retrieve and write to output slots independent of other outputters. Furthermore, an extra latch is inserted inside of `PostgresPacketWriter::WriteDataRow` to serialize concurrent `OutputBuffers` from writing to the network.

The full change-list includes the following:
- Tracking resource usage for parallel query execution
- Tracking counters with parallel pipelines
- `CREATE INDEX` instrumentation
- Additional mini-runner instrumentation
- Setting ExecutionSettings from settings flags
- Parsing indexes of CSV files
- Renaming of settings flags
- Altered metrics column order and names
- Fix #1116


17zhangw:
Temporarily blocked by #1210.

benchmark/runner/mini_runners.cpp
```diff
@@ -681,42 +681,82 @@ static void GenUpdateDeleteIndexArguments(benchmark::internal::Benchmark *b) {
...
+      }
+    }
+  }
+}
+
+static void GenCreateIndexMixedArguments(benchmark::internal::Benchmark *b) {
+  // 0 is a special argument used to indicate serial
+  auto num_threads = {0, 1, 2, 4, 8, 16};
+  std::vector<int64_t> row_nums = {1,    3,    5,     7,     10,    50,     100,    200,    500,    1000,
+                                   2000, 5000, 10000, 20000, 50000, 100000, 200000, 500000, 1000000};
```
malin1993ml, line 734: These values repeat quite a lot. I think we would extract them into a common vector and also pass them into `TableGenerator` when we generate test tabels for `MiniRunners`. For now just use a global variable here (I'm thinking of creating another config file when I clean up the code). And then we can special case for the runners that don't have these exact parameters. Also, create a `GenerateMiniRunnerTables` function in `TableGenerator` to seperate the logic to create mini runner tables from to logic to create other test tables.

benchmark/runner/mini_runners.cpp
```diff
@@ -681,42 +681,82 @@ static void GenUpdateDeleteIndexArguments(benchmark::internal::Benchmark *b) {
...
  * 4 - row
  * 5 - cardinality
+ * 6 - mixed type is a varchar or not
+ * 7 - # possible threads
  */
 static void GenCreateIndexArguments(benchmark::internal::Benchmark *b) {
-  auto num_cols = {1, 3, 5, 7, 9, 11, 13, 15};
+  // 0 is a special argument used to indicate serial
+  auto num_threads = {0, 1, 2, 4, 8, 16};
+  auto num_cols = {1, 2, 3, 4, 5, 7, 9, 11, 13, 15};
```
malin1993ml, line 695: Similar here. Extract this into a common global vector. Also, define the column combination for the mini runner tables in this file and pass them into `GenerateMiniRunnerTables` as argument. In general, remove hard-coded values in `table_generator.cpp` (also applies to `GenerateMiniRunnerIndexTables`). Put all the mini-runner related parameters in this file.

malin1993ml (commented):
The modeling side looks good! I guess the majority of this PR is about the instrumentation that needs @lmwnshn to review though.

lmwnshn (commented):
I'm still digging at the PR and half the comments are comments to myself probably. But I won't be able to go further into it today, switching to work on messenger. So posting a quarter-to-half review for now

script/model/info/data_info.py
```diff
@@ -41,14 +41,17 @@
...
+
 # Index for tuple_num in the model input feature vector
 TUPLE_NUM_INDEX = 0
 
 # Index for cardinality in the model input feature vector
 CARDINALITY_INDEX = 3
 
 # Index for execution mode in the model input feature vector
-EXECUTION_MODE_INDEX = 6
+EXECUTION_MODE_INDEX = 7
```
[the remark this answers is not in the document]
malin1993ml: I think Wan is correct here. Let's parse as many indexes in this file as possible directly from the csv file (including the indexes for the `Target`s). We probably need to define a new `ExecutionFeature` enum class (similar to the [`ArithmeticFeature`](https://github.com/cmu-db/noisepage/blob/master/script/model/type.py#L59)) and construct the index map based on the csv header.

src/execution/compiler/operator/index_create_translator.cpp
```diff
@@ -118,7 +153,15 @@ void IndexCreateTranslator::PerformPipelineWork(WorkContext *context, FunctionBu
       auto *record =
           codegen->CallBuiltin(ast::Builtin::ExecutionContextSetMemoryUseOverride, {GetExecutionContext(), get_mem});
       function->Append(codegen->MakeStmt(record));
+      RecordCounters(*GetPipeline(), function);
     }
+  } else if (IsPipelineMetricsEnabled()) {
+    // For parallel, just record 0 --- for the memory use.
```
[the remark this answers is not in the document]
17zhangw: This generates the function that gets invoked by `TableVectorIterator` during a parallel build.

src/execution/compiler/operator/hash_aggregation_translator.cpp
```diff
@@ -250,24 +263,131 @@ void HashAggregationTranslator::TearDownAggregationHashTable(FunctionBuilder *fu
...
+  return builder.Finish();
+}
+
 void HashAggregationTranslator::InitializePipelineState(const Pipeline &pipeline, FunctionBuilder *function) const {
-  if (IsBuildPipeline(pipeline) && build_pipeline_.IsParallel()) {
-    InitializeAggregationHashTable(function, local_agg_ht_.GetPtr(GetCodeGen()));
+  if (IsBuildPipeline(pipeline)) {
+    if (build_pipeline_.IsParallel()) {
+      InitializeAggregationHashTable(function, local_agg_ht_.GetPtr(GetCodeGen()));
+      CounterSet(function, agg_count_, 0);
```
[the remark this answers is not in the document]
17zhangw: See the above discussion on `agg_count_`.

src/execution/compiler/operator/hash_aggregation_translator.cpp
```diff
@@ -250,24 +263,131 @@ void HashAggregationTranslator::TearDownAggregationHashTable(FunctionBuilder *fu
...
+}
+
+void HashAggregationTranslator::EndParallelPipelineWork(const Pipeline &pipeline, FunctionBuilder *function) const {
+  RecordCounters(pipeline, function);
+
+  if (IsBuildPipeline(pipeline) && IsCountersEnabled()) {
+    auto *codegen = GetCodeGen();
+    const auto &agg_ht = local_agg_ht_;
+    auto *tuple_count = codegen->CallBuiltin(ast::Builtin::AggHashTableGetInsertCount, {agg_ht.GetPtr(codegen)});
+    function->Append(codegen->Assign(agg_count_.Get(codegen), tuple_count));
```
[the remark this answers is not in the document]
17zhangw: That is accurate. This is mainly done since TBB can assign multiple tasks using the same thread-local state so it is very likely to change if and when TBB gets replaced. This value is still a little bit dubious in general --- since it tracks the number of unique inserts from the last "state". That is also why `agg_count_` cannot be set to 0 in `InitializeCounters` (since there is also an implicit guarantee that `BeginParallelPipelineWork` will call `InitializeCounters` which might also benefit from more visible documentation).

I think I will add documentation around the `agg_count_` variable in `src/include/execution/compiler/operator/hash_aggregation_translator.h` (along with references) unless you have a better place in mind.

src/execution/compiler/executable_query.cpp
```diff
@@ -154,14 +154,18 @@ void ExecutableQuery::Run(common::ManagedPointer<exec::ExecutionContext> exec_ct
...
   exec_ctx->SetExecutionMode(static_cast<uint8_t>(mode));
   exec_ctx->SetPipelineOperatingUnits(GetPipelineOperatingUnits());
+  exec_ctx->SetQueryId(query_id_);
 
   // Now run through fragments.
   for (const auto &fragment : fragments_) {
     fragment->Run(query_state.get(), mode);
   }
+
+  exec_ctx->SetQueryState(nullptr);
```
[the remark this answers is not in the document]
17zhangw: I forgot whether there was a discussion to re-use ExecutionContexts? But I will add a comment here.

src/brain/operating_unit_util.cpp
```diff
@@ -0,0 +1,36 @@
...
+      return ExecutionOperatingUnitType::HASHJOIN_BUILD;
+    case ExecutionOperatingUnitType::PARALLEL_MERGE_AGGBUILD:
+      return ExecutionOperatingUnitType::AGGREGATE_BUILD;
+    case ExecutionOperatingUnitType::PARALLEL_SORT_STEP:
+    case ExecutionOperatingUnitType::PARALLEL_SORT_MERGE_STEP:
+      return ExecutionOperatingUnitType::SORT_BUILD;
+    case ExecutionOperatingUnitType::CREATE_INDEX_MAIN:
+      return ExecutionOperatingUnitType::CREATE_INDEX;
+    default:
+      return ExecutionOperatingUnitType::INVALID;
```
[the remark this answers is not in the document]
17zhangw: This function sort of serves a dual purpose: (1) checking if the ExecutionOperatingUnitType represents a parallel OU and (2) returning the non-parallel type if it is a parallel OU. So with the current way the function is written, an assert can't be inserted since the function is supposed to test.

src/brain/operating_unit_recorder.cpp
```diff
@@ -166,6 +163,7 @@ void OperatingUnitRecorder::AggregateFeatures(brain::ExecutionOperatingUnitType
   size_t num_rows = 1;
   size_t cardinality = 1;
   size_t num_loops = 0;
+  size_t num_concurrent = 0;  // Will concurrent data be baked into plan nodes?
```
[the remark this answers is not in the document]
17zhangw: I think this is more so of a question of whether going forwards, will plan information carry instructions of how many threads to use? Or will that information still only be available at runtime?

src/execution/compiler/operator/hash_aggregation_translator.cpp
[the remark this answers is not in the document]
17zhangw: Docs have been added.

src/execution/compiler/codegen.cpp
```diff
@@ -584,11 +584,10 @@ ast::Expr *CodeGen::TableIterClose(ast::Expr *table_iter) {
...
 }
 
-ast::Expr *CodeGen::IterateTableParallel(catalog::table_oid_t table_oid, ast::Identifier col_oids,
-                                         ast::Expr *query_state, ast::Expr *exec_ctx, ast::Identifier worker_name) {
-  ast::Expr *call = CallBuiltin(
-      ast::Builtin::TableIterParallel,
-      {Const32(table_oid.UnderlyingValue()), MakeExpr(col_oids), query_state, exec_ctx, MakeExpr(worker_name)});
+ast::Expr *CodeGen::IterateTableParallel(catalog::table_oid_t table_oid, ast::Expr *col_oids, ast::Expr *query_state,
+                                         ast::Expr *exec_ctx, ast::Identifier worker_name) {
+  ast::Expr *call = CallBuiltin(ast::Builtin::TableIterParallel, {Const32(table_oid.UnderlyingValue()), col_oids,
```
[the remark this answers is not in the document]
17zhangw: This was done to de-duplicate some code in `IndexCreateTranslator`. `col_oids` can't be an `ast::Identifier` in that case since the col_oids comes from the QueryState.

src/include/settings/settings_defs.h
```diff
@@ -276,3 +284,37 @@ SETTING_string(
...
+SETTING_bool(
+    use_counters,
+    "Whether to use counters (default: false)",
+    false,
+    true,
+    terrier::settings::Callbacks::NoOp
+)
+
+SETTING_bool(
+    use_pipeline_metrics,
```
malin1993ml, line 315: Can you consolidate this with `metrics_pipeline`? Also I think we can rename this to `pipeline_metrics_enable` to be consistent with our other flag namings. Also, please apply the naming to other metrics, e.g., `logging_metrics_enable`, and `counters_enable` as well.

src/include/metrics/pipeline_metric.h
```diff
@@ -77,8 +78,8 @@ class PipelineMetricRawData : public AbstractRawData {
    * Note: This includes the columns for the input feature, but not the output (resource counters)
    */
   static constexpr std::array<std::string_view, 1> FEATURE_COLUMNS = {
-      "query_id, pipeline_id, exec_mode, num_features, features, est_output_rows, key_sizes, num_keys, "
-      "est_cardinalities, mem_factor, num_loops"};
+      "query_id, pipeline_id, num_features, features, exec_mode, est_output_rows, key_sizes, num_keys, "
```
githubxxcc, line 81: I believe this is no longer consistent with data_info.py 's `RAW_FEATURES_VECTOR_INDEX = 4`

17zhangw:
That is correct. I have some unpushed local changes aimed at refactoring
away all the indexes.

src/include/metrics/pipeline_metric.h
17zhangw: That is correct. I have some unpushed local changes aimed at refactoring away all the indexes.

src/include/settings/settings_defs.h
```diff
@@ -276,3 +284,29 @@ SETTING_string(
...
+
+SETTING_int(
+    num_threads,
+    "Number of threads for parallel query execution (default: 1)",
+    1,
+    1,
+    128,
+    true,
+    terrier::settings::Callbacks::NoOp
+)
```
malin1993ml, line 304: Personally I prefer not to have two parameters for them. I think this can lead to confusion and it's hard for people to reason about the expected behavior when you only set one of them. I prefer to just have one parameter called `num_parallel_execution_threads`, and just make the default value the same as you use in `Common::Constants`. And add a comment in `Constants` to say that the value will be overrode when the `SettingsManager` is enabled. Also add comments to `IS_COUNTERS_ENABLED` and `IS_PIPELINE_METRICS_ENABLED` which you may modify in `UpdateFromSettingsManager`.

script/model/type.py
```diff
@@ -54,6 +39,51 @@ class OpUnit(enum.IntEnum):
...
+    CPU_ID = 18,
+    CPU_CYCLES = 19,
+    INSTRUCTIONS = 20,
+    CACHE_REF = 21,
+    CACHE_MISS = 22,
+    REF_CPU_CYCLES = 23,
+    BLOCK_READ = 24,
+    BLOCK_WRITE = 25,
+    MEMORY_B = 26,
+    ELAPSED_US = 27
```
malin1993ml, line 86: If it's not too hard to refactor, can we still have another `Target` enum class for all of these? They should not really be called features (*labels*, rather) and they're not specific to the execution engine either. I think it's more accurate if we can separate them.

script/model/training_util/data_transforming_util.py
```diff
@@ -66,13 +66,13 @@ def _tuple_num_linear_log_predict_transform(x, y):
 
 def _tuple_num_log_train_transform(x, y):
     # Transform down the target according to the log tuple num value in the input
-    tuple_num = np.copy(x[:, data_info.TUPLE_NUM_INDEX])
+    tuple_num = np.copy(x[:, data_info.INPUT_CSV_INDEX[ExecutionFeature.EST_OUTPUT_ROWS]])
     return y / (np.log2(tuple_num) + 1e-6)[:, np.newaxis]
```
malin1993ml, line 70: Change the `1e-6` to `1` here. This is a minor model bug fix I have in my branch. But I anticipate that my branch will take even more time to be eventually merged than your branch, so let's just fix it here.

script/model/training_util/data_transforming_util.py
```diff
@@ -66,13 +66,13 @@ def _tuple_num_linear_log_predict_transform(x, y):
...
 
 
 def _tuple_num_log_predict_transform(x, y):
     # Transform up the target according to the log tuple num value in the input
-    tuple_num = np.copy(x[:, data_info.TUPLE_NUM_INDEX])
+    tuple_num = np.copy(x[:, data_info.INPUT_CSV_INDEX[ExecutionFeature.EST_OUTPUT_ROWS]])
     return y * (np.log2(tuple_num) + 1e-6)[:, np.newaxis]
```
malin1993ml, line 76: Same here. Change `1e-6` to `1`.

malin1993ml (changes requested):
The refactoring looks pretty good! I have a few more cleanup requests and then I think we're good on the modeling side!

script/model/type.py
```diff
@@ -54,6 +39,51 @@ class OpUnit(enum.IntEnum):
...
+    QUERY_ID = 0,
+    PIPELINE_ID = 1,
+
+    # # features
+    NUM_FEATURES = 2,
+    FEATURES = 3,
+
+    # input features
+    EXEC_MODE = 4,
+    EST_OUTPUT_ROWS = 5,
```
malin1993ml, line 60: Oh another minor thing, can you rename this to just `NUM_ROWS`? Since it really represents the number of rows processed but not the number of output rows. I guess you need to modify the c++ side as well.

mbutrovich:
My only comment right now is that the PR description is missing a brief "how". It says what it does, but not how or what changes roll downstream to everyone else. Are there changes to the semantics of the execution engine, the metrics framework, etc.?

script/model/type.py
17zhangw: Separated!

src/execution/compiler/compilation_context.cpp
```diff
@@ -161,7 +164,15 @@ void CompilationContext::GeneratePlan(const planner::AbstractPlanNode &plan) {
     codegen_.GetPipelineOperatingUnits()->RecordOperatingUnit(pipeline->GetPipelineId(), std::move(features));
 
     pipeline->Prepare(query_->GetExecutionSettings());
-    pipeline->GeneratePipeline(&main_builder, query_id_t{unique_id_});
+    {
+      util::RegionVector<ast::FunctionDecl *> pipeline_decls(query_->GetContext()->GetRegion());
+      for (auto &[_, op] : ops_) {
+        (void)_;
```
lmwnshn, line 170: C++ sucks. You can ignore this comment.

util/execution/tpl.cpp
```diff
@@ -118,7 +120,11 @@ static void CompileAndRun(const std::string &source, const std::string &name = "
 
   // Generate test tables
   sql::TableGenerator table_generator{&exec_ctx, db_main->GetStorageLayer()->GetBlockStore(), ns_oid};
-  table_generator.GenerateTestTables(IS_MINI_RUNNER);
+  table_generator.GenerateTestTables();
+  if (IS_MINI_RUNNER) {
```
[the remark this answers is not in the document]
lmwnshn: If we don't, I would like to keep mini runners stuff out of tpl.

src/include/execution/compiler/operator/output_translator.h
```diff
@@ -37,18 +37,23 @@ class OutputTranslator : public OperatorTranslator {
    */
   void DefineHelperStructs(util::RegionVector<ast::StructDecl *> *decls) override;
 
-  /** Set up counters for Lin's models. */
-  void InitializeQueryState(FunctionBuilder *function) const override;
-
   /**
-   * Perform the main work of the translator.
+   * @param pipeline The current pipeline.
```
lmwnshn, line 41: I suppose these comments aren't actually necessary since they don't add anything over the override (if we're going to be consistent with TeardownPipelineState not having any)

src/include/execution/compiler/operator/operator_translator.h
```diff
@@ -146,6 +154,42 @@ class OperatorTranslator : public ColumnValueProvider {
    */
   virtual void BeginPipelineWork(const Pipeline &pipeline, FunctionBuilder *function) const {}
 
+  /**
+   * Function to initialize relevant countfers.
```
lmwnshn, line 158: Typo: counters

lmwnshn: Same with RecordCounters below.
And optionally if you want to throw a period at the end of pipeline's parameter comment that'd be nice.

src/include/execution/compiler/operator/hash_aggregation_translator.h
```diff
@@ -196,6 +214,29 @@ class HashAggregationTranslator : public OperatorTranslator, public PipelineDriv
 
   // The number of output rows from the aggregation.
   StateDescriptor::Entry num_agg_outputs_;
+
+  // TBB can run multiple tasks using the same thread local state. For counter
```
lmwnshn, line 218: me see big descriptive comment, me like

Thanks! Good writeup

src/include/execution/compiler/codegen.h
```diff
@@ -1157,12 +1180,12 @@ class CodeGen {
    * Call \@sorterInit(). Initialize the provided sorter instance using a memory pool, comparison
    * function and the struct that will be materialized into the sorter instance.
    * @param sorter The sorter instance.
-   * @param mem_pool The memory pool instance.
+   * @param exec_ctx ExecutionContext
```
lmwnshn, line 1183: Nit: comment

src/include/execution/compiler/codegen.h
```diff
@@ -580,7 +580,7 @@ class CodeGen {
    * @param table_oid The OID of the table to be scanned.
    * @param col_oids The column OIDs from the table that should be scanned.
    * @param query_state The query state pointer.
-   * @param exec_ctx The execution context that we are running in.
+   * @param exec_ctx ExecutionContext
```
lmwnshn, line 583: Nit: comment

src/include/execution/ast/type.h
```diff
@@ -10,6 +10,7 @@
 #include "execution/sql/storage_interface.h"
 #include "execution/util/region.h"
 #include "execution/util/region_containers.h"
+#include "metrics/metrics_manager.h"
```
lmwnshn, line 13: Unnecessary include?

src/include/brain/operating_unit_util.h
```diff
@@ -167,13 +167,27 @@ class OperatingUnitUtil {
...
+   * Determines whether the operating unit type is a blocking OU
+   * @param feature OperatingUnitType to consider
+   * @returns blocking or not
+   */
+  static bool IsOperatingUnitTypeBlocking(ExecutionOperatingUnitType feature);
+
+  /**
+   * Gets the non-parallel type for the OU f
+   * @param feature Parallel OU
+   * @returns Corresponding non-parallel OU or INVALID
```
lmwnshn, line 187: return not returns

src/include/brain/operating_unit_util.h
```diff
@@ -167,13 +167,27 @@ class OperatingUnitUtil {
 
   /**
    * Whether or not an operating unit type can be merged
-   * @param f OperatingUnitType to consider
+   * @param feature OperatingUnitType to consider
    * @returns mergeable or not
```
lmwnshn, line 171: return not returns

src/include/brain/operating_unit.h
```diff
@@ -247,6 +364,13 @@ class PipelineOperatingUnits {
     return itr->second;
   }
 
+  /**
+   * Checks whether a certain pipeline exists
+   * @param pipeline Pipeline Identifier
+   * @returns if exist or not
```
lmwnshn, line 370: return not returns

src/include/brain/operating_unit.h
```diff
@@ -135,6 +198,11 @@ class ExecutionOperatingUnitFeature {
     return sum / mem_factors_.size();
   }
 
+  /**
+   * @returns number of iterations as a reference
```
lmwnshn, line 202: return not returns

src/include/brain/operating_unit.h
```diff
@@ -116,11 +164,26 @@ class ExecutionOperatingUnitFeature {
...
    */
   size_t GetCardinality() const { return cardinality_; }
 
+  /**
+   * @returns num concurrent as a reference
+   */
+  size_t &GetNumConcurrent() { return num_concurrent_; }
+
+  /**
+   * @returns num concurrent
```
lmwnshn, line 183: return not returns

src/include/brain/operating_unit.h
```diff
@@ -116,11 +164,26 @@ class ExecutionOperatingUnitFeature {
...
+   */
+  size_t &GetCardinality() { return cardinality_; }
+
   /**
    * @returns estimated cardinality
    */
   size_t GetCardinality() const { return cardinality_; }
 
+  /**
+   * @returns num concurrent as a reference
```
lmwnshn, line 178: return not returns

lmwnshn: wait, we return a reference? if this is to modify, consider making it use a setter instead?

CLion doesn't seem to pick up any users of this..

lmwnshn: ditto comments for the other getters above in this file

src/execution/sema/sema_builtin.cpp
```diff
@@ -1229,14 +1258,14 @@ void Sema::CheckBuiltinTableIterParCall(ast::CallExpr *call) {
     return;
   }
 
-  // The fourth argument is the execution context.
+  // The second argument must be a pointer to the execution context.
```
lmwnshn, line 1261: I think this is really the fourth

src/execution/sema/sema_builtin.cpp
```diff
@@ -1229,14 +1258,14 @@ void Sema::CheckBuiltinTableIterParCall(ast::CallExpr *call) {
...
   const auto exec_ctx_kind = ast::BuiltinType::ExecutionContext;
-  if (!IsPointerToSpecificBuiltin(call_args[3]->GetType(), exec_ctx_kind)) {
+  if (!IsPointerToSpecificBuiltin(call->Arguments()[3]->GetType(), exec_ctx_kind)) {
     ReportIncorrectCallArg(call, 3, GetBuiltinType(exec_ctx_kind)->PointerTo());
     return;
   }
 
-  // The fifth argument is the scanner function.
+  // The fourth argument is the scanner function.
```
lmwnshn, line 1268: And this the fifth

src/execution/sema/sema_builtin.cpp
```diff
@@ -3089,19 +3127,85 @@ void Sema::CheckBuiltinCall(ast::CallExpr *call) {
...
     case ast::Builtin::ExecutionContextStartPipelineTracker:
     case ast::Builtin::ExecutionContextEndPipelineTracker:
-    case ast::Builtin::ExecutionContextGetFeature:
-    case ast::Builtin::ExecutionContextRecordFeature: {
+    case ast::Builtin::ExecOUFeatureVectorInitialize: {
       CheckBuiltinExecutionContextCall(call, builtin);
       break;
     }
+    case ast::Builtin::ExecOUFeatureVectorReset: {
+      if (!CheckArgCount(call, 1)) {
```
lmwnshn, line 3149: Could you move this into its own function above?

src/execution/exec/execution_context.cpp
```diff
@@ -67,78 +99,89 @@ void ExecutionContext::EndPipelineTracker(query_id_t query_id, pipeline_id_t pip
...
 const parser::ConstantValueExpression &ExecutionContext::GetParam(const uint32_t param_idx) const {
   return (*params_)[param_idx];
 }
 
+void ExecutionContext::RegisterHook(size_t hook_idx, HookFn hook) {
+  TERRIER_ASSERT(hook_idx < hooks_.capacity(), "Incorrect number of reserved hooks");
+  hooks_[hook_idx] = hook;
+}
+
+void ExecutionContext::InvokeHook(size_t hookIndex, void *tls, void *arg) {
```
lmwnshn, line 179: snake_pls notCamel

I thought the linter would catch this, actually. @lmwnshn check

src/execution/sql/sorter.cpp
```diff
@@ -188,9 +196,24 @@ void Sorter::SortParallel(const ThreadStateContainer *thread_state_container, co
     Sort();
 
     // Finish
+    exec_ctx_->InvokeHook(post_hook, tls, this);
     return;
   }
 
+  // For the purpose of metrics recording, we assume that computing the splitters and
```
lmwnshn, line 203: This is useful. Thanks

lmwnshn (changes requested):
LGTM! Nice work. Minor nitpicks / clarifications, and we should be good to go

(I assume it works since @malin1993ml is using it, I haven't tested minirunners with it on in dev4)

githubxxcc:
Is it just me or someone else also experiencing build errors on mac for target like `mini_runners`?

util/execution/tpl.cpp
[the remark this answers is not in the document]
17zhangw: Purged.

src/include/main/db_main.h
```diff
@@ -730,13 +731,14 @@ class DBMain {
...
       std::unique_ptr<metrics::MetricsManager> metrics_manager = std::make_unique<metrics::MetricsManager>();
-      if (metrics_query_trace_) metrics_manager->EnableMetric(metrics::MetricsComponent::QUERY_TRACE, 0);
-      if (metrics_pipeline_) metrics_manager->EnableMetric(metrics::MetricsComponent::EXECUTION_PIPELINE, 9);
-      if (metrics_transaction_) metrics_manager->EnableMetric(metrics::MetricsComponent::TRANSACTION, 0);
-      if (metrics_logging_) metrics_manager->EnableMetric(metrics::MetricsComponent::LOGGING, 0);
-      if (metrics_gc_) metrics_manager->EnableMetric(metrics::MetricsComponent::GARBAGECOLLECTION, 0);
-      if (metrics_bind_command_) metrics_manager->EnableMetric(metrics::MetricsComponent::BIND_COMMAND, 0);
-      if (metrics_execute_command_) metrics_manager->EnableMetric(metrics::MetricsComponent::EXECUTE_COMMAND, 0);
+      if (query_trace_metrics_) metrics_manager->EnableMetric(metrics::MetricsComponent::QUERY_TRACE, 0);
+      // TODO(wz2): If using parallel, we might want to record all per-task recordings.
```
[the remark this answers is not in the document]
17zhangw: This is actually a bit of an old comment --- patched with a setting to control sampling.

benchmark/runner/mini_runners.cpp
```diff
@@ -2207,7 +1721,9 @@ void InitializeRunnersState() {
                              .SetUseExecution(true)
                              .SetUseTrafficCop(true)
                              .SetUseNetwork(true)
-                             .SetNetworkPort(terrier::runner::port);
+                             .SetUseSettingsManager(true)
+                             .SetSettingsParameterMap(std::move(param_map))
+                             .SetNetworkPort(terrier::runner::settings.port_);
 
   db_main = db_main_builder.Build().release();
```
githubxxcc, line 1729: If I run `mini_runners --benchmark_filter=SEQ0`, it will take forever to run (I actually never got the chance to see it finish). 

As Lin suggested, I could comment these lines out. Just wondering if it would be this going forward, or there is other way to generate the traces.

src/include/execution/sql/functions/runners_functions.h
```diff
@@ -25,10 +25,10 @@ class EXPORT MiniRunnersFunctions {
     if (num_tuples.val_ < 2) return;
 
     // We are already in an output slot
+    auto *output_buffer = ctx->OutputBufferNew();
 
     static_assert(sizeof(Integer) == sizeof(Real));
-    auto output_buffer = ctx->GetOutputBuffer();
-    for (auto row = 0; row < num_tuples.val_ - 2; row++) {
+    for (auto row = 0; row < num_tuples.val_ - 1; row++) {
```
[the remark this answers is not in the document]
17zhangw: Yes. `num_tuples.val_ - 1` is the correct terminating condition.

src/brain/operating_unit_recorder.cpp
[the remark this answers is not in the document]
17zhangw: Opened as Issue #1241.

src/include/settings/settings_param.h
```diff
@@ -60,6 +64,7 @@ class ParamInfo {
         callback_(callback) {}
 
  private:
+  friend void terrier::runner::InitializeRunnersState();
```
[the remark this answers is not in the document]
17zhangw: So the problem with using the settings manager is that `use_metrics_thread` and `wal_enable` currently do not have callbacks to change them. I added the friend to be able to override the underlying values directly. The alternative could potentially be to modify the `argc/argv` to add the necessary arguments for gflags to set the values. Thoughts @lmwnshn?

src/include/execution/exec/execution_context.h
```diff
@@ -40,31 +50,34 @@ class EXPORT ExecutionContext {
    * @param schema the schema of the output
    * @param accessor the catalog accessor of this query
    * @param exec_settings The execution settings to run with.
+   * @param metrics_manager The metrics manager for recording metrics
    */
   ExecutionContext(catalog::db_oid_t db_oid, common::ManagedPointer<transaction::TransactionContext> txn,
-                   const OutputCallback &callback, const planner::OutputSchema *schema,
+                   OutputCallback callback, const planner::OutputSchema *schema,
```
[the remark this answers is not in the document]
17zhangw: Due to life-cycle reasons, I constructed the `OutputCallback` prior to constructing the `ExecutionContext`.

benchmark/runner/mini_runners.cpp
17zhangw: I'm not sure of the use-case exactly but there is this flag `--mini_runner_rows_limit` which will control the maximum sizes of tables generated (which could potentially speed up the loading phase).

src/brain/operating_unit.cpp
```diff
@@ -4,4 +4,61 @@ namespace terrier::brain {
...
+        }
+        default:
+          TERRIER_ASSERT(false, "Invalid feature attribute.");
+          return;
+      }
+    }
+  }
+
+  TERRIER_ASSERT(did_find, "No matching feature was found");
+  (void)did_find;  // silence release warning
```
lmwnshn, line 61: We prefer UNUSED_ATTRIBUTE. Only (void)blah if it is part of a structured binding.

src/execution/sql/table_vector_iterator.cpp
```diff
@@ -138,7 +138,7 @@ bool TableVectorIterator::ParallelScan(uint32_t table_oid, uint32_t *col_oids, u
   timer.Start();
 
   // Execute parallel scan
-  size_t num_threads = std::max(exec_ctx->GetExecutionSettings().GetNumberofThreads(), 0);
+  size_t num_threads = std::max(exec_ctx->GetExecutionSettings().GetNumberofParallelExecutionThreads(), 0);
```
lmwnshn, line 141: GetNumberOfParallelExecutionThreads?

src/include/settings/settings_defs.h
```diff
@@ -182,71 +182,81 @@ SETTING_bool(
...
+    pipeline_metrics_enable,
     "Metrics collection for the ExecutionEngine pipelines (default: false).",
     false,
     true,
     terrier::settings::Callbacks::MetricsPipeline
 )
 
+SETTING_int(
+    pipeline_metrics_interval,
+    "Sampling rate of metrics collection for the ExecutionEngine pipelines (default: 0).",
```
lmwnshn, line 242: What does 0/1/2/... mean? 0 = nothing, 10 = 100%? Could you update the comment here?

src/include/execution/exec/output.h
```diff
@@ -90,7 +91,7 @@ class EXPORT OutputBuffer {
   uint32_t num_tuples_;
   uint32_t tuple_size_;
   byte *tuples_;
-  OutputCallback callback_;
+  const OutputCallback &callback_;
```
lmwnshn, line 94: Could you document the conversation we had about const &'ing the outputcallback here?

lmwnshn (changes requested):
Nice! we are so close to being free

src/include/execution/exec/output.h
17zhangw: While const &'ing the OutputCallback is fine here, the way OutputCallback is setup might need to change. From a local build of master, it would appear that `std::function<>` copies the `OutputWriter` thus leading to an incorrect `OutputWriter::NumRows()` count (i.e always 0).

(comment here for reference)

17zhangw:
This branch does have a bug with parallel memory tracking (see #1250). From a brief discussion with @lmwnshn, the plan is to first get #1208 in and then patch #1250 in a separate PR.

lmwnshn (approved):
LGTM. we are free^!

^ Terms and conditions apply. Subject to merge conflicts.

src/include/execution/exec/output.h
```diff
@@ -91,6 +91,18 @@ class EXPORT OutputBuffer {
...
+  /**
+   * This is defined as a const OutputCallback & since the caller is able to guarantee the
+   * life-time of the OutputCallback. Furthermore, this relies on the assumption that a
+   * std::function<> can be invoked by multiple threads without corrupting any data internal
+   * to the std::function<>. If this assumption is proven wrong at a future date, this
+   * const &-ing will need to be revisited.
+   *
+   * An additional constraint is that if and only if multiple OutputBuffers are used (i.e
+   * parallel output to network), the OutputCallback function must be thread-safe. There is
+   * no serialization in invoking of callback_ between multiple OutputBuffers.
```
lmwnshn, line 104: Though I wonder if you meant synchronization here.

mbutrovich:
Thanks for adjusting that output latch!

test/include/execution/sql_test.h
```diff
@@ -51,17 +51,11 @@ class SqlBasedTest : public TplTest {
   common::ManagedPointer<storage::BlockStore> BlockStore() { return block_store_; }
 
   std::unique_ptr<exec::ExecutionContext> MakeExecCtx(exec::OutputCallback *callback = nullptr,
-                                                      const planner::OutputSchema *schema = nullptr,
-                                                      bool force_serial = false) {
```
lmwnshn: ah, a default parameter. that's why it escaped noticed. :eyes: 

(fyi we try to avoid adding default parameters too)

lmwnshn (approved):
LGTM^

^ :eyes:

mbutrovich:
I get a memory leak in `compiler_test` when I run this branch locally. Not sure why ASAN didn't symbolize anything though...
```
=================================================================
==109718==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 6168 byte(s) in 3 object(s) allocated from:
    #0 0x7f402a1aab47 in operator new[](unsigned long) (/lib/x86_64-linux-gnu/libasan.so.5+0x10fb47)
    #1 0x7f4024f9e66f  (/lib/x86_64-linux-gnu/libtbb.so.2+0x2766f)

SUMMARY: AddressSanitizer: 6168 byte(s) leaked in 3 allocation(s).
```

mbutrovich:
Here's a heap use-after-free on re-run:
[asan.txt](https://github.com/cmu-db/noisepage/files/5418716/asan.txt)

src/include/settings/settings_defs.h
```diff
@@ -185,63 +185,82 @@ SETTING_bool(
...
 
+SETTING_int(
+    pipeline_metrics_interval,
+    "Sampling rate of metrics collection for the ExecutionEngine pipelines with 0 = 100%, 1 = 50%, "
+    "9 = 10%, X = 1/(X+1)% (default: 0 for 100%).",
+    0,
+    0,
+    10,
+    true,
+    terrier::settings::Callbacks::NoOp
```
malin1993ml, line 251: I think you need to add a real callback for this one. Otherwise this only have effects when the system starts, and won't do anything when you change it dynamically. You probably need to add a new function in `MetricsManager` for the callback to call and change the sample interval accordingly.

All of my previous comments on the modeling part are addressed!

17zhangw: Right. This makes sense. Added the callback.

17zhangw:
> I get a memory leak in `compiler_test` when I run this branch locally. Not sure why ASAN didn't symbolize anything though...
> 
> ```
> =================================================================
> ==109718==ERROR: LeakSanitizer: detected memory leaks
> 
> Direct leak of 6168 byte(s) in 3 object(s) allocated from:
>     #0 0x7f402a1aab47 in operator new[](unsigned long) (/lib/x86_64-linux-gnu/libasan.so.5+0x10fb47)
>     #1 0x7f4024f9e66f  (/lib/x86_64-linux-gnu/libtbb.so.2+0x2766f)
> 
> SUMMARY: AddressSanitizer: 6168 byte(s) leaked in 3 allocation(s).
> ```

Talked with this with @lmwnshn. Believe this to be the known tbb leak.
In order to run the `compiler_test` with suppressions locally:
```
LSAN_OPTIONS=suppressions=PATH_TO_BUILD_SUPPORT_DATA/lsan_suppressions.txt ./test/compiler_test
```

The heap use-after-free has to do with races in the test output consumers.

src/include/settings/settings_defs.h
malin1993ml: Thanks! But this is getting a bit tricky now because `Callbacks::MetricsPipeline` will enable the pipeline metrics with 0 sample interval blindly and ignoring the `pipeline_metrics_interval` we set. To get this really working, I think you need to add a new `SetMetricsSampleInterval(const MetricsComponent component, uint32_t sample_interval)` function in `MetricsManager` for the callback to call to change `sample_interval_`. And then add a new `EnableMetric(const MetricsComponent component)` function in `MetricsManager` for `Callbacks::MetricsPipeline` to call, which does not change `sample_interval_` in `MetricsManager`. Make sure you call `SetMetricsSampleInterval` in `DBMain` if `SettingsManager` is enabled, because there can be the case that we do not start the system with pipeline metrics, but enable it later.

I would recommend adding test cases for this in `metrics_test` if you want to add this feature. There's already a `ToggleSettings` test case where you can augment to test the sample interval setting in different scenarios. You can also add a `GetMetricsSampleInterval` function in `MetricsManager` to let you probe the `MetricsManager` state. Also, we happen not to have test for the pipeline metrics in `metrics_test`, so this is a good opportunity to add some now. We don't need to test the actual features being recorded because we already have tests for that, but we can test whether the pipeline metrics are successfully recorded and whether the number of records is correct.

src/include/settings/settings_defs.h
```diff
@@ -185,63 +185,82 @@ SETTING_bool(
...
     false,
     true,
     terrier::settings::Callbacks::MetricsPipeline
 )
 
+SETTING_int(
+    pipeline_metrics_interval,
+    "Sampling rate of metrics collection for the ExecutionEngine pipelines with 0 = 100%, 1 = 50%, "
+    "9 = 10%, X = 1/(X+1)% (default: 0 for 100%).",
+    0,
```
malin1993ml, line 247: I think we should keep this as 9? Let's not change this frequently without specific reason.

src/include/settings/settings_defs.h
```diff
@@ -279,4 +298,22 @@ SETTING_string(
     true,
     terrier::settings::Callbacks::NoOp
 )
+
+SETTING_int(
+    num_parallel_execution_threads,
+    "Number of threads for parallel query execution (default: 0)",
+    0,
```
malin1993ml, line 305: What does 0 mean... no thread to execute the parallel task? It's a bit confusing to me.

src/include/main/db_main.h
```diff
@@ -639,13 +639,14 @@ class DBMain {
...
-    bool metrics_query_trace_ = false;
-    bool metrics_pipeline_ = false;
-    bool metrics_transaction_ = false;
-    bool metrics_logging_ = false;
-    bool metrics_gc_ = false;
-    bool metrics_bind_command_ = false;
-    bool metrics_execute_command_ = false;
+    bool query_trace_metrics_ = false;
+    bool pipeline_metrics_ = false;
+    uint32_t pipeline_metrics_interval_ = 0;
```
malin1993ml, line 644: Same here. I would keep this as 9.

malin1993ml (commented):
The changes look good! Some minor things.

src/include/main/db_main.h
17zhangw: Reverted to 9.

src/include/settings/settings_defs.h
17zhangw: Reverted to 9.

src/include/settings/settings_defs.h
17zhangw: I changed this to 1 (which matches the behavior of 0 on my system).

malin1993ml (approved):

Issue #1210, opened 2020-09-28
[BUG] TPL struct offset computations do not account for padding when compiled.


lmwnshn:
Found by @17zhangw.

## Summary
TPL structs add padding in between elements. This padding does not appear to be present in LLVM's defined structs. This breaks any `Lea` bytecode. This bug is present on master.

## Environment

**OS**: Ubuntu 20.04 (LTS).

**Compiler**: GCC 9.3.0

**LLVM version**: LLVM 8.0.1

## Steps to Reproduce

@17zhangw found this bug when recording to a custom `ExecOUFeatureVector` type. We boiled it down to this minimal example.

Run this TPL program with `tpl -sql -print-tbc`.

```
struct State {
    count  : uint32
    sorter : Sorter
    count1 : uint32
    count2 : int32
}
fun sizers(a : uint32, b : Sorter, c : uint32, d: ExecutionContext, e : State) -> nil {}
fun setUpState(execCtx: *ExecutionContext, state: *State) -> nil {
    state.count = 0
    state.count1 = 42
}
fun tearDownState(state: *State) -> nil {
}
fun main(execCtx: *ExecutionContext) -> int32 {
    var state: State
    state.count2 = 500
    setUpState(execCtx, &state)
    tearDownState(&state)
    return state.count2
}
```

`state.count2` is set to 500 and never modified thereafter, so we should always return 500.

### Expected Behavior

Output should be the same in all three modes.

```
[2020-09-28 15:40:06.098] [execution_logger] [info] VM main() returned: 500
[2020-09-28 15:40:06.098] [execution_logger] [info] ADAPTIVE main() returned: 500
[2020-09-28 15:40:06.441] [execution_logger] [info] JIT main() returned: 500
```

### Actual Behavior

Compiled mode is different.

```
[2020-09-28 15:40:06.098] [execution_logger] [info] VM main() returned: 500
[2020-09-28 15:40:06.098] [execution_logger] [info] ADAPTIVE main() returned: 500
[2020-09-28 15:40:06.441] [execution_logger] [info] JIT main() returned: 42
```

### Notes

The TPL struct is {4 byte int + 4 byte padding, 160 byte sorter, 4 byte int + 4 byte padding, 4 byte int + 4 byte padding}.

However, LLVM says the struct has size 176 (4 + 160 + 4 + 4, and 4 bytes padding to make it divisible by 8).

[...]

PR #1214, opened 2020-09-29
Fixes struct padding for TPL #1210

17zhangw:
Fixes Issue #1210 by inserting extra padding members when constructing structs.


pmenon (changes requested):
First pass. You also need to add tests.

src/execution/ast/context.cpp
```diff
@@ -277,6 +277,34 @@ MapType *MapType::Get(Type *key_type, Type *value_type) {
   return map_type;
 }
 
+Field StructType::CreatePaddingElement(uint32_t size, Context *ctx, uint32_t *remainder) {
+  // We only need a placeholder name. The identifier for the placeholder
+  // does not have to be unique since TPL code will never refer to a
+  // placeholder field and LLVM IR does not rely on field names.
+  ast::Identifier name = ctx->GetIdentifier("__field$0$");
```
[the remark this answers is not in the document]
lmwnshn: I think the comment is incorrect but the `GetIdentifier` function makes names unique automatically within the context.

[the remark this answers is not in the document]
pmenon: `ast::Context::GetIdentifier()` doesn't unique-ify names.

src/execution/ast/type.cpp
```diff
@@ -105,9 +105,10 @@ MapType::MapType(Type *key_type, Type *val_type)
 // ---------------------------------------------------------
 
 StructType::StructType(Context *ctx, uint32_t size, uint32_t alignment, util::RegionVector<Field> &&fields,
-                       util::RegionVector<uint32_t> &&field_offsets)
+                       util::RegionVector<Field> &&unpadded_fields, util::RegionVector<uint32_t> &&field_offsets)
```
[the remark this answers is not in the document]
17zhangw: Created Issue #1218 to track.

pmenon (changes requested):
Much cleaner! A few small changes, then it should be good.

sample_tpl/struct-padding.tpl
```diff
@@ -0,0 +1,20 @@
...
+    state.count1 = 42
+}
+fun tearDownState(state: *State) -> nil {
+}
+fun main(execCtx: *ExecutionContext) -> int32 {
+    var state: State
+    state.count2 = 500
+    setUpState(execCtx, &state)
+    tearDownState(&state)
+    return state.count2
```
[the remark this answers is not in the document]
17zhangw: So I tried to simplify this test into primitive types by removing the Sorter instance and using basic types, arrays, and nested structures but they did not return incorrect results when run against the master branch. I noticed that it's only when using certain non-primitives (i.e Sorter or TableVectorIterator for instance) that an incorrect result will be returned.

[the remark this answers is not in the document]
pmenon: I suppose that's good news.

Strip this down further to the bare essentials. For example, `ExecutionContext` isn't needed. Also, add checks in this same file for arrays and nested structures containing `Sorters`.

src/execution/ast/context.cpp
```diff
@@ -30,6 +30,16 @@
 // #include "execution/util/csv_reader.h" Fix later.
 #include "execution/util/execution_common.h"
 
+namespace {
+
+terrier::execution::ast::Field CreatePaddingElement(uint32_t id, uint32_t size, terrier::execution::ast::Context *ctx) {
+  terrier::execution::ast::Identifier name = ctx->GetIdentifier("__pad$" + std::to_string(id) + "$");
+  auto *pad_type = terrier::execution::ast::BuiltinType::Get(ctx, terrier::execution::ast::BuiltinType::Int8);
+  return terrier::execution::ast::Field(name, terrier::execution::ast::ArrayType::Get(size, pad_type));
```
pmenon, line 38: Move this into the `terrier::execution::ast` namespace directly above `StructType::Get()` where it's needed. Refactoring this way should clean it up.

test/execution/ast_type_test.cpp
```diff
@@ -19,6 +19,27 @@ class TypeTest : public TplTest {
 
   ast::Identifier Name(const std::string &s) { return Ctx()->GetIdentifier(s); }
 
+  ast::Identifier PaddingName(uint32_t offset) { return Ctx()->GetIdentifier("__pad$" + std::to_string(offset) + "$"); }
+
+  void CheckArrayType(StructType *struct_type, ast::Identifier name, uint32_t offset, uint32_t length,
```
pmenon, line 24: This interface seems a little unintuitive w.r.t. what it does. Types don't have names, and the name can be derived from the offset, so isn't required in the interface. But, at a higher level, I think what you want is:

```
/** Checks if the element at the given offset is padding of the given size. */
void CheckIsPadding(StructType *type, std::size_t offset, std::size_t expected_padding_size);
```

The implementation pulls out the field by name (using the name derived from the offset) and (1) checks it exists and (2) checks the type is indeed an array of the given size and whose elements are bytes using an another `void CheckIsArrayType(Type *type, std::size_t num_elems, Type *elem_type);`

src/execution/ast/context.cpp
```diff
@@ -300,18 +310,27 @@ StructType *StructType::Get(Context *ctx, util::RegionVector<Field> &&fields) {
...
+        size = new_size;
       }
 
       // Update size and calculate alignment
       field_offsets.push_back(size);
-      size += field.type_->GetSize();
-      alignment = std::max(alignment, field.type_->GetAlignment());
+      all_fields.emplace_back(field);
+      size += field_type->GetSize();
+      alignment = std::max(alignment, field_type->GetAlignment());
```
pmenon, line 333: Either use `field_align` or remove that variable.

src/execution/ast/context.cpp
```diff
@@ -300,18 +310,27 @@ StructType *StructType::Get(Context *ctx, util::RegionVector<Field> &&fields) {
...
+    util::RegionVector<Field> all_fields(ctx->GetRegion());
     util::RegionVector<uint32_t> field_offsets(ctx->GetRegion());
+    all_fields.reserve(fields.size());
+    field_offsets.reserve(fields.size());
     for (const auto &field : fields) {
+      auto *field_type = field.type_;
+
       // Check if the type needs to be padded
-      uint32_t field_align = field.type_->GetAlignment();
+      uint32_t field_align = field_type->GetAlignment();
```
pmenon, line 321: Nit: `const`-ify?

pmenon (changes requested):
Small refactors.

test/execution/ast_type_test.cpp
```diff
@@ -19,6 +19,25 @@ class TypeTest : public TplTest {
 
   ast::Identifier Name(const std::string &s) { return Ctx()->GetIdentifier(s); }
 
+  void CheckIsArrayType(Type *type, size_t num_elem, Type *elem_type) {
+    EXPECT_EQ(type->GetTypeId(), Type::TypeId::ArrayType);
```
pmenon, line 23: Change to `EXPECT_TRUE(type->IsArrayType());`

src/execution/ast/context.cpp
```diff
@@ -32,6 +32,12 @@
 
 namespace terrier::execution::ast {
 
+Field CreatePaddingElement(uint32_t id, uint32_t size, Context *ctx) {
```
pmenon, line 35: I should've been clearer. Move this into an anonymous namespace **within** `terrier::execution::ast` like [this](https://github.com/cmu-db/noisepage/blob/master/src/execution/ast/type_printer.cpp#L11). Also, move this right above `StructType::Get()`

pmenon (changes requested):
Much better. Last few changes, then it's good to go.

sample_tpl/struct-padding.tpl
```diff
@@ -0,0 +1,52 @@
...
+    state.count2 = 500
+    state.embed.a = 42
+    return state.count2
+}
+
+fun main() -> int32 {
+    var sum : int32
+    sum = test1() + test2() + test3()
+    return sum
+}
```
pmenon, line 52: Can you add some comments on what's going on here?

17zhangw: Added comments.

pmenon (approved):
LGTM.

PR #1217, opened 2020-10-01
Simplify select

jrolli:
The existing code for `SelectIntoBuffer` had not been updated since we corrected a logic error in aborts (the referenced ABA problem).  Since aborts are unlinked and GC'd in the same manner as commits, we no longer have to worry about having to do multiply reads of the in-place data to guard against a dirty read.  If we have a dirty read, we are guaranteed to see and apply an UNDO record that corrects them when checking the version pointer.


mbutrovich:
At a high level we're arguing that the version chain is no longer a deque (which it sorta could have been if a txn aborted and unlinked itself from the insertion end). It's basically just a queue, append-only at one end, prune from the other end.

mbutrovich (approved):
LGTM. Just need Jenkins to stop misbehaving.

PR #1224, opened 2020-10-04
CMake refactor.

lmwnshn:
## TL;DR

- 2.8x speedup on average (me Ubuntu, Matt Mac).
- Unit tests are run in parallel now. **This requires specifying resource groups if unit tests conflict, e.g., if [two tests](https://github.com/cmu-db/noisepage/pull/1224/files#diff-af3b638bc2a3e6c650974192a53c7291R782) try to listen on port 15721**. We can run 100/142 of our tests in < 10 seconds, 138/142 in 1 minute, all tests in 3 minutes.
- Old CMake infrastructure out except for `FindLibevent` and `FindTBB`.
- All new CMake logic in one file, root [CMakeLists](https://github.com/lmwnshn/terrier/blob/die_cmake_dieeee/CMakeLists.txt).
- You should be running `cmake .. -GNinja -DNOISEPAGE_USE_ASAN=On -DNOISEPAGE_UNITY_BUILD=On` for best performance. If you find yourself doing incremental builds a lot, consider disabling unity builds: `cmake .. -GNinja -DNOISEPAGE_USE_ASAN=On`.

## CMake Options

```
option(NOISEPAGE_GENERATE_COVERAGE
        "Enable C++ code coverage."
        OFF)

option(NOISEPAGE_USE_ASAN
        "Enable ASAN, a fast memory error detector. https://clang.llvm.org/docs/AddressSanitizer.html"
        OFF)

option(NOISEPAGE_USE_JEMALLOC
        "Link jemalloc instead of system malloc. https://github.com/jemalloc/jemalloc"
        OFF)

option(NOISEPAGE_UNITTEST_OUTPUT_ON_FAILURE
        "Verbose output for unittests when they fail. Warning: on jumbo, this is VERY verbose!"
        OFF)

option(NOISEPAGE_UNITY_BUILD
        "Enable Unity builds for much faster compilation. https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html"
        OFF)
```

## Details

This PR rips out most of the Apache Arrow stuff that we yoinked early on with our own handwritten CMake.

This was motivated by having to modify CMake to add a third party dependency. It was not clear where to start, what to read, or what to modify. Moreover, previously every single third party dependency was globbed into our own `TERRIER_SRCS`, which is a problem because we compile with `-Wall -Werror`. People have spent time fixing up the bwtree and libpg_query code in the past to deal with this, which defeats the purpose of calling something third party. This also meant that third party was sharing all our compilation flags whether we liked it or not.

At a high level, CMake has "old style" CMake and "modern" CMake.
- Old style: Global variables such as CMAKE_CXX_FLAGS that are applied to everything you build. Typically requires CMake 2.8 (released 2010).
- Modern: Explicitly designate what you want to build as targets, apply the compile and link flags you want explicitly. Targets can inherit public flags from other targets, there can be interface and private flags as well (basically OOP). Typically requires CMake 3.1x.

By switching to modern CMake, we:

- Decouple compiling dependencies from compiling NoisePage : You can compile dependencies with what you want. In this case, we're currently compiling dependencies with `-fPIC -fsanitize=address -fvisibility=hidden`. We were already doing this, but involuntarily and with more flags in the past.
- Discover weird coupling : Tests used to depend on benchmark headers and benchmarks used to depend on test headers. It turns out the benchmark header was being brought in for silly reasons, so we don't have that dependency now. All dependencies are now explicitly spelled out instead of accidentally leaking include directories between compilation targets.
- Get smaller CMake footprint: With relatively detailed comments, there's 1000 lines of CMake in the root CMakeLists.txt. It was designed to be readable top to bottom. If you read that top to bottom and understand it, you understand our entire build system. Also, adding most dependencies should only take one line of `add_noisepage_dep`, perhaps with a `third_party/deps_override/NAME_CMakeLists.txt`.
- Enable the use of Unity builds. At a high level, globs multiple translation units into one so that it compiles much faster.

**To future maintainers, it takes up to a week to get familiar with modern CMake. Resources are linked at the top of CMakeLists.txt.**

Caveats:

- I didn't bother with generator expressions that distinguish between the libraries being built, installed, etc.

## Performance

I expect most people to be aiming to build the DBMS binary. The suggested configuration is with ninja (**requires you to pass -GNinja to CMake**), lld (default on Ubuntu, disabled on Mac), and unity builds (**requires you to pass -DNOISEPAGE_UNITY_BUILD=On to CMake**).

**Max speed build DBMS binary: ninja + lld + unity builds**
Wan on Ubuntu: 1135 seconds to 381 seconds (2.97x speedup)
Matt on Mac: 1793 seconds to 625 seconds (2.86x speedup)

The **huge win is unity builds**, but that **may suck for incremental recompiles**. So that's a flag, see above.
lld is pretty good. Apple's linker does a decent job of linking fast as well.
Ninja shaves off two to four minutes, but you don't really need to install it if you don't want to.

Older preliminary timings on my desktop below with notes.

- `ninja` can go higher than `-j 9` without crashing my computer, `make` can't. `ninja` should be a little faster in practice.
- I was convinced by @gonzalezjo that FetchContent is the way to go. Basically, we fetch our dependencies at configure time instead of embedding them into `third_party`. This is considered idiomatic and it does prevent us from mucking around with external dependencies too much, also makes tracking and upgrading versions really easy. (If anyone has a bandwidth quota, let me know?) But this means the new Ninja setup is building a couple more libraries than the old setup.

**DBMS binary : ninja + lld**
Old, `cmake .. && time make -j9 terrier`: 847 seconds aka 14 min
New, `cmake .. -GNinja && time ninja -j9 noisepage`: 740 seconds aka 12 min

**Build everything : ninja + lld**
Old, `cmake .. && time make -j9`: 2651 seconds aka 44 min
New, `cmake .. -GNinja && time ninja -j9`: 1771 seconds aka 29 min

**Max speed build everything : ninja + lld + unity builds**
Max speed (unity builds, default ninja parallel jobs is 14 on my machine):
New, `cmake .. -GNinja && time ninja noisepage`: 345 seconds aka 5 min
New, `cmake .. -GNinja && time ninja`: 1546 seconds aka 26 min
The bottleneck with unity builds is having so many different test files.
Globbing test files together produces further speedup.


build-support/run-clang-tidy.py
```diff
@@ -261,6 +261,9 @@ def main():
     files = [make_absolute(entry['file'], entry['directory'])
              for entry in database]
 
+    # NoisePage hack: don't check third party dependencies.
```
lmwnshn, line 264: Added this to run-clang-tidy.py because clang-tidy seems to barf if you stick a .clang-tidy in the `_deps/src/DEPENDENCY_NAME` folder.

build-support/tpl_bytecode_handlers_ir_compiler.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
+"""
+Usage: tpl_bytecode_handlers_ir_compiler.py LLVM_COMPILER CMAKE_BINARY_DIR BCH_CPP BCH_OUT
```
lmwnshn, line 3: See [this](https://github.com/cmu-db/noisepage/pull/1224/files#diff-af3b638bc2a3e6c650974192a53c7291R500-R505).

src/common/strong_typedef.cpp
```diff
@@ -47,8 +47,10 @@ template class StrongTypeAlias<terrier::catalog::tags::trigger_oid_t_typedef_tag
 template class StrongTypeAlias<terrier::catalog::tags::type_oid_t_typedef_tag, uint32_t>;
 template class StrongTypeAlias<terrier::catalog::tags::view_oid_t_typedef_tag, uint32_t>;
 template class StrongTypeAlias<terrier::common::tags::action_id_t_typedef_tag, uint64_t>;
-template class StrongTypeAlias<terrier::execution::tags::query_id_t_typedef_tag, uint32_t>;
+template class StrongTypeAlias<terrier::execution::tags::feature_id_t_typedef_tag, uint32_t>;
```
lmwnshn, line 50: These were missing. I guess it didn't blow up before because the definitions were being leaked around.

src/include/execution/tpl.h
```diff
@@ -0,0 +1,4 @@
+#pragma once
```
lmwnshn, line 1: I removed the configure_file step that used tpl.h.in to build this file at configure time.

src/execution/sql/aggregation_hash_table.cpp
```diff
@@ -10,6 +10,7 @@
 
 #include "common/error/exception.h"
 #include "common/math_util.h"
+#include "count/hll.h"
```
lmwnshn, line 13: libcount is now count.

third_party/versions.txt
```diff
@@ -15,38 +15,7 @@
 # specific language governing permissions and limitations
 # under the License.
 
-# Toolchain library versions
-
-GTEST_VERSION=1.8.0
-GFLAGS_VERSION=2.2.0
-GBENCHMARK_VERSION=1.4.1
-
 # Library, Version, Commit Hash
```
lmwnshn, line 18: We are missing versions for bwtree, flatbuffers, libpg_query.

lmwnshn (commented):
All interesting changes outside of the top-level CMakeLists.txt.

CMakeLists.txt
```diff
@@ -472,7 +472,7 @@ target_link_libraries(noisepage_shared PUBLIC noisepage_objlib)             # Co
 
 add_executable(noisepage src/main/terrier.cpp)
 target_link_libraries(noisepage noisepage_static)
-set_target_properties(noisepage PROPERTIES CXX_EXTENSIONS OFF)
+set_target_properties(noisepage PROPERTIES CXX_EXTENSIONS OFF ENABLE_EXPORTS ON)
```
lmwnshn, line 475: I have no idea why ENABLE_EXPORT is necessary for the LLVMEngine's TPLMemoryManager to find its symbols. Any explanations I can throw into CMakeLists.txt would be greatly appreciated.

third_party/versions.txt
lmwnshn: Per Matt: bwtree is whatever was in Peloton.
I'll add that we have since modified it: prior to this PR I added instrumentation for Lin's stuff, with this PR I disable ASAN on the BwTree.

gonzalezjo:
I'm still not sure why Unity builds even work by default. What happens if you crank up the batch size?

pmenon:
Nice clean up! Ninja is legit. Just FYI, both GCC and Clang won't automatically colorize output if they detect output isn't going to a terminal. Since Ninja buffers output from the compiler, GCC/Clang won't generate colored text. To solve, you need to explicitly request colorization. I have this option in my CMake file:

```
if (${TPL_COLORIZE_OUTPUT})
    if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fdiagnostics-color=always")
    elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fcolor-diagnostics")
    else ()
        message(WARNING "Cannot colorize output for unknown compiler: ${CMAKE_CXX_COMPILER_ID}")
    endif ()
endif ()
```

You can add a similar default-on option, or just always enable color.

src/include/optimizer/operator_node_contents.h
```diff
@@ -126,7 +122,7 @@ class OperatorNodeContents : public BaseOperatorNodeContents {
    * Utility method for applying visitor pattern on the underlying operator
    * @param v operator visitor for visitor pattern
    */
-  void Accept(common::ManagedPointer<OperatorVisitor> v) const override;
+  void Accept(common::ManagedPointer<OperatorVisitor> v) const override { v->Visit(reinterpret_cast<const T *>(this)); }
```
lmwnshn, line 125: This seems to address the problems that @gonzalezjo found.

mbutrovich:
Just watching this in CI: need to update the Jenkinsfile for the new environment variable names, along with whatever the ideal build commands are now.

lmwnshn (commented):
@17zhangw want to glance over this commit? Optimizer things.

https://github.com/cmu-db/noisepage/pull/1224/commits/5abb65b1ed00b9ac17826773a1d0893bb7256ddb

jasonbak:
Yo wtf this is insane.

jasonbak:
I give it a 🚢

third_party/versions.txt
jasonbak: Yup, +1 on what Matt said.

PR #1228, opened 2020-10-09
Compiled storage accessors

jrolli:
# Description

Implements the `InsertInto` function as a compiled, per-table
function that is created at instantiation time.  Currently only
updates the DataTableTests to ensure correctness to allow input
on larger code structure and standards for this project.

### Remaining Tasks
Again, you should only create PR once you are reasonably confident you are near completion. However, if there are some tasks still remaining before the PR is ready to merge, please create a checklist to track active progress. An [example](https://github.com/cmu-db/noisepage/issues/1031) from a PR by @thepinetree follows:

:pushpin: TODOs:
- [ ] Update remaining tests & benchmarks to initialize LLVM
- [ ] Create `DELETE` function
- [ ] Create `UPDATE` and `SELECT` functions (partial interpretation base on columns present in projection)

## Performance
This PR has the potential to greatly affect (hopefully improve) performance for inserts and may affect other transactions based off other storage layer decision (notably nullmap elisions for `NOT NULL` columns could significantly improve writes).  A more specific performance evaluation will occur once those decisions are solidified.

### Index Maintenance done in `DataTable` accessors

Since our general discussion is that concurrent modifications to a `DataTable` and associated indexes is generally unsafe, we could solve two problems in one shot by having inserts & deletes to all indexes managed in the compiled function.  This will eventually force a write-lock on the table in the catalog (once we move the compiled functions into there) and reduce the complexity required by the execution layer.

### Exposing helpers to enable query-time compilation of `Select` and `Update` accessors

These functions rely on additional query-time inputs (which columns are touched) that cannot be a priori made into straight-line code.  However, with some cooperation with the execution layer, we could enable compiled queries to generate specific implementations of these functions that leverage the "static" nature of their usage of these functions.

### Compactor optimization

The compactor currently materializes a tuple into a projected row during its move logic.  With these compiled accessors, we should be able to adapt the logic to create code that moves data between tuple slots without materializing the tuple in between.


src/include/storage/data_table.h
```diff
@@ -302,8 +309,8 @@ class DataTable {
   bool SelectIntoBuffer(common::ManagedPointer<transaction::TransactionContext> txn, TupleSlot slot,
                         RowType *out_buffer) const;
 
-  void InsertInto(common::ManagedPointer<transaction::TransactionContext> txn, const ProjectedRow &redo,
-                  TupleSlot dest);
+  void (*InsertInto)(common::ManagedPointer<transaction::TransactionContext> txn, const ProjectedRow &redo,
+                     TupleSlot dest);
```
jrolli, line 313: This is a style point for discussion.  Clang-Tidy treats this as a member rather than a method.  In this case it is private and not much of an issue, but I think in general we should keep function pointers which will be used as methods named with the same convention as regular methods.

src/include/transaction/transaction_context.h
```diff
@@ -101,9 +101,10 @@ class TransactionContext {
    * @param slot the TupleSlot inserted
    * @return a persistent pointer to the head of a memory chunk large enough to hold the undo record
    */
-  storage::UndoRecord *UndoRecordForInsert(storage::DataTable *const table, const storage::TupleSlot slot) {
-    byte *const result = undo_buffer_.NewEntry(sizeof(storage::UndoRecord));
-    return storage::UndoRecord::InitializeInsert(result, finish_time_.load(), slot, table);
+  static storage::UndoRecord *UndoRecordForInsert(common::ManagedPointer<TransactionContext> txn,
+                                                  storage::DataTable *const table, const storage::TupleSlot slot) {
+    byte *const result = txn->undo_buffer_.NewEntry(sizeof(storage::UndoRecord));
+    return storage::UndoRecord::InitializeInsert(result, txn->finish_time_.load(), slot, table);
```
jrolli, line 107: I changed this to make the arguments needed to call from LLVM IR explicit.  I think it is fine to revert it, but I think whatever we do should be consistent across all the `UndoRecordFor[...]` functions.

src/storage/tuple_accessor.cpp
```diff
@@ -0,0 +1,222 @@
...
+    throw std::runtime_error(err);
+  }
+  TERRIER_ASSERT(engine_ != nullptr, "Engine initialization should not fail");
+
+  datatable->InsertInto = (void (*)(terrier::common::ManagedPointer<terrier::transaction::TransactionContext>,
+                                    const terrier::storage::ProjectedRow &,
+                                    terrier::storage::TupleSlot))engine_->getFunctionAddress(INSERT_INTO_NAME);
+}
+
+void TupleAccessor::BuildInsertInto(llvm::Module *module, DataTable *datatable) {
```
jrolli, line 47: This is what I think needs the most attention.  If this were ASM, I'd have a comment on every line explaining what that line is doing in addition to guideposts.  I am thinking that is equally appropriate here, but I wanted to get input before investing that level of energy.

jrolli (commented):
Highlighting points of interest.

jrolli:
Closing because this will be refactored to use TPL for the code generation and will probably end up as several incremental PRs.

PR #1231, opened 2020-10-12
Ubuntu 20.04 fixes for packages.sh and Jenkinsfile

mbutrovich:
- Packages.sh is now checking against 20.04.
- Add `libpqxx-dev` to the Linux package list rather than `wget`ing it.
- Jenkins will use the `terrier:focal` image now.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -35,7 +35,7 @@ pipeline {
                 stage('ubuntu-18.04/gcc-7.3.0 (Debug/format/lint/censored)') {
                     agent {
                         docker {
-                            image 'ubuntu:bionic'
+                            image 'terrier:focal'
                         }
                     }
                     steps {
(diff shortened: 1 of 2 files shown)


Jenkinsfile
```diff
@@ -32,10 +32,10 @@ pipeline {
                     }
                 }
 
-                stage('ubuntu-18.04/gcc-7.3.0 (Debug/format/lint/censored)') {
+                stage('ubuntu-20.04/gcc-9.3 (Debug/format/lint/censored)') {
```
crd477, line 35: Good.  I like this convention of major.minor better as it's less likely to be inconsistent with what the package manager installs throughout the 20.04 support cycle.

script/installation/packages.sh
```diff
@@ -41,10 +39,10 @@ LINUX_BUILD_PACKAGES=(\
   "cmake" \
   "doxygen" \
   "git" \
-  "g++-7" \
```
crd477, line 44: Is `g++-9` guaranteed to be picked up by something else, like `build-essential`?

crd477 (approved):
I made two small changes affecting only comments in `packages.sh`, otherwise LGTM.

script/installation/packages.sh
mbutrovich: https://packages.ubuntu.com/focal/build-essential

Looks like it's >=9.2.

PR #1233, opened 2020-10-12
Ubuntu 20.04 OLTPBench testing update

bialesdaniel:
# OLTPBench test script update due to python upgrade

## Description
With the Ubuntu upgrade the python version got updated to 3.8.5. We were using `platform.linux_distribution()` to get information about the OS that the tests were run on. We were storing this metadata along with the test results. `platform.linux_distribution()` has been deprecated. I switched over to using the python package [distro](https://pypi.org/project/distro/) which provides the same functionality. I tested this locally and it looks good. Without this fix the nightly OLTPBenchmarks will not report their results correctly. There is no impact to the main terrier build job. 

## Remaining tasks
None

## Performance
N/A
## Further work
None


mbutrovich (approved):
LGTM

PR #1234, opened 2020-10-13
SQL Explain

abigalekim:
# SQL Explain

## Description
Implements a version of SQL Explain (see https://www.postgresql.org/docs/9.1/sql-explain.html for reference) so that EXPLAIN ran on queries outputs the physical execution plan, in JSON. I implemented this feature by modifying the traffic cop module and several other files in the postgres folder to ensure that the physical plan (in JSON) is extracted and then outputted properly via the Postgresql network protocol.

This PR adds functionality by implementing a new feature, and is planned to be used to verify an index population.

### Remaining Tasks
:pushpin: TODOs:
- [x] Implement SQL EXPLAIN to output the JSON physical plan
- [x] Add basic unit tests in Junit to verify that EXPLAIN runs without error.
- [ ] Write unit tests in the modules I made changes to.

## Performance
This feature does not deal with performance features.

## Further Work
One suggested improvement was adding more comprehensive testing with EXPLAIN to determine whether the physical plans of the database match the query optimizer.

Notes: Jenkins currently doesn't work because a jumbotest fails but I don't think our code intersects.


src/traffic_cop/traffic_cop.cpp
```diff
@@ -275,6 +278,43 @@ TrafficCopResult TrafficCop::ExecuteDropStatement(
                                                common::ErrorCode::ERRCODE_DATA_EXCEPTION)};
 }
 
+TrafficCopResult TrafficCop::ExecuteExplain(const common::ManagedPointer<network::ConnectionContext> connection_ctx,
+                                            const common::ManagedPointer<network::PostgresPacketWriter> out,
+                                            const common::ManagedPointer<planner::AbstractPlanNode> physical_plan,
+                                            const terrier::network::QueryType query_type) const {
+  TERRIER_ASSERT(connection_ctx->TransactionState() == network::NetworkTransactionStateType::BLOCK,
```
mbutrovich, line 285: Let's refactor the body down to:
```
  TERRIER_ASSERT(connection_ctx->TransactionState() == network::NetworkTransactionStateType::BLOCK,
                 "Not in a valid txn. This should have been caught before calling this function.");
  // Need a single column to write back to the client
  std::vector<planner::OutputSchema::Column> output_columns;
  output_columns.emplace_back("QUERY PLAN", type::TypeId::VARCHAR, nullptr);
  out->WriteRowDescription(output_columns, {network::FieldFormat::text});

  // Dump to JSON string, wrap in StringVal, write the data row to the client
  const std::string plan_string = physical_plan->ToJson().dump(4);
  const execution::sql::StringVal plan_string_val =
      execution::sql::StringVal(plan_string.c_str(), plan_string.length());
  out->WriteDataRow(reinterpret_cast<const byte *const>(&plan_string_val), output_columns,
                    {network::FieldFormat::text});

  return {ResultType::COMPLETE, 0};
```
This cuts down on dummy variables, eliminates the switch with a single case, omits the `query_type` check that doesn't buy us much, and clarifies the logic.

mbutrovich: Note that passing the expression in as `nullptr` requires us to change postgres_packet_writer.cpp::78 to
```
    const auto &name = columns[i].GetExpr() == nullptr || columns[i].GetExpr()->GetAlias().empty()
                           ? columns[i].GetName()
                           : columns[i].GetExpr()->GetAlias();
```
To handle that expression being `nullptr`. To be honest, I don't understand why we have expressions in the output columns at all.

src/traffic_cop/traffic_cop_util.cpp
```diff
@@ -25,7 +26,13 @@ std::unique_ptr<planner::AbstractPlanNode> TrafficCopUtil::Optimize(
     std::unique_ptr<optimizer::AbstractCostModel> cost_model, const uint64_t optimizer_timeout) {
   // Optimizer transforms annotated ParseResult to logical expressions (ephemeral Optimizer structure)
   optimizer::QueryToOperatorTransformer transformer(accessor, db_oid);
-  auto logical_exprs = transformer.ConvertToOpExpression(query->GetStatement(0), query);
+  common::ManagedPointer<parser::SQLStatement> query_statement = query->GetStatement(0);
+
+  if (query_statement->GetType() == parser::StatementType::EXPLAIN) {
```
mbutrovich, line 31: Can you add a comment what's happening here? Presumably you're extracting the underlying statement for optimization?

mbutrovich (changes requested):
This is off to a great start! I mostly have style nits, though we do have one major bug. If you try to `EXPLAIN` on a statement that shouldn't bind, the system crashes in the optimizer. For example:
```
terrier=# explain insert into foo values (1),(2),(3);
Time: 169.740 ms
!> 
```
I suspect you have to bind the underlying SQL statement like you did with the optimizer. Let me know if you have any questions about my feedback. Thanks!

src/include/network/network_util.h
```diff
@@ -51,6 +51,15 @@ class NetworkUtil {
 
   /**
    * @param type query type from the parser
+<<<<<<< HEAD
```
mbutrovich, line 54: Weird artifact from a merge? I think the function is dead code anyway right now.

src/network/postgres/postgres_packet_writer.cpp
```diff
@@ -71,8 +71,9 @@ void PostgresPacketWriter::WriteRowDescription(const std::vector<planner::Output
     const auto field_format = field_formats[i < field_formats.size() ? i : 0];
 
     // TODO(Matt): Figure out how to get table oid and column oids in the OutputSchema (Optimizer's job?)
-    const auto &name =
-        columns[i].GetExpr()->GetAlias().empty() ? columns[i].GetName() : columns[i].GetExpr()->GetAlias();
+    const auto &name = columns[i].GetExpr() == nullptr || columns[i].GetExpr()->GetAlias().empty()
```
mbutrovich, line 74: Will you add a comment why it can now be `nullptr` so we remember why we did this?

src/include/network/network_defs.h
```diff
@@ -115,6 +115,7 @@ enum class QueryType : uint8_t {
   QUERY_DROP_TRIGGER,
   QUERY_DROP_SCHEMA,
   QUERY_DROP_VIEW,
+  QUERY_EXPLAIN,
```
mbutrovich, line 118: Do we believe `EXPLAIN` should fall into non-transactional instead? I guess I can test this.

mbutrovich (changes requested):
Minor stuff right now. Need to explore transactional semantics of EXPLAIN.

src/include/network/network_defs.h
abigalekim: Yes, it should be non-transactional because since SQL EXPLAIN just generates the query plan, it does not actually modify or access the database and thus I think it being non-transactional is fine.

mbutrovich:
So `EXPLAIN` is transactional, by which I mean: an `EXPLAIN` on an invalid query (e.g., doesn't bind) will force a running transaction to abort, and you cannot issue an `EXPLAIN` within a failed transaction. I'll see if we follow these semantics.

For example, here's `EXPLAIN` unable to run in a failed transaction on Postgres:
```sql
matt=# begin;
BEGIN
Time: 0.084 ms
matt=*# explain select * from foo;
                      QUERY PLAN                       
-------------------------------------------------------
 Seq Scan on foo  (cost=0.00..35.50 rows=2550 width=4)
(1 row)

Time: 1.322 ms
matt=*# create table foo (id int);
2020-12-01 12:18:44.933 EST [6243] ERROR:  relation "foo" already exists
2020-12-01 12:18:44.933 EST [6243] STATEMENT:  create table foo (id int);
ERROR:  relation "foo" already exists
Time: 1.132 ms
matt=!# explain select * from foo;
2020-12-01 12:18:49.553 EST [6243] ERROR:  current transaction is aborted, commands ignored until end of transaction block
2020-12-01 12:18:49.553 EST [6243] STATEMENT:  explain select * from foo;
ERROR:  current transaction is aborted, commands ignored until end of transaction block
Time: 0.141 ms
matt=!# 
```
And here's `EXPLAIN` triggering transaction failure:
```sql
matt=# begin;
BEGIN
Time: 0.146 ms
matt=*# explain select * from foo;
2020-12-01 12:20:18.814 EST [6243] ERROR:  relation "foo" does not exist at character 23
2020-12-01 12:20:18.814 EST [6243] STATEMENT:  explain select * from foo;
ERROR:  relation "foo" does not exist
LINE 1: explain select * from foo;
                              ^
Time: 1.070 ms
matt=!# create table foo (id int);
2020-12-01 12:20:25.873 EST [6243] ERROR:  current transaction is aborted, commands ignored until end of transaction block
2020-12-01 12:20:25.873 EST [6243] STATEMENT:  create table foo (id int);
ERROR:  current transaction is aborted, commands ignored until end of transaction block
Time: 0.173 ms
matt=!# 
```

mbutrovich:
I think we're good. Nice job!

mbutrovich (approved):
LGTM, will merge once this commit gets through CI. Nice job!

mbutrovich:
We should put that tracefile test back because it caught a legit bug. `EXPLAIN DELETE from temp where true;` crashes the system in what looks like new `stats_calculator` logic.

mbutrovich:
Blocked on #1545.

src/network/postgres/postgres_network_commands.cpp
```diff
@@ -461,6 +463,8 @@ Transition DescribeCommand::Exec(const common::ManagedPointer<ProtocolInterprete
     } else if (portal->GetStatement()->GetQueryType() == network::QueryType::QUERY_SELECT) {
       out->WriteRowDescription(portal->OptimizeResult()->GetPlanNode()->GetOutputSchema()->GetColumns(),
                                portal->ResultFormats());
+    } else if (portal->GetStatement()->GetQueryType() == network::QueryType::QUERY_EXPLAIN) {
+      out->WriteExplainRowDescription();
     } else {
       out->WriteNoData();
```
lmwnshn, line 469: ^ @abigalekim this is your enemy

mbutrovich:
We were barfing on DESCRIBE in extended query protocol. [#aee6483 ](https://github.com/cmu-db/noisepage/pull/1234/commits/aee6483b34150e1ad0206e27d1d32a8b93d23f0c) resolves it.

mbutrovich (approved):
We did it everyone! LGTM!

PR #1235, opened 2020-10-13
WAL: Remove redundant call between serializer and disk consumer

mbutrovich:
Probably a merge error somewhere in the 10,000 line diff of #774.

Also renamed a variable to make its purpose more clear. This is part of a more exhaustive review of WAL performance.


lmwnshn (approved):
So this removes an if statement that is always false? LGTM.

mbutrovich:
> So this removes an if statement that is always false? LGTM.
Exactly.

Issue #1230, opened 2020-10-09
TPC-H : Nested query optimization stuck at DeriveStats task


xinzhu-cai:
# Bug Report

## Summary
TPCH Query 20 does not stop executing. It happens at DeriveStats tasks after applying TopDown and BottonUp rewrites. It keeps running DeriveStats task for groups 22 and 23.

## Environment
I don't think it's environment-specific. I run on macOS 10.14+ with Clang 8.0+

## Steps to Reproduce
- Compile with the following args: -DCMAKE_BUILD_TYPE=Release ..
- Run terrier executable
- Connect to the terrier terminal by executing : psql -h localhost -p 15721 -d terrier
- Create the tables from "https://github.com/oltpbenchmark/oltpbench/blob/master/src/com/oltpbenchmark/benchmarks/tpch/ddls/tpch-postgres-ddl.sql" .
- Run the query
```
-- TPC-H Query 20

select
        s_name,
        s_address
from
        supplier,
        nation
where
        s_suppkey in (
                select
                        ps_suppkey
                from
                        partsupp
                where
                        ps_partkey in (
                                select
                                        p_partkey
                                from
                                        part
                                where
                                        p_name like 'forest%'
                        )
                        and ps_availqty > (
                                select
                                        0.5 * sum(l_quantity)
                                from
                                        lineitem
                                where
                                        l_partkey = ps_partkey
                                        and l_suppkey = ps_suppkey
                                        and l_shipdate >= date '1994-01-01'
                                        and l_shipdate < date '1995-01-01'
                        )
        )
        and s_nationkey = n_nationkey
        and n_name = 'CANADA'
order by
        s_name;
```
This query never stops running. We expect it to return an empty result.

PR #1236, opened 2020-10-15
 DeriveStats Support for Semi Join

xinzhu-cai:
This is to fix Issue #1230 

- Add ChildStatsDeriver::Visit for LogicalSemiJoin

- Add StatsCalculator::Visit for LogicalSemiJoin

- Enable ChildStatsDeriver::Visit for LogicalAggregateAndGroupBy


src/optimizer/statistics/stats_calculator.cpp
```diff
@@ -149,6 +149,55 @@ void StatsCalculator::Visit(const LogicalInnerJoin *op) {
   // TODO(boweic): calculate stats based on predicates other than join conditions
 }
 
+void StatsCalculator::Visit(const LogicalSemiJoin *op) {
+  // Check if there's join condition
```
preetansh, line 153: Is the functionality exactly same to the LogicalInnerJoin? Can we move the common code to a function rather than repeating the code or do we expect them to be different later?

preetansh (changes requested):

src/optimizer/statistics/stats_calculator.cpp
xinzhu-cai: This might be different in the future.

src/optimizer/statistics/stats_calculator.cpp
```diff
@@ -149,6 +149,58 @@ void StatsCalculator::Visit(const LogicalInnerJoin *op) {
...
+            (left_child_group->HasColumnStats(right_col) && right_child_group->HasColumnStats(left_col))) {
+          curr_rows /= std::max(std::max(left_child_group->GetNumRows(), right_child_group->GetNumRows()), 1);
+        }
+      }
+    }
+    root_group->SetNumRows(static_cast<int>(curr_rows));
+  }
+
+  size_t num_rows = root_group->GetNumRows();
+  // Uncomment when required_cols_ is supported in system
```
lmwnshn, line 181: Add a TODO here, otherwise people might not see it?

lmwnshn (changes requested):

lmwnshn (approved):
LGTM.

src/optimizer/statistics/stats_calculator.cpp
```diff
@@ -149,6 +149,58 @@ void StatsCalculator::Visit(const LogicalInnerJoin *op) {
...
+        if ((left_child_group->HasColumnStats(left_col) && right_child_group->HasColumnStats(right_col)) ||
+            (left_child_group->HasColumnStats(right_col) && right_child_group->HasColumnStats(left_col))) {
+          curr_rows /= std::max(std::max(left_child_group->GetNumRows(), right_child_group->GetNumRows()), 1);
+        }
+      }
+    }
+    root_group->SetNumRows(static_cast<int>(curr_rows));
+  }
+
+  // TODO(elenac): For now we comment this because required_cols_ is always empty in the system
```
mbutrovich, line 180: Why is it always empty? When will it not always be empty? What's the timeline on all of this not being dead code? Next PR? Please elaborate this comment for people other than the original author.

mbutrovich (changes requested):
Is there a test case that can be written for this?

src/optimizer/statistics/stats_calculator.cpp
xinzhu-cai: This is not always empty. I wrongly ignored the following case
https://github.com/cmu-db/noisepage/blob/f0a6e9f2cbc5ad3c9fe79846d9193806460b5403/src/optimizer/optimizer_task.cpp#L228
Uncommented this block to handle such cases. 

These test cases in nested-query.test covers it:
`SELECT pno FROM part WHERE pno in (SELECT pno FROM part WHERE pno in (SELECT shipment.pno FROM shipment where sno >= 1));
`
`
SELECT * FROM shipment WHERE pno <= (SELECT MAX(pno) FROM part) order by pno;
`

mbutrovich (approved):
Addressed my concerns, thanks! Not sure if anyone else needs another pass.

PR #1239, opened 2020-10-17
Abort Action Cleanup functor

preetansh:
Using functors for resource deletion action on abort instead of using RegisterAbortAction(). Also added a functor for efficient addition of abort functions using function pointer instead of std::function().


src/include/transaction/transaction_context.h
```diff
@@ -173,13 +176,27 @@ class TransactionContext {
...
+   */
+  template <typename T>
+  void RegisterAbortCleanupAction(T *resource) {
+    abort_actions_.emplace_back(std::make_unique<TransactionEndCleanupFunctor>(
+        [](void *ptr) {
+          auto specialized = static_cast<T *>(ptr);
+          delete specialized;
+        },
+        resource));
+  }
```
jrolli, line 191: While this change makes calling the framework much easier, I think this comes at a large cost to debugging.  Specifically, it would have made debugging and fixing #1267 significantly harder because this has the side effect of hiding where the deferral was generated/registered.

One advantage of using a lambda at the call-site every time is that stack traces will effectively point to where the action was registered.  On stack traces that registered deletes with this function, there will be no useful information about how/where the action was registered other than the typename (which may not be unique).

lmwnshn:
Is this ready for review or in progress?

PR #1190, opened 2020-09-17
psutil for pid and check port

jazzplato:
Use `psutil` for
- check if a pid exists
- kill processes listening to certain port
Dependency
- added `psutil` to the python package section in the installation script
Jenkinsfile
- added `python3 kill_server_on_port.py 15721` to kill all the lingering processes for macos

PR #1242, opened 2020-10-19
Prevent pyc bytecode generation when sudo in Jenkins.

lmwnshn:
# Prevent pyc bytecode generation when sudo in Jenkins.

## Description

Merging #1190 creates .pyc files as root, preventing Jenkins `deleteDir()` from functioning correctly. Unsure why this was not caught in CI for that.


bialesdaniel:
I think you might need to add the same changes into nightly-jenkinsfile

lmwnshn:
Codecov is drunk.

PR #1243, opened 2020-10-19
collect memory consumption info during OLTPBench tests

jazzplato:
# Collect memory consumption info during OLTPBench tests
This PR is about memory info collection during the OLTPBench tests

## Summary
This PR include
- A standalone python script `collect_mem_info.py` to be run with `sudo` to collect the memory info of the given pid.
- A thread for each of the test case in the TestCase class which run the memory info collection function periodically and accumulate the results.
- For memory info collection, only the following metrics will be collected
  - Resident set size: `rss`
  - Virtual Memory size: `vms`
- After finishing each test case, the memory info will be reported to the performance storage system Django API as part of the incremental metrics.
- This memory info collection feature can be used for **all** kinds of tests as long as the base test class is extended from the `TestServer` class.
  - It is enabled by default for all OLTPBench tests
  - It is disabled by default for all the other kinds of tests
- The memory info collection interval can be configured via command line arguments `--collect-mem-freq`, whose default value is 5 (i.e. collect memory info every 5 seconds), which is of the same frequency of the OLTPBench incremental results.


script/testing/oltpbench/reporting/report_result.py
```diff
@@ -4,22 +4,44 @@
...
-def send_results(api_url,username,password,results):
-    result = requests.post(api_url + '/oltpbench/', json=results,auth=(username,password))
+def send_results(api_url, username, password, results):
+    result = requests.post(api_url + '/oltpbench/',
+                           json=results,
+                           auth=(username, password))
     result.raise_for_status()
+
+
+def add_mem_info_to_incremental(metrics, mem_info_dict):
```
bialesdaniel, line 39: This method seems like it would be more appropriate in the parse_data file

script/testing/util/common.py
```diff
@@ -12,6 +12,12 @@
 from util import constants
 
 
+class MemoryInfo:
```
bialesdaniel, line 15: Why do we need a class for memory Info? Its just two data attributes.

script/testing/util/test_server.py
```diff
@@ -153,6 +159,20 @@ def run_test(self, test_case: TestCase):
...
+            self.collect_mem_thread = threading.Timer(
+                self.collect_mem_freq, update_mem_info, [
+                    self.db_process.pid, self.collect_mem_freq,
+                    test_case.mem_info_dict
+                ])
+            # collect the initial memory info
+            update_mem_info(self.db_process.pid, self.collect_mem_freq,
+                            test_case.mem_info_dict)
+            # collect the subsequent memory info
+            self.collect_mem_thread.start()
```
bialesdaniel, line 174: I'm pretty sure this isn't going to do what you want. We need to break up the oltpbench test execution commands. There is a data loading stage in OLTPBench and we don't want to gather the memory info during that, we just want to gather the memory info during the test. I think the flow would be 1. oltpbench load data 2. create mem_info thread 3. start OLTPBench test 4. start mem_info thread 5. When OLTPBench test complete make sure the mem_info is also complete

bialesdaniel: Checkout the `--load` and `--execute` commands https://github.com/oltpbenchmark/oltpbench/wiki I could be wrong but I think we are doing all the commands at once. So I think this might require some more significant changes to the OLTPBenchTestCase code.

script/testing/util/test_server.py
```diff
@@ -153,6 +159,20 @@ def run_test(self, test_case: TestCase):
         # run the pre test tasks
         test_case.run_pre_test()
 
+        # start a thread to collect the memory info if needed
+        if self.collect_mem_info:
+            # spawn a thread to collect memory info
+            self.collect_mem_thread = threading.Timer(
```
bialesdaniel, line 165: This feels like it might help to break into a class or function. I was thinking class just cause I was thinking you instantiate it with the collection frequency int and it creates this timer and then you call `start` and it will gather all the data and store it in something like a `results` attribute and then you can have functions like `get_average_rss()` `get_average_vms()` `populate_incremental_metrics(incremental_oltpbench_results_arr)`. I don't think it needs to be a class but its just how I was thinking about it.

bialesdaniel (changes requested):
I think this will end up gathering memory usage data while OLTPBench is loading the database with data. See my comments in the `test_server.py` file for more details.

script/testing/oltpbench/reporting/report_result.py
jazzplato: done

script/testing/util/common.py
jazzplato: The point here is to organize the info. I can do it in a tuple or just 2 vars, but since these 2 vars are used in multiple functions and files, it is important to keep track of which is which and what do they mean. If you think putting them into a class is a bit overdoing, I can also do them with a `namedtuple`, which serves the same purpose but a bit more flexible and succinct.

jazzplato: Changed to `namedtuple`

script/testing/util/test_server.py
jazzplato: Created a `MemoryMetrics` class for this purpose

script/testing/util/test_server.py
jazzplato: Done. Moved the create and load DB actions into pretest stage.

jazzplato:
A few updates:
- Created a `MemoryMetrics` class for statistics
- Moved the creating and loading DB actions into pretest stage, in order to make the memory consumption metrics more accurate.
- Fixed the bugs of the `threading.Timer`, changed it to `PeriodicTask` class that is newly added

script/testing/util/periodic_task.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        Initialize the PeriodicTask object and schedule the initial Timer
+        """
+        self._timer = None
+        self.function = function
+        self.interval = interval
+        self.args = args
+        self.kwargs = kwargs
+        self.is_running = False
+        self.start()
```
bialesdaniel, line 19: I would think we want to separate the initialization and the start command so that we can initialize and then start when we see fit.

script/testing/util/periodic_task.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def _run(self):
+        """
+        Execute the periodic task and rescheule the next Timer thread
+        """
+        self.is_running = False
```
bialesdaniel, line 25: This isn't intuitive. By setting is_running = False the if statement in `start` is kinda useless

script/testing/util/test_server.py
```diff
@@ -36,6 +38,10 @@ def __init__(self, args):
         self.continue_on_error = self.args.get(
             "continue_on_error", constants.DEFAULT_CONTINUE_ON_ERROR)
 
+        # memory info collection
+        self.collect_mem_info = self.args.get("collect_mem_info", False)
+        self.collect_mem_freq = self.args.get("collect_mem_freq",
```
bialesdaniel, line 43: What if we used the `buckets` attribute from the OLTPBench and brought that into the TestServer?

script/testing/util/test_server.py
```diff
@@ -153,6 +159,16 @@ def run_test(self, test_case: TestCase):
...
         test_case.run_pre_test()
 
+        # start a thread to collect the memory info if needed
+        if self.collect_mem_info:
+            # spawn a thread to collect memory info
+            self.collect_mem_thread = PeriodicTask(
+                self.collect_mem_freq, update_mem_info, self.db_process.pid,
+                self.collect_mem_freq, test_case.mem_metrics.mem_info_dict)
+            # collect the initial memory info
+            update_mem_info(self.db_process.pid, self.collect_mem_freq,
```
bialesdaniel, line 169: Do we need this? It seems like it would collect info when the OLTPBench test isn't running

bialesdaniel (changes requested):
I think you need to merge in the upstream changes cause the mircobenchmarks changed a couple of these files. Overall nice work. This will be great stuff. Don't forget to run some experiments to see if the memory collection affects the performance numbers. Maybe run one of the benchmarks 10 times with the master branch and 10 times with your changes and see if there is a big difference

script/testing/collect_mem_info.py
```diff
@@ -0,0 +1,18 @@
...
+        description="Collect the memory info of the given process IDs.")
+    aparser.add_argument("pids", type=int, nargs="+", help="Pids to collect")
+    args = vars(aparser.parse_args())
+
+    for pid in args.get("pids", []):
+        mem_info = collect_mem_info(pid)
+        rss = mem_info.rss if mem_info else ""
+        vms = mem_info.vms if mem_info else ""
+        res = "{RSS},{VMS}".format(RSS=rss, VMS=vms)
+        print(res)
```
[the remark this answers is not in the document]
jazzplato: yeah that is right. it is a subprocess script whose stdout output will be used and we dont want timstamp or anything else in it

script/testing/oltpbench/run_oltpbench.py
```diff
@@ -16,39 +16,44 @@
...
-
     test_suite_json = load_test_suite(configfile_path)
 
-    max_connection_threads = int(test_suite_json.get('server_args',{}
-                                                ).get('connection_thread_count',
-                                                    str(constants.OLTPBENCH_DEFAULT_CONNECTION_THREAD_COUNT)))
+    # All OLTPBench test collect memory info by default.
+    disable_mem_info = args.get("disable_mem_info", False)
+    args["collect_mem_info"] = not disable_mem_info
+    args["collect_mem_freq"] = constants.OLTPBENCH_DEFAULT_BUCKETS
```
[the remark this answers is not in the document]
jazzplato: no they are 2 distinct values but can be the same only for otlpbench. currently what i did was to use the value of `OLTPBENCH_DEFAULT_BUCKETS` for both the `args.bucket` (which is for oltpbench) and `args.collect_mem_freq1 (which is for memory consumption collection interval)

script/testing/oltpbench/test_case_oltp.py
```diff
@@ -32,34 +31,40 @@ def set_oltp_attributes(self,args):
...
             args.get("loader_threads",
-                          constants.OLTPBENCH_DEFAULT_LOADER_THREADS))
+                     constants.OLTPBENCH_DEFAULT_LOADER_THREADS))
         self.client_time = int(
             args.get("client_time", constants.OLTPBENCH_DEFAULT_TIME))
         self.weights = str(args.get("weights"))
         self.transaction_isolation = str(
             args.get("transaction_isolation",
-                          constants.OLTPBENCH_DEFAULT_TRANSACTION_ISOLATION))
+                     constants.OLTPBENCH_DEFAULT_TRANSACTION_ISOLATION))
```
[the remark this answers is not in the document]
jazzplato: I guess Yapf did it when doing the formating upon save

script/testing/oltpbench/test_case_oltp.py
```diff
@@ -112,20 +114,45 @@ def init_test_case(self):
     def run_pre_test(self):
         self.config_xml_file()
         self.create_result_dir()
+        self.create_and_load_db()
```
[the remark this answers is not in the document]
jazzplato: No, I think we will still do that for every each test case, because this is for `run_pre_test` not `run_pre_suite`

script/testing/oltpbench/test_case_oltp.py
```diff
@@ -112,20 +114,45 @@ def init_test_case(self):
...
+        Create the directory for the result output files if not exists.
+        """
         if not os.path.exists(self.test_result_dir):
             os.makedirs(self.test_result_dir)
 
+    def create_and_load_db(self):
+        """
+        Create the database and load the data before the actual test execution.
+        """
+        cmd = "{BIN} -c {XML} -b {BENCHMARK} --create={CREATE} --load={LOAD}".format(
```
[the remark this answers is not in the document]
jazzplato: Hmm it make sense. But according to the log output, I the creating and loading DB only happened once
```
10-31-2020 08:14:03,624 [common.py:36] INFO : 08:14:03,602 (DBWorkload.java:271) INFO  - ======================================================================
10-31-2020 08:14:03,695 [common.py:36] INFO : 08:14:03,695 (DBWorkload.java:533) INFO  - Creating new TATP database...
10-31-2020 08:14:04,005 [common.py:36] INFO : 08:14:04,004 (DBWorkload.java:535) INFO  - Finished!
10-31-2020 08:14:04,025 [common.py:36] INFO : 08:14:04,022 (DBWorkload.java:536) INFO  - ======================================================================
10-31-2020 08:14:04,034 [common.py:36] INFO : 08:14:04,033 (DBWorkload.java:559) INFO  - Loading data into TATP database with 1 threads...
10-31-2020 08:14:07,283 [common.py:36] INFO : 08:14:07,282 (DBWorkload.java:563) INFO  - Finished!
10-31-2020 08:14:07,310 [common.py:36] INFO : 08:14:07,283 (DBWorkload.java:564) INFO  - ======================================================================
10-31-2020 08:14:07,323 [common.py:36] INFO : 08:14:07,319 (DBWorkload.java:598) INFO  - Skipping benchmark workload execution
```

script/testing/util/constants.py
```diff
@@ -51,6 +52,11 @@
 # OS family
 OS_FAMILY_DARWIN = "darwin"
 
+# Memory info collection
+COLLECT_MEM_FREQ = 5  # collect mem info every 5 seconds by default
+MEM_INFO_SPLITTER = ","
+MEM_INFO_INVALID_TIME = -1
```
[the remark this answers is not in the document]
jazzplato: Ah gotcha, I will remove this.

script/testing/util/mem_metrics.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self):
+        self.mem_info_dict = {}
+        self.INVALID_TIME = MEM_INFO_INVALID_TIME
+
+    def get_avg(self):
+        """
+        Obtain the average value of all the memory info data
+        """
+        # The average cannot be calculated since there is no mem info data
+        if not self.mem_info_dict:
```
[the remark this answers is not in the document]
jazzplato: Yes. In Python all the following will be implicit `False`
- `None` type
- Empty list/set/dict
- `0`
- Empty string `""`

script/testing/util/periodic_task.py
jazzplato: Yes we can, but I think this one is more accurate in the face of time drift. Because every line of code takes time to execute, the 5 sec would be 5 sec + time needed for spawning the thread in real runtime. I tried a number of ways to see which resulted in the least time drift, this method won by only incurring like 10+ ms drift.

script/testing/util/test_server.py
jazzplato: This is tricky, because this is `test_server.py` not `test_oltpbench.py`, other tests will also use this part of code, and junit, for example, does not have this `buckets`. So I think maybe it is better to separate those 2 concepts.

script/testing/util/test_server.py
jazzplato: Yes it is needed. Because the 1st execution of the `PeriodicTask` takes place after 5 sec, and we will miss the 0 sec data point if this line of code is removed.

script/testing/oltpbench/reporting/parse_data.py
```diff
@@ -51,8 +54,47 @@ def parse_files(results_dir):
...
+
+
+def add_incremental_mem_metrics(metrics, mem_metrics):
+    """
+    Add the memory info to the incremental metrics by time
+    """
+    incremental_metrics = metrics.get('incremental_metrics', [])
+    for metrics in incremental_metrics:
+        mem_info = mem_metrics.mem_info_dict.get(
+            metrics.get('time', mem_metrics.INVALID_TIME))
```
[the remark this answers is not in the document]
jazzplato: Gotcha! I will remove it.

script/testing/oltpbench/reporting/report_result.py
```diff
@@ -1,25 +1,36 @@
...
 
 import requests
-from oltpbench.reporting.parse_data import parse_data
+from oltpbench.reporting.parse_data import parse_data, add_mem_metrics
 from oltpbench.constants import PERFORMANCE_STORAGE_SERVICE_API
 
-def report(api_url, server_data, results_dir, username, password, query_mode='simple'):
-    metadata, timestamp, type, parameters, metrics = parse_data(results_dir)
+
+def report(api_url,
```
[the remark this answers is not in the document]
jazzplato: Ahhh... Merge conflicts here I come

script/testing/oltpbench/run_oltpbench.py
```diff
@@ -16,39 +16,44 @@
...
-
     configfile_path = get_configfile_path(args)
-
     test_suite_json = load_test_suite(configfile_path)
 
-    max_connection_threads = int(test_suite_json.get('server_args',{}
-                                                ).get('connection_thread_count',
-                                                    str(constants.OLTPBENCH_DEFAULT_CONNECTION_THREAD_COUNT)))
+    # All OLTPBench test collect memory info by default.
+    disable_mem_info = args.get("disable_mem_info", False)
```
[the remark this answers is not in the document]
jazzplato: Sure, putting the default value in either argparse or here will be fine with me

script/testing/oltpbench/reporting/parse_data.py
```diff
@@ -51,8 +54,47 @@ def parse_files(results_dir):
...
+    add_overall_mem_metrics(metrics, mem_metrics)
+    add_incremental_mem_metrics(metrics, mem_metrics)
+
+
+def add_overall_mem_metrics(metrics, mem_metrics):
+    """
+    Add the average memory metrics to the overal reporting metrics
+    """
+    avg_mem_info = mem_metrics.get_avg()
+    metrics['memory_info'] = {
```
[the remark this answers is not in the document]
jazzplato: Wait, I dont really get it. Do you mean the following JSON schema?
```
metrics['memory_info'] = {
    'avg_rss' : xxx,
    'avg_vms': yyy,
}
```

script/testing/util/test_server.py
bialesdaniel: but wouldn't the 0 sec point be wrong here cause OLTPBench wont be running yet. I wonder what throughput is for 0 sec interval cause it should be 0 right. If its not zero it must be the first chance it gets to gather data after OLTPBench starts.

script/testing/util/test_server.py
bialesdaniel: Yah I'm suggesting moving buckets into the test_server because it now has meaning everywhere. It would mean the memory collection interval (maybe we rename buckets) then we can remove that OLTPBench specific part and use the test_server buckets

script/testing/oltpbench/reporting/parse_data.py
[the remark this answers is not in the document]
bialesdaniel: yah that was my first suggestion. The other was 
```
metrics['memory_info'] = {
    'rss':{
        avg: xxx,
    },
    'vms':{
        avg: xxx,
    },
}
```

script/testing/util/test_server.py
bialesdaniel: What if the PeriodicTask `start` looked like:
```
    def start(self):
        """
        Schedule and start the initial Timer thread
        """
        
        if not self.is_running:
            self.function(*self.args, **self.kwargs)    #This handles the second 0

        self._timer = Timer(self.interval, self._run)

    def _run(self):
        self.is_running = True
        self.start()
        self.function(*self.args, **self.kwargs)
```

script/testing/util/periodic_task.py
bialesdaniel: Really I would have thought something like this would have had less latency between the start and measuring 
```
            self.collect_mem_thread = PeriodicTask(
                self.collect_mem_freq, update_mem_info, self.db_process.pid,
                self.collect_mem_freq, test_case.mem_metrics.mem_info_dict)

        with open(test_case.test_output_file, "w+") as test_output_fd:
            ret_val, _, _ = run_command(test_case.test_command,
                                        test_case.test_error_msg,
                                        stdout=test_output_fd,
                                        stderr=test_output_fd,
                                        cwd=test_case.test_command_cwd)

        self.collect_mem_thread.start() # or is the run_command blocking I forgot. If it is this definitely doesn't work.
```

script/testing/oltpbench/reporting/parse_data.py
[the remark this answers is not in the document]
jazzplato: Thanks! I think this is brilliant!

script/testing/oltpbench/reporting/parse_data.py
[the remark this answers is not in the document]
jazzplato: removed

script/testing/util/constants.py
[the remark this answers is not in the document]
jazzplato: removed

script/testing/util/periodic_task.py
jazzplato: No this would not work. Because the `run_command` is blocking. We need to `start` the thread before the `run_command`

script/testing/oltpbench/reporting/parse_data.py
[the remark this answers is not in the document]
jazzplato: changed as requested.

script/testing/oltpbench/reporting/report_result.py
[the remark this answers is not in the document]
jazzplato: Merged with the master and the conflicts resolved

script/testing/oltpbench/run_oltpbench.py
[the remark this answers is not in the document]
jazzplato: Oh I think argparse make whichever uses `store_true` default value as `False` implicitly already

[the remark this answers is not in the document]
jazzplato: removed the default `False` directly

script/testing/oltpbench/test_case_oltp.py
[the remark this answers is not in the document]
jazzplato: I am keeping it the current way now because I have tested the current approach did create and load the database for every each test case.

script/testing/util/test_server.py
jazzplato: Hmm I dont think so. Because we have 3 tests now and only OLTPBench has bucket concept, JUnit and MicroBench do not even have this concept nor have any incremental metrics. I prefer to keep it the current way to keep the bucket concept within the OLTPBench.

script/testing/util/test_server.py
jazzplato: Good approach! I will do it this way.

jazzplato: Oh wait, there is a problem about the `is_running`. The `is_running` does not mean `is_not_started`, but means whether there is an existing Timer scheduled.

script/testing/util/periodic_task.py
jazzplato: No, the `is_running` means whether the PeriodicTask has any existing Timer thread already schedule or not. It is important not to double schedule a Timer which may mess up the interval and lose track of the lifecycle of one thread (but many more threads whenever the `.start()` is called). It is a safeguard measurement.

script/testing/util/test_server.py
jazzplato: replaced it by `incremental_metric_freq`

script/testing/oltpbench/run_oltpbench.py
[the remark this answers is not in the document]
jazzplato: replaced by `incremental_metric_freq`

bialesdaniel (approved):

script/testing/reporting/parsers/parse_data.py
```diff
@@ -128,13 +131,10 @@ def parse_db_metadata():
...
-    db_metadata = {
-        'noisepage': {
-            'db_version': UNKNOWN_RESULT
-        }
-    }
+    db_metadata = {'noisepage': {'db_version': UNKNOWN_RESULT}}
     try:
-        with open(CURRENT_DIR + '/../../../../src/include/common/version.h') as version_file:
+        with open(CURRENT_DIR +
+                  '/../../../../src/include/common/version.h') as version_file:
```
apavlo, line 137: At some point this is going to break if we ever move this file. I would add a `FIXME` here

script/testing/util/mem_metrics.py
```diff
@@ -0,0 +1,27 @@
+#!/usr/bin/python3
```
apavlo, line 1: This should be `#!/usr/bin/env python3`

script/testing/util/periodic_task.py
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/python3
```
apavlo, line 1: This should be `#!/usr/bin/env python3`

script/testing/util/test_case.py
```diff
@@ -1,9 +1,12 @@
+#!/usr/bin/python3
```
apavlo, line 1: This should be `#!/usr/bin/env python3`

apavlo (changes requested):
Minor tweaks. After this, then this is ready.

script/testing/reporting/parsers/parse_data.py
bialesdaniel: Yah. The good news is we currently don't use this data for anything and it will gracefully fail. I think we should add a comment in the version.h file as well so people know the impact when they make the change.

script/testing/util/mem_metrics.py
jazzplato: fixed

script/testing/util/periodic_task.py
jazzplato: fixed

script/testing/util/test_case.py
jazzplato: fixed

script/testing/reporting/parsers/parse_data.py
jazzplato: - Added a `FIXME`
- Changed the path concatenation to `os.path.join`

script/testing/collect_mem_info.py
```diff
@@ -0,0 +1,18 @@
+#!/usr/local/bin/python3
```
lmwnshn, line 1: I thought /usr/bin/env python3 was generally more reliable, could be mistaken though

jazzplato (commented):
changed default python interpreter to `#!/usr/bin/env python3`

PR #1122, opened 2020-08-25
DISTINCT aggregation fix

githubxxcc:
#1048
### Problem 
For aggregation with distinct, e.g. `SELECT COUNT(DISTINCT x)` or `SELECT SUM(DISTINCT x) FROM y GROUP BY c`, there is currently no mechanism put in place in the `HashAggregationTranslaor` and `StaticAggregationTranslator` to prevent the same `x` to be advanced to aggregation more than once. 

### Solution 
The high level idea is to have a hash table so that for each tuple value being aggregated, if it needs to be distinct, it should not be advanced to aggregation if it already exists in the hash table. 

Also, the build pipeline for both the `StaticAggregationTranslator` and  `HashAggregationTranslator` would need to be serial since I don't think the hashtable is thread safe.

PR #1244, opened 2020-10-20
Enable some DISTINCT-related tests that were disabled during the initial TPL2 port.

lmwnshn:
# Enable some DISTINCT-related tests.

## Description

Enable some DISTINCT-related tests that aren't currently being run. They got fixed with #1122.

Also remove some weird "run as main program" stuff at the bottom of compiler_test.cpp.

@17zhangw want to look at the expected feature vectors?


17zhangw (approved):
Feature vectors look good.

PR #1246, opened 2020-10-20
Fix SELECT NULL and SELECT * crashing the DBMS.

### LINKED ISSUE #1020: SELECT NULL fails
If you have `NULL` in the output list of a `SELECT` statement, then the server crashes. It doesn't matter whether you have a `FROM` clause or not:

**Client**
```
terrier=# CREATE TABLE xxx (val int);
CREATE TABLE
terrier=# INSERT INTO xxx VALUES (1);
INSERT 0 1
terrier=# SELECT NULL FROM xxx;
```

**Server**
```
terrier: /home/pavlo/Documents/Peloton/Github/terrier/src/include/planner/plannodes/output_schema.h:40:
 terrier::planner::OutputSchema::Column::Column(std::__cxx11::string, terrier::type::TypeId, std::unique_ptr<terrier::parser::AbstractExpression>): Assertion `(type_ != type::TypeId::INVALID) && ("Attribute type cannot be INVALID.")' failed.
Aborted (core dumped)
```

### LINKED ISSUE #1191: SELECT *; crashes the server.
Found by @varsosemblem on twitch.

Should be similar to #1020 in nature.

```
terrier=# SELECT *;
```

lmwnshn:
# Fix SELECT NULL and SELECT * crashing the DBMS.

## Description

Fix SELECT NULL and SELECT * crashing the DBMS at the binder level.

SELECT NULL: set the type of NULLs to text. Fix #1020 and further documented therein.
SELECT *: Check for the presence of a tablename, throw an error otherwise. Fix #1191.

Also make the check_github_labels script executable since it seems to be missing that.


mbutrovich (approved):
LGTM!

PR #1248, opened 2020-10-20
Fix for SELECT Table Qualifier

### LINKED ISSUE #780: Select Star with Table Qualifier Does Not Work

17zhangw:
# Fix for SELECT Table Qualifier #780 

## Description

Fix #780.

Historically, `parser::StarExpression` was used for both `COUNT(*)` and `SELECT *`. The PR creates a new `parser::TableStarExpression` used exclusively to handle the cases of `SELECT *` and `SELECT t.*`. During binding, the function `BinderContext::GenerateAllColumnExpressions` is then able to consider the qualified table name (if specified) when generating all column value expressions from a given table.

## Queries that now work
```
CREATE TABLE t (a INT);
SELECT t.* FROM t;
SELECT tt.* FROM t tt;
SELECT t.* FROM t tt; # throws a binder exception
```


src/binder/bind_node_visitor.cpp
```diff
@@ -79,8 +80,10 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::CopyStatement> node)
     node->GetCopyTable()->Accept(common::ManagedPointer(this).CastManagedPointerTo<SqlNodeVisitor>());
 
     // If the table is given, we're either writing or reading all columns
+    auto table_star = parser::TableStarExpression();
```
lmwnshn, line 83: `parser::TableStarExpression table_star`?

lmwnshn (approved):
Minor change, otherwise LGTM. Let's get @apavlo to weigh in though.

PR #1249, opened 2020-10-20
Modify check_github_labels script to scrape the PR page if rate-limited.

lmwnshn:
# Modify check_github_labels script to scrape the PR page if rate-limited.

## Description

Unauthenticated rate-limit is 60 requests per hour. Authenticated would be 5000 per hour.
Web scraping like this is unlimited and Jenkins is too fragile for me to want to think about OAuth tokens, so...

### DIFF
--- a/build-support/check_github_labels.py
+++ b/build-support/check_github_labels.py
@@ -19,17 +19,37 @@ def get_pr_num():
     return None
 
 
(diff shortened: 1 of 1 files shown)


mbutrovich (approved):
LGTM

PR #1254, opened 2020-10-21
Fix the PR label check

bialesdaniel:
# Fix the PR label check 

## Description
Currently, the check_github_labels.py errors out if there is no regex match. This is a hotfix so that builds can complete. when `match` is `None` it will just continue with the build.

Here is proof that it works if there are no matches http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/pr-check-fix/1/pipeline/9


lmwnshn:
Closing in favor of #1256.

PR #1257, opened 2020-10-21
Add explicit dependencies to CMake and -GNinja to nightly.

lmwnshn:
# Heading

Add explicit dependencies to CMake and -GNinja to nightly Jenkinsfile.

## Description

Dependencies added:
- noisepage_objlib (spdlog)
- test_util libs (gtest gtest_main gmock gmock_main)
- benchmark_util libs (gtest gtest_main gmock gmock_main)

Dependencies are built in release because the debug versions have unacceptable performance for coverage builds. Especially spdlog. `make` appears to struggle to figure these dependencies out on its own.

spdlog is not added as a true dependency because the library name changes between debug mode and release mode, namely libspdlogd.a versus libspdlog.a. Due to the different names, the build system gets confused.

Therefore the dependency is manually added.

The nightly Jenkinsfile is an omission I made.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -625,6 +625,13 @@ target_link_libraries(noisepage_objlib PUBLIC       # PUBLIC: all consumers of t
         ${TBB_LIBRARIES_RELEASE}
         )
 
+# Dependencies are built in release because the debug versions have unacceptable performance for coverage builds.
+# make appears to struggle to figure these dependencies out on its own.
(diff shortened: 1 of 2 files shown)


lmwnshn:
@17zhangw If this fixes your issue, I will merge. CI won't catch it since it uses ninja.

17zhangw:
Merging this branch allows me to do a clean build of `mini_runners`.

PR #1259, opened 2020-10-21
Add all weird CMake dependencies to noisepage_objlib instead of real targets.

lmwnshn:
# Add all weird CMake dependencies to noisepage_objlib instead of real targets.

## Description

The dependencies have to be added to an object library because adding the dependencies to a target causes a linking attempt that will fail. The linking fails because the linker is looking for the debug version.


lmwnshn:
Now _this_ is something that could have been caught if I waited three hours for #1257 to stroll leisurely through CI. Oh well.

And the reason #1257 exists is because `make` and `ninja` are aggravatingly ever so slightly different in how smartly they resolve dependencies.

PR #1262, opened 2020-10-21
CMake: Apply +x to build-support for CLion's remote host compatibility.

mbutrovich:
## Description
CLion's remote toolchain functionality seems to lose `+x` attributes on files when they're rsync'd to the remote host. This adds a step to the CMake file that ensures the required scripts in `build-support` have `+x`.


lmwnshn (approved):
LGTM.

PR #1264, opened 2020-10-21
Abort (instead of failing) builds that are not labeled "ready-for-ci".

jkosh44:
I just want to test and see if this works on Jenkins

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,18 +6,14 @@ pipeline {
     }
     stages {
         stage('Ready For CI') {
-        agent { label 'macos' }
-            environment {
-                LIBRARY_PATH="$LIBRARY_PATH:/usr/local/opt/libpqxx/lib/"
-                LLVM_DIR=sh(script: "brew --prefix llvm@8", label: "Fetching LLVM path", returnStdout: true).trim()
-                CC="${LLVM_DIR}/bin/clang"
-                CXX="${LLVM_DIR}/bin/clang++"
-            }
+        agent any
             steps {
-                int ready_for_build = sh(script: 'python3 ./build-support/check_github_labels.py', returnStatus: true)
-                if(status != 0) {
-                    currentBuild.result = 'ABORTED'
-                    error('Not ready for CI')
+                script {
+                   ready_for_build = sh script: 'python3 ./build-support/check_github_labels.py', returnStatus: true
+                   if(ready_for_build != 0) {
+                        currentBuild.result = 'ABORTED'
+                        error('Not ready for CI')
+                   }
                 }
             }
         }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -21,6 +21,7 @@ pipeline {
                    ready_for_build = sh script: 'python3 ./build-support/check_github_labels.py', returnStatus: true
                    if(ready_for_build != 0) {
                         currentBuild.result = 'ABORTED'
+                        return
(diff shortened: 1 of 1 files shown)


jkosh44:
This will cause the build to look like this
Screenshot from 2020-10-21 13-54-38
Note that it's gray with a dashed line and not a red X. This will help differentiate failed builds from aborted not ready builds.
I have 2 pending tasks with this (if anyone has an answer that would be helpful)

1. Currently I'm running this check on a single agent
```groovy
        agent {
            docker {
                image 'terrier:focal'
            }
        }
```
This will cause this agent to become a bottleneck in all builds. I tried to do something like this
```groovy
agent any
```
which should grab any available agent (https://www.jenkins.io/doc/book/pipeline/syntax/#agent-parameters). However I got the following error (http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-1264/2/pipeline)
```
‘Jenkins’ is reserved for jobs with matching label expression
‘h1.jenkins.cmudb-ci.narwhal.pdl.cmu.edu’ is reserved for jobs with matching label expression
```
My best guess is that some agents are reserved for certain tasks so when I grabbed a random agent it refused to perform this check. So it would be nice to somehow specify use any agent that isn't reserved for specific tasks
2. (Optional) I'm much more familiar with scripted Jenkins pipelines than declarative Jenkins pipelines. So I just put a script into the steps to do this. It's not very cohesive with the rest of the file. So if anyone knows how to implement this declaratively, let me know, I think it's probably some combination of steps and whens. I would play around with different variations but Jenkins takes way to long to just test random configurations.

Jenkinsfile
```diff
@@ -5,7 +5,27 @@ pipeline {
...
+            }
+            steps {
+                script {
+                   ready_for_build = sh script: 'python3 ./build-support/check_github_labels.py', returnStatus: true
+                   if(ready_for_build != 0) {
+                        currentBuild.result = 'ABORTED'
+                        abort('Not ready for CI. Please add ready-for-ci tag in Github when you are ready to build your PR.')
+                   }
+                }
+            }
```
lmwnshn, line 27: I think we may want to cleanup the directory:

```
                    post {
                        cleanup {
                            deleteDir()
                        }
                    }
```

lmwnshn (changes requested):
Essentially LGTM, awesome work!

Just one possible change -- cleaning up the build directory, I think this isn't being done right now?

Good luck on the midterm!!!!

Jenkinsfile
jkosh44: I added it. I'm still slightly concerned about the random agent bottleneck that I mentioned. Not sure what a good fix is though.

lmwnshn: I'm not particularly concerned with that agent bottleneck; we have a lot of those terrier:focal's lying around. Unless you have opinions, I would be OK with merging as is.

jkosh44:
@lmwnshn I made the changes you suggested and I confirmed with Dan that the Ready For CLI stage will use any available agent that has docker on it, so I'm not too concerned about that anymore.

Jenkinsfile
jkosh44: Yeah, sorry didn't see this comment. After talking with Dan I think this is fine.

lmwnshn (approved):
LGTM.

PR #1268, opened 2020-10-23
'Terrier' -> 'NoisePage' Name Change

apavlo:
# DBMS Name Change

## Description
Rename files / strings / namespaces / env variable from `terrier` to `noisepage`. 
https://youtu.be/kAPOKpnU-_U

## Remaining Tasks
The only thing that I think remains is to rename the docker images on Jenkins, but that is very low priority.


crd477:
I just did the following on the Docker agents:
```
docker image tag terrier:focal noisepage:focal
```
so it should be safe to rename the image it in the Jenkins pipeline too.

lmwnshn:
I added a quick fix to CMakeLists that wasn't being picked up before the image rename since we cache stuff more aggressively on the mac iirc. Will merge if passes CI.

lmwnshn (approved):
LGTM.

PR #1271, opened 2020-10-25
Fix double parsing

### LINKED ISSUE #1253: Cast srting to float test fails in Jenkins
## Summary
Jenkins build failed: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-1161/36/tests/ on sql_vector_operations_vector_cast_test. Right now it's unclear what the cause was and how to reproduce, but as I investigate I will fill in the details.

The PR where this came up is here: https://github.com/cmu-db/noisepage/pull/1161/files

## Environment

**OS**: macOS 10.14

**Compiler**: Clang 8.0. 

## Steps to Reproduce
I was able to reproduce by running the following commands on a mac
```
mkdir build
cd build
cmake -GNinja -DCMAKE_BUILD_TYPE=Release -DTERRIER_USE_ASAN=OFF -DTERRIER_BUILD_BENCHMARKS=OFF ..
ninja noisepage
cmake --build . --target sql_vector_operations_vector_cast_test -- 
./test/sql_vector_operations_vector_cast_test
```
Note: I had to wrap the test in a loop because it doesn't always fail. The way to make it fail is if there is an 'E' or 'e' in memory right after the `VarlenEntry`. There's no good way to force this scenario, so you just have to run it a bunch of times.

### Expected Behavior
Casting a string to a float should be successful

### Actual Behavior

- The cast throws an exception
- The cast is of type Double not Float
- The precision of the results are not what we expect

jkosh44:
# Fix Double and Float Parsing

## Description
We currently use the [`fast_double_parser` library](https://github.com/lemire/fast_double_parser) when parsing strings to doubles and floats. Specifically when casting `VarlenEntry` to a `float` or `double`. 
https://github.com/cmu-db/noisepage/blob/12c2203713fe1ae423409a38b9baa00a2158e0d7/src/execution/sql/operators/cast_operators.cpp#L7-L18
The issue is that `fast_double_parser` expects the string to be null terminated and `VarlenEntry` is not null terminated. The solution is to switch to a library that can parse non null terminated strings. the writer of `fast_double_parser` has a different library exactly for this use case called [`fast_float`](https://github.com/lemire/fast_float).

Fixes #1253 

## Remaining tasks

- [X] Investigate whether `fast_double_parser` needs to be removed from `CSVReader`
https://github.com/cmu-db/noisepage/blob/12c2203713fe1ae423409a38b9baa00a2158e0d7/src/execution/util/csv_reader.cpp#L76-L80
- [X] Remove fast double parser code
- [X] Benchmark performance differences

## Performance
To benchmark performance I forked this library: [https://github.com/lemire/simple_fastfloat_benchmark](https://github.com/lemire/simple_fastfloat_benchmark) and added fast_double_parser and changed the range of numbers to [-1, 1): [https://github.com/jkosh44/simple_fastfloat_benchmark](https://github.com/jkosh44/simple_fastfloat_benchmark)

The results on my computer (Ubuntu 20.04, 24GB Memory, Intel® Core™ i7-10510U CPU @ 1.80GHz × 8)
```bash
# parsing random integers in the range [-1,1)
volume = 2.14603 MB 
netlib                                  :   250.75 MB/s (+/- 2.7 %)    11.68 Mint/s      30.98 i/B   697.08 i/f (+/- 0.0 %)      0.22 bm/B     4.96 bm/f (+/- 0.4 %)     16.31 c/B   366.99 c/f (+/- 1.1 %) 
strtod                                  :   204.33 MB/s (+/- 3.5 %)     9.52 Mint/s      48.60 i/B  1093.58 i/f (+/- 0.0 %)      0.05 bm/B     1.22 bm/f (+/- 0.4 %)     20.00 c/B   450.10 c/f (+/- 1.5 %) 
abseil                                  :   397.04 MB/s (+/- 3.1 %)    18.50 Mint/s      26.82 i/B   603.47 i/f (+/- 0.0 %)      0.04 bm/B     1.00 bm/f (+/- 0.3 %)     10.12 c/B   227.74 c/f (+/- 1.8 %) 
fastfloat                               :   886.57 MB/s (+/- 5.0 %)    41.31 Mint/s      12.98 i/B   292.05 i/f (+/- 0.0 %)      0.02 bm/B     0.51 bm/f (+/- 0.5 %)      4.61 c/B   103.64 c/f (+/- 2.9 %) 
fast_double_parser                      :  1177.99 MB/s (+/- 8.5 %)    54.89 Mint/s      10.96 i/B   246.56 i/f (+/- 0.0 %)      0.02 bm/B     0.51 bm/f (+/- 0.4 %)      3.47 c/B    78.16 c/f (+/- 5.9 %) 
# You can also provide a filename: it should contain one string per line corresponding to a number
```
So unfortunately we are taking a hit on performance by switching from `fast_double_parser` to `fast_float`, however it still seems like it's better than most alternatives. Also unless we want to switch `VarlenEntry` to be null terminated we don't really have a choice, because `fast_double_parser` doesn't work correctly unless the string is properly terminated.


jkosh44:
I'm getting a decent amount of intermittent failures on Jenkins builds. They don't seem to be related to these changes though.

test/execution/sql_vector_operations_vector_cast_test.cpp
```diff
@@ -179,4 +179,32 @@ TEST_F(VectorCastTest, CastStringToFloat) {
...
+  EXPECT_TRUE(a->IsNull(3));
+  EXPECT_EQ(GenericValue::CreateDouble(0.8), a->GetValue(4));
+  EXPECT_EQ(GenericValue::CreateDouble(910), a->GetValue(5));
+}
+
+// NOLINTNEXTLINE
+TEST_F(VectorCastTest, CastStringToFloatParseError) {
+  exec::ExecutionSettings exec_settings{};
+
+  // a = [NULL, "-123.45", "6.75", NULL, "0.8", "910"]
```
lmwnshn, line 204: Technically this should be 6.75E, and document that's why it throws an exception, etc. But this is short enough that people should figure it out.

lmwnshn (approved):
LGTM if passes CI.

test/execution/sql_vector_operations_vector_cast_test.cpp
jkosh44: It's such a quick fix, that I'd rather fix it

pmenon:
Does the build work when you compile `CSVReader`, too? After a cursory look, `fast_float` looks like it's probably missing some `inlines` that'll cause duplicate symbols if included in multiple CUs. I had to modify the `fast_double_parser.h` in a similar manner.

jkosh44:
> Does the build work when you compile `CSVReader`, too? After a cursory look, `fast_float` looks like it's probably missing some `inlines` that'll cause duplicate symbols if included in multiple CUs. I had to modify the `fast_double_parser.h` in a similar manner.

@pmenon I haven't actually tried to compile `CSVReader` to be honest, but now that you mention it I probably should (and will tomorrow). At one point I had actually moved the `TryCast` into the .h file (in an attempt to combine the double and float) and I started getting some linker errors that may have been about duplicate symbols. If that's the case I may need to reach out to you for an explanation on the fix.

pmenon:
Yes, please test all changes.

Also, don't move the casting into the header files. They're in CPP files to reduce header bloat, which impacts JIT time (which is pretty trash as it is).

jkosh44:
@pmenon So `CSVReader` is actually not compiling because of a missing include of the header file: https://github.com/cmu-db/noisepage/blob/12c2203713fe1ae423409a38b9baa00a2158e0d7/src/execution/util/csv_reader.cpp#L3-L10
After adding the include locally though it compiles fine. Not sure what the linker error I saw earlier was.

If you think it's a good idea I can add the include to this PR

lmwnshn:
CSVReader was commented out pending CI upgrades for some missing includes, IIRC. CI has since been upgraded but we haven't uncommented that yet.

lmwnshn:
I am merging this PR first as it helps overall CI stability. We can add stuff in a followup PR.

PR #1272, opened 2020-10-26
TPL Atomic Operations

jrolli:
# TPL Atomic Operations

## Description

Adds atomic `and`, `or`, and `compare-exchange` operations for `uint8`, `uint16`, `uint32`, and `uint64` types.  This is a minimal implementation that allows the manipulation of "concurrent bitmaps" and version pointers in TPL which is sufficient for implementing accessors for the storage layer.

### Codegen Storage Layer Accessors

This is the primary motivation for this PR.  Specifically, this should provide enough TPL functionality that we can turn the current storage layer accessors into codegen'd TPL that treats a block as a struct and does straightline execution without any interpretation for insert, delete, and update.  Select can also be better codegen'd but will still need some degree of interpretation for applying `UndoRecord`s.

### Additional Atomics

This does not add other atomics such as add/subtract or 16-byte compare-exchange which may also be useful.  Those changes should be fairly easy to implement if necessary.


src/execution/sema/sema_builtin.cpp
```diff
@@ -1840,6 +1840,83 @@ void Sema::CheckMathTrigCall(ast::CallExpr *call, ast::Builtin builtin) {
...
+      ReportIncorrectCallArg(call, 0, GetBuiltinType(operand_kind));
+      return;
+    }
+    if (!call_args[2]->GetType()->IsSpecificBuiltin(operand_kind)) {
+      ReportIncorrectCallArg(call, 1, GetBuiltinType(operand_kind));
+      return;
+    }
+  }
+
+  call->SetType(GetBuiltinType(ast::BuiltinType::Nil));
```
lmwnshn, line 1917: I thought CAS usually returned a bool, but if we don't need it then I guess it is fine / can be added later.

src/include/execution/ast/builtins.h
```diff
@@ -312,6 +312,20 @@ namespace noisepage::execution::ast {
   F(OffsetOf, offsetOf)                                                 \
   F(PtrCast, ptrCast)                                                   \
                                                                         \
+  /* Low-level Atomics*/                                                \
```
pmenon, line 315: Do you need a builtin for each type, or is it sufficient to have just one of each type and generate the appropriate bytecode based on the type of input after SA?

src/include/execution/vm/bytecode_handlers.h
```diff
@@ -1536,6 +1536,70 @@ VM_OP_WARM void OpPow(noisepage::execution::sql::Real *result, const noisepage::
   noisepage::execution::sql::ArithmeticFunctions::Pow(result, *base, *val);
 }
 
+// ---------------------------------------------------------
+// Atomic memory operations
+// ---------------------------------------------------------
+
+// NOLINTNEXTLINE (clang tidy incorrectly assesses dest as pointing to const)
+VM_OP_HOT void OpAtomicAnd1(uint8_t *dest, uint8_t val) {
+  reinterpret_cast<std::atomic<uint8_t> *>(dest)->fetch_and(val);
```
pmenon, line 1545: You can't assume that `std::atomic<T>` and `T` have the same size or layout. You can, instead, use one of the GCC/Clang atomic builtins which use raw types. See [here](https://github.com/cmu-db/noisepage/blob/master/src/include/execution/sql/chaining_hash_table.h#L199) for an example. You'll need to figure out the parameters to pass to achieve the memory semantics you intended with the original `fetch_and()` call.

test/execution/atomics_test.cpp
```diff
@@ -0,0 +1,673 @@
...
+  const SourcePosition &EmptyPos() const { return pos_; }
+
+  util::Region region_;
+  SourcePosition pos_;
+};
+
+// NOLINTNEXTLINE
+TEST_F(AtomicsTest, AtomicAndOr1) {
+  // Setup the compilation environment
+  sema::ErrorReporter error_reporter(&region_);
```
pmenon, line 36: Is there a reason you're handcrafting the AST rather than using raw TPL source and running it through the compilation pipeline?

There's a bit of code duplication that might benefit from a refactor. Also, add some comments between the test that, at least, show the text you intend to generate.

pmenon (changes requested):
Quick whirlwind review. Generally, looks good.

test/execution/atomics_test.cpp
jrolli: Does the main binary use a different parser than the standalone `tpl` utility?  The standalone utility's parser was crashing and burning (re: previous Slack conversation) so I assumed that the same TPL would not work here.  I would absolutely prefer for it to be plain TPL as that is more sustainable.

src/include/execution/ast/builtins.h
jrolli: Do you have a good example builtin I can mimic for that behavior?  I agree that would be nicer but didn't see a clear way to do that (ties into the TODO of making the compare-exchange ops not require the type to be `uint` as well).

src/include/execution/vm/bytecode_handlers.h
jrolli: I can make that change, but for at least the 8- and 64-bit cases we have static asserts that this is the case (see `concurrent_bitmap.h`).

src/include/execution/ast/builtins.h
pmenon: Take a look at the `@hash()` builtins.

src/include/execution/vm/bytecode_handlers.h
pmenon: Please do, I'd rather not have that restriction.

test/execution/atomics_test.cpp
pmenon: Yes, it is the same parser. IIRC, it was crashing when you try to dereference. But, since you're now using a builtin function, it should work, right?

jrolli:
@pmenon & @lmwnshn, I've incorporated the feedback above.  Additionally, I've attempted to add the return values for the atomic arithmetic (before value) and compare-exchange (boolean) but apparently messed that up pretty bad.  If you all have ideas or can point me in the right direction, that would be appreciated.

lmwnshn (approved):
LGTM, though waiting on Prashanth's comments as well

src/execution/vm/bytecode_generator.cpp
```diff
@@ -1887,6 +1887,100 @@ void BytecodeGenerator::VisitBuiltinArithmeticCall(ast::CallExpr *call, ast::Bui
...
+  LocalVar ret;
+  auto return_used = GetExecutionResult() != nullptr;
+  if (return_used) {
+    ret = GetExecutionResult()->GetOrCreateDestination(call->GetType());
+    GetExecutionResult()->SetDestination(ret);
+  } else {
+    ret = GetCurrentFunction()->NewLocal(call->GetType());
+  }
+
+  switch (args[1]->GetType()->GetSize()) {
```
pmenon, line 1903: Switches for sizes seems odd for two reasons:
1. It isn't consistent with the rest of the style in this file.
2. It forces you to have a default clause despite type resolution having completed.

Use `if-then-else` with an assert in the last `else` clause.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -1887,6 +1887,100 @@ void BytecodeGenerator::VisitBuiltinArithmeticCall(ast::CallExpr *call, ast::Bui
   GetExecutionResult()->SetDestination(dest);
 }
 
+void BytecodeGenerator::VisitBuiltinAtomicAndCall(ast::CallExpr *call) {
+  auto args = call->Arguments();
+  LocalVar dest = VisitExpressionForRValue(args[0]);
+  LocalVar val = VisitExpressionForRValue(args[1]);
+  LocalVar ret;
+  auto return_used = GetExecutionResult() != nullptr;
```
pmenon, line 1895: `auto` isn't more concise than `bool`; I'd prefer the latter. Also, make it `const`.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -1887,6 +1887,100 @@ void BytecodeGenerator::VisitBuiltinArithmeticCall(ast::CallExpr *call, ast::Bui
...
+      break;
+    case 8:
+      GetEmitter()->Emit(Bytecode::AtomicAnd8, dest, val, ret);
+      break;
+    default:
+      UNREACHABLE("Unexpected operand size for atomicAnd!");
+  }
+}
+
+void BytecodeGenerator::VisitBuiltinAtomicOrCall(ast::CallExpr *call) {
```
pmenon, line 1921: I think you can combine this with the previous `BuiltinAtomicAndCall()`.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -1887,6 +1887,100 @@ void BytecodeGenerator::VisitBuiltinArithmeticCall(ast::CallExpr *call, ast::Bui
   GetExecutionResult()->SetDestination(dest);
 }
 
+void BytecodeGenerator::VisitBuiltinAtomicAndCall(ast::CallExpr *call) {
+  auto args = call->Arguments();
```
pmenon, line 1891: You're making a copy of the arguments vector here, you need to make this `auto &args`. Or, better yet, make it `const auto &args`.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -1887,6 +1887,100 @@ void BytecodeGenerator::VisitBuiltinArithmeticCall(ast::CallExpr *call, ast::Bui
...
+      GetEmitter()->Emit(Bytecode::AtomicCompareExchange2, dest, expected, desired, ret);
+      break;
+    case 4:
+      GetEmitter()->Emit(Bytecode::AtomicCompareExchange4, dest, expected, desired, ret);
+      break;
+    case 8:
+      GetEmitter()->Emit(Bytecode::AtomicCompareExchange8, dest, expected, desired, ret);
+      break;
+    default:
+      UNREACHABLE("Unexpected operand size for atomicOr!");
```
pmenon, line 1980: Fix message.

src/execution/vm/vm.cpp
```diff
@@ -2224,6 +2224,41 @@ void VM::Interpret(const uint8_t *ip, Frame *frame) {  // NOLINT
 #undef BINARY_REAL_MATH_OP
 #undef UNARY_REAL_MATH_OP
 
+#define ATOMIC_BINARY_OP(AOP, T)                       \
+  OP(Atomic##AOP) : {                                  \
+    auto *dest = frame->LocalAt<T *>(READ_LOCAL_ID()); \
+    auto val = frame->LocalAt<T>(READ_LOCAL_ID());     \
+    auto *ret = frame->LocalAt<T *>(READ_LOCAL_ID());  \
+    OpAtomic##AOP(dest, val, ret);                     \
```
pmenon, line 2232: The semantics we've been using for all operations is that the first argument is the return argument.

src/execution/sema/sema_builtin.cpp
```diff
@@ -1840,6 +1840,69 @@ void Sema::CheckMathTrigCall(ast::CallExpr *call, ast::Builtin builtin) {
   call->SetType(GetBuiltinType(return_kind));
 }
 
+void Sema::CheckAtomicCall(ast::CallExpr *call, ast::Builtin builtin) {
+  const auto &call_args = call->Arguments();
+
+  int arg_count;
```
pmenon, line 1846: This was a little difficult to follow and I think can benefit from refactoring. I'd prefer collapsing all logic into one `switch`, one block for the `And` and `Or`, and another for the `CompareExchange`. Then check argument counts appropriately. Also change the size check to `IsIntegral() and Size() <= 8` for clarity (rather than the verbose checks across switches and branches - make it a lambda check if you want).

You also need to check for integer types in the compare-exchange case.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -1887,6 +1887,100 @@ void BytecodeGenerator::VisitBuiltinArithmeticCall(ast::CallExpr *call, ast::Bui
...
+    case 8:
+      GetEmitter()->Emit(Bytecode::AtomicOr8, dest, val, ret);
+      break;
+    default:
+      UNREACHABLE("Unexpected operand size for atomicOr!");
+  }
+}
+
+void BytecodeGenerator::VisitBuiltinAtomicCompareExchangeCall(ast::CallExpr *call) {
+  auto args = call->Arguments();
```
pmenon, line 1953: Vector copy.

test/execution/atomics_test.cpp
```diff
@@ -0,0 +1,513 @@
...
+  ast::AstNodeFactory factory(&region_);
+  ast::Context context(&region_, &error_reporter);
+
+  auto src = std::string(
+      "fun atomic_and(dest: *uint8, mask: uint8) -> uint8 {\n"
```
pmenon, line 41: Style nit: you can use string literals to avoid the need to manually add new lines to text. See [here](https://github.com/cmu-db/noisepage/blob/master/test/execution/vm_bytecode_generator_test.cpp#L46) for an example.

test/execution/atomics_test.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+  // Compile it...
+  auto input = compiler::Compiler::Input("Atomic Definitions", &context, &src);
+  auto module = compiler::Compiler::RunCompilationSimple(input);
+  EXPECT_STREQ(error_reporter.SerializeErrors().c_str(), "");
+  ASSERT_FALSE(module == nullptr);
+
+  // The function should exist
+  std::function<uint8_t(uint8_t *, uint8_t)> atomic_and;
+  EXPECT_TRUE(module->GetFunction("atomic_and", vm::ExecutionMode::Interpret, &atomic_and));
```
pmenon, line 58: Have you tested these in compiled mode, too?

test/execution/atomics_test.cpp
```diff
@@ -0,0 +1,513 @@
...
+      "}\n"
+      "fun atomic_or(dest: *uint8, mask: uint8) -> uint8 {\n"
+      "  var x = @atomicOr(dest, mask)\n"
+      "  return x\n"
+      "}");
+
+  // Compile it...
+  auto input = compiler::Compiler::Input("Atomic Definitions", &context, &src);
+  auto module = compiler::Compiler::RunCompilationSimple(input);
+  EXPECT_STREQ(error_reporter.SerializeErrors().c_str(), "");
```
pmenon, line 53: Use [`ErrorReporter::HasErrors()`](https://github.com/cmu-db/noisepage/blob/master/src/include/execution/sema/error_reporter.h#L66)

test/execution/atomics_test.cpp
```diff
@@ -0,0 +1,513 @@
...
+      EXPECT_TRUE(success);
+    };
+
+    MultiThreadTestUtil::RunThreadsUntilFinish(&thread_pool, num_threads, workload);
+    EXPECT_EQ(target.load(), num_threads);
+  }
+}
+
+// NOLINTNEXTLINE
+TEST_F(AtomicsTest, AtomicCompareExchange8) {
```
pmenon, line 462: These tests are very very similar. With the exception of the types (both C++ and TPL) they operate on, they look identical. Can you think of a way to refactor them?

pmenon (changes requested):
Tests look much better and maintainable. Some minor refactoring + style fixes.

src/execution/vm/bytecode_generator.cpp
jrolli: I couldn't think of an easy way to combine this with the other call without essentially creating a switch/if statement inside of the combined function that personally I think would be harder to read because it leads to a nested if/switch (the bytecodes are specialized to both operator and operand size).

src/execution/sema/sema_builtin.cpp
jrolli: Acknowledge all but the last.

I deliberately did not check for integer on the compare-exchange case to allow for it to be used for pointer swapping that is still type-checked.  For compare-exchange, I think swapping in pointers (specifically undo records) will likely be the most common use case and I think we will open ourselves up to unnecessary errors if we have to convert to an integral before swapping as you may end up swapping in an incompatible type.

test/execution/atomics_test.cpp
jrolli: I can remove these now.  This was a hack to get the error printed out in the log when there were errors in the test.  The other TPL-based tests just have the assert on not having a nullptr.

test/execution/atomics_test.cpp
jrolli: I might be able to make the inner part of each "class" of test templated on the integer type and takes the TPL code as the argument?

test/execution/atomics_test.cpp
jrolli: I have not, but none of the execution unittests use compiled mode at the moment.

src/execution/sema/sema_builtin.cpp
pmenon: Compare-exchanging pointers makes sense to me. I brought it up because the PR description only listed integer types.

If you want pointers in this PR, add some tests for them, too. And add comments in the refactored function stating this.

test/execution/atomics_test.cpp
pmenon: We have an issue for that [here](https://github.com/cmu-db/noisepage/issues/1165). In the mean time, check here to be sure.

test/execution/atomics_test.cpp
jrolli: How does it look now?

test/execution/atomics_test.cpp
jrolli: Acknowledge.  If you're fine with the general style of the test code now, I'll add a parameter to the templated functions to indicate whether or not to compile the code.  And test everything in both interpreted and compiled mode.

src/execution/vm/bytecode_generator.cpp
pmenon: Since the only difference is the bytecode you're using, can you combine them and choose between the two codes?
```
if (size == N) {
  auto bytecode = builtin == Builtin::AtomicOr ? Bytecode::AtomicOr[N] : Bytecode::AtomicAnd[N];
  GetEmitter()->Emit(bytecode, ....);
}
```

A good hint is if you're type-checking shares code, then bytecode generation can probably share code.

test/execution/atomics_test.cpp
pmenon: Much better. Now, rather than passing the whole source to the testing functions, can you pass just the primitive TPL type as a string and format the source based on this type using `fmt::format()`?

pmenon (changes requested):
Last nits and questions. Should be good to go after this.

test/execution/atomics_test.cpp
pmenon: Okay with me.

src/execution/vm/bytecode_generator.cpp
jrolli: My only hesitation here is that I'm assuming at some point we'll add `add` and `subtract` (and possibly xor?) at which case that approach gets awkward.  I'm fine doing what you suggested, or I can also bite the bullet and add those atomics as well (should go faster now).

test/execution/atomics_test.cpp
jrolli: Will do.

src/execution/sema/sema_builtin.cpp
jrolli: On second look, I don't think TPL currently supports the primary use case, atomic singly-linked list, because it does not support recursive data structures at this time.  I've inserted a TODO and implemented the integral enforcement for `CompareExchange`.

test/execution/atomics_test.cpp
jrolli: I am able to run the tests in compiled mode as long as I copy the `bytecode_handlers_ir.bc` file into my working directory before execution.  This seems like there is a larger issue with how that bytecode is stored and loaded into the library that is mostly outside the scope of this PR.

What are your thoughts on having the templated tests instrumented for both execution modes but commenting out the compiled tests until we settle on the right way to handle that file dependency?

src/execution/sema/sema_builtin.cpp
```diff
@@ -1840,6 +1840,68 @@ void Sema::CheckMathTrigCall(ast::CallExpr *call, ast::Builtin builtin) {
...
+  // Permissive arg check to protect the following argument dereference
+  if (!CheckArgCountAtLeast(call, 1)) return;
+
+  auto operand_type = call_args[0]->GetType()->GetPointeeType();
+  if (operand_type == nullptr) {
+    ReportIncorrectCallArg(call, 0, "'dest' must be pointer to an integral type");
+    return;
+  }
+
+  /* TODO(John): Ideally we want to support CMPXCHG on pointers as well, but the
```
pmenon, line 1855: Nit: although I understand this issue, I'm not sure it's appropriate here. Compare-exchange on pointers and recursive data types are unrelated. Either state that compare-exchange on pointers is future work, or add support here now.

pmenon (changes requested):
Last tweaks, then it's good to go.

src/execution/sema/sema_builtin.cpp
jrolli: Removed the comment and left the code as integral only.

pmenon (approved):
LGTM.

PR #815, opened 2020-03-05
Disable Broken Tests / Microbenchmarks

apavlo:
This PR is unfortunately a necessary step to stabilize our builds.

* ~~Disable `CompilerTest::SimpleSortTest` (#811)~~
* Disable all TPCCTests (#791)
* Disable `FilterManagerTests` (???)
* Set the failure threshold for `DataTableBenchmarks` to 75% (#812)

https://youtu.be/79fDI5iA3xo

PR #1275, opened 2020-10-27
Re-enable TPCC tests.

lmwnshn:
# Re-enable TPCC tests.

## Description

We used to run TPCC tests in CI, disabled in #815. Maybe we should run this again.


lmwnshn:
How does coverage go _down_ when enabling extra tests? Anyways, going to go ahead and just merge this one.

PR #1276, opened 2020-10-28
Fix RecoveryTest::UnrecoverableTransactionTest

### LINKED ISSUE #1205: DeferredActionManager: Unlink on a table that was already dropped
## Summary
The DAManager freed a table that still had records to be unlinked from it.

## Environment
To address the bug, especially if it environment specific, we need to know what kind of configuration you are running on. Please include the following:

**OS**: Ubuntu 18.04 (LTS), Jenkins CI

**Compiler**: GCC 7.3

## Steps to Reproduce
1. Spotted in Jenkins [here](http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-1161/38/tests/).

It looks like the freeing of a table happens in the DAManager's backlog, but later there's still a GC operation to unlink a version from that table. The DAManager should process an action that is not safe to be processed. Unclear if this is a bug in the logic that enqueued the dropping of the table (needs another deferral?). It looks like it was a catalog table based on the stack trace for the initial allocation.

### Expected Behavior
Don't free a table before everything is unlinked.

### Actual Behavior
It freed a table before everything was unlinked.

### LINKED ISSUE #1267: jumbotest_storage_other test failure
# jumbotest_storage_other Test Failed

## Summary
The jumbotest_storage_other failed once when running in Jenkins. I was not able to reproduce and I am not familiar with the tests so I am not sure of a possible cause.  
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/collect-microbenchmark-results/75/pipeline

## Environment

**OS**: ubuntu-20.04

**Compiler**: clang-8.0

## Steps to Reproduce
I was not able to reproduce

### Expected Behavior
I'm not sure, other than the test passing.

jrolli:
# Fix GC issue in UnrecoverableTransactionTest

Fixes #1267
Fixes #1205 (duplicate of 1267)

## Description

The `UnrecoverableTransactionTest` fails to fully clear the GC queue prior to tearing down the catalog.  This causes two deferrals to attempt to deallocate the database object created on line 463 and manifests as an ASAN use-after-free error on one of the catalog tables.  Doing a restart of the "old" system after aborting has the side-effect of forcing a flush of the GC queue.

## Additional Discussion (Summary of Slack conversation)

@lmwnshn, commented that this PR is somewhat hacky as it takes a heavy-handed response to a very specific problem which likely is lurking elsewhere.  I advocated for fixing it here in the test as the long term solution is likely to change once #1157 makes it into master.


lmwnshn (approved):
LGTM. Will merge when CI passes and after we throw in the slack discussion

mbutrovich:
My only concern: doesn't this test use `DBMain`? I thought `DBMain` was set up in such a way to make sure everything gets cleared in the correct order. Is there a fix to be made in `DBMain` rather than in the test itself?

mbutrovich:
More specifically: DBMain's destructor should already call what the shutdown part of `ShutdownAndRestartSystem` call that @jrolli added does. Is there something about how the deferral of aborted catalog bootstraps changes the minimum number of GC invocations required to cleanly shut down the system? See `transaction_defs.h` `MIN_GC_INVOCATIONS`.

jrolli:
I’ll take another look at it and see if I can see if this is an issue in DBMain.

jrolli:
@mbutrovich, relooking the code it is I'm not sure whether it is an issue with DBMain or not.  The "promise" to catalog Teardown function is that there are no active transactions and all Undo records have been pruned.

We could solve this by doing GC-Teardown-GC in the catalog layer's destructor, but I think the intent was it is on the traffic cop (which does not exist in this test) to decide when it's done pruning transactions and invoke GC since the catalog cannot reason about that state.

In my view this PR forces the test to play by the rules, but it's an easy adjustment if we want to play it safe.

mbutrovich (approved):
Nice fix and a clear explanation in the PR and docs in the code! Thanks @jrolli!

PR #1278, opened 2020-10-28
Disable messenger by default.

lmwnshn:
# Disable messenger by default.

## Description

Also fix a settings network port fuckup that came in the Messenger PR.

### DIFF
--- a/src/include/settings/settings_defs.h
+++ b/src/include/settings/settings_defs.h
@@ -316,4 +316,12 @@ SETTING_bool(
     true,
     noisepage::settings::Callbacks::NoOp
 )
+
+SETTING_bool(
+    use_messenger,
+    "Whether to enable the messenger (default: false)",
+    false,
+    false,
+    noisepage::settings::Callbacks::NoOp
+)
     // clang-format on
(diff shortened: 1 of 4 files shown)


src/main/noisepage.cpp
```diff
@@ -80,7 +80,7 @@ int main(int argc, char *argv[]) {
                      .SetUseExecution(true)
                      .SetUseTrafficCop(true)
                      .SetUseNetwork(true)
-                     .SetUseMessenger(true)
+                     .SetUseMessenger(false)
```
mbutrovich, line 83: Shouldn't need this since we're not overriding the default value.

src/include/settings/settings_defs.h
```diff
@@ -316,4 +316,12 @@ SETTING_bool(
     true,
     noisepage::settings::Callbacks::NoOp
 )
+
+SETTING_bool(
+    use_messenger,
```
mbutrovich, line 321: I think the convention we standardized on is `<component>_enable` for settings name.

mbutrovich (changes requested):
Nits. Thanks for the quick turnaround on this!

mbutrovich (approved):
LGTM!

PR #1279, opened 2020-10-29
fix typo in SetIndexedUpdate in Update Plannode

mush-zhang:
# Fix Typo in SetIndexedUpdate in Update Plannode

## Description:
  
The indexed_update_ variable is not set according to the passed in parameter, but always to true. Fixed it by letting indexed_update_ adopt the boolean value passed into this function.


lmwnshn (approved):
LGTM. Thanks for the catch!

mbutrovich:
I think this is blocked on #1287.

PR #1282, opened 2020-10-29
Forward declare more stuff in expressions, plan nodes, error data.

lmwnshn:
# Forward declare more stuff in expressions, plan nodes, error data.

## Description

We do **not** see a speed improvement with unity builds on. I think the reason unity builds help us so much is because we're not forward declaring things properly - globbing everything together lets the `#pragma once` in our headers kind of save us there.

Though I expect to see a speed improvement with unity builds off.

Given the lack of improvement, I don't know if we want to merge this.


mbutrovich (approved):
Even if Unity builds aren't affected much, this will help things like clang-tidy. Thanks!

PR #1285, opened 2020-10-29
Fix clang-tidy.

lmwnshn:
# Fix clang-tidy.

## Description

clang-tidy is busted with unity builds. To reproduce, grab current master: clang-tidy blows up on messenger with unity builds off, doesn't blow up with unity builds on. I don't understand why because in theory it scans over the same code anyway.

This PR:

- moves clang-tidy to the format stage. This should be good because (1) clang-tidy is deterministic, our tests are not, so this saves us time when restarting Test stage; (2) we have a ready-for-ci tag, so CI won't reach the format stage unless we want it to, so we don't have too much wasted resource consumption there
- fixes clang-tidy stuff that was not caught.

Thanks @jkosh44 for pointing out stuff was busted

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,6 +51,7 @@ pipeline {
                         sh 'cd build && timeout 20m ninja check-format'
                         sh 'cd build && timeout 20m ninja check-lint'
                         sh 'cd build && timeout 20m ninja check-censored'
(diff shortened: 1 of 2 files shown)


mbutrovich (approved):
LGTM!

PR #1286, opened 2020-10-29
Fix UPDATE foo SET a=a in the binder.

lmwnshn:
# Fix UPDATE foo SET a=a in the binder.

## Description

There was a binding error for updates of the form `UPDATE foo SET a=a` because the check for `a`'s type happened before `a` was given a type from the schema.

Comes with a test.

Thanks @17zhangw for pointing out the bug.


mbutrovich (approved):
That's a good catch. Thanks for the test, too! LGTM

PR #1287, opened 2020-10-29
Fix update.test Tracefile

### LINKED ISSUE #1284: Testing: floating point precision causing trace file failures
# Bug Report
## Summary
CI trace file tests causing false failures due to floating point precision differences.

## Environment
Jenkins CI.

## Steps to Reproduce
See: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-1279/12/pipeline

### Expected Behavior
Trace file tests shouldn't fail for floating point precision differences.

### Actual Behavior
Trace file tests fail for floating point precision differences. See link above. If I'm reading that correctly, the results have the same raw values, but they hashed differently so the test failed.

jrolli:
# Fix `update.test` Tracefile

Fixes #1284 

## Description

This is another set of tests that falsely assumed that Noisepage and Postgres would return identical tuple orders on unsorted data.  To fix this, this PR removes hashed results which are both hard to debug and implicitly assume a sort order, and instead relies on the "rowsort" option (replacing "nosort") to force the sorting of both the result set and expected in Java.

Additionally, this PR includes the original SQL file along with the command to generate it to facilitate any future updates or new tests in the future.

## Further work

This appears to be a larger issue throughout all of the tracefiles and probably needs a to be replicated across the other JUnit tests.

### DIFF
--- a/script/testing/junit/traces/update.test
+++ b/script/testing/junit/traces/update.test
@@ -17,7 +17,7 @@ statement ok
 UPDATE update1 SET c2 = '2020-01-03 11:22:33.721058' WHERE c1 = 2;
 
 query I rowsort
-SELECT * from update1 ORDER BY c1 ASC;
+SELECT * from update1;
 ----
 1
 2020-01-02 12:23:34.567893
@@ -49,14 +49,14 @@ statement ok
 UPDATE update2 SET c2 = 4 WHERE c1 = 22;
 
 query II rowsort
-SELECT * FROM update2 ORDER BY c2;
+SELECT * FROM update2;
 ----
 1
 1
-22
-4
 23
 33
+22
+4
 
 query II rowsort
 SELECT * FROM update2 WHERE c2=22;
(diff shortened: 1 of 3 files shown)


script/testing/junit/traces/update.test
```diff
@@ -155,12 +201,13 @@ INSERT INTO update5 (a) VALUES (1),(2),(3);
 statement ok
 UPDATE update5 SET a=a;
 
-query I nosort
-SELECT a FROM update5;
+query I rowsort
+SELECT a FROM update5 ORDER BY a;
```
lmwnshn, line 205: Both rowsort and order by?

jrolli: I botched a save... Will fix and push the updated version.

mbutrovich (approved):
LGTM, thanks for looking at this @jrolli!

PR #1288, opened 2020-10-30
Model Server Manager Infrastructure for Pilot

githubxxcc:
# Model Server Infrastructure 

## Description
Model server infrastructure to : 
1. Test mini_runner/mini_trainer pipeline 
2. Allow ModelServerManager(c++) to interact with ModelServer(python)


githubxxcc:
@lmwnshn could you please take a look?  If you run `model_server_test`, the test will try to send a message to the python model, but apparently it is triggering `Host unreachable` when sending the message. (You should see that I catch that exception in SendMessage function) 

I don't think it is caused by the python side since it doesn't even get to sending, so it must be something wrong with regards to how I am using the messenger. I was wondering if there is some race between sending the message and initializing the socket.

src/model_server/model_server_manager.cpp
```diff
@@ -0,0 +1,132 @@
...
+    // Wait for the child
+    wait_pid = waitpid(py_pid_, &status, 0);
+
+    if (wait_pid < 0) {
+      NETWORK_LOG_ERROR("Failed to wait for the child process...");
+      return;
+    }
+  } else {
+    // Run the script in in a child
+    std::string python3_bin("/usr/local/bin/python3");
```
lmwnshn, line 85: This path is not guaranteed to work, e.g. does not work on my system (located at /usr/bin/python3, and macs will be different).

script/installation/packages.sh
```diff
@@ -72,6 +72,8 @@ PYTHON_PACKAGES=(\
   "requests" \
   "psutil" \
   "distro"  \
```
lmwnshn, line 74: I had to manually install `tqdm`.

lmwnshn: And `lightgbm`.

script/model/model_server.py
```diff
@@ -0,0 +1,249 @@
...
+from mini_trainer import MiniTrainer
+from util import logging_util
+import logging
+
+
+logging_util.init_logging('info')
+# Initialize ZMQ connection with the c++ side
+end_point = sys.argv[1]
+context = zmq.Context()
+socket = context.socket(zmq.DEALER)
```
lmwnshn, line 40: Seems to work if you replace with the following:

```
socket.set_string(zmq.IDENTITY, 'model')
logging.info(f"Python model trying to connect to manager at {end_point}")
socket.connect(f"ipc://{end_point}")
logging.info(f"Python model connected to {end_point}")
socket.send_multipart([''.encode('utf-8'), '0-0-mario'.encode('utf-8')])
```

Inside your MSM, you are asking it to send a message to "model". If you do not set the identity of the Python MS to "model", then it says host unreachable because there is nobody called "model".

src/include/main/db_main.h
```diff
@@ -815,8 +844,8 @@ class DBMain {
       gc_metrics_ = settings_manager->GetBool(settings::Param::gc_metrics_enable);
       bind_command_metrics_ = settings_manager->GetBool(settings::Param::bind_command_metrics_enable);
       execute_command_metrics_ = settings_manager->GetBool(settings::Param::execute_command_metrics_enable);
-
-      use_messenger_ = settings_manager->GetBool(settings::Param::messenger_enable);
```
githubxxcc, line 819: hmmm this should not be deleted.

script/model/model_server.py
```diff
@@ -0,0 +1,254 @@
+#!/usr/bin/env python3
+"""
+This file contains the python model server implementation.
+
+The server should be stateless but with caching of models.
+The message format that the model-server expects should be kept consistent with Messenger class in
```
lmwnshn, line 6: nit: can we consistently find and replace

model-server
modelserver
model server

etc

to be a single spelling? Maybe ModelServer?

script/model/model_server.py
```diff
@@ -0,0 +1,254 @@
...
+
+TODO(Ricky):
+- Encapsulate the modelserver in a class
+- Modelserver side of callbacks (once we are more clear on the actions)
+- Model invocation could be improved.
+
+"""
+
+from __future__ import annotations
+import zmq
```
lmwnshn, line 19: nit: arrange includes to be standard python include order? generally,

system packages before custom packages
alphabetical

src/include/main/db_main.h
```diff
@@ -424,6 +425,12 @@ class DBMain {
         messenger_layer = std::make_unique<MessengerLayer>(common::ManagedPointer(thread_registry), messenger_port_,
                                                            messenger_identity_);
       }
+      std::unique_ptr<model::ModelServerManager> model_server_manager = DISABLED;
+      if (pilot_enable_) {
```
lmwnshn, line 429: rename all instances of Pilot to ModelServerManager? model_server_manager_enable_, ...

src/include/main/db_main.h
```diff
@@ -714,6 +722,25 @@ class DBMain {
       return *this;
     }
 
+    /**
+     *
```
lmwnshn, line 726: delete this line?

src/include/main/db_main.h
```diff
@@ -714,6 +722,25 @@ class DBMain {
...
+     * @param value with pilot
+     * @return  self reference for chaining
```
lmwnshn, line 728: slightly weird spacing, either have both param and return tab aligned or just use one space

src/include/main/db_main.h
```diff
@@ -714,6 +722,25 @@ class DBMain {
...
+     */
+    Builder &SetWithPilot(const bool value) {
+      pilot_enable_ = value;
+      return *this;
+    }
+
+    /**
+     * @param value model_server_path
+     * @return  self reference for chaining
```
lmwnshn, line 737: slightly weird spacing, either have both param and return tab aligned or just use one space

src/include/main/db_main.h
```diff
@@ -714,6 +722,25 @@ class DBMain {
       return *this;
     }
 
+    /**
+     *
+     * @param value with pilot
+     * @return  self reference for chaining
+     */
+    Builder &SetWithPilot(const bool value) {
```
lmwnshn, line 730: rename pilot to MSM?

src/include/main/db_main.h
```diff
@@ -787,7 +816,6 @@ class DBMain {
         wal_persist_threshold_ =
             static_cast<uint64_t>(settings_manager->GetInt64(settings::Param::wal_persist_threshold));
       }
```
lmwnshn, line 818: put this line back?

src/include/main/db_main.h
```diff
@@ -815,8 +843,9 @@ class DBMain {
       gc_metrics_ = settings_manager->GetBool(settings::Param::gc_metrics_enable);
       bind_command_metrics_ = settings_manager->GetBool(settings::Param::bind_command_metrics_enable);
       execute_command_metrics_ = settings_manager->GetBool(settings::Param::execute_command_metrics_enable);
```
lmwnshn, line 845: put this line back too?

src/include/model_server/model_server_manager.h
```diff
@@ -0,0 +1,127 @@
+#pragma once
+
+#include <filesystem>
```
lmwnshn, line 3: as far as possible, forward declare these includes and put them into msm.cpp?

src/include/model_server/model_server_manager.h
```diff
@@ -0,0 +1,127 @@
...
+#include "loggers/messenger_logger.h"
+#include "messenger/messenger.h"
+#include "messenger/messenger_defs.h"
+
+namespace noisepage::model {
+static constexpr const char *MODEL_CONN_ID_NAME = "model-server-conn";
+static constexpr const char *MODEL_TARGET_NAME = "model";
+static constexpr const char *MODEL_IPC_PATH = "/tmp/model-server-ipc";
+static constexpr const char *MODEL_TCP_HOST = "127.0.0.1";
+static constexpr const int MODEL_TCP_PORT = 15645;
```
lmwnshn, line 14: this is more of a note to me, but we need to start a definitive "these are the ports we are using"

src/include/model_server/model_server_manager.h
```diff
@@ -0,0 +1,127 @@
...
+static constexpr const int INVALID_PID = 0;
+
+/*
+ * Use 128 as convention to indicate failure in a subprocess:
+ * https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html
+ */
+static constexpr const unsigned char MODEL_ERROR_BINARY = 128;
+
+/**
+ * Inteface for pilot related operations
```
lmwnshn, line 24: typo: interface

also rename pilot -> modelserver

src/include/settings/settings_defs.h
```diff
@@ -324,4 +324,20 @@ SETTING_bool(
     false,
     noisepage::settings::Callbacks::NoOp
 )
+
+SETTING_bool(
+    pilot_enable,
```
lmwnshn, line 329: pilot -> msm?

src/loggers/model_logger.cpp
```diff
@@ -0,0 +1,16 @@
+#include "loggers/model_logger.h"
+
+#include <memory>
+
+namespace noisepage::model {
```
lmwnshn, line 5: be consistent on whether model or model_server? based on your folder names

src/messenger/messenger.cpp
```diff
@@ -461,12 +461,17 @@ void Messenger::SendMessage(common::ManagedPointer<ConnectionRouter> router_id,
 
   // Build and send the message. Note that ConnectionRouter is a ROUTER socket.
   zmq::message_t router_data(recv_id.data(), recv_id.size());
-  if (router_id->socket_->send(router_data, zmq::send_flags::sndmore).has_value()) {
-    ZmqMessage reply = ZmqMessage::Build(send_msg_id, recv_cb_id, router_id->identity_, message);
-    ZmqUtil::SendMsgIdentity(common::ManagedPointer(router_id->socket_.get()), router_id->identity_);
-    ZmqUtil::SendMsgPayload(common::ManagedPointer(router_id->socket_.get()), reply);
-    MESSENGER_LOG_TRACE(fmt::format("[PID={}] Messenger ({}) SENT-TO {}: {} ", ::getpid(), router_id->identity_,
-                                    recv_id, reply.GetRawPayload()));
+  try {
```
lmwnshn, line 464: This is the first thing where I think we should discuss briefly. I think it makes more sense for the **caller** of this function to catch the exception so that they have a choice. If we need to expose some debugging information I'm OK with putting all relevant debug information into the exception, see the exceptions that currently take error codes.

lmwnshn (changes requested):
Generally looks pretty good, nice job!

I have one concern being the wrapping of SendMessage in an exception, I think we want all exceptions to propagate to the caller with all the relevant debugging information embedded inside.

The rest is nitpicking over style/naming/grammar/etc.

I'll take a second pass later, brain a bit fried. overall looks good though!

script/installation/packages.sh
```diff
@@ -72,6 +72,10 @@ PYTHON_PACKAGES=(\
...
+  "pyzmq" \
+  "tqdm" \
+  "lightgbm" \
+  "skylearn" \
```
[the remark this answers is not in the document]
githubxxcc: oops good catch!

src/include/main/db_main.h
githubxxcc: Is @rabbit721 Pilot gonna be something the MSM should depend on or vice versa, or they are independent components of the system?
I saw in her demo today that she had a `pilot_enable` flag, not sure if we want to use that to enable MSM as well?

src/messenger/messenger.cpp
githubxxcc: I see, having the caller handle any exceptions makes sense. I will revert the change.

src/include/main/db_main.h
rabbit721: From Lin's previous messages on slack, Pilot should be able to invoke the mini_trainer and do inference of models through MSM. So I think we would need MSM enabled when the Pilot is enabled.

malin1993ml: The pilot is dependent on the MSM since the pilot needs to perform inference through MSM. The training might be separate though. We might directly invoke the trainers once at system start time or just directly import some pre-trained models (by specifying the model path during inference).

script/model/model_server_test.py
```diff
@@ -0,0 +1,166 @@
...
+
+    def test_inference(self):
+        """
+        Load the test model maps, and test if inference could be performed with the model
+        """
+
+        # Test parameters
+        file_path = self.generate_seq_file()
+        result_path = MINI_TRAINER_RESULT_DIR
+        model_map = self.train_model(str(TEST_SEQ_FILES_DIR), result_path)
```
malin1993ml, line 129: I would load the model from the dumped path instead of directly use the in-memory object. This will test the model saving and loading.

src/include/model_server/model_server_manager.h
```diff
@@ -0,0 +1,127 @@
...
+   */
+  void TrainWith(const std::string &model_name, const std::string &seq_files_dir);
+
+  /**
+   * Perform inference on the given data file.
+   *
+   * This function should expect some return (Not figured out yet)
+   * @param data_file Seq file
+   * @param model_map_path Model map (should probably be obtained from TrainWith call)
+   * @bug What to do with return and callback?
```
malin1993ml, line 92: We do need to get the predictions back... what's the convention to get the results back in our message framework?

script/model/model_server.py
```diff
@@ -0,0 +1,254 @@
...
+    test_ratio = 0.9
+    trim = 0.2
+    error_bias = 1
+    data = opunit_data.get_mini_runner_data(data_file, result_path, {}, {}, trim)[0]
+    x_train, x_test, y_train, y_test = model_selection.train_test_split(data.x, data.y,
+                                                                        test_size=test_ratio,
+                                                                        random_state=0)
+    # Load the model map
+    with open(model_map_path, 'rb') as f:
+        model_map = pickle.load(f)
```
malin1993ml, line 188: I would add the cache now.

src/include/model_server/model_server_manager.h
```diff
@@ -0,0 +1,127 @@
...
+ * Inteface for pilot related operations
+ */
+class ModelServerManager {
+ public:
+  /**
+   * Construct a ModelServerManager with the given executable script to the python ModelServer
+   * @param model_bin python script path
+   * @param messenger Messenger pointer
+   */
+  ModelServerManager(const std::string &model_bin, const common::ManagedPointer<messenger::Messenger> &messenger);
```
malin1993ml, line 33: nit: rename `model_bin` to `model_server_bin`

src/model_server/model_server_manager.cpp
```diff
@@ -0,0 +1,135 @@
...
+                                       const common::ManagedPointer<messenger::Messenger> &messenger)
+    : messenger_(messenger),
+      router_(ListenAndMakeConnection(messenger, MODEL_IPC_PATH, ModelServerHandler)),
+      thd_(std::thread([this, model_bin] {
+        while (!shut_down_) {
+          this->StartModelServer(model_bin);
+        }
+      })) {}
+
+void ModelServerManager::StartModelServer(const std::string &model_path) {
```
malin1993ml, line 59: nit: `model_path` to `model_server_path`. Change the naming to be explicit in other places as well.

malin1993ml (changes requested):
In general the code quality looks pretty good! There're a few functionality change requests based on what we really need. I also think we do need to have c++ test on the end-to-end communication (we can discuss).

test/model_server/model_server_test.cpp
```diff
@@ -0,0 +1,79 @@
...
+                       .SetUseTrafficCop(true)
+                       .SetModelServerPath("../../script/model/model_server.py")
+                       .Build();
+
+    return db_main;
+  }
+};
+
+// NOLINTNEXTLINE
+TEST_F(ModelServerTest, DISABLED_TerminalTest) {
```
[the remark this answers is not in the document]
malin1993ml: Following up on this. You probably need to generate the sequence files first by invoking the `mini_runners` directly. We're not going to type any input from the command line in our testing pipeline on Jenkins.

src/messenger/messenger.cpp
```diff
@@ -516,9 +519,8 @@ void Messenger::ServerLoop() {
         bool has_custom_serverloop = poll_items.server_callbacks_[i] != nullptr;
         MESSENGER_LOG_TRACE("[PID={}] Messenger RECV-FR {} (custom serverloop: {}): {}", ::getpid(), msg.GetRoutingId(),
                             has_custom_serverloop, msg.GetRawPayload());
-        if (!has_custom_serverloop) {
-          ProcessMessage(msg);
-        } else {
+        ProcessMessage(msg);
```
githubxxcc, line 522: This change is needed if the user of Messenger wants to : 
1. Have its own router 
2. Register callback with the Messenger's Loop 

The issue for the current Messenger API is:
If I want to register a callback when `SendMsg`, I need to register it with the Messenger's main loop. But because I have a customized serverloop, I also need to evaluate and invoke the callback. 

I changed this to fa temp fix. I think there could be better ways to solve this, such as allowing the server's callback to get the `recv_cb_id` so that it could operate on the callback id.

githubxxcc:
Updates: 
- Make synchronous API for `ModelServerManager::DoInference`, and synchronous support for `ModelServerManager::TrainWith`
- Update the `model_server_test.cpp` to be end-to-end testing by running `mini_runners` 
- Refactor the code to be under `self_driving` 

Discussion: 
- Proposed some changes to the Messenger's (@lmwnshn) 
  1. message format should allow `-` in the message body
  2. how does a serverloop evaluate callbacks?


(Will be addressing some of the review comments that got collapsed (I just saw) and not resolved)
The PR is ready for review again.

src/messenger/messenger.cpp
```diff
@@ -148,7 +148,10 @@ class ZmqMessage {
       UNUSED_ATTRIBUTE int check =
           std::sscanf(payload_.c_str(), "%" SCNu64 "-%" SCNu64 "-", &send_msg_id_, &recv_cb_id_);
       NOISEPAGE_ASSERT(2 == check, "Couldn't parse the message header.");
-      message_.remove_prefix(message_.find_last_of('-') + 1);
+
```
githubxxcc, line 151: This allows for `-` in the message payload.

script/model/mini_trainer.py
```diff
@@ -50,7 +53,7 @@ def _train_specific_model(self, data, y_transformer_idx, method_idx):
         regressor.train(data.x, data.y)
         self.model_map[data.opunit] = regressor
 
-    def _train_data(self, data, summary_file):
+    def train_data(self, data, summary_file):
```
githubxxcc, line 56: Exposing these functions to be used by `model_server.py`

test/model_server/model_server_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+  };
+
+  // Send a message
+  std::string msg = "ModelServerTest";
+  ms_manager->PrintMessage(msg);
+
+  // Perform a training
+  std::vector<std::string> models{"lr"};
+  auto seq_files_dir = std::filesystem::current_path();
+  MODEL_SERVER_LOG_INFO("Training with {}, and input dir: {}", models[0], seq_files_dir);
```
malin1993ml, line 113: nit: we should print out all the models in the vector if you want this log information.

test/model_server/model_server_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+  ModelServerFuture<std::string> future;
+  ms_manager->TrainWith(models, seq_files_dir, &future);
+  auto res = future.Wait();
+  ASSERT_EQ(res.second, true);  // Training succeeds
+
+  // Perform inference
+  auto result = ms_manager->DoInference("OP_INTEGER_PLUS_OR_MINUS", features);
+  ASSERT_GT(result.size(), 0);
```
malin1993ml, line 121: Should't we check that the results have the same size as the features?

script/model/model_server.py
```diff
@@ -0,0 +1,355 @@
...
+    `model_server.py <ZMQ_ENDPOINT>`
+
+The server should be stateless but with caching of models.
+The message format that the ModelServer expects should be kept consistent with Messenger class in
+the noisepage source code.
+The command format should be kept with ModelServerManager in the noisepage source
+code.
+
+TODO(Ricky):
+- Encapsulate the ModelServer in a class
```
malin1993ml, line 15: I think this is done?

script/model/model_server.py
```diff
@@ -0,0 +1,355 @@
...
+        txn_sample_interval = ModelServer.TXN_SAMPLE_INTERVAL
+        trainer = MiniTrainer(seq_files_dir, result_path, ml_models, test_ratio, trim, expose_all,txn_sample_interval)
+        if seq_files_dir:
+            # Perform training from MiniTrainer and input files directory
+            model_map = trainer.train()
+        else:
+            """
+            NOTE:
+            This branch has not been tested yet because the MiniTrainer now heavily relies on the CSV file. 
+            So training with data directly is not yet possible without refactoring the MiniTrainer
```
malin1993ml, line 235: I don't really think we need this branch... when do you think we need to train with data directly? Where this "direct" data is going to come from?

script/model/model_server.py
```diff
@@ -0,0 +1,355 @@
...
+        return pprint.pformat(self.__dict__)
+
+
+class ModelServer:
+    """
+    ModelServer(MS) class that runs in a loop to handle commands from the ModelServerManager from C++
+    """
+
+    # Training parameters
+    RESULT_PATH = '/tmp'
```
malin1993ml, line 118: I think this should be managed by the c++ side. Both the `train` and `infer` command should include this field as a parameter. `train` only saves the models on disk at the specified path. And `infer` checks whether the models from the specified path have been loaded (cached). If not, load the models from the path.

script/model/model_server.py
```diff
@@ -0,0 +1,355 @@
...
+            opunit_data_list = self._transform_opunit_data(raw_data)
+            for data in opunit_data_list:
+                best_y_transformer, best_method = trainer.train_data(data, None)
+                if trainer.expose_all:
+                    trainer.train_specific_model(data, best_y_transformer, best_method)
+
+            # Directly get the model map
+            model_map = trainer.get_model_map()
+
+        self.g_model_map = model_map
```
malin1993ml, line 251: This would make the `model_server` stateful. I think we can remove this line with the above suggested refactor.

malin1993ml (changes requested):
Thanks for the changes, Ricky! I think the new API generally satisfies our needs. Just a few more small tweaks.

malin1993ml:
@githubxxcc Also, is this ready for CI? You need to add the ready-for-ci label in order to kick start the Jenkins tests.

test/model_server/model_server_test.cpp
githubxxcc: Good idea!

script/model/model_server.py
githubxxcc: Yeah, I thought we could get rid of the `mini_runners` dependency by just feeding the model server with structurally identical fake data, because I feel having a test that depends on a benchmark might involve too much coupling. 
But I think feeding fake direct data to the model server couldn't be done without much refactoring, so I did not go down this path.

script/model/model_server.py
githubxxcc: I think one question I have is: would there be multiple models trained at the same time?

script/model/model_server.py
githubxxcc: Oh, I think I might have made false assumptions about the use case. So the models trained would be persistent across model server lifetimes? Like the pilot would train the model A, and then expects to re-use it after the model server reboots?
I was assuming the models will be trained every ModelServer's lifetime, like train->infer->infer->infer, and nothing is persisted. 
But yeah, I could have a persisted model abstraction with filename as the handle.

test/model_server/model_server_test.cpp
```diff
@@ -0,0 +1,128 @@
...
+                       .Build();
+
+    return db_main;
+  }
+
+  static bool GenerateMiniTrainerCSV() {
+    pid_t pid;
+    if ((pid = ::fork()) == 0) {
+      std::string mini_runner = MINI_RUNNER_PATH;
+      std::string bch_filter = "--benchmark_filter=SEQ0";
```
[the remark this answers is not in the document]
githubxxcc: what could be a combination of parameters that make `mini_runners` run in one minute, or dozen of seconds? I have tried 
```
 --mini_runner_rows_limit=10 
--rerun=1 
--warm_num=0 
--create_index_small_limit=10 
--updel_limit=10 
--skip_large_rows_runs=true
```
but it did not finish in 10min

script/model/model_server.py
malin1993ml: Right I think we can remove this branch then.

script/model/model_server.py
malin1993ml: Yes. The trained models would expect to be persistent across model server lifetimes.

script/model/model_server.py
malin1993ml: There can be multiple models trained at the same time, but they would be in the same model map.

src/messenger/messenger.cpp
lmwnshn: I thought we were going to roll it into #1331 but I don't see my changes there any more.

I noticed this while prototyping replication and changed it so that the callback function has direct access to the entire `ZmqMessage`, see [here](https://github.com/lmwnshn/noisepage/blob/replication/src/include/messenger/messenger_defs.h). Roughly speaking, every different component gets its own serverloop and implements its own callbacks rather than piggybacking on the main messenger loop.

I think the replication framework PR has been languishing for a few weeks. I can open a separate PR to bring in my changes, I think this would work for you?

src/messenger/messenger.cpp
lmwnshn: yep, good catch

malin1993ml (approved):
Just a small thing otherwise LGTM! Thanks for the changes!

src/include/self_driving/model_server/model_server_manager.h
```diff
@@ -0,0 +1,239 @@
...
+
+/**
+ * Interface for ModelServerManager related operations
+ */
+class ModelServerManager {
+  enum class Callback : uint64_t { NOOP = 0, CONNECTED, DEFAULT };
+
+ public:
+  /**
+   * Construct a ModelServerManager with the given executable script to the python ModelServer
```
lmwnshn, line 129: Generic nit for file: capitalize Python

src/include/self_driving/model_server/model_server_manager.h
```diff
@@ -0,0 +1,239 @@
...
+  ~ModelServerManager() { StopModelServer(); }
+
+  /**
+   * Stop the python-ModelServer daemon by sending a message to the Python model server
+   */
+  void StopModelServer();
+
+  /**
+   * Check if the model server has started.
+   * The user of this function should poll this until the pid changes.
```
lmwnshn, line 147: Wait, why does the user also need to poll the PID? Shouldn't the PID only be changed if connected_ is changed as well?

src/include/self_driving/model_server/model_server_manager.h
```diff
@@ -0,0 +1,239 @@
...
+
+  /*******************************************************
+   * ModelServer <-> ModelServerManager logic routines
+   *******************************************************/
+
+  /**
+   * Ask the model server to print a message
+   * This function does not expect a callback
+   *
+   * @warning The message should not contain '-' as per current protocol
```
lmwnshn, line 169: Can this warning be removed now with your fix?

src/include/self_driving/model_server/model_server_manager.h
```diff
@@ -0,0 +1,239 @@
...
+   * @param msg Message to print
+   */
+  void PrintMessage(const std::string &msg);
+
+  /**
+   * Train a model with the given seq files directory
+   *
+   *  This function will be invoked asynchronously.
+   *  The caller should poll from ModelServerManager::Trained() to check if the training has been done.
```
lmwnshn, line 178: I don't see this function, and also I think this is encapsulated by the future?

src/include/self_driving/model_server/model_server_manager.h
```diff
@@ -0,0 +1,239 @@
...
+   * @param seq_files_dir Seq files's enclosing directory (seq files generated by mini_runner)
+   * @param future A future object which the caller waits for
+   */
+  void TrainWith(const std::vector<std::string> &methods, const std::string &seq_files_dir,
+                 const std::string &save_path, ModelServerFuture<std::string> *future);
+
+  /**
+   * Perform inference on the given data file.
+   *
+   * This function should expect some return (Not figured out yet)
```
lmwnshn, line 193: Sorry, can you document this a bit more? What is currently being returned and what is being figured out?

src/include/self_driving/model_server/model_server_manager.h
```diff
@@ -0,0 +1,239 @@
...
+   */
+  void StartModelServer(const std::string &model_path);
+
+  static void ModelServerHandler(common::ManagedPointer<messenger::Messenger> messenger, std::string_view sender_id,
+                                 std::string_view message, uint64_t recv_cb_id);
+
+  std::string IPCPath() const { return MODEL_IPC_PATH; }
+  std::string TCPPath() const { return fmt::format("{}:{}", MODEL_TCP_HOST, MODEL_TCP_PORT); }
+
+  /** Messenger handler **/
```
lmwnshn, line 217: Same `/** */` comment as above

lmwnshn: Or just use `///<`

src/include/self_driving/model_server/model_server_manager.h
lmwnshn: Also, this function is blocking, right? Can you add a quick note to all blocking functions' header comments, that they block until the result is available?

lmwnshn (changes requested):
Minor nits, overall looks pretty good. I like the way modeling_test reads, though I have some concerns about signaling the presence of errors.

test/messenger/messenger_test.cpp
```diff
@@ -289,7 +289,7 @@ TEST_F(MessengerTests, BasicListenTest) {
 
     // Send "replica1" to the primary to let them know who we are.
     messenger->SendMessage(common::ManagedPointer(&con_primary), "replica1", CallbackFns::Noop,
-                           static_cast<uint8_t>(Messenger::BuiltinCallback::ECHO));
+                           static_cast<uint8_t>(Messenger::BuiltinCallback::NOOP));
```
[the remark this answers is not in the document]
githubxxcc: Yes, since the current temp fix at https://github.com/cmu-db/noisepage/pull/1288/files#diff-5c77dbc0c8f92240246000b920bc808448e1e240d6f2475633133dc2e9fd892dR523 will evaluate a message both in `ProcessMsg` and in the customized server callback. 
The original code will not trigger this echo callback at all since it doesn't go through the main loop's callback evaluation. This is just to keep the test behaves the same way (And in fact, without this change, the temp fix above doesn't work) 

It might be better to have your changes to the messenger merged in the master, but this is what I need now to pass tests.

script/installation/packages.sh
```diff
@@ -73,6 +73,10 @@ PYTHON_PACKAGES=(\
   "requests" \
   "psutil" \
   "distro"  \
```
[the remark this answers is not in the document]
githubxxcc: This is hard. Brain is small.

script/model/model_server.py
```diff
@@ -0,0 +1,394 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
githubxxcc: Good idea!

src/include/settings/settings_defs.h
```diff
@@ -324,4 +324,20 @@ SETTING_bool(
...
+    model_server_enable,
+    "Whether to enable the ModelServerManager (default: false)",
+    false,
+    false,
+    noisepage::settings::Callbacks::NoOp
+)
+
+SETTING_string(
+    model_server_path,
+    "The python model server script to invoke (default: INVALID_PATH)",
```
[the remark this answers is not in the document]
githubxxcc: Was thinking the default path is usually a relative path to where the binary is invoked, so I was wanted to force user to supply the path explicitly. 
But I guess using a default relative path should also work.

src/self_driving/model_server/model_server_manager.cpp
```diff
@@ -0,0 +1,214 @@
...
+  }
+}
+
+}  // namespace noisepage::modelserver
+
+namespace noisepage::modelserver {
+
+ModelServerManager::ModelServerManager(const std::string &model_bin,
+                                       const common::ManagedPointer<messenger::Messenger> &messenger)
+    : messenger_(messenger), thd_(std::thread([this, model_bin] {
```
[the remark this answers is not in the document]
githubxxcc: I remember changing it to reference would have the `model_bin` being claimed by GC somehow.. (Or something that I ran into where this string gets corrupted)

I am not entirely sure how/why this happens.

src/include/self_driving/model_server/model_server_manager.h
```diff
@@ -0,0 +1,239 @@
...
+ *
+ *      auto result = future.Wait();
+ *      bool success = result.first;
+ *      auto data = result.second;
+ * ```
+ *
+ * @tparam Result the future's result
+ */
+template <class Result>
+class ModelServerFuture {
```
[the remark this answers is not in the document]
githubxxcc: Yes, caller of any async API such as `TrainWith` needs to provide a future pointer. So it will be used outside of ModelServerManager.

test/self_driving/modeling_test.cpp
```diff
@@ -0,0 +1,101 @@
...
+
+  // Perform inference
+  auto result = ms_manager->DoInference("OP_INTEGER_PLUS_OR_MINUS", save_path, features);
+  ASSERT_EQ(result.size(), features.size());
+  result = ms_manager->DoInference("OP_DECIMAL_COMPARE", save_path, features);
+  ASSERT_EQ(result.size(), features.size());
+  result = ms_manager->DoInference("OP_INTEGER_MULTIPLY", save_path, features);
+  ASSERT_EQ(result.size(), features.size());
+
+  // Model at another path should not exist
```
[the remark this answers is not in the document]
githubxxcc: Really good point!

Jenkinsfile
```diff
@@ -522,7 +522,12 @@ pipeline {
             }
         }
         stage('End-to-End Performance') {
-            agent { label 'benchmark' }
+            agent {
+                docker {
+                    image 'noisepage:focal'
+                    args '--cap-add sys_ptrace -v /jenkins/ccache:/home/jenkins/.ccache'
+                }
+            }
```
malin1993ml, line 530: Why are we changing the performance test machine?

Jenkinsfile
```diff
@@ -569,6 +569,31 @@ pipeline {
                  }
              }
         }
+        stage('Self-driving end-to-end test') {
+            agent { label 'benchmark' }
```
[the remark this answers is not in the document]
malin1993ml: How much speed different are there between the benchmark machines and other machines? In general it's better to assign a reasonable machine to the self-driving test because the runners would consume some time to generate the training data.

Jenkinsfile
githubxxcc: Do we need to keep track of performance of this pipeline?

malin1993ml: What do you mean by `keep track of performance`? I guess my point is more about that you're changing the machine for the `End-to-End Performance` stage, which is not the stage we're adding.

Jenkinsfile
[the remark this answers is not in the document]
lmwnshn: The benchmark machines are actually not necessarily the strongest machines, e.g., 

- benchmark: 2.40GHz, 24 CPUs
- newer swarm machine: 3.10GHz, 4 CPUs

What makes the benchmark machines special is that jobs get exclusive use of the machine.
This helps (but also not really) to reduce variance in End-To-End Performance.

Jenkinsfile
```diff
@@ -581,6 +586,8 @@ pipeline {
                 cmake -GNinja -DNOISEPAGE_UNITY_BUILD=ON -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DNOISEPAGE_USE_ASAN=OFF -DNOISEPAGE_USE_JEMALLOC=ON -DNOISEPAGE_BUILD_TESTS=OFF  -DNOISEPAGE_BUILD_SELF_DRIVING_TESTS=ON -DNOISEPAGE_UNITTEST_OUTPUT_ON_FAILURE=ON ..
                 ninja mini_runners''', label: 'Self-driving tests (Compile mini_runners)'
 
+                // mini_runners parameters are arbitrary picked so that it finished within reasonable time (<10min) and
```
lmwnshn, line 589: Sorry, could you elaborate more?

1. Most people do not touch mini_runners, so they do not know what any of these parameters are.
2. Furthermore, if this is just an end-to-end functionality test that does not measure performance, why does it matter how big the data generated is? If the data size does not matter, shouldn't the parameters be picked so that all components are exercised (rather than aiming to hit some arbitrary testing time)?

Template suggestion:

```
// The parameters to the mini_runners target are (arbitrarily picked to complete tests within a reasonable time / picked to exercise all OUs).
// Specifically, the parameters chosen are:
// - mini_runner_rows_limit does (DESCRIPTION). This is set to 1000 because (REASON).
// - rerun does (DESCRIPTION). This is set to 0 because (REASON).
// - warm_num does (DESCRIPTION). This is set to 1 because (REASON).
```

Jenkinsfile
```diff
@@ -589,6 +596,7 @@ pipeline {
                 sh script: '''
                 cd build
                 export BUILD_ABS_PATH=`pwd`
+                sudo ln -s /usr/bin/python3 /usr/bin/python
```
lmwnshn, line 599: Why would this be necessary? This seems dangerous and will break random scripts on the target machine that assume `python` is `python2`.

also @lmwnshn to check later whether the benchmark machines have had `python` permanently modified, I remember it doesn't reset cleanly sometimes.

lmwnshn (changes requested):
Some quick comments mostly pertaining to Jenkins, the diff of addressed changes looks ok but I will probably need to review the whole PR again for better context.

Jenkinsfile
githubxxcc: Right, I am still figuring out how to run this correctly with Jenkins. I ran into a  `python` not found with the `shabang` here: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/PR-31/13/pipeline

Jenkinsfile
githubxxcc: Yeah, I think this is why @malin1993ml picked those parameters. I could elaborate more on this.

CMakeLists.txt
```diff
@@ -907,6 +907,7 @@ function(add_noisepage_test
     add_dependencies(${TEST_LABEL} ${TEST_NAME})
 endfunction()
 
+# Jumbo tests and self-driving end-to-end tests will be other test targets, so not included in unittests test sources
```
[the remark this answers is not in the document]
githubxxcc: Oh shoot, right.

Jenkinsfile
githubxxcc: Oh yes, this was a good catch.

Jenkinsfile
```diff
@@ -569,6 +569,46 @@ pipeline {
...
+                mkdir build
+                cd build
+                cmake -GNinja -DNOISEPAGE_UNITY_BUILD=ON -DCMAKE_CXX_COMPILER_LAUNCHER=ccache -DCMAKE_BUILD_TYPE=Release -DNOISEPAGE_USE_ASAN=OFF -DNOISEPAGE_USE_JEMALLOC=ON -DNOISEPAGE_BUILD_TESTS=OFF  -DNOISEPAGE_BUILD_SELF_DRIVING_TESTS=ON ..
+                ninja mini_runners''', label: 'Self-driving tests (Compile mini_runners)'
+
+                // The parameters to the mini_runners target are (arbitrarily picked to complete tests within a reasonable time / picked to exercise all OUs).
+                // Specifically, the parameters chosen are:
+                // - mini_runner_rows_limit=100, which sets the maximal number of rows/tuples processed to be 100 (small table)
+                // - rerun=0, which skips rerun since we are not testing benchmark performance here
+                // - warm_num=0, which skips the warmup iterations.
```
malin1993ml, line 593: How about setting `warm_num` to 1? This way we can also test the warmup part of the code. I've tried that increasting the rerun iterations would multiply the runtime. But having a warmup iteration does not seem too bad.

githubxxcc: Oh sure, thought it is just data generation for running the `mini_runners`.

lmwnshn (approved):
LGTM! Thanks for addressing the comments.

lmwnshn:
I need to investigate what's going on with the messenger_test. I realize that locally it is not outputting the same termination messages, looks like something weird happened at teardown.

PR #1290, opened 2020-11-01
Rohan-style Fixed-Point Decimal Data Type

rohanaggarwal7997:
This pull request aims to provide a fixed point decimal package for noisepage aiming to provide fixed point support with techniques used from Hacker's Delight. Addition and Subtraction are natively fast due to the format chosen. Multiplication and Division are made faster by techniques and optimizations from Hacker's Delight. There are some specific optimizations that have not been done and are specifically mentioned in comments why they were avoided at this point. 

The package supports 38 digit decimals with upto 38 digits after the decimal point and throws overflow exceptions whenever anything bad happens. The package has been tested multiple times with 10^7 test cases for multiplication and 3.8 * 10^7 test cases for division in all possible digits after the decimal point.

The API calls for the various package operations are properly documented in runtimetypes.h
The package has been fairly tested and certain boundary test cases have been kept. 

Three additional files are added in the scripts folder - 

magic_number_generator.py - Allows the user to generate the magic number for any constant to speed up the division process, instructions are provided in the script to add the magic number in the package.

decimal_multiplication_generator.py - Generates a regression test case file to check the multiplication algorithm

decimal_division_generator.py - Generates a regression test case file to check the division algorithm


src/include/common/error/error_code.h
```diff
@@ -179,6 +179,7 @@ enum class ErrorCode : uint16_t {
   ERRCODE_TOO_MANY_JSON_ARRAY_ELEMENTS,                     // 2203D
   ERRCODE_TOO_MANY_JSON_OBJECT_MEMBERS,                     // 2203E
   ERRCODE_SQL_JSON_SCALAR_REQUIRED,                         // 2203F
+  ERRCODE_DECIMAL_RESULT_OVERFLOW,                          // 22040
```
mbutrovich, line 182: Out of curiosity: where did this come from? We pulled an updated error code list from Postgres when this file was generated a few months ago.

src/include/execution/sql/decimal_magic_numbers.h
```diff
@@ -0,0 +1,269 @@
+#pragma once
+
+#include <unordered_map>
+
+#include "storage/storage_defs.h"
```
mbutrovich, line 5: What does this file need `storage_defs.h` for?

src/include/common/error/error_code.h
rohanaggarwal7997: I searched overflow and didnt find anything

src/include/execution/sql/decimal_magic_numbers.h
```diff
@@ -0,0 +1,269 @@
...
+  /// High Middle 64 bits
+  uint128_t B;
+  /// Low Middle 64 bits
+  uint128_t C;
+  /// Lowest 64 bits
+  uint128_t D;
+  /// value - p in magic division
+  unsigned p;
+  /// Algo type
+  unsigned algo;
```
mbutrovich, line 99: Should this be a more descriptive type if it can only be 0 or 1, like an enum?

src/include/common/error/error_code.h
rohanaggarwal7997: I think the reason is postgres supports arbitrary precision decimals through strings

src/include/execution/sql/decimal_magic_numbers.h
```diff
@@ -0,0 +1,269 @@
...
+  /// Highest 64 bits
+  uint128_t A;
+  /// High Middle 64 bits
+  uint128_t B;
+  /// Low Middle 64 bits
+  uint128_t C;
+  /// Lowest 64 bits
+  uint128_t D;
+  /// value - p in magic division
+  unsigned p;
```
mbutrovich, line 97: More descriptive type? `uint32_t`?

src/include/execution/sql/decimal_magic_numbers.h
rohanaggarwal7997: It doesn't i'll fix this

src/include/common/error/error_code.h
mbutrovich: So did we make up that error code?

src/include/execution/sql/decimal_magic_numbers.h
rohanaggarwal7997: isnt unsigned and uint32_t same?

src/include/common/error/error_code.h
rohanaggarwal7997: Yes

src/include/execution/sql/decimal_magic_numbers.h
rohanaggarwal7997: ill fix this

src/include/execution/sql/decimal_magic_numbers.h
mbutrovich: Prefer the more descriptive type.

rohanaggarwal7997: cool i'll fix this

mbutrovich: That comment applies throughout this PR as I skim it. `unsigned` -> `uint32_t`, etc.

We use the latter for NoisePage.

rohanaggarwal7997: cool i'll fix everywhere

src/include/execution/sql/decimal_magic_numbers.h
```diff
@@ -0,0 +1,269 @@
...
+    {5, {0xcccccccccccccccc, 0xcccccccccccccccc, 0xcccccccccccccccc, 0xcccccccccccccccd, 258, 0}},
+    {7, {0x2492492492492492, 0x4924924924924924, 0x9249249249249249, 0x2492492492492493, 259, 1}},
+    {777, {0xa8b098e00a8b098e, 0x00a8b098e00a8b09, 0x8e00a8b098e00a8b, 0x098e00a8b098e00b, 265, 0}},
+    {999, {0x6680a40106680a4, 0x0106680a40106680, 0xa40106680a401066, 0x80a40106680a4011, 266, 1}},
+};
+
+/* Map of powers of 2
+ * This map stores powers of two to be used during
+ * constant division of a decimal with a power of 2*/
+std::unordered_map<uint128_t, unsigned> PowerTwo = {
```
mbutrovich, line 203: Isn't this hash map equivalent to a `ctz` or `ffs` instruction on the key?

mbutrovich: Hm, maybe such a thing doesn't exist for 128-bit? I still think there has to be a faster way than hashing the key and chasing pointers.

rohanaggarwal7997: Yes but that instruction is not uniformly supported by all machines
I have a nlz function to do for arbritary functions in runtime_type.cpp
I do not use it for powers of 2 because the map will be faster

mbutrovich: What machines are unable to support it? What's the feature set required on the CPU?

rohanaggarwal7997: Also to ur point i dont think it exists for 128 bit

rohanaggarwal7997: Ill follow up on this

mbutrovich: I guess I'm generally uncomfortable with the algorithm relying on STL `unordered_map`s for constant lookups. If it can't be done as an instruction, then maybe it should be something that's compile-time constant (i.e. macros though we don't prefer to use those).

@pmenon may have more accurate comments about performance considerations for this sort of stuff though.

mbutrovich: That said, for readability, correctness testing, and basic proof-of-concept, I believe these are absolutely the right ways to start the implementation.

rohanaggarwal7997: Well so what actually happened was that the hacker's delight book mentions that the clz function is not supported everywhere and not even by all gcc implementations, if i remember correctly. 
Then i searched online for a 128 bit implementation using the clz and i got the result that uses it on each 64 bit half

https://stackoverflow.com/questions/28423405/counting-the-number-of-leading-zeros-in-a-128-bit-integer

Then I implemented the nlz function for arbritary numbers in runtime_types.cpp
and used the map here

I think we can change it to whatever @pmenon and you think is best

src/include/common/error/error_code.h
mbutrovich: Adding new error codes has larger implications for the system and Postgres compatibility. We should decide as a group what to do if a specific error code we want doesn't exist. Typically you would just defer to the general error code for that class if a more specific one doesn't exist. In this case, it would be `ERRCODE_DATA_EXCEPTION,       // 22000`.

mbutrovich:
This is really impressive, FYI.

benchmark/storage/decimal_benchmark.cpp
```diff
@@ -0,0 +1,74 @@
...
+
+  // Read buffers pointers for concurrent reads
+  std::vector<execution::sql::Decimal128> decimals_;
+  std::vector<int128_t> floats_;
+};
+
+// Insert the num_inserts_ of tuples into a DataTable concurrently
+// NOLINTNEXTLINE
+BENCHMARK_DEFINE_F(DecimalBenchmark, AddDecimal)(benchmark::State &state) {
+  // NOLINTNEXTLINE
```
lmwnshn, line 32: Is this line necessary?

benchmark/storage/decimal_benchmark.cpp
```diff
@@ -0,0 +1,74 @@
...
+      for (unsigned i = 0; i < decimals_.size(); i++) {
+        result += decimals_[i];
+      }
+  }
+  state.SetIterationTime(static_cast<double>(elapsed_ms) / 1000.0);
+}
+
+// NOLINTNEXTLINE
+BENCHMARK_DEFINE_F(DecimalBenchmark, AddFloat)(benchmark::State &state) {
+  // NOLINTNEXTLINE
```
lmwnshn, line 47: Is this line necessary?

script/magic_numbers/Decimal_Division_Generator.java
```diff
@@ -0,0 +1,75 @@
...
+	{
+
+		int no_of_test_cases = 1000000;
+
+		Random rand = new Random(); //instance of random class
+        //generate random values from 0-10
+      	int upperbound = 10;
+      	
+
+      	for(int i = 0; i < no_of_test_cases; i++) {
```
lmwnshn, line 16: could you please run intellij's java autoformat or something similar on this and other java files?

script/magic_numbers/decimal_multiplication_generator.py
```diff
@@ -0,0 +1,47 @@
+import math
+
+from random import seed
+from random import randint
+# seed random number generator
```
lmwnshn, line 5: could you similarly please run intellij/clion/something autoformat on this file? thanks!

src/execution/compiler/compiler.cpp
```diff
@@ -98,6 +100,8 @@ void Compiler::Run(Compiler::Callbacks *callbacks) {
 
   callbacks->EndPhase(Phase::BytecodeGeneration, this);
 
+  ast::AstPrettyPrint::Dump(std::cerr, this->GetAST());  // NOLINT
```
lmwnshn, line 103: make sure this gets removed before merge

lmwnshn: the include above too

src/include/common/error/error_code.h
lmwnshn: was consensus reached?

mbutrovich: > was consensus reached?

I don't recall it being brought up in a meeting.

src/include/execution/sql/runtime_types.h
```diff
@@ -412,6 +412,16 @@ class EXPORT Timestamp {
 template <typename T>
 class EXPORT Decimal {
  public:
+  /** This function converts an input decimal string to the
```
pmenon, line 415: Format these comments to be consistent with the others in this file.

src/include/execution/sql/runtime_types.h
```diff
@@ -488,9 +498,93 @@ class EXPORT Decimal {
...
+   * @param value the decimal to be multiplied with
+   * @param precision Number of digits after decimal point.*/
+  void MultiplyAndSet(const Decimal<T> &value, unsigned precision);
+
+  /** Signed version of MultiplyAndSet
+   * @param input the decimal to be multiplied with
+   * @param lower_precision Number of digits after decimal point
+   * of the decimal with lower precision. Note the result will be stored in
+   * the higher precision value.*/
+  void SignedMultiplyWithDecimal(Decimal<T> input, unsigned lower_precision);
```
pmenon, line 571: Any reason `input` isn't a `const &`?

src/include/execution/sql/decimal_magic_numbers.h
```diff
@@ -0,0 +1,269 @@
...
+                                                           {0x84ec3c97da624ab4, 0xbd5af13bef0b113f, 240, 0},
+                                                           {0xd4ad2dbfc3d07787, 0x955e4ec64b44e865, 244, 0},
+                                                           {0x5512124cb4b9c969, 0x6ef285e8eae85cf5, 246, 0},
+                                                           {0x881cea14545c7575, 0x7e50d64177da2e55, 250, 0},
+                                                           {0x6ce3ee76a9e3912a, 0xcb73de9ac6482511, 253, 0}};
+
+/*Power map of powers of 10 for multiplying
+ * This map is used when we need to multiply with denominator
+ * precision in the divide routine*/
+uint128_t PowerOfTen[39][2] = {{0, 0},
```
pmenon, line 148: 1. Fix comment format.
2. This isn't really a map, it's a list of powers-of-ten separated into upper and lower 128-bit words indexed by exponent. I'd add this clarity into the comment.
3. `constexpr`.

src/include/execution/sql/decimal_magic_numbers.h
```diff
@@ -0,0 +1,269 @@
...
+  /// value - p in magic division
+  unsigned p;
+  /// Algo type
+  unsigned algo;
+};
+
+/* Magic Array for 128 bit division with powers of 10
+ * This map is used after multiplication of decimals to get
+ * the correct result*/
+struct Magic_Number_128 MAGIC_MAP_128_BIT_POWER_TEN[39] = {{0, 0, 0, 0},
```
pmenon, line 105: 1. Remove `struct`.
2. Comment format.
3. `constexpr` if you make the `Magic_Number_128` constructor `constexpr`, IIRC.

pmenon (changes requested):
First pass.

rohanaggarwal7997:
1370

src/include/common/error/error_code.h
rohanaggarwal7997: Resolved in latest PR

src/include/execution/sql/decimal_magic_numbers.h
rohanaggarwal7997: resolved in latest PR

src/include/execution/sql/decimal_magic_numbers.h
rohanaggarwal7997: Resolved everywhere

benchmark/storage/decimal_benchmark.cpp
rohanaggarwal7997: resolved

script/magic_numbers/Decimal_Division_Generator.java
rohanaggarwal7997: resolved

script/magic_numbers/decimal_multiplication_generator.py
rohanaggarwal7997: done

src/execution/compiler/compiler.cpp
rohanaggarwal7997: done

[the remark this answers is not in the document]
rohanaggarwal7997: done

src/include/execution/sql/runtime_types.h
```diff
@@ -488,9 +498,93 @@ class EXPORT Decimal {
...
+   * Get the underlying value of the decimal
+   * @return This decimal value.
+   */
+  T GetValue() const { return value_; }
+
+  /**
+   * Set the underlying value of the decimal
+   * @param value The value to be set.
+   */
+  void SetValue(T value) { this->value_ = value; }
```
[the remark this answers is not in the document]
rohanaggarwal7997: resolved

src/execution/sql/runtime_types.cpp
```diff
@@ -655,4 +657,666 @@ Timestamp Timestamp::FromYMDHMSMU(int32_t year, int32_t month, int32_t day, int3
...
 
+//===----------------------------------------------------------------------===//
+//
+// Decimal
+//
+//===----------------------------------------------------------------------===//
+
+template <typename T>
+void Decimal<T>::RoundUpAndSet(std::string input, unsigned int precision) {
+  this->value_ = 0;
```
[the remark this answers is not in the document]
rohanaggarwal7997: resolved

src/execution/sql/runtime_types.cpp
```diff
@@ -655,4 +657,666 @@ Timestamp Timestamp::FromYMDHMSMU(int32_t year, int32_t month, int32_t day, int3
...
+
+  if (input.size() == 0) return;
```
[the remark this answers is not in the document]
rohanaggarwal7997: resolved

src/execution/sql/runtime_types.cpp
```diff
@@ -655,4 +657,666 @@ Timestamp Timestamp::FromYMDHMSMU(int32_t year, int32_t month, int32_t day, int3
...
+
+  unsigned pos = 0;
```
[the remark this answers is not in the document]
rohanaggarwal7997: resolved

benchmark/storage/decimal_benchmark.cpp
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  void TearDown(const benchmark::State &state) final {}
+
+  // Read buffers pointers for concurrent reads
+  std::vector<execution::sql::Decimal128> decimals_;
+  std::vector<int128_t> floats_;
+};
+
+// Insert the num_inserts_ of tuples into a DataTable concurrently
```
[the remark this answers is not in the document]
rohanaggarwal7997: resolved

src/include/execution/sql/decimal_magic_numbers.h
```diff
@@ -0,0 +1,269 @@
...
+                               {0x4ee2d6d415b, 0x85acef8100000000},
+                               {0x314dc6448d93, 0x38c15b0a00000000},
+                               {0x1ed09bead87c0, 0x378d8e6400000000},
+                               {0x13426172c74d82, 0x2b878fe800000000},
+                               {0xc097ce7bc90715, 0xb34b9f1000000000},
+                               {0x785ee10d5da46d9, 0x00f436a000000000},
+                               {0x4b3b4ca85a86c47a, 0x098a224000000000}};
+
+/* Magic map for 128 bit division with constants*/
+std::unordered_map<uint128_t, struct Magic_Number_128> MagicMap128BitConstantDivision = {
```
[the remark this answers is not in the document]
rohanaggarwal7997: required when more constants are added

PR #1294, opened 2020-11-03
Fix use-after-free in DDL aborts

### LINKED ISSUE #1267: jumbotest_storage_other test failure

jrolli:
# Fix use-after-free in DDL aborts

Fixes #1267 (for real this time...)

## Description

The root cause of #1267 was a rare race between an aborting transaction and the GC that could cause deferred abort actions to occur before the transaction is unlinked (see thread schedule in the changes).  The proper way to handle an implicit dependency between a deferred action (which conceptually unlinking is) and another deferred action is to double-defer the dependent action.

### DIFF
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -915,10 +915,30 @@ bool DatabaseCatalog::SetTablePointer(const common::ManagedPointer<transaction::
       write_lock_.load() == txn->FinishTime(),
       "Setting the object's pointer should only be done after successful DDL change request. i.e. this txn "
       "should already have the lock.");
-  // We need to defer the deletion because their may be subsequent undo records into this table that need to be GCed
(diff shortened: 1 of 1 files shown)


mbutrovich (approved):
LGTM!

PR #1300, opened 2020-11-06
Stopping trackers on exceptional execution

17zhangw:
# Stop Resource Trackers on Exceptional Execution

## Description
This behavior was observed by @rabbit721. When query execution enters exceptional control flow (i.e inserting into an unique index fails which leads to AbortTxn which throws an exception), an assert gets tripped in OpCheckTrackersStopped in debug builds.

The PR fixes this by checking the transaction must abort flag and stops the resource trackers if they are still running. This is determined to be correct as trackers will all be stopped if query execution proceeds normally.

The PR also adds a configuration to run the junit traces with pipeline metrics enabled in 1 configuration.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -51,6 +51,7 @@ pipeline {
                         sh 'cd build && timeout 20m ninja check-format'
                         sh 'cd build && timeout 20m ninja check-lint'
                         sh 'cd build && timeout 20m ninja check-censored'
+                        sh 'cd build && ninja check-clang-tidy'
                     }
                     post {
                         cleanup {
@@ -74,6 +75,7 @@ pipeline {
                         sh 'cd build && timeout 20m ninja check-format'
                         sh 'cd build && timeout 20m ninja check-lint'
                         sh 'cd build && timeout 20m ninja check-censored'
+                        sh 'cd build && ninja check-clang-tidy'
                     }
                     post {
                         cleanup {
@@ -101,6 +103,7 @@ pipeline {
                         sh 'cd build && timeout 20m ninja check-format'
                         sh 'cd build && timeout 20m ninja check-lint'
                         sh 'cd build && timeout 20m ninja check-censored'
+                        sh 'cd build && ninja check-clang-tidy'
                     }
                     post {
                         cleanup {
@@ -132,7 +135,6 @@ pipeline {
                         cmake -GNinja -DNOISEPAGE_UNITY_BUILD=ON -DNOISEPAGE_TEST_PARALLELISM=1 -DCMAKE_BUILD_TYPE=Debug -DNOISEPAGE_USE_ASAN=ON -DNOISEPAGE_BUILD_BENCHMARKS=OFF -DNOISEPAGE_USE_JUMBOTESTS=OFF ..
                         ninja''', label: 'Compiling'
 
-                        sh 'cd build && ninja check-clang-tidy'
                         sh script: 'cd build && timeout 10s sudo python3 -B ../script/testing/kill_server.py 15721', label: 'Kill PID(15721)'
                         sh script: 'cd build && timeout 10s sudo python3 -B ../script/testing/kill_server.py 15722', label: 'Kill PID(15722)'
                         sh script: 'cd build && timeout 10s sudo python3 -B ../script/testing/kill_server.py 15723', label: 'Kill PID(15723)'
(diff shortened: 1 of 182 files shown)


mbutrovich:
Are we ever going to be interested in pipeline metrics for aborted transactions?

Jenkinsfile
```diff
@@ -177,6 +177,7 @@ pipeline {
                         sh 'cd build && timeout 1h ninja jumbotests'
                         sh 'cd build && timeout 1h ninja check-tpl'
                         sh script: 'cd build && timeout 20m python3 ../script/testing/junit/run_junit.py --build-type=debug --query-mode=simple', label: 'UnitTest (Simple)'
+                        sh script: 'cd build && timeout 20m python3 ../script/testing/junit/run_junit.py --build-type=debug --query-mode=simple --server-args="--pipeline_metrics_enable=True --pipeline_metrics_interval=0 --counters_enable=True"', label: 'UnitTest (Simple)'
```
lmwnshn, line 180: Change the label.

src/execution/exec/execution_context.cpp
```diff
@@ -41,6 +42,21 @@ void ExecutionContext::RegisterThreadWithMetricsManager() {
 }
 
 void ExecutionContext::CheckTrackersStopped() {
+  // Resource trackers are not automatically terminated at the end of query execution. If an
+  // exception is thrown during execution between StartPipelineTracker and EndPipelineTracker,
+  // then the trackers will keep on running (assuming the ThreadContext stays alive).
+  //
+  // If a transaction has aborted through @abortTxn, then it is very probable that EndPipelineTracker
```
lmwnshn, line 49: I think you may need to do `\@abortTxn` to prevent doxygen from complaining?

src/execution/exec/execution_context.cpp
```diff
@@ -41,6 +42,21 @@ void ExecutionContext::RegisterThreadWithMetricsManager() {
...
+  // was not called to stop the resource tracker. This check here terminates the resource trackers
+  // if they are still running (with the caveat that no metrics will be recorded).
+  if (GetTxn()->MustAbort() && noisepage::common::thread_context.resource_tracker_.IsRunning()) {
+    noisepage::common::thread_context.resource_tracker_.Stop();
+  }
+
+  // Codegen is responsible for guaranteeing that StartPipelineTrackers and EndPipelineTrackers
+  // are properly matched (if a thread calls StartPipelineTracker, it will call EndPipelineTracker
```
lmwnshn, line 57: nit: it will -> it must

lmwnshn (changes requested):
LGTM pending minor nits, though it leaves wider questions that @mbutrovich has (he'll be taking a pass too).

17zhangw:
> Are we ever going to be interested in pipeline metrics for aborted transactions?

I am inclined to believe that at some future date, we would want to know about the queries that abort from pipeline metrics. But that also does leave a larger discussion about recording (1) whether a pipeline succeeded/aborted partially and (2) some information about the abort. A more correct alternative might be to call `ExecutionContext::EndPipelineTracker` so at least the pipeline metrics would still get written.

Maybe @malin1993ml can also weigh in on this?

malin1993ml:
@17zhangw I actually need to think about this a little bit... let me get back to this later.

malin1993ml:
@17zhangw @mbutrovich Chatted with Andy about this briefly today. While I can't conclude definitely, I think it's probably not worth it to track the aborted OUs, build models for them, and use these models with the predictions of what and how many OUs may abort. I think it's going to be hard to make accurate predictions about it, and modeling those aborted OUs may not bring us significant benefit in actuality either. So I think we can assume we don't need to track metrics for the aborted OUs. We can still collect metrics for the earlier pipelines that do cause the abort through, which I believe we already support.

malin1993ml (approved):
LGTM once others' comments are addressed.

script/testing/junit/traces/select.test
```diff
@@ -1175,3 +1175,16 @@ SELECT tb.* FROM t, tt tb;
...
+DROP TABLE IF EXISTS t;
+
+statement ok
+CREATE TABLE t (a INT PRIMARY KEY);
+
+statement ok
+INSERT INTO t VALUES (1);
+
+statement error
+INSERT INTO t VALUES (1);
```
lmwnshn, line 1189: Drop your tables.

src/execution/exec/execution_context.cpp
```diff
@@ -41,6 +42,21 @@ void ExecutionContext::RegisterThreadWithMetricsManager() {
 }
 
 void ExecutionContext::CheckTrackersStopped() {
```
mbutrovich, line 44: `CheckTrackersStopped` sounds like it's just returning a true/false rather than mutating state. Can we call this something like `EnsureTrackersStopped`vof `GuaranteeTrackersStopped` since that's what it's actually doing?

mbutrovich (changes requested):
Minor nit.

src/execution/exec/execution_context.cpp
17zhangw: Changed all instances of `CheckTrackersStopped` -> `EnsureTrackersStopped`.

mbutrovich (approved):
LGTM, merge once through CI.

PR #1315, opened 2020-11-09
Improve format of query trace recording.

rabbit721:
This PR makes the following fixes to the original PR #1161 for query trace and query text recording, which ultimately address #1113:

1. Added query parameter in query trace recording

2. Fixed query text recording by wrapping text with quotation marks. 

3. Added database oid and query parameter types recording in query text recording

In addition, to achieve the above, utility code in type::TypeUtil and parser::ConstantValueExpression are been updated:

1. The TypeIdFromString function is added as the inverse function of TypeIdToString in TypeUtil

2. The FromString function is added as the inverse function of ToString in ConstantValueExpression

### DIFF
--- a/src/include/metrics/metrics_store.h
+++ b/src/include/metrics/metrics_store.h
@@ -181,8 +181,10 @@ class MetricsStore {
 
   /**
    * Record queries generated
+   * @param db_oid database oid
    * @param query_id id of the query
    * @param query_text text of the query
+   * @param param parameter associated with this query
    * @param timestamp time of query generation
    */
   void RecordQueryText(catalog::db_oid_t db_oid, const execution::query_id_t query_id, const std::string &query_text,
@@ -198,7 +200,6 @@ class MetricsStore {
    * @param query_id id of the query
    * @param timestamp time of the query execution
    * @param param parameter associated with this query
-   * @param oid database oid
    */
   void RecordQueryTrace(const execution::query_id_t query_id, const uint64_t timestamp,
                         common::ManagedPointer<const std::vector<parser::ConstantValueExpression>> param) {
--- a/src/include/metrics/metrics_store.h
+++ b/src/include/metrics/metrics_store.h
@@ -185,10 +185,12 @@ class MetricsStore {
    * @param query_text text of the query
    * @param timestamp time of query generation
    */
-  void RecordQueryText(const execution::query_id_t query_id, const std::string &query_text, const uint64_t timestamp) {
+  void RecordQueryText(catalog::db_oid_t db_oid, const execution::query_id_t query_id, const std::string &query_text,
+                       common::ManagedPointer<const std::vector<parser::ConstantValueExpression>> param,
+                       const uint64_t timestamp) {
     NOISEPAGE_ASSERT(ComponentEnabled(MetricsComponent::QUERY_TRACE), "QueryTraceMetric not enabled.");
     NOISEPAGE_ASSERT(query_trace_metric_ != nullptr, "QueryTraceMetric not allocated. Check MetricsStore constructor.");
-    query_trace_metric_->RecordQueryText(query_id, query_text, timestamp);
+    query_trace_metric_->RecordQueryText(db_oid, query_id, query_text, param, timestamp);
   }
 
   /**
@@ -199,11 +201,10 @@ class MetricsStore {
    * @param oid database oid
    */
   void RecordQueryTrace(const execution::query_id_t query_id, const uint64_t timestamp,
-                        common::ManagedPointer<const std::vector<parser::ConstantValueExpression>> param,
-                        catalog::db_oid_t oid) {
+                        common::ManagedPointer<const std::vector<parser::ConstantValueExpression>> param) {
     NOISEPAGE_ASSERT(ComponentEnabled(MetricsComponent::QUERY_TRACE), "QueryTraceMetric not enabled.");
     NOISEPAGE_ASSERT(query_trace_metric_ != nullptr, "QueryTraceMetric not allocated. Check MetricsStore constructor.");
-    query_trace_metric_->RecordQueryTrace(query_id, timestamp, param, oid);
+    query_trace_metric_->RecordQueryTrace(query_id, timestamp, param);
   }
 
   /**
--- a/src/include/metrics/metrics_store.h
+++ b/src/include/metrics/metrics_store.h
@@ -22,6 +22,7 @@
 #include "metrics/pipeline_metric.h"
 #include "metrics/query_trace_metric.h"
 #include "metrics/transaction_metric.h"
+#include "parser/expression/constant_value_expression.h"
 
 namespace noisepage::metrics {
 
@@ -194,11 +195,15 @@ class MetricsStore {
    * Record query execution history
    * @param query_id id of the query
    * @param timestamp time of the query execution
+   * @param param parameter associated with this query
+   * @param oid database oid
    */
-  void RecordQueryTrace(const execution::query_id_t query_id, const uint64_t timestamp) {
+  void RecordQueryTrace(const execution::query_id_t query_id, const uint64_t timestamp,
+                        common::ManagedPointer<const std::vector<parser::ConstantValueExpression>> param,
+                        catalog::db_oid_t oid) {
     NOISEPAGE_ASSERT(ComponentEnabled(MetricsComponent::QUERY_TRACE), "QueryTraceMetric not enabled.");
     NOISEPAGE_ASSERT(query_trace_metric_ != nullptr, "QueryTraceMetric not allocated. Check MetricsStore constructor.");
-    query_trace_metric_->RecordQueryTrace(query_id, timestamp);
+    query_trace_metric_->RecordQueryTrace(query_id, timestamp, param, oid);
   }
 
   /**
--- a/src/include/metrics/query_trace_metric.h
+++ b/src/include/metrics/query_trace_metric.h
@@ -4,13 +4,17 @@
 #include <chrono>  //NOLINT
 #include <fstream>
 #include <list>
+#include <sstream>
 #include <string>
 #include <utility>
 #include <vector>
 
 #include "catalog/catalog_defs.h"
+#include "common/managed_pointer.h"
+#include "execution/exec_defs.h"
 #include "metrics/abstract_metric.h"
 #include "metrics/metrics_util.h"
+#include "parser/expression/constant_value_expression.h"
 #include "transaction/transaction_defs.h"
 
 namespace noisepage::metrics {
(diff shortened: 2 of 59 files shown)


src/include/metrics/query_trace_metric.h
```diff
@@ -92,10 +98,18 @@ class QueryTraceMetricRawData : public AbstractRawData {
...
+    QueryTrace(const execution::query_id_t query_id, const uint64_t timestamp, std::string param_string,
+               std::string type_string, const catalog::db_oid_t oid)
+        : query_id_(query_id),
+          timestamp_(timestamp),
+          param_string_(std::move(param_string)),
+          type_string_(std::move(type_string)),
+          oid_(oid) {}
     const execution::query_id_t query_id_;
     const uint64_t timestamp_;
+    std::string param_string_;
```
mbutrovich, line 110: These should be `const` like the other fields.

src/include/metrics/query_trace_metric.h
```diff
@@ -111,10 +125,24 @@ class QueryTraceMetric : public AbstractMetric<QueryTraceMetricRawData> {
...
-  void RecordQueryTrace(const execution::query_id_t query_id, const uint64_t timestamp) {
-    GetRawData()->RecordQueryTrace(query_id, timestamp);
+  void RecordQueryTrace(const execution::query_id_t query_id, const uint64_t timestamp,
+                        common::ManagedPointer<const std::vector<parser::ConstantValueExpression>> param,
+                        catalog::db_oid_t oid) {
+    std::ostringstream param_stream;
+    std::ostringstream type_stream;
+    for (const auto &val : (*param)) {
+      if (val.IsNull()) {
+        return;
```
mbutrovich, line 137: We just bail if a parameter is NULL?

src/parser/expression/constant_value_expression.cpp
```diff
@@ -250,6 +251,59 @@ std::string ConstantValueExpression::ToString() const {
   }
 }
 
+ConstantValueExpression ConstantValueExpression::FromString(const std::string &val, uint64_t type_id) {
```
mbutrovich, line 254: Is this used anywhere? Also the parameter type should be `TypeId`. Force the caller to do a `static_cast` if necessary. Doesn't make sense to allow it to be a `uint64_t` and then cast it on the first line.

src/parser/expression/constant_value_expression.cpp
```diff
@@ -250,6 +251,59 @@ std::string ConstantValueExpression::ToString() const {
...
+    case type::TypeId::VARBINARY: {
+      auto string_val = execution::sql::ValueUtil::CreateStringVal(val);
+      return ConstantValueExpression(cve_type, string_val.first, std::move(string_val.second));
+    }
+    default:
+      UNREACHABLE("Invalid TypeId.");
+  }
+}
+
+std::string ConstantValueExpression::GetTypeAsString() const {
```
mbutrovich, line 287: Why is this logic specific to CVE? This doesn't seem like it belongs here. I would think `TypeUtil` with the other method to convert `TypeId` to string.

src/parser/expression/constant_value_expression.cpp
```diff
@@ -250,6 +251,59 @@ std::string ConstantValueExpression::ToString() const {
   }
 }
 
+ConstantValueExpression ConstantValueExpression::FromString(const std::string &val, uint64_t type_id) {
+  const auto cve_type = static_cast<type::TypeId>(type_id);
+
+  switch (cve_type) {
```
mbutrovich, line 257: What about NULL? If you're adding general purpose code to CVE it needs to handle all appropriate use cases, or uncovered uses need to be documented.

src/include/metrics/query_trace_metric.h
rabbit721: The null valued parameters are observed to only come from the database loading phase. For the pilot's purpose, we only need the queries from the execution phase. It would be better if we can directly detect if the query is from the loading phase and ignore it. But I haven't found a way to do so.

src/parser/expression/constant_value_expression.cpp
rabbit721: Moved to TypeUtil in lastest push.

src/include/metrics/query_trace_metric.h
rabbit721: Resolved in the latest push.

src/parser/expression/constant_value_expression.cpp
rabbit721: Resolved in the latest push.

src/include/metrics/query_trace_metric.h
mbutrovich: That sounds like a design decision based on oltpbench. We want to solve the general case. Couldn't query trace metrics be turned on for any arbitrary workload?

src/include/type/type_util.h
```diff
@@ -92,6 +92,46 @@ class TypeUtil {
       }
     }
   }
+
+  static type::TypeId TypeIdFromString(const std::string &type_string) {
+    if (type_string == "INVALID") {
```
mbutrovich, line 97: This should be a switch.

rabbit721: I don't think c++ supports switch on strings?

mbutrovich: Right, forget you were looking at strings...

src/include/metrics/query_trace_metric.h
malin1993ml: Maybe we can just record `NULL` here, since that's how postgres specifies NULL value in a SQL query.

src/parser/expression/constant_value_expression.cpp
malin1993ml: Then here we can check whether the string is `NULL` and create a NULL CVE using `ConstantValueExpression(const type::TypeId type)`. Double-check with postgres to see whether they accept `null`, `nUlL`, etc.

rabbit721: I have added the null checking in the latest push. Let me know if there's any other issues.

src/include/metrics/query_trace_metric.h
rabbit721: Fixed in the latest push.

src/parser/expression/constant_value_expression.cpp
```diff
@@ -250,6 +251,41 @@ std::string ConstantValueExpression::ToString() const {
...
+      if (val.empty()) return ConstantValueExpression(type_id, execution::sql::TimestampVal::Null());
+      return ConstantValueExpression(type_id, execution::sql::TimestampVal(execution::sql::Timestamp::FromString(val)));
+    }
+    case type::TypeId::DATE: {
+      if (val.empty()) return ConstantValueExpression(type_id, execution::sql::DateVal::Null());
+      return ConstantValueExpression(type_id, execution::sql::DateVal(execution::sql::Date::FromString(val)));
+    }
+    case type::TypeId::VARCHAR:
+    case type::TypeId::VARBINARY: {
+      auto string_val = execution::sql::ValueUtil::CreateStringVal(val);
```
mbutrovich, line 281: We seems to be missing NULL for variable length types.

rabbit721: Updated using the general constructor.

src/include/metrics/metrics_store.h
```diff
@@ -194,11 +195,15 @@ class MetricsStore {
...
+   * @param oid database oid
    */
-  void RecordQueryTrace(const execution::query_id_t query_id, const uint64_t timestamp) {
+  void RecordQueryTrace(const execution::query_id_t query_id, const uint64_t timestamp,
+                        common::ManagedPointer<const std::vector<parser::ConstantValueExpression>> param,
+                        catalog::db_oid_t oid) {
     NOISEPAGE_ASSERT(ComponentEnabled(MetricsComponent::QUERY_TRACE), "QueryTraceMetric not enabled.");
     NOISEPAGE_ASSERT(query_trace_metric_ != nullptr, "QueryTraceMetric not allocated. Check MetricsStore constructor.");
-    query_trace_metric_->RecordQueryTrace(query_id, timestamp);
+    query_trace_metric_->RecordQueryTrace(query_id, timestamp, param, oid);
```
malin1993ml, line 206: nit: change `oid` to `db_oid` to be more explicit. Also, could you put `db_oid` at the front? It's easier to follow with higher level concept in front followed by decreasing level of concepts.

src/include/metrics/query_trace_metric.h
```diff
@@ -79,23 +84,32 @@ class QueryTraceMetricRawData : public AbstractRawData {
...
   }
 
-  void RecordQueryTrace(const execution::query_id_t query_id, const uint64_t timestamp) {
-    query_trace_.emplace_back(query_id, timestamp);
+  void RecordQueryTrace(const execution::query_id_t query_id, const uint64_t timestamp, const std::string &param_string,
+                        const std::string &type_string, const catalog::db_oid_t oid) {
+    query_trace_.emplace_back(query_id, timestamp, param_string, type_string, oid);
   }
 
   struct QueryText {
```
malin1993ml, line 92: We need to add `db_oid` to query text as well.

src/include/metrics/query_trace_metric.h
```diff
@@ -111,10 +125,24 @@ class QueryTraceMetric : public AbstractMetric<QueryTraceMetricRawData> {
...
+      if (val.IsNull()) {
+        param_stream << "";
+      } else {
+        param_stream << val.ToString();
+      }
+      type_stream << type::TypeUtil::TypeIdToString(val.GetReturnValueType());
+      param_stream << ";";
+      type_stream << ";";
+    }
+    GetRawData()->RecordQueryTrace(query_id, timestamp, param_stream.str(), type_stream.str(), oid);
```
malin1993ml, line 145: I think we should move the type recording from the `QueryTrace` metric to the `QueryText` metric. The types should be fixed with the same query template and we only need to record them once and later load them once. Let me know whether this is doable.

malin1993ml (changes requested):
Generally looks good! A few small changes.

@mbutrovich Do you have any more comments? In fact, do you have any suggestions on how should we test the query trace recording with parameters? AFAIK `pqxx` uses the simple query protocol and we cannot trigger the prepared statements?

src/include/metrics/metrics_store.h
rabbit721: Yeah sure, resolved in latest push.

malin1993ml:
Okay according to the Slack discussion the way to test this right now is just to enable the query trace recording under the extended query mode in our junit tests. @17zhangw will probably add that test stage in #1300.

malin1993ml (approved):
Should be good once the latest tests have passed.

PR #1317, opened 2020-11-10
Embed Cardinalities into Plan Nodes

xinzhu-cai:
## Description
 Embed the cardinality information in the generated physical plan nodes during plan generation.

### Remaining Tasks

:pushpin: TODOs:
- [x] Add cardinality fields to plan nodes
- [ ] Add tests


lmwnshn:
I think I am missing the context here. Do we still plan on reusing plan nodes? If yes, I don't think we should embed cardinalities into plan nodes. This seems like state that should live elsewhere. @apavlo

lmwnshn:
I believe we said we would move to the auxiliary data structure approach.

PR #1321, opened 2020-11-10
Add support for aliases in ORDER BY.

### LINKED ISSUE #1220: TPC-H : Alias column is not referenced in an order by clause
# Feature Request
## Summary
For a column defined with an alias, we cannot use it in the same query in the order by.
Further investigation is required to understand the problem and possible solutions.

## Environment

## Steps to Reproduce
2. Run terrier executable
5. Run the query 

```SQL
select
        sum(c_acctbal * (1 - c_acctbal)) as revenue
from
        customer
group by
        c_name
order by
        revenue desc
limit 20;

Error:
 ERROR:  column "revenue" does not exist
```

gautam20197:
Fix #1220 

Changes in the PR:

1. Binder: Added binding of order by columns using alias by picking from the select list. If the alias does not exist in the select list, the binder will throw an error of the column not existing.

2. Optimizer: Create a derived value expression from children expression of the aggregate expression that will be sorted by the order by.


src/binder/bind_node_visitor.cpp
```diff
@@ -516,6 +517,7 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::SelectStatement> node
 
   if (node->GetSelectOrderBy() != nullptr) {
     UnifyOrderByExpression(node->GetSelectOrderBy(), node->GetSelectColumns());
+//    bool alias_unified = UnifyAliasNameOrderByExpression(node->GetSelectOrderBy(), node->GetSelectColumns());
```
lmwnshn, line 520: Dead code.

lmwnshn (changes requested):
Could you remove the dead code? Will help you generate the trace files later.

src/binder/bind_node_visitor.cpp
gautam20197: Yup I'll do that.

lmwnshn:
Additionally, you may need to run format/tidy/etc first.

lmwnshn (approved):
LGTM if passes CI.

PR #1323, opened 2020-11-11
Add documentation for preventing Run / Debug from CLion from crashing on ASAN Container Overflow

turingcompl33t:
# Prevent Run / Debug from CLion from Crashing on Asan Container Overflow

## Description

I was experiencing issues wherein attempting to disable Asan's container overflow checks with an environment variable would fail when running or debugging the application from within CLion - the IDE appears not to respect the environment variable for some reason (no matter how hard I tried to make it...). Fortunately, CLion provides another way to set options for various sanitizers that I just came across. This PR just adds this finding to the troubleshooting documentation.

## Remaining tasks

None.

## Performance

N/A

## Further work

None


lmwnshn (approved):
LGTM.

PR #1326, opened 2020-11-13
Add support for SQL IN operator in WHERE clauses.

### LINKED ISSUE #1221: TPC-H : Constants in the "in" clause give a unsupported type ExprTransform error
# Feature Request
## Summary
We get a expression transform error when we have constants for the in clause.

## Environment

## Steps to Reproduce
2. Run terrier executable
5. Run the query 

```SQL
select
        p_brand,
        p_type,
        p_size
from
        partsupp,
        part
where
        p_size in (49, 14, 23, 45, 19, 3, 36, 9);

Error:
 ERROR:  ExprTransform: unsupported type
```

lmwnshn:
# Add support for SQL IN operator in WHERE clauses.

## Description

Fix #1221 by adding support for the SQL IN operator in WHERE clauses.

- `... WHERE foo IN (1)` will be treated the same way as `foo = 1`, including vectorized stuff.
- `... WHERE foo IN (1,2,...,N)` will be treated as `(foo = 1) OR (foo = 2) OR ... OR (foo = N)`.

Also fixes some tables not getting dropped in `select.test`. And adds a couple of tests.

Just picking random things off the TPC-H list. We're actually very close.

Will flip to ready-for-review if passes CI.

### DIFF
--- a/src/execution/compiler/expression/comparison_translator.cpp
+++ b/src/execution/compiler/expression/comparison_translator.cpp
@@ -12,9 +12,14 @@ namespace noisepage::execution::compiler {
 ComparisonTranslator::ComparisonTranslator(const parser::ComparisonExpression &expr,
                                            CompilationContext *compilation_context)
     : ExpressionTranslator(expr, compilation_context) {
-  // Prepare the left and right expression subtrees for translation.
-  compilation_context->Prepare(*expr.GetChild(0));
-  compilation_context->Prepare(*expr.GetChild(1));
+  // Prepare all expression subtrees for translation.
+  NOISEPAGE_ASSERT(expr.GetChildrenSize() == 2 || expr.GetExpressionType() == parser::ExpressionType::COMPARE_IN,
+                   "Every ComparisonExpression should have exactly two children, except for COMPARE_IN.");
+  NOISEPAGE_ASSERT(expr.GetChildrenSize() >= 2 || expr.GetExpressionType() == parser::ExpressionType::COMPARE_IN,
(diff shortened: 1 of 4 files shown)


src/execution/compiler/expression/comparison_translator.cpp
```diff
@@ -12,9 +12,14 @@ namespace noisepage::execution::compiler {
 ComparisonTranslator::ComparisonTranslator(const parser::ComparisonExpression &expr,
                                            CompilationContext *compilation_context)
     : ExpressionTranslator(expr, compilation_context) {
-  // Prepare the left and right expression subtrees for translation.
-  compilation_context->Prepare(*expr.GetChild(0));
-  compilation_context->Prepare(*expr.GetChild(1));
+  // Prepare all expression subtrees for translation.
+  NOISEPAGE_ASSERT(expr.GetChildrenSize() == 2 || expr.GetExpressionType() == parser::ExpressionType::COMPARE_IN,
+                   "Every ComparisonExpression should have exactly two children, except for COMPARE_IN.");
+  NOISEPAGE_ASSERT(expr.GetChildrenSize() >= 2 || expr.GetExpressionType() == parser::ExpressionType::COMPARE_IN,
```
lmwnshn, line 18: This should be (!COMPARE_IN) || (COMPARE_IN && 2 children). I'll fix that.

pmenon (approved):
LGTM.

Side note: `IN` comparisons of *any* size can be trivially decomposed to utilize vectorized selection builtins on each value. We can go HAM and utilize `FilterManager` to get dynamic reordering on them, too. Easy to implement, but not necessary for this PR.

PR #1327, opened 2020-11-13
Add support for type cast expressions.

### LINKED ISSUE #1219: TPC-H : Casting a string to date
# Feature Request
## Summary
For any query working on columns with data type date, there is no way to enter dates into the table or have predicates consisting dates. 
The main problem is the cast operator transform is missing in the system.

## Environment

## Steps to Reproduce
2. Run terrier executable
5. Run the query 

```SQL
select
        o_orderpriority
from
        orders
where
        o_orderdate >= date '1993-07-01'
        and o_orderdate < date '1993-10-01'


or 

select
        o_orderpriority
from
        orders
where
        o_orderdate >= '1993-07-01'::date
        and o_orderdate < '1993-10-01'::date

or 

select
        o_orderpriority
from
        orders
where
        o_orderdate >= TO_DATE('1993-07-01','YYYYMMDD');
        and o_orderdate < TO_DATE('1993-10-01','YYYYMMDD');

Error:
 what():  Code generation for expression type 'OPERATOR_CAST' not supported.
```

lmwnshn:
# Add support for type cast expressions.

## Description

Fix #1219.

This enables `::` casting and also allows for `select true` which I believe @tanujnay112 may have wanted?

I think `::`-casting worked in the past because we pulled out the TypeCastExpression's contents, which is behavior that was probably rolled back with the binder cleanup (sherpa). This PR brings that back with the justification of Postgres not allowing you to do anything interesting in casts, as far as I can tell:

```
postgres=# select '1+1'::bigint;
ERROR:  invalid input syntax for type bigint: "1+1"
LINE 1: select '1+1'::bigint;
               ^
postgres=# select cast('1+1' as bigint);
ERROR:  invalid input syntax for type bigint: "1+1"
LINE 1: select cast('1+1' as bigint);
                    ^
postgres=# \errverbose
ERROR:  22P02: invalid input syntax for type bigint: "1+1"
LINE 1: select cast('1+1' as bigint);
                    ^
LOCATION:  scanint8, int8.c:126
```

With this PR:

```
noisepage=# select true;
 ?column? 
----------
 t
(1 row)

noisepage=# SELECT '2020-11-07'::date;
  ?column?  
------------
 2020-11-07
(1 row)

noisepage=# CREATE TABLE foo (a date);
CREATE TABLE
noisepage=# INSERT INTO foo VALUES ('2020-11-07'::date);
INSERT 0 1
noisepage=# INSERT INTO foo VALUES ('2020-11-08'::date);
INSERT 0 1
noisepage=# select * from foo where a > '2020-11-07'::date;
     a      
------------
 2020-11-08
(1 row)

noisepage=# SELECT * FROM foo WHERE a >= '2020-11-07'::date;
     a      
------------
 2020-11-07
 2020-11-08
(2 rows)
```

## Gripes

I'm still not a very big fan of the visitor pattern strewn all over the frontend, wherein we have the problem of

1. you visit, visit...
2. you visited a TypeCastExpression
3. you want to replace a TypeCastExpression with its child now
4. but you only have a pointer to the TypeCastExpression, not its parent -- how do you replace it? as far as I can tell, right now you can't.

This PR also hops on the general optimizer "when in doubt, make a copy" wagon in EvaluateExpression.

Marking as in-progress until it passes CI.

### DIFF
--- a/src/binder/bind_node_visitor.cpp
+++ b/src/binder/bind_node_visitor.cpp
@@ -645,6 +645,14 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::ComparisonExpression>
   sherpa_->CheckDesiredType(expr.CastManagedPointerTo<parser::AbstractExpression>());
   sherpa_->SetDesiredTypePair(expr->GetChild(0), expr->GetChild(1));
   SqlNodeVisitor::Visit(expr);
+
+  // If any of the operands are typecasts, the typecast children should have been casted by now. Pull the children up.
+  for (size_t i = 0; i < expr->GetChildrenSize(); ++i) {
+    auto child = expr->GetChild(i);
+    if (parser::ExpressionType::OPERATOR_CAST == child->GetExpressionType()) {
+      expr->SetChild(i, child->GetChild(0));
+    }
+  }
 }
 
 void BindNodeVisitor::Visit(common::ManagedPointer<parser::ConjunctionExpression> expr) {
@@ -656,6 +664,14 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::ConjunctionExpression
     sherpa_->SetDesiredType(child, type::TypeId::BOOLEAN);
   }
   SqlNodeVisitor::Visit(expr);
+
+  // If any of the operands are typecasts, the typecast children should have been casted by now. Pull the children up.
+  for (size_t i = 0; i < expr->GetChildrenSize(); ++i) {
+    auto child = expr->GetChild(i);
+    if (parser::ExpressionType::OPERATOR_CAST == child->GetExpressionType()) {
+      expr->SetChild(i, child->GetChild(0));
+    }
+  }
 }
 
 void BindNodeVisitor::Visit(common::ManagedPointer<parser::ConstantValueExpression> expr) {
@@ -742,6 +758,8 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::SubqueryExpression> e
 
 void BindNodeVisitor::Visit(UNUSED_ATTRIBUTE common::ManagedPointer<parser::TypeCastExpression> expr) {
   BINDER_LOG_TRACE("Visiting TypeCastExpression...");
+  NOISEPAGE_ASSERT(1 == expr->GetChildrenSize(), "TypeCastExpression should have exactly 1 child.");
+  sherpa_->SetDesiredType(expr->GetChild(0), expr->GetReturnValueType());
   SqlNodeVisitor::Visit(expr);
 }
 
--- a/src/binder/bind_node_visitor.cpp
+++ b/src/binder/bind_node_visitor.cpp
@@ -664,14 +664,6 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::ConjunctionExpression
     sherpa_->SetDesiredType(child, type::TypeId::BOOLEAN);
   }
   SqlNodeVisitor::Visit(expr);
-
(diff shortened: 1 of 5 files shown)


src/binder/binder_util.cpp
```diff
@@ -90,13 +90,20 @@ void BinderUtil::CheckAndTryPromoteType(const common::ManagedPointer<parser::Con
             break;
           }
           case type::TypeId::TINYINT: {
+            size_t size;
```
lmwnshn, line 93: We used to double-template (template calls a template) this code, but I think we went back to this style for readability.

We could conceivably macro the whole thing instead. It is a little repetitive. No strong opinions though.

src/binder/bind_node_visitor.cpp
```diff
@@ -645,6 +645,14 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::ComparisonExpression>
   sherpa_->CheckDesiredType(expr.CastManagedPointerTo<parser::AbstractExpression>());
   sherpa_->SetDesiredTypePair(expr->GetChild(0), expr->GetChild(1));
   SqlNodeVisitor::Visit(expr);
+
+  // If any of the operands are typecasts, the typecast children should have been casted by now. Pull the children up.
```
lmwnshn, line 649: I find it hacky to manually look for typecasts here, but we do a similar thing in select statements.

Might need to look for typecasts elsewhere, this will unblock TPC-H though I think.

In general, if your visiting needs to knock out the thing X, you need to make that change at all callers of visit(X) rather than at the implementation of visit(X) itself. Please correct me if I'm wrong.

src/binder/bind_node_visitor.cpp
```diff
@@ -645,6 +645,14 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::ComparisonExpression>
...
+  for (size_t i = 0; i < expr->GetChildrenSize(); ++i) {
+    auto child = expr->GetChild(i);
+    if (parser::ExpressionType::OPERATOR_CAST == child->GetExpressionType()) {
```
tanujnay112, line 652: We should leave an assertion to make sure we are pulling up a constant value expression because this would break later in the exec engine with column value expressions I think.

src/include/parser/expression/type_cast_expression.h
```diff
@@ -9,6 +9,9 @@
 namespace noisepage::parser {
 /**
  * TypeCastExpression represents cast expressions of the form CAST(expr) or expr::TYPE.
+ * TypeCastExpression should not exist beyond the optimizer, and the child should be used instead.
+ * The role of a TypeCastExpression is to annotate the type of expr from above in the BinderSherpa.
+ * For example, Postgres does not allow CAST('1+1' AS BIGINT) nor '1+1'::BIGINT, throwing error 22P02.
```
tanujnay112, line 14: This doesn't account for the following trace that works on Postgres

`create table sample (x INTEGER);`
`insert into sample values (1);`
`select x::DECIMAL from sample;`

tanujnay112 (changes requested):
Let's add some assertions to catch when people try to do TypeCasts on column value expressions.

src/include/parser/expression/type_cast_expression.h
lmwnshn: You're right. This actually fails earlier in the binder.

src/binder/bind_node_visitor.cpp
lmwnshn: thanks!

tanujnay112 (approved):
LGTM for now after CI, wishing the binder let us replace visited nodes though.

PR #1329, opened 2020-11-13
Hotfix: fix the nightly reporting decimal precision of memory summary issue

jazzplato:
Limit the `avg` of `rss` and `vms` to be 4 decimal places upon reporting to the django api.


apavlo (approved):
LGTM

PR #1330, opened 2020-11-14
Artifact Stats

bialesdaniel:
# Artifact Stats

## Description
This PR adds a framework for collecting "artifact stats". These are collected as part of the nightly build and currently include compilation time and memory on startup. The idea behind this was to make it easily extensible. See the README.md in the artifact_stats directory for more details on how it works and how to add new metrics.

Here is a sample result from the Artifact Stats stage: ` Artifact stats: {'compile_time_sec': 184.91079518012702, 'rss_on_start': 39628800, 'vms_on_start': 2355093504}`
Example of a passing nightly pipeline: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier-nightly/detail/the-branch-formerly-known-as-binary-metrics/12/pipeline/6

This merge also includes work to delete the WAL after each OLTPBench testcase. 

## Further work
Once this is merged we are going to have members of the CMU-DB team try adding a collector to understand if it is actually easy to work with. Our current plan is to have @apavlo add a collector for binary size and @pmenon add a collector for CPU utilization on startup. 

Also, there is some work to simplify the compile +build commands which I will make part of a separate PR


bialesdaniel:
There were a couple issues that broke the build but those are now fixed

apavlo (changes requested):
1. My major complaint is about embedding an extra build command for the DBMS inside of the Python code. People are not going to know to change this. Is this just to get the compilation time? If yes, why not add a build command in the Jenkins stage but have it pipe out the build time (using [`time`](https://ss64.com/bash/time.html)) to a file. Then have your script look for that file and parse it?
2. Maybe rename "artifact_stats_collectors" dir to be just "collectors"?

script/testing/artifact_stats/common_collector_functions.py
```diff
@@ -0,0 +1,28 @@
...
+        use_cache - Whether to increase the speed by using the build cache
+        is_debug - Whether to output the logs of the compilation to stdout
+    Retrun:
+        int - The exit code of the compilation (0=success)
+    """
+    cmake_cmd = f'cmake -GNinja -DNOISEPAGE_UNITY_BUILD=ON \
+        {"-DCMAKE_CXX_COMPILER_LAUNCHER=ccache" if use_cache else ""} -DCMAKE_BUILD_TYPE=Release \
+        -DNOISEPAGE_USE_ASAN=OFF -DNOISEPAGE_USE_JEMALLOC=ON ..'
+    ninja_cmd = 'ninja noisepage'
+    compile_commands = [cmake_cmd, ninja_cmd]
```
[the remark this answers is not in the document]
bialesdaniel: Yah I think there are a couple of options.
1) I could pass a string as a command-line arg to the script which contains the build commands
       - Its not the most elegant solution and we might have to pass multiple args to be able to turn on and off caching
2) I could create a build script
       - This way if we wanted to make changes to the build commands there is only one place to make those changes. 
          The normal build, nightly build, and these tests could use the build script.
       - You still could have the problem where if there are drastic changes to the arguments then these tests might not 
          be updated. 
3) We could set it up so the nightly build will send a slack message when it fails so people are aware if things like this get broken. 
       - I could do something where I read through the Jenkinsfile to see if the build command matches a build command 
          in the Jenkinsfile and throw an error if it doesn't. 
4) I could do all the metric collection directly in the Jenkinsfile-nightly.
       - This would make the file pretty unwieldy and much harder for people to test before committing changes

I can keep thinking about if there are better solutions.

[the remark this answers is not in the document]
bialesdaniel: I could also store the build command in an environment variable in the Jenkinsfile and use that in the script

bialesdaniel:
@apavlo This is ready for review again. I've added the logic to delete the WAL after each OLTPBench test case. I updated the way we are collecting the build + compile time. Here is an example of a passing nightly build http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier-nightly/detail/the-branch-formerly-known-as-binary-metrics/23/pipeline

I did not include creating the shared Jenkinfile build+compile function. I will make that part of a separate PR.

bialesdaniel:
I made changes to the Jenkinsfile so this won't pass the terrier build but here is proof that it can pass: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/PR-27/9/pipeline/160

bialesdaniel:
Still passing after last commit: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/the-branch-formerly-known-as-binary-metrics/4/pipeline/484

Jenkinsfile-nightly
```diff
@@ -16,6 +16,33 @@ pipeline {
...
+            agent {
+                docker {
+                    image 'noisepage:focal'
+                }
+            }
+            steps {
+                sh 'echo $NODE_NAME'
+                sh script:'echo y | sudo ./script/installation/packages.sh all', label: 'Installing packages'
+
+                sh script:'''
```
apavlo, line 29: Minor: Can you just add a comment in here that explains what this command does and why we need it?

build-support/check_github_labels.py
```diff
@@ -7,17 +7,19 @@
...
 
-
 def get_pr_num():
     match = re.match(r'.*terrier_PR-([^@/]*).*', os.getcwd())
     if match and len(match.groups()) == 1:
         return int(match.groups()[0])
     return None
 
+def is_correct_build_folder():
+    return getenv('JOB_NAME').startswith('terrier/')
```
apavlo, line 22: It should be `noisepage` not `terrier`? Or is this a Jenkins thing?

build-support/check_github_labels.py
```diff
@@ -37,6 +39,10 @@ def check_labels_impolite(pr_num):
 
 
 if __name__ == '__main__':
+    if not is_correct_build_folder():
+        print('This build originated from a jenkins subfolder. Skipping ci-label check')
```
apavlo, line 43: Shouldn't this be `LOG.info` ?

script/testing/oltpbench/test_case_oltp.py
```diff
@@ -127,6 +127,8 @@ def run_post_test(self):
                 os.path.join(os.getcwd(), "oltp_result",
                              self.filename_suffix), self.publish_username,
                 self.publish_password, self.mem_metrics, self.query_mode)
+        
+
```
apavlo, line 131: Unnecessary newlines. Are we using a linter for the Python code?

apavlo (changes requested):
Minor changes requested.

build-support/check_github_labels.py
bialesdaniel: Yah, this is a Jenkins thing. If we rename the Jenkins job this will need to change. I created this cause any build from the testing team folder wouldn't pass this first step. Cause it would try to lookup the PRs in our repo on Noisepage.

build-support/check_github_labels.py
bialesdaniel: This whole script uses `print()`. I can switch it to `LOG` if you want.

apavlo (approved):
LGTM

PR #1331, opened 2020-11-14
Replication Framework Setup

tpan496:
# Replication Infrastructure
## Description
- Adds a working end to end replication infrastructure using the new messenger system. The current replication framework supports the serialization/deserialization of log records.
- Working test for sync/async replication.

## Remaining tasks
- [x] Refactor tests
- [x] Cleanup

## Further work


src/include/messenger/messenger.h
```diff
@@ -189,6 +189,14 @@ class Messenger : public common::DedicatedThreadTask {
   /** @return The ConnectionRouter with the specified router_id. Created by ListenForConnection. */
   common::ManagedPointer<ConnectionRouter> GetConnectionRouter(const std::string &router_id);
 
+  void ProcessDefaultMessage(std::string_view &sender_id, std::string_view &msg, uint64_t recv_cb_id) {
```
tpan496, line 192: I am not really sure where I am supposed to invoke the callbacks that are stored in the messenger. Let me know if I should do this inside the replication manager or use something else.

src/replication/replication_manager.cpp
```diff
@@ -0,0 +1,240 @@
...
+
+void ReplicationManager::ReplicaAck(const std::string &replica_name) {
+  std::unique_lock<std::mutex> lock(mutex_);
+  REPLICATION_LOG_INFO("Sending ack to: " + replica_name);
+  messenger_->SendMessage(GetReplicaConnection(replica_name), ack_message_identifier_, nullptr,
+                          static_cast<uint64_t>(ReplicationManager::MessageType::ACK));
+  lock.unlock();
+}
+
+void ReplicationManager::EventLoop(common::ManagedPointer<noisepage::messenger::Messenger> messenger,
```
tpan496, line 121: I am not able to get the receiver ids in this case, only the sender ids, so I can't find a way to just use the defined enums for splitting cases. Is there another way for me to do that? Otherwise, I would just do something similar to this. (Using the enums instead of those temporary strings).

PR #1336, opened 2020-11-17
Make DataTable::accessor_ private

jrolli:
# Make DataTable::accessor_ private

The `accessor_` member of `DataTable` was made public during the course of rewriting the `SlotIterator` which exposes internal storage layer APIs to the rest of the system.

## Description

This makes `accessor_` private again and adds `AccessObserver` and `ArrowSerializer` as "friends" to `DataTable` in order to allow them the privileged access they currently use in their implementations.


apavlo (commented):
LGTM

PR #1337, opened 2020-11-17
Fix the casting when accessing atttypid in pg_attribute.

mbutrovich:
#fix 1322. We were accessing the memory as a `type::TypeId` (uint8_t) but it's actually a uint32_t in the schema for pg_attribute.

```
noisepage=# create table t1(c1 int, c2 varchar, c3 int, c4 int, c5 timestamp);
CREATE TABLE
Time: 6.804 ms
noisepage=# SELECT attname,atttypid,nextcoloid FROM pg_attribute JOIN pg_class ON attrelid = reloid WHERE relname = 't1';
 attname | atttypid | nextcoloid 
---------+----------+------------
 c1      |        4 |          6
 c2      |        9 |          6
 c3      |        4 |          6
 c4      |        4 |          6
 c5      |        7 |          6
(5 rows)

Time: 13.904 ms
noisepage=# 

```


lmwnshn (approved):
LGTM, nice.

PR #1339, opened 2020-11-18
Add Plan Node Metadata in Optimizer

xinzhu-cai:
Refactor execution plan generation APIs in the optimizer to return both the generated execution plan and its metadata to other components. The plan metadata contains metadata for all plan nodes. Cardinality is contained in the plan node metadata.

TODO:
- [x] Add plan node id
- [x] Define new data structures including `OptimizeResult` and `PlanMetaData`
- [x] Refactor optimizer APIs to return `OptimizeResult` for `BuildPlanTree` instead of `PlanNode`
- [x] Add unit test


malin1993ml:
I'm removing myself for now since the PR is still in-progress (to avoid the Slack spam). @xinzhu-cai Please tag me as reviewer again once it's ready-for-review!

src/optimizer/plan_generator.cpp
```diff
@@ -364,6 +375,7 @@ void PlanGenerator::Visit(const Limit *op) {
     }
 
     output_plan_ = order_build.Build();
+    plan_meta_data_->AddPlanNodeMetaData(output_plan_->GetPlanNodeId(), plan_node_meta_data_);
```
malin1993ml, line 378: Why we're explicitly adding the `PlanNodeMetaData` for `Limit` but not for others?

src/optimizer/optimizer.cpp
```diff
@@ -98,19 +101,18 @@ std::unique_ptr<planner::AbstractPlanNode> Optimizer::ChooseBestPlan(
...
 
   // Derive root plan
   auto *op = new OperatorNode(gexpr->Contents(), {}, txn);
 
-  PlanGenerator generator;
-  auto plan = generator.ConvertOpNode(txn, accessor, op, required_props, required_cols, output_cols,
-                                      std::move(children_plans), std::move(children_expr_map));
+  planner::PlanMetaData::PlanNodeMetaData plan_node_meta_data(context_->GetMemo().GetGroupByID(id)->GetNumRows());
+  auto plan = generator->ConvertOpNode(txn, accessor, op, required_props, required_cols, output_cols,
+                                       std::move(children_plans), std::move(children_expr_map), plan_node_meta_data);
```
malin1993ml, line 115: You can move the `plan_node_meta_data` into the function, similar to other ephemeral objects.

malin1993ml (changes requested):
Generally looks good! I have a few small questions/suggestions.

src/optimizer/plan_generator.cpp
xinzhu-cai: Adding plan node meta data of other nodes is handled in `ConvertOpNode`. We need to do this for `Limit` in `Visit` because the limit node is not the `output_plan_` but an additional node generated in `Visit`.

malin1993ml: Got it! I think it's helpful to add comments to document this to avoid confusion.

malin1993ml: I still think it's good to add some comments in the code.

src/optimizer/plan_generator.cpp
```diff
@@ -52,25 +52,28 @@
...
+    : plan_id_counter_(0), plan_meta_data_(plan_meta_data) {}
 
 std::unique_ptr<planner::AbstractPlanNode> PlanGenerator::ConvertOpNode(
     transaction::TransactionContext *txn, catalog::CatalogAccessor *accessor, AbstractOptimizerNode *op,
     PropertySet *required_props, const std::vector<common::ManagedPointer<parser::AbstractExpression>> &required_cols,
     const std::vector<common::ManagedPointer<parser::AbstractExpression>> &output_cols,
-    std::vector<std::unique_ptr<planner::AbstractPlanNode>> &&children_plans,
-    std::vector<ExprMap> &&children_expr_map) {
+    std::vector<std::unique_ptr<planner::AbstractPlanNode>> &&children_plans, std::vector<ExprMap> &&children_expr_map,
+    const planner::PlanMetaData::PlanNodeMetaData &plan_node_meta_data) {
```
malin1993ml, line 63: This should be `planner::PlanMetaData::PlanNodeMetaData &&plan_node_meta_data` if you use the move semantics.

src/optimizer/plan_generator.cpp
```diff
@@ -52,25 +52,28 @@
...
   required_props_ = required_props;
   required_cols_ = required_cols;
   output_cols_ = output_cols;
   children_plans_ = std::move(children_plans);
   children_expr_map_ = children_expr_map;
   accessor_ = accessor;
+  plan_node_meta_data_ = plan_node_meta_data;
```
malin1993ml, line 70: Here you can use `plan_node_meta_data_ = std::move(plan_node_meta_data);` to save the copy.

malin1993ml (changes requested):
Still some small tweaks.

src/include/planner/plannodes/plan_meta_data.h
```diff
@@ -0,0 +1,62 @@
...
+   private:
+    int cardinality_ = -1;
+  };
+
+  /**
+   * Add meta data for a plan node
+   * @param plan_node_id plan node id
+   * @param meta_data plan node meta data
+   */
+  void AddPlanNodeMetaData(const plan_node_id_t &plan_node_id, const PlanNodeMetaData &meta_data) {
```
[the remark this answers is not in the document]
xinzhu-cai: clang-tidy generates an error when I move the  `plan_node_meta_data` into the function call because this structure is small .

src/optimizer/plan_generator.cpp
xinzhu-cai: Clang-tidy does not allow this.

src/include/planner/plannodes/plan_meta_data.h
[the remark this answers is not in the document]
malin1993ml: Oh interesting! That's fine then I didn't know that restriction.

src/optimizer/optimizer.cpp
```diff
@@ -52,11 +53,12 @@ std::unique_ptr<planner::AbstractPlanNode> Optimizer::BuildPlanTree(transaction:
   }
 
   try {
-    auto best_plan = ChooseBestPlan(txn, accessor, root_id, phys_properties, output_exprs);
-
+    PlanGenerator generator(optimize_result->GetPlanMetaData());
```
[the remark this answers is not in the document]
xinzhu-cai: I think the optimizeResult shall be something constructed by the optimizer. I tried to do what you proposed. That involves too many changes than I expected.

src/optimizer/plan_generator.cpp
xinzhu-cai: I didn't see this comment. Yep, I'll add some.

src/optimizer/optimizer.cpp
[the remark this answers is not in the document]
malin1993ml: I see that's fine.

malin1993ml (approved):
LGTM. Thanks, @xinzhu-cai!

src/include/network/postgres/statement.h
```diff
@@ -141,7 +155,7 @@ class Statement {
   // The following objects can be "cached" in Statement objects for future statement invocations. Though they don't
   // relate to the Postgres Statement concept, these objects should be compatible with future queries that match the
   // same query text. The exception to this that DDL changes can break these cached objects.
-  std::unique_ptr<planner::AbstractPlanNode> physical_plan_ = nullptr;                // generated in the Bind phase
+  std::unique_ptr<optimizer::OptimizeResult> optimize_result_ = nullptr;              // generated in the Optimize phase
```
mbutrovich, line 158: This should still be Bind. There's no such thing as an Optimize phase. Postgres protocol supports Parse, Bind, Execute.

mbutrovich:
What is the scope (unique across what boundaries, if any?) of a `plan_node_id` and how is it intended to be used?

xinzhu-cai:
The `plan_node_id` is embedded in plan nodes.  So it is unique across the lifetime of a physical plan tree generated by the optimizer. We maintain a `plan_node_id` to `PlanNodeMetaData` mapping inside of `PlanMetaData` and let `Optimizer::BuildPlanTree` return both the `PlanMetaData` and the root plan node. Therefore, the rest of the system has access to the meta data (e.g. cardinality) of a specific plan node with the help of its `plan_node_id`.

mbutrovich:
> So it is unique across the lifetime of a physical plan tree generated by the optimizer.
Right that's my question: unique among the other plan nodes in its plan tree? Are they globally unique for the lifetime of the system (I suspect not since I think the Optimizer is instantiated for each invocation). I see something called `plan_node_id` getting stashed in plan nodes and that seems ripe for misuse if its uniqueness and intended uses cases is not documented at its declaration.

xinzhu-cai:
Yes, they are unique in the plan tree and are not globally unique during the lifetime of the system. The plan generator is reinitialized every time we generate a new plan tree. Added some documentation for plan_node_id. Is that enough?

mbutrovich:
That's perfect, thanks! Will merge once it gets through CI.

PR #1341, opened 2020-11-19
Refactor self-driving namespace and directories

malin1993ml:
1. Change the `brain` namespace to `selfdriving`.
2. Change the `brain` folder to `self_driving`.
3. Create `self_driving/modeling` folder to move the current modeling related files into.
4. Change `brain_defs.h` to `operating_unit_defs.h`.
5. Merge `BrainUtil` into `OperatingUnitUtil`.


malin1993ml:
I'm currently having the underscore `_` in the folder name `self_driving` and omitting the underscore in the namespace `selfdriving`, similar to how we handle the `traffic_cop` folder and the `trafficcop` namespace. Any comments on whether that's a convention that we should _follow / change_ are welcome.

17zhangw (approved):
Refactor looks good to me!

PR #1345, opened 2020-11-22
Refactor pg_language out of DatabaseCatalog.

lmwnshn:
Refactor `pg_language` logic away from `DatabaseCatalog` into its own `PgLanguageImpl` object.

Same deal as #1344 which should be merged first.

Just opening it to keep track of merge order.

PR #1077, opened 2020-08-13
Add Constraint Support

Poojita-Raj:
Adding constraint support for the following - 
1. Primary Key 
2. Unique 
3. Foreign Key (Delete Cascade)
4. Update 
[design document](https://github.com/YingjingLu/721-P3-Constraints/blob/fk_cascade/P3DesignDoc.md)

PR #1347, opened 2020-11-23
Refactor pg_constraint out of DatabaseCatalog.

lmwnshn:
Refactor `pg_constraint` logic away from `DatabaseCatalog` into its own `PgConstraintImpl` object.

Same deal as #1345 which should be merged first.

Just opening it to keep track of merge order.

Note that this does NOT contain the changes from #1077. That will come in a future PR, where we decide if modularizing things is good enough to add that code in, or if we need to go for the DSL.


src/include/catalog/postgres/pg_constraint_impl.h
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * Bootstrap:
+   *    pg_constraint
+   *    pg_constraint_oid_index
+   *    pg_constraint_name_index
+   *    pg_constraint_namespace_index
+   *    pg_constraint_table_index
+   *    pg_constraint_index_index
+   *    pg_constraint_foreigntable_index
```
jrolli, line 51: Add dependency info.

src/include/catalog/postgres/pg_constraint_impl.h
```diff
@@ -0,0 +1,86 @@
...
+  /**
+   * Prepare to create pg_constraint.
+   * Does NOT create anything until the relevant bootstrap functions are called.
+   *
+   * @param db_oid          The OID of the database that pg_constraint should be created in.
+   */
+  explicit PgConstraintImpl(db_oid_t db_oid);
+
+  /** Bootstrap the projected row initializers for pg_constraint. */
+  void BootstrapPRIs();
```
jrolli, line 41: Make private and call from `Bootstrap`

src/include/catalog/postgres/pg_constraint_impl.h
```diff
@@ -0,0 +1,86 @@
...
+   * Get all of the expressions used for constraints, used in TearDown.
+   * A buffer is passed in to allow for buffer reuse optimizations.
+   *
+   * @param txn             The transaction to obtain all the expressions in.
+   * @param buffer          The buffer to use for getting function contexts back out.
+   * @param buffer_len      The length of buffer. Only used in debug mode to assert correctness.
+   * @return All the expressions used for constraints.
+   */
+  std::vector<parser::AbstractExpression *> TearDownGetExpressions(
+      common::ManagedPointer<transaction::TransactionContext> txn, byte *buffer, UNUSED_ATTRIBUTE uint64_t buffer_len);
```
jrolli, line 69: See question in #1344 of whether this should return an opaque function that encapsulates the deletions instead.

src/include/catalog/postgres/pg_constraint_impl.h
```diff
@@ -0,0 +1,86 @@
...
+  const db_oid_t db_oid_;
+
+  storage::SqlTable *constraints_;
+  storage::index::Index *constraints_oid_index_;
+  storage::index::Index *constraints_name_index_;  // indexed on namespace OID and name
+  storage::index::Index *constraints_namespace_index_;
+  storage::index::Index *constraints_table_index_;
+  storage::index::Index *constraints_index_index_;
+  storage::index::Index *constraints_foreigntable_index_;
+};
```
jrolli, line 84: Where are the pre-generated projected row initializers?

jrolli (changes requested):
Pretty much same comments as #1344 and #1345.

src/include/catalog/postgres/pg_constraint_impl.h
lmwnshn: wip

src/include/catalog/postgres/pg_constraint_impl.h
lmwnshn: Per previous conversation, needs builder refactor

src/include/catalog/postgres/pg_constraint_impl.h
lmwnshn: Refactored, good idea

lmwnshn:
Will merge with #1358.

PR #1348, opened 2020-11-23
Refactor pg_type out of DatabaseCatalog.

lmwnshn:
Refactor `pg_type` logic away from `DatabaseCatalog` into its own `PgTypeImpl` object.

Same deal as #1347 which should be merged first.

Just opening it to keep track of merge order.

PR #1349, opened 2020-11-23
Refactor pg_attribute out of DatabaseCatalog.

lmwnshn:
Refactor `pg_attribute` logic away from `DatabaseCatalog` into its own `PgAttributeImpl` object.

Same deal as #1348 which should be merged first.

Just opening it to keep track of merge order.


src/include/catalog/postgres/pg_attribute_impl.h
```diff
@@ -0,0 +1,120 @@
...
+  /**
+   * Prepare to create pg_attribute.
+   * Does NOT create anything until the relevant bootstrap functions are called.
+   *
+   * @param db_oid          The OID of the database that pg_attribute should be created in.
+   */
+  explicit PgAttributeImpl(db_oid_t db_oid);
+
+  /** Bootstrap the projected row initializers for pg_attribute. */
+  void BootstrapPRIs();
```
jrolli, line 45: Hide and call from `Boostrap`

src/include/catalog/postgres/pg_attribute_impl.h
```diff
@@ -0,0 +1,120 @@
...
+   * Bootstrap:
+   *    pg_attribute
+   *    pg_attribute_oid_index
+   *    pg_attribute_name_index
+   *
+   * @param dbc             The catalog object to bootstrap in.
+   * @param txn             The transaction to bootstrap in.
+   */
+  void Bootstrap(common::ManagedPointer<DatabaseCatalog> dbc,
+                 common::ManagedPointer<transaction::TransactionContext> txn);
```
jrolli, line 57: I'm not sure this one separates out very cleanly and may need to be apart of the "core" catalog implementation that still gets handled as a unit (along with `pg_class`, `pg_index`, and `pg_namespace`) since you cannot fully define any table or index (to include these tables and their associated indexes) without all four of these tables.

jrolli (commented):
@mbutrovich, I've had some side discussions with @lmwnshn in Slack about these PRs to cleanup the catalog organization, and I think that `pg_attribute`, `pg_class`, `pg_index`, and `pg_namespace` should still be consolidated (either in `database_catalog` or a new `core_impl`) because it is hard to cleanly separate their bootstrapping logic.  Do you have any thoughts or concerns on this approach.

lmwnshn:
Closed in favor of pg_core PR (coming soon).

PR #1351, opened 2020-11-23
Fix SET parallel_execution which was not hooked up to execution.

lmwnshn:
# Fix SET parallel_execution which was not hooked up to execution.

## Description

We had a `parallel_execution` setting described as "Whether parallel execution for scans is enabled" which does not work as described. This is because it wasn't connected to the `SettingsManager`.

I notice some CI pipelines are setting this flag -- specifically `tpcc_parallel_disabled` -- parallel scans should really be disabled for real now.

I don't know how to write tests for this. If we had EXPLAIN tpl, this would work:

```
CREATE TABLE foo (a int);
/** By default, parallel execution is on. */
EXPLAIN tpl SELECT * FROM foo;  /** Test that ParallelWork in output. */
SET parallel_execution TO false;
EXPLAIN tpl SELECT * FROM foo;  /** Test that SerialWork in output. */
SET parallel_execution TO true;
EXPLAIN tpl SELECT * FROM foo;  /** Test that ParallelWork in output. */
```

Right now, tested manually.

I don't think there's a point in running this through CI but we're not starved on resources right now, so we might as well.

@17zhangw 指詰め

Also thanks to @rohanaggarwal7997 for asking the question which opened up this rabbit hole! good catch

### DIFF
--- a/src/execution/exec/execution_settings.cpp
+++ b/src/execution/exec/execution_settings.cpp
@@ -5,6 +5,7 @@ namespace noisepage::execution::exec {
 
 void ExecutionSettings::UpdateFromSettingsManager(common::ManagedPointer<settings::SettingsManager> settings) {
   if (settings) {
+    is_parallel_execution_enabled_ = settings->GetBool(settings::Param::parallel_execution);
     number_of_parallel_execution_threads_ = settings->GetInt(settings::Param::num_parallel_execution_threads);
     is_counters_enabled_ = settings->GetBool(settings::Param::counters_enable);
     is_pipeline_metrics_enabled_ = settings->GetBool(settings::Param::pipeline_metrics_enable);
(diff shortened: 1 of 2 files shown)


17zhangw (approved):
LGTM

mbutrovich:
In reviewing this, it looks like #1208 makes it so we read the execution settings from the `SettingsManager` on _every_ query execution. 

See: https://github.com/cmu-db/noisepage/blob/5efcc29e5d07b0fc5692ae25a4bd44264a5d474d/src/traffic_cop/traffic_cop.cpp#L356

That wasn't really the intended design, Looking up values from the `SettingsManager` was never meant to be on the critical path of query execution. Settings are meant to be cached in classes that use them, and then use the `SettingsManager`'s support for callbacks to orchestrate an update on change.

PR #1358, opened 2020-11-29
Refactor pg core tables out of DatabaseCatalog + general catalog cleanup.

lmwnshn:
# Refactor pg core tables out of DatabaseCatalog + general cleanup.

## Description

PR:
- Separates `pg_namespace`, `pg_class`, `pg_index`, and `pg_attribute` out into a `PgCoreImpl` container.
- Reorders functions in `DatabaseCatalog{.h,.cpp}` to be roughly in order of `Bootstrap`.
- Strong opinions on documentation? Switches to aligned documentation (WIP - 2020-11-29) since I find it easier to skim. Also start using more doxygen features in general, for example `///<` documents the member to the left, and `/** You can comment multiple things at once with */ ///@{ ///@}`.
- Introduce a `CatalogColumnDef` meant to prevent fixes like #1337 from being necessary again.

Other random things:
- Strictly scoping stuff with `{}` is pretty good. Found at least three copypaste bugs.
- Eliminate some unused or duplicate functions from `DatabaseCatalog`.
- `pg_proc` no longer hardcodes inbuilt oids for every proc, which was noted to overflow into the internal OID range. Instead, we use the same `next_oid++` that would be used for users directly for **all** procs.
- Where possible, add `static_assert` to enforce that e.g., only `RecoveryManager` can call some functions. I would use `std::enable_if` but getting compiled out makes testing a pain, and the function signature is weirder for people to pick up `template <typename CallerClass> auto SetTableSchemaPointer(...) -> std::enable_if_v<std::is_same_v<CallerClass, RecoveryManager>, old_return_type>` is a mouthful. This does mean that callers of these protected functions are now required to call `SetTableSchemaPointer<RecoveryManager>(...)` instead of `SetTableSchemaPointer(...)`, but that doesn't seem too bad -- this seems more secure than relying on people reading a `@warning`.

Non-goals:
- We were discussing refactoring the builder. I tried doing some template things but they haven't worked out. I am a little tired of the catalog as of right now and I am inclined to leave it as it is; especially since it is mostly standalone and will probably be never modified. Though I don't feel strongly either way about it and if you feel strongly about it I'm happy to revisit this.


src/include/catalog/database_catalog.h
```diff
@@ -36,490 +27,233 @@ class Index;
...
-   * @param table being queried
-   * @return vector of OIDs for all of the constraints that apply to this table
+   * @brief Rename a table.
+   *
+   * @param txn         The transaction to rename the table in.
+   * @param table       The table to be renamed.
+   * @param name        The new name for the table.
+   * @return            True if the rename succeeded. False otherwise.
+   *
+   * TODO(WAN): if this logic can be pushed to PgCoreImpl, update this comment to match style
```
jrolli, line 119: So this should be primarily implemented in `PgCoreImpl` with only the DDL lock taken in the wrapper.  That said, I got a good laugh when I looked at the current implementation...

src/include/catalog/database_catalog.h
```diff
@@ -36,490 +27,233 @@ class Index;
...
+   * @param table       The table whose schema should be updated.
+   * @param new_schema  The new schema to update the table to.
+   * @return            True if the update succeeded. False otherwise.
+   *
+   * @warning           The catalog accessor assumes it takes ownership of the schema object that is passed.
+   *                    As such, there is no guarantee that the pointer is still valid when this function returns.
+   *                    If the caller needs to reference the schema object after this call, the caller should use
+   *                    the GetSchema function to obtain the authoritative schema for this table.
+   *
+   * TODO(WAN): if this logic can be pushed to PgCoreImpl, update this comment to match style
```
jrolli, line 141: Unlike the other function with this comment, I think this one ultimately has to stay at the `DatabaseCatalog`-level because it may have wider impacts then just `Core` (e.g., updating `Constraints` and possibly other tables).

src/include/catalog/database_catalog.h
```diff
@@ -529,151 +263,85 @@ class DatabaseCatalog {
...
-   * @param oid oid to object
-   * @return pair of ptr to schema and ClassKind of object requested. ptr will be nullptr if no entry was found for the
-   * given oid
-   */
-  std::pair<void *, postgres::ClassKind> GetClassSchemaPtrKind(
-      common::ManagedPointer<transaction::TransactionContext> txn, uint32_t oid);
+  /** @brief Create a column entry in pg_attribute. @see PgAttributeImpl::CreateColumn */
+  template <typename Column, typename ClassOid, typename ColOid>
+  bool CreateColumn(common::ManagedPointer<transaction::TransactionContext> txn, ClassOid class_oid, ColOid col_oid,
+                    const Column &col);
```
jrolli, line 302: Does this actually need to be exposed outside of `Core`?  The original intent was that `pg_attribute` was never updated in isolation and that things like "creating a column" were only achievable through providing an updated schema object.  When we get to DDL, exposing this strikes me as an amazing footgun for the cached schema pointer to be out-of-sync with the underlying data.

src/include/catalog/database_catalog.h
```diff
@@ -529,151 +263,85 @@ class DatabaseCatalog {
...
+   * @param txn             The transaction to perform the schema change in.
+   * @param oid             The OID of the index.
+   * @param schema          The new index schema to set.
+   * @return                True if the index schema was set successfully. False otherwise.
    */
+  template <typename CallerType>
   bool SetIndexSchemaPointer(common::ManagedPointer<transaction::TransactionContext> txn, index_oid_t oid,
-                             const IndexSchema *schema);
+                             const IndexSchema *schema) {
+    static_assert(std::is_same_v<CallerType, storage::RecoveryManager>, "Only recovery should call this.");
```
jrolli, line 338: This is slick!  This may deserve some documentation in the `docs` folder as a technique for enforcing caller constraints elsewhere in the codebase.  I think we have warnings about this in other places but lack actual enforcement mechanisms.

src/include/catalog/index_schema.h
```diff
@@ -23,18 +23,19 @@ class TpccPlanTest;
 
 namespace noisepage::storage {
 class RecoveryManager;
-}
+}  // namespace noisepage::storage
 
 namespace noisepage::tpcc {
 class Schemas;
-}
+}  // namespace noisepage::tpcc
```
jrolli, line 30: I know this only popped in the diff because of a `clang-format` run, but this falls into the same category of the TPC-H friendship on `proc`.  I'm fine with leaving a TODO and issue, but I do think we need to clearly mark this as the "wrong" approach for future benchmarks/tests.

src/include/catalog/index_schema.h
```diff
@@ -419,6 +421,7 @@ class IndexSchema {
 
   friend class Catalog;
   friend class postgres::Builder;
+  friend class postgres::PgCoreImpl;
```
jrolli, line 424: nit: Can you consolidate this cluster with `friend class DatabaseCatalog;` at the start of the `private` section (exactly where they sit doesn't matter so much, but they should be together)?

src/include/catalog/postgres/pg_constraint_impl.h
```diff
@@ -0,0 +1,97 @@
...
+   * Cleaned up by: DatabaseCatalog::TearDown, where the scans from pg_class and pg_index pick these up.
+   */
+  ///@{
+  storage::SqlTable *constraints_;                         ///< The constraints table.
+  storage::index::Index *constraints_oid_index_;           ///< Indexed on: conoid
+  storage::index::Index *constraints_name_index_;          ///< Indexed on: connamespace, conname
+  storage::index::Index *constraints_namespace_index_;     ///< Indexed on: connamespace
+  storage::index::Index *constraints_table_index_;         ///< Indexed on: conrelid
+  storage::index::Index *constraints_index_index_;         ///< Indexed on: conindid
+  storage::index::Index *constraints_foreigntable_index_;  ///< Indexed on: confrelid
```
jrolli, line 93: Should these (and all of the other table/index pointers in other implementation files) actually be `ManagedPointer`s?  While the catalog as a whole "owns" them, it's actually the `pg_class` storage layer that truly owns them.  More specifically, the only safe way to delete them is through the teardown logic.

src/include/catalog/schema.h
```diff
@@ -16,17 +16,18 @@
 
 namespace noisepage {
 class StorageTestUtil;
-}
+}  // namespace noisepage
 
 namespace noisepage::tpcc {
 class Schemas;
-}
+}  // namespace noisepage::tpcc
```
jrolli, line 23: See comment in `index_schema.h`

src/include/type/type_id.h
```diff
@@ -32,6 +32,7 @@ enum class TypeId : uint8_t {
   VARBINARY,
   PARAMETER_OFFSET,
   VARIADIC,
+  HACK_PG_TYPE_VAR_ARRAY,  ///< pg_type requires a distinct type for var_array.
```
jrolli, line 35: Is it a hack, then?

src/catalog/database_catalog.cpp
```diff
@@ -30,856 +29,89 @@
...
+  auto teardown_pg_core = pg_core_.GetTearDownFn(txn, common::ManagedPointer(this));
+  auto teardown_pg_constraint = pg_constraint_.GetTearDownFn(txn);
+  auto teardown_pg_proc = pg_proc_.GetTearDownFn(txn);
 
-  BootstrapLanguages(txn);
+  auto dbc_nuke = [=]() {
+    // Order matters!
+    teardown_pg_core();
+    teardown_pg_constraint();
+    teardown_pg_proc();
```
jrolli, line 52: If order matters, then I'm surprised to see them in this order rather than reverse bootstrap order.  I'm pretty sure order does not matter here since the deferrals should ensure any live references across the catalog are dead before the object deletions occur.  If I'm wrong, this needs a more explicit comment on the ordering.

src/catalog/database_catalog.cpp
```diff
@@ -30,856 +29,89 @@
...
+  BootstrapPRIs();
 
-  // pg_language
-  const std::vector<col_oid_t> pg_language_all_oids{postgres::PG_LANGUAGE_ALL_COL_OIDS.cbegin(),
-                                                    postgres::PG_LANGUAGE_ALL_COL_OIDS.cend()};
-  pg_language_all_cols_pri_ = languages_->InitializerForProjectedRow(pg_language_all_oids);
-  pg_language_all_cols_prm_ = languages_->ProjectionMapForOids(pg_language_all_oids);
+  bool UNUSED_ATTRIBUTE retval;
+  retval = TryLock(txn);
+  NOISEPAGE_ASSERT(retval, "Bootstrap operations should not fail");
```
jrolli, line 77: If this assert were to ever fail, I suspect we'll need as much help for our future selves as possible...
```suggestion
  NOISEPAGE_ASSERT(retval, "Bootstrap should not fail to get write-lock: another thread grabbed it early? Check recovery logic (most probable cause)");
```

jrolli (changes requested):
For the most part it looks good and should greatly help the catalog code's readability.

Mostly nits on naming and comments.

src/include/catalog/catalog_column_def.h
```diff
@@ -0,0 +1,79 @@
...
+  /** @see ProjectedRow::SetNotNull */
+  void SetNotNull(common::ManagedPointer<storage::ProjectedRow> pr, const storage::ProjectionMap &pm) const {
+    pr->SetNotNull(pm.at(oid_));
+  }
+
+  /** @see ProjectedRow::Set */
+  void Set(common::ManagedPointer<storage::ProjectedRow> pr, const storage::ProjectionMap &pm, CatalogType val) const {
+    if constexpr (std::is_pointer_v<CatalogType>) {  // NOLINT
+      pr->Set<StorageType, false>(pm.at(oid_), reinterpret_cast<StorageType>(val), false);
+    } else {  // NOLINT
```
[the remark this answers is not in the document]
lmwnshn: The NOLINTs are because we're using cpplint that doesn't recognize constexpr, so it sees "else without if".

src/include/catalog/database_catalog.h
lmwnshn: :hankey: Will move

src/include/catalog/catalog_accessor.h
```diff
@@ -355,14 +355,6 @@ class EXPORT CatalogAccessor {
...
   common::ManagedPointer<execution::functions::FunctionContext> GetProcCtxPtr(proc_oid_t proc_oid);
 
-  /**
-   * Gets a functions context object for a given proc if it is null for a valid proc id then the functions context
-   * object is reconstructed, put in pg_proc and returned
-   * @param proc_oid The proc_oid whose FunctionContext object we are returning here
-   * @return nullptr if proc_oid is invalid else a valid functions context object for this proc_oid
-   */
-  common::ManagedPointer<execution::functions::FunctionContext> GetFunctionContext(proc_oid_t proc_oid);
-
```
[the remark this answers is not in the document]
lmwnshn: Renamed

src/include/catalog/catalog_column_def.h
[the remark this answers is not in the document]
lmwnshn: Actually, you're right -- weird. I did get the compiler to segfault yesterday because I was running low on memory. I'll try again to see which NOLINTs are really necessary.

src/include/catalog/catalog_column_def.h
```diff
@@ -0,0 +1,79 @@
...
+  }
+
+  /** @see ProjectedRow::SetNotNull */
+  void SetNotNull(common::ManagedPointer<storage::ProjectedRow> pr, const storage::ProjectionMap &pm) const {
+    pr->SetNotNull(pm.at(oid_));
+  }
+
+  /** @see ProjectedRow::Set */
+  void Set(common::ManagedPointer<storage::ProjectedRow> pr, const storage::ProjectionMap &pm, CatalogType val) const {
+    if constexpr (std::is_pointer_v<CatalogType>) {  // NOLINT
```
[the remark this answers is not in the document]
lmwnshn: Added.

src/include/catalog/catalog_column_def.h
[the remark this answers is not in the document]
lmwnshn: OK, so you don't need the NOLINTs on the `if constexpr`, but you need them on the `else` otherwise it blows up in spectacular fashion. Pushed.

src/include/catalog/database_catalog.h
lmwnshn: Took away their guns (specifically, removed `CreateColumn` and `DeleteColumns`).

Is exposing `GetColumns` okay? `RecoveryManager` uses it. Leaving unresolved for now.

jrolli: `GetColumns` is fine since it's read-only and I think needed in some other logic as well.

src/include/catalog/database_catalog.h
lmwnshn: Removed TODO.

src/include/catalog/database_catalog.h
lmwnshn: Marking resolved

src/include/catalog/database_catalog.h
lmwnshn: Added a bit to `cpp_guidelines_code_style`, take a look? thanks!

src/include/type/type_id.h
lmwnshn: Good question. I am concerned that we do not have support for this anywhere else in the system, though -- I suppose/hope it will blow up as "UNKNOWN TYPE" if people try to use it elsewhere.

src/catalog/database_catalog.cpp
lmwnshn: thanks

src/catalog/database_catalog.cpp
lmwnshn: You are right, it teardowns fine if I flip ordering around. I don't remember why I wrote that (whoops).

lmwnshn: removed

src/include/type/type_id.h
lmwnshn: renamed to VAR_ARRAY

src/include/catalog/index_schema.h
lmwnshn: moved

src/include/catalog/schema.h
lmwnshn: Happily removable without further work.

src/include/catalog/index_schema.h
lmwnshn: Happily removable without further work.

src/include/catalog/postgres/pg_constraint_impl.h
lmwnshn: I've done the refactor. I think you run into a bit of weirdness in `SetIndexPointer` and `SetTablePointer` as those expect a raw pointer because they need to double-defer a delete, but, those are also only called from recovery so I guess that's within the weirdness budget there.

jrolli (approved):
LGTM.  Thanks for changing the held pointers in the catalog over to `ManagedPointers` and adding the static assert trick to the C++ guidelines documentation!

PR #1368, opened 2020-12-03
Common Jenkinsfile Build Function

bialesdaniel:
# Common Jenkinsfile Build Function

## Description
This PR makes several changes to the build pipelines.
1. Create a common build function that accepts arguments to determine which compiler options to select. This will simplify future changes to the compile and build commands because now they only need to be changed in a single location. I also added docs that describe the current options for each build stage. For the most part, I did not change any options just a couple of small improvements especially for the OLTPBench tests where we were building more than we needed to.
2. I changed the End-to-End Performance test in the terrier build to now report the results to the performance-storage-service. This will allow our Github bot to compare PR results to the master results.
3. I added a 600-sec YCSB test to the nightly build so that we can see the effect of a read-only workload on the memory utilization. 

Since this PR primarily contains changes to Jenkinsfiles we tested it in our Jenkins folder. It passed all builds
Terrier: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/common-compile-function/37/pipeline/
Nightly: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier-nightly/detail/common-compile-function/7/pipeline/99

I need to merge in changes to master but I figured I would do that after a review because it would be easier to do the review this way.


Jenkinsfile
```diff
@@ -128,12 +147,13 @@ pipeline {
...
-                        cd build
-                        cmake -GNinja -DNOISEPAGE_UNITY_BUILD=ON -DNOISEPAGE_TEST_PARALLELISM=1 -DCMAKE_BUILD_TYPE=Debug -DNOISEPAGE_USE_ASAN=ON -DNOISEPAGE_BUILD_BENCHMARKS=OFF -DNOISEPAGE_USE_JUMBOTESTS=OFF ..
-                        ninja''', label: 'Compiling'
+                        
+                        script{
+                            if(!utils){
+                                utils = load("Jenkinsfile-utils.groovy")
+                            }
+                            utils.noisePageBuild(os:"macos",useASAN:true)
+                        }
```
apavlo, line 156: nit: Can you run this through a formatter or make sure the spacing matches up with the rest of the file? Maybe something like this:

```
# Build System
script {
    if (!utils) utils = load("Jenkinsfile-utils.groovy")
    utils.noisePageBuild(os:"macos", useASAN:true)
}
```

Jenkinsfile-utils.groovy
```diff
@@ -0,0 +1,109 @@
+/*
```
apavlo, line 1: Does this file have to have a `.groovy` extension? Maybe we can name the file `Jenkinsfile-utils` and then add this to the first line of the file:

```
#!/usr/bin/env groovy
```

Jenkinsfile-utils.groovy
```diff
@@ -0,0 +1,109 @@
...
+        isCodeCoverage: false,
+        isJumboTest: false,
+        isRecordTime: false
+    ]
+    def Map config = defaultArgs << args
+    def String compileCmd = generateCompileCmd(config)
+    def String buildCmd = generateBuildCmd(config)
+    def String buildScript = generateBuildScript(compileCmd, buildCmd, config.isRecordTime)
+
+    sh script:buildScript, label: "Compile & Build"
```
apavlo, line 24: The label `Compile & Build` is redundant. Why not just `Build`?

Jenkinsfile-utils.groovy
apavlo, line 1: Run this file through a formatter.

Jenkinsfile-utils.groovy
```diff
@@ -0,0 +1,109 @@
...
+
+    if(!config.isBuildBenchmarks){
+        compileCmd += " -DNOISEPAGE_BUILD_BENCHMARKS=OFF"
+    }
+
+    if(!config.isBuildTests){
+        compileCmd += " -DNOISEPAGE_BUILD_TESTS=OFF"
+    }
+
+    if(config.buildType=="Release" && !config.isBuildTests){
```
apavlo, line 61: Add an additional check to prevent them from using jemalloc if `config.useASAN` is true.

Jenkinsfile-utils.groovy
```diff
@@ -0,0 +1,109 @@
...
+    def String buildScript = generateBuildScript(compileCmd, buildCmd, config.isRecordTime)
+
+    sh script:buildScript, label: "Compile & Build"
+}
+
+/*
+generateCompileCmd creates the cmake command string. It is based on the
+config passed into the function. The config arguments are the same as the
+defaultArgs in noisePageBuild.
+*/
```
apavlo, line 31: Add leading `*` prefix to function comments to match rest of the system:

```
/**
 * generateCompileCmd creates the cmake command string. It is based on the
 * config passed into the function. The config arguments are the same as the
 * defaultArgs in noisePageBuild.
 */
```

Example: https://github.com/cmu-db/noisepage/blob/master/src/include/main/db_main.h#L52-L56

Jenkinsfile-utils.groovy
```diff
@@ -0,0 +1,109 @@
...
+file.
+*/
+def generateBuildScript(compileCmd, buildCmd, isRecordTime){
+    def String script = '''
+    mkdir build
+    cd build
+    '''
+    if(isRecordTime){
+        script += """
+        /usr/bin/time -o /tmp/compiletime.txt -f %e sh -c \"$compileCmd
```
apavlo, line 98: NIT: Can you rename the file to `noisepage-compiletime.txt`? We should be using the `noisepage-` prefix to any temp files that we create to make easier for people to clean up.

apavlo (changes requested):
Overall this is much cleaner. Thanks for doing this. The documentation is really nice.

I only have minor change requests.

bialesdaniel:
I made the updates. Here are the two new passing builds.
Terrier: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/common-compile-function/42/pipeline/172Nightly: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier-nightly/detail/common-compile-function/13/pipeline/75

Jenkinsfile-nightly
```diff
@@ -91,31 +92,24 @@ pipeline {
...
+                sh script:'''
+                cd script/testing
+                python3 microbench/run_microbench.py --run --num-threads=4 --benchmark-path $(pwd)/../../build/benchmark --logfile-path=/mnt/ramdisk/benchmark.log --publish-results=prod --publish-username=${PSS_CREATOR_USR} --publish-password=${PSS_CREATOR_PSW}
+                ''', label:'Microbenchmark'
 
-                }
+                archiveArtifacts 'script/testing/*.json'
+                junit 'script/testing/*.xml'
+                // This next line can be removed when the loop is removed
+                sh 'rm script/testing/*.json'
```
skyfishjy, line 111: Do we need to run this rm command? Base on the comment, I think we can delete them and I also think deleteDir() in post steps will do the similar thing

skyfishjy (commented):
LGTM

Jenkinsfile-nightly
bialesdaniel: Good catch!

Jenkinsfile-utils
```diff
@@ -0,0 +1,110 @@
...
+ * config passed into the function. The config arguments are the same as the
+ * defaultArgs in noisePageBuild.
+ */
+String generateCompileCmd(Map config = [:]) {
+    String compileCmd = "cmake -GNinja -DCMAKE_BUILD_TYPE=$config.buildType"
+
+    if (config.isCodeCoverage) {
+        compileCmd += ' -DNOISEPAGE_GENERATE_COVERAGE=ON'
+    } else {
+        compileCmd += ' -DNOISEPAGE_UNITY_BUILD=ON'
```
lmwnshn, line 40: Nit: add a comment that UNITY_BUILD messes with the accurate tracking of code coverage.

Jenkinsfile-utils
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    if (config.os != 'macos' && config.useCache) {
+        compileCmd += ' -DCMAKE_CXX_COMPILER_LAUNCHER=ccache'
```
lmwnshn, line 44: Nit: add a comment on why macOS doesn't use ccache

Jenkinsfile-utils
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    if (config.os == 'ubuntu' && config.isBuildTests) {
```
lmwnshn, line 47: Q: why is this ubuntu specific?

Jenkinsfile-utils
```diff
@@ -0,0 +1,110 @@
...
+
+    sh script:buildScript, label: 'Build'
+}
+
+/**
+ * generateCompileCmd creates the cmake command string. It is based on the
+ * config passed into the function. The config arguments are the same as the
+ * defaultArgs in noisePageBuild.
+ */
+String generateCompileCmd(Map config = [:]) {
```
lmwnshn, line 34: Nit: it seems like you want to always generate a complete command, rather than relying on the CMake noisepage defaults. Otherwise someone can change those and not know of this function, then mystery things happen in CI.

lmwnshn: So to be explicit, you would have every `-DNOISEPAGE_BLA_BLA` and explicitly `ON/OFF` them.

docs/build_stage_compile_settings.md
```diff
@@ -0,0 +1,33 @@
+# Compile Settings for Build Stages
+This doc was last updated: 11/3/20
+
+This doc describes the different compiler options for the Jenkins builds. If a cell is left blank it means that the option is not specified thus uses the default.
+
+## Terrier build
```
lmwnshn, line 6: Nit: terrier -> NoisePage. Though the URL has to stay the same, yeah.

docs/build_stage_compile_settings.md
```diff
@@ -0,0 +1,33 @@
+# Compile Settings for Build Stages
+This doc was last updated: 11/3/20
+
+This doc describes the different compiler options for the Jenkins builds. If a cell is left blank it means that the option is not specified thus uses the default.
```
lmwnshn, line 4: Nit: relying on default behavior is bugprone, it seems better to be explicit.

docs/build_stage_compile_settings.md
```diff
@@ -0,0 +1,33 @@
...
+| Release/unittest | macos | clang-8.0 | ON |  |  |  | OFF | Release |  |  |  |  |
+| Release/jumbotests | ubuntu | gcc-9.3 | ON | $(nproc) | ccache |  | OFF | Release | ON |  |  |  |
+| Release/jumbotests | ubuntu | clang-8.0 | ON | $(nproc) | ccache |  | OFF | Release | ON |  |  |  |
+| Debug/e2etest/oltpbench | macos | clang-8.0 | ON |  |  | ON | OFF | Debug |  |  |  | OFF |
+| Debug/e2etest/oltpbench | ubuntu | gcc-9.3 | ON |  | ccache | ON | OFF | Debug |  |  |  | OFF |
+| End-to-End Performance | benchmark | gcc-9.3 | ON |  | ccache |  | OFF | Release |  | ON |  | OFF |
+| Microbenchmark | benchmark | gcc-9.3 | ON |  | ccache |  | ON | Release |  | ON |  | OFF |
+
+
+## Terrier Nightly build
```
lmwnshn, line 25: Nit: terrier -> NoisePage

docs/build_stage_compile_settings.md
```diff
@@ -0,0 +1,33 @@
+# Compile Settings for Build Stages
```
lmwnshn, line 1: I am generally dubious that we will keep this document up to date. It seems better to have a single source of truth (the Jenkinsfile that runs). Can you push this to the Jenkinsfile, define an array or something?

Jenkinsfile-utils
```diff
@@ -0,0 +1,110 @@
+#!/usr/bin/env groovy
```
lmwnshn, line 1: See comment in doc/ file below.

If we are going this route, I am inclined to define an array of all the possible configurations here in this file.
Then there is a single source of truth here.

lmwnshn (changes requested):
I think separating it out into a groovy file is a nice idea. However, since we're already doing that, can we then instead just define all the possible stages and configurations as one configuration matrix in `Jenkinsfile-utils`?

I am hesitant about the current approach because it splits up "where do I go to see what is happening, how do I add something" into even more separate files:

- `CMakeLists.txt`, to figure out defaults (we currently have to do this)
- `Jenkinsfile`, to figure out what stage (we currently have to do this)
- `Jenkinsfile-utils`, to figure out what the Jenkinsfile is actually doing (this would be a new step. And it requires people to understand your generateSomethingCmd functions)
- `docs/build_stage_compile_settings.md`, to see what we think the config is (this would be a new step which has to be kept in sync with all three files above)

If you were to instead push all this information into a single matrix of configurations in `Jenkinsfile-utils`, then people only have to read and update that one place.

Jenkinsfile-utils
bialesdaniel: Why doesn't it use ccache?

Jenkinsfile-utils
bialesdaniel: cause macos doesn't support `$(nproc)`.

Jenkinsfile-utils
bialesdaniel: That makes sense. That would also simplify the case you mentioned in your overall comment cause developers would no longer have to check `CMakeLists.txt`

docs/build_stage_compile_settings.md
bialesdaniel: I think we should probably call it terrier build until we actually change the name in Jenkins, otherwise, I think it would be confusing.

docs/build_stage_compile_settings.md
bialesdaniel: I think defining an array semi defeats the purpose of this. Which was to make it so change to the build would only require a single change. I included a last updated line in the docs because I anticipated that issue as well. I'm not entirely sure the docs are necessary.

Another potential solution would be to build documentation rules into the build. AKA if the Jenkins-utils file changes and this one does not then fail the build. (That might be a little strict but there could be good rules to enforce documentation)

Jenkinsfile-utils
bialesdaniel: I responded in the doc file but I don't think that is the most elegant solution.

docs/build_stage_compile_settings.md
bialesdaniel: Another problem of defining an array is then there is a lack of transparency into the build from the Jenkinsfile. Currently, by passing in just the args that pertain to the stage we can see what is most important/unique about the build for that stage.

lmwnshn: I don't follow, sorry. For example, if you define a base config `C` with all the defaults and then each entry of the array is just `C.SetParam1(true).SetParam2(false)`, etc., that seems to be the best of both worlds.

As it stands in this PR, a change to the build system requires changing all the files mentioned above including the documentation here.

Jenkinsfile-utils
lmwnshn: Not something I know off the top of my head, I'm guessing we should refer to #830.

Jenkinsfile-utils
lmwnshn: Could this be documented then?

docs/build_stage_compile_settings.md
bialesdaniel: Can you checkout the changes I pushed to `Jenkinsfile-utils`. I'm kinda doing that but not in an array. I have a default config and then rules that apply the changes. I think one question is whether the a developer needs to be aware of the distinct aspects of the build when looking at the Jenkinsfile/Jenkinsfile-nightly. If they do then the current solution seems good (if I delete the docs) because it highlights the key attributes as arguments to the function called in the Jenkinsfile/Jenkinsfile-nightly (i.e. `utils.noisePageBuild(buildType:utils.RELEASE_BUILD, isBuildTests:false)`). However, if when changing the Jenkinsfile the developer does not need to know about the build config then we could make that method `utils.noisePageBuild('End-to-End Performance')` and then do a map of stage->config in this utils. Thoughts?

bialesdaniel:
Fun fact these changes speed up the total build time of the terrier build by 10 min. (adding test parallelization on macos + excluding tests in the OLTPBench builds.)

bialesdaniel:
Passing Build: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/common-compile-function/82/pipeline
Passing Nightly: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier-nightly/detail/common-compile-function/29/pipeline/

bialesdaniel:
Latest passing build: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/common-compile-function/84/pipeline/268

bialesdaniel:
Passing build: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/common-compile-function/95/pipeline/340

bialesdaniel:
Latest passing build: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/common-compile-function/97/pipeline

apavlo (approved):
LGTM

bialesdaniel:
Passing Nightly Build: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier-nightly/detail/common-compile-function/45/pipeline/
Passing Terrier Build: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/common-compile-function/99/pipeline/

lmwnshn (approved):
Rubber stamping.

PR #1369, opened 2020-12-03
Fix UPDATE behavior when SET clauses reference other columns.

### LINKED ISSUE #1366: Incorrect UPDATE behavior when SET clauses reference other columns
## Summary
While poking around the mini-runners yesterday, I found this update bug. The issue comes when the `SET` clauses of an update statement reference another column that is also being updated.

## Steps to Reproduce
Executing the following on the master branch produces the wrong result:
```
noisepage=# create table t (a INT, b INT, c INT);
CREATE TABLE
noisepage=# insert into t values (1, 2, 3);
INSERT 0 1
noisepage=# update t set a = b + 1, b = a + 1, c = c;
UPDATE 1
noisepage=# select * from t;
 a | b | c 
---+---+---
 3 | 4 | 3
(1 row)
```

### Expected Behavior
Expected behavior from postgres:
```
postgres=# create table t (a INT, b INT, c INT);
CREATE TABLE
postgres=# insert into t values (1, 2, 3);
INSERT 0 1
postgres=# update t set a = b + 1, b = a + 1, c = c;
UPDATE 1
postgres=# select * from t;
 a | b | c 
---+---+---
 3 | 2 | 3
(1 row)
```

### Actual Behavior
TPL for `update t set a = b + 1, b =  a + 1, c = c`
```
var update_pr: *ProjectedRow
update_pr = @getTablePR(&updater)
@prSetIntNull(update_pr, 0, @vpiGetIntNull(vpi, 2))
@prSetIntNull(update_pr, 1, @vpiGetIntNull(vpi, 1))
@prSetIntNull(update_pr, 2, @vpiGetIntNull(vpi, 0))
@prSetIntNull(update_pr, 0, @prGetIntNull(update_pr, 0))
@prSetIntNull(update_pr, 1, @prGetIntNull(update_pr, 1) + @intToSql(1))
@prSetIntNull(update_pr, 2, @prGetIntNull(update_pr, 2))
if (!@tableUpdate(&updater, &slot)) {
     @abortTxn(queryState.execCtx)
} 
```

The problem arguably is that the SET clauses are using `update_pr` and mutating `update_pr` which means all the `SET` clauses are not applied "at-once" but are individually applied (as if multiple updates).

17zhangw:
## Description

Fix #1366, which indicates an error with respect to how set clauses are codegen'ed. This PR fixes the issue by codegen'ing the SET clauses to refer to the original tuple (i.e the tuple presented by the child plan node). For columns that are not updated by any SET clause, the value from the original tuple is copied into the new tuple.

### DIFF
--- a/script/testing/junit/traces/update.test
+++ b/script/testing/junit/traces/update.test
@@ -211,3 +211,41 @@ SELECT a FROM update5;
 statement ok
 DROP TABLE update5
 
+statement ok
+CREATE TABLE update6 (a INT, b INT, c INT);
+
+statement ok
+INSERT INTO update6 VALUES (1, 2, 3);
+
+statement ok
+UPDATE update6 SET a = b + 1, b = a + 1, c = c;
+
+query III rowsort
+SELECT * FROM update6;
+----
+3
+2
+3
+
+statement ok
+UPDATE update6 SET b = a + 1, c = a + b, a = b + 1;
+
+query III rowsort
+SELECT * FROM update6;
+----
+3
+4
+5
+
+statement ok
+UPDATE update6 SET b = a + c;
+
+query III rowsort
+SELECT * FROM update6;
+----
+3
+8
+5
+
+statement ok
+DROP TABLE update6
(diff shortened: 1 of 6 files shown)


lmwnshn (approved):
LGTM, ありがとう

PR #1370, opened 2020-12-03
Fixed Decimal package with noisepage - current master level

rohanaggarwal7997:
This is the updated version of #1290. After fixing all the conflict errors(more than 100) with the latest master which was pulled recently into by branch.
This PR provides all fixed decimal operation functionality to noisepage in the SQL queries.

There are still a few comments left to be addressed from 1290 and some small changes to perfectly clean this PR up. I am noting them down here and would be fixing them. This shouldn't take more than one day.

1) Hash function used by the PR has to be updated according to discussion with Matt in general, currently we are using https://github.com/vnmakarov/mum-hash/blob/95b65eb25635bab8f0801b09a18f50749dc70a99/src/City.h#L87.
Also unordered map in decimal package has to be restored with this hash.
2) There is a small code commented out in sema_builtin.cpp and bind_node_visitor.cpp that has to be corrected.
3) To Json and From Json functions do not support 128 bit values and discussion needs to be done on how to handle them, currently a static cast to 64 bits is being used.
4) AstPrettyPrint and logger level set has to be removed.
5) Some comments from 1290 have to be addressed. 
6) Prashant's comment to change the builtin name from upgradeprecision to rescaleprecision has to be addressed.


mbutrovich:
Closing in favor of #1434.

PR #1373, opened 2020-12-04
Modeling Index Inserts and Deletes

17zhangw:
## Description
PR augments the operating unit models with index inserts and deletes. Previously inserts would only capture the raw cost of inserting into a table; similarly, updates and deletes would implicitly capture an index-insert or index-delete operation. This PR introduces functionality to explicitly capture inserting into an index or deleting from an index.

A list of changes are as follows:
- Index-Insert and Index-Delete models and runners (in TPL)
- Explicitly separate in-place updates and updates that delete then insert
- For updates that modify indexes, record `insert`, `delete` and corresponding index updates
- Update Runners now are in-place sequential scan updates
- `InsertPlanNode`, `UpdatePlanNode`, and `DeletePlanNode` properly capture `index_oids` metadata
- PerfMonitor no longer records `bus_cycles_` (recording all would cause all counters to zero)
- Warming up INSERTS
- Match varchar key size in runners with recorder


src/include/common/perf_monitor.h
```diff
@@ -61,8 +61,12 @@ class PerfMonitor {
     // uint64_t branch_misses_;
     /**
      * Bus cycles, which can be different from total cycles.
+     *
+     * TODO(wz2): Dated Nov 5th, 2020 on dev4. Recording {cycle,instr,cache-ref,cache-miss,bus,ref-cpu} causes all
```
17zhangw, line 65: I disabled `bus_cycles_` since adding `bus_cycles` would cause all counters to be zeroed. (this is already `kernel.perf_event_paranoid=-1` and `nmi_watchdog=0`). I disabled recording bus_cycles since we currently don't use this in the models.

There is also a new ebpf metrics collection framework under-way, so this `perf_monitor.h` might end up getting replaced with a more robust solution.

@mbutrovich

src/self_driving/modeling/operating_unit_recorder.cpp
```diff
@@ -49,10 +49,69 @@
...
+    tracking[std::make_tuple(index->GetSize(), key_size, num_keys)] += 1;
+  }
+
+  // Record each distinct group
+  for (auto &track : tracking) {
+    // TODO(wz2): Eventually need to modify track.second to consider num_rows/cardinality
+    pipeline_features_.emplace(
+        type, ExecutionOperatingUnitFeature(current_translator_->GetTranslatorId(), type, std::get<0>(track.first),
+                                            std::get<1>(track.first), std::get<2>(track.first), track.second, 1, 0, 0));
+  }
```
malin1993ml, line 112: I actually don't think we need to group the indexes here. Typically a table won't have a hundred different indexes. I think it's simpler to just leave the indexes separately.

malin1993ml (changes requested):
This generally looks good! The main question is about the syntax that we use for the index-insert/delete runners.

benchmark/runner/mini_runners.cpp
```diff
@@ -713,7 +718,7 @@ void NetworkQueriesOutputRunners(pqxx::work *txn) {
   auto types = {type::TypeId::INTEGER, type::TypeId::DECIMAL};
   std::vector<int64_t> row_nums = {1, 3, 5, 7, 10, 50, 100, 500, 1000, 2000, 5000, 10000};
 
-  bool metrics_enabled = true;
+  bool metrics_enabled = db_main->GetSettingsManager()->GetBool(settings::Param::pipeline_metrics_enable);
```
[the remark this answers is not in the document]
17zhangw: It is possible that at this point, metrics are already enabled. We use `metrics_enabled` to ensure that the first `iters - 1` iterations run with pipeline metrics off and the last iteration runs with pipeline metrics on.

benchmark/runner/mini_runners.cpp
```diff
@@ -780,8 +784,7 @@ void NetworkQueriesCreateIndexRunners(pqxx::work *txn) {
   // Extract to restore
   int original_threads = db_main->GetSettingsManager()->GetInt(settings::Param::num_parallel_execution_threads);
   bool counters = db_main->GetSettingsManager()->GetBool(settings::Param::counters_enable);
-
-  bool metrics_enabled = true;
+  bool metrics_enabled = db_main->GetSettingsManager()->GetBool(settings::Param::pipeline_metrics_enable);
```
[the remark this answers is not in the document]
17zhangw: See above comment.

benchmark/runner/mini_runners.cpp
```diff
@@ -1239,95 +1622,61 @@ void MiniRunners::ExecuteUpdate(benchmark::State *state) {
...
+  auto int_size = type::TypeUtil::GetTypeTrueSize(type::TypeId::INTEGER);
+  auto bigint_size = type::TypeUtil::GetTypeTrueSize(type::TypeId::BIGINT);
+  auto tuple_size = int_size * tbl_ints + bigint_size * tbl_bigints;
+  auto tuple_col = tbl_ints + tbl_bigints;
+  auto update_size = int_size * num_integers + bigint_size * num_bigints;
+  auto update_col = num_integers + num_bigints;
+
+  // UPDATE [] SET [col] = [col]
+  // By doing an update in this way, we simulate a seq_scan more closely
+  // by requiring the seq_scan to produce the column value.
```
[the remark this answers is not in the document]
17zhangw: Right. I just double-checked this against TPCC. I will need to restructure the `UPDATE` so it doesn't modify any indexed columns (so it does not do a delete followed by an insert).

benchmark/runner/mini_runners.cpp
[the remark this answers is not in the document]
malin1993ml: Oh I thought the `metrics_enabled` variable is redundant for that. But now I realize that we cannot disable the metrics again if it's already disabled. I think this makes sense.

benchmark/runner/mini_runners.cpp
```diff
@@ -949,6 +963,242 @@ BENCHMARK_DEFINE_F(MiniRunners, SEQ0_OutputRunners)(benchmark::State &state) {
...
+    auto units = std::make_unique<selfdriving::PipelineOperatingUnits>();
+
+    std::stringstream query;
+    query << "CREATE INDEX idx" << i << " ON " << tbl_name << " (" << cols << ")";
+    auto equery = OptimizeSqlStatement(query.str(), std::make_unique<optimizer::TrivialCostModel>(), std::move(units),
+                                       PassthroughPlanChecker, nullptr, nullptr, &settings);
+    BenchmarkExecQuery(1, equery.first.get(), equery.second.get(), true, &empty_params, &settings);
+  }
+
+  if (!is_insert) {
```
malin1993ml, line 994: Why do we have to insert a tuple and the delete this specific tuple right away? I think deleting any tuple would be fine, since we're aborting the transaction anyway, right? For example, you can just delete the first tuple (`offset` = 0) from any `RawBlock` in the `DataTable`. I believe the `TupleSlot` (`RawBlock` + `offset`) is all you need for deletion.

benchmark/runner/mini_runners.cpp
```diff
@@ -1457,9 +1719,8 @@ BENCHMARK_DEFINE_F(MiniRunners, SEQ3_SortRunners)(benchmark::State &state) {
   units->RecordOperatingUnit(execution::pipeline_id_t(1), std::move(pipe1_vec));
 
   std::stringstream query;
-  auto cols = ConstructColumns("", type::TypeId::INTEGER, type::TypeId::DECIMAL, num_integers, num_decimals);
-  auto tbl_name =
-      ConstructTableName(type::TypeId::INTEGER, type::TypeId::DECIMAL, tbl_ints, tbl_decimals, row, table_car);
+  auto cols = ConstructColumns("", type::TypeId::INTEGER, type::TypeId::REAL, num_integers, num_reals);
+  auto tbl_name = ConstructTableName(type::TypeId::INTEGER, type::TypeId::REAL, tbl_ints, tbl_reals, row, table_car);
```
malin1993ml, line 1723: I think we can combine `GenerateTableName` and `GenerateMixedTableName` from `TableGenerator` to a single function call. The function can always take in vectors of types and column counts, evaluate whether any type's column count is 0, and generate the correct table name.

benchmark/runner/mini_runners.cpp
```diff
@@ -1068,7 +1324,7 @@ BENCHMARK_DEFINE_F(MiniRunners, SEQ2_0_IndexScanRunners)(benchmark::State &state
   std::stringstream query;
   auto table_name = execution::sql::TableGenerator::GenerateTableIndexName(type, num_rows);
```
malin1993ml, line 1325: I forgot that why for some runners we're using this set of tables and for other runners we're using the default set of tables? The comments for `GenerateMiniRunnersData` and `GenerateMiniRunnerIndexTables` in `TableGenerator` are very light in terms of explaining their differences.

script/model/training_util/data_transforming_util.py
```diff
@@ -90,18 +91,53 @@ def _tuple_num_linear_log_predict_transform(x, y):
...
+    tuple_num = np.copy(x[:, data_info.INPUT_CSV_INDEX[ExecutionFeature.NUM_ROWS]])
+    cardinality = np.copy(x[:, data_info.INPUT_CSV_INDEX[ExecutionFeature.EST_CARDINALITIES]])
+    return y * ((cardinality * np.log2(tuple_num)) + _TRANSFORM_EPSILON)[:, np.newaxis]
+
+
+# Transform the target in a linear log scale (cardinality * log tuple_num)
+_tuple_num_cardinality_linear_log_transformer = (_tuple_num_cardinality_linear_log_train_transform,
+                                                 _tuple_num_cardinality_linear_log_predict_transform)
+
+def _tuple_cardinality_linear_train_transform(x, y):
```
malin1993ml, line 125: Remove the "tuple" and just call this  `_cardinality_linear_train_transform`. Apply to others as well.

script/model/training_util/data_transforming_util.py
```diff
@@ -90,18 +91,53 @@ def _tuple_num_linear_log_predict_transform(x, y):
...
+    # Transform down the target according to the log cardinality in the input
```
malin1993ml, line 126: Typos in comments as well (log).

malin1993ml (changes requested):
I think this version is much better! Just have some small questions.

benchmark/runner/mini_runners.cpp
17zhangw: This makes sense. I think was mostly legacy from the TPL delete code (in case where table cardinality is not 0).

benchmark/runner/mini_runners.cpp
17zhangw: `GenerateTableName` and `GenerateMixedTableName` have been combined.

benchmark/runner/mini_runners.cpp
17zhangw: I think this is a legacy decision (I don't remember why this was the case either). Either way, I have removed `GenerateMiniRunnerIndexTables` and unified it such that all runners use the same tables.

script/model/training_util/data_transforming_util.py
```diff
@@ -44,63 +45,98 @@ def _tuple_num_memory_cardinality_linear_predict_transform(x, y):
...
+_num_cardinality_linear_log_transformer = (_num_cardinality_linear_log_train_transform,
+                                           _num_cardinality_linear_log_predict_transform)
+
+def _cardinality_linear_train_transform(x, y):
+    # Transform down the target according to the cardinality in the input
+    cardinality = np.copy(x[:, data_info.INPUT_CSV_INDEX[ExecutionFeature.EST_CARDINALITIES]])
+    return y / (cardinality + _TRANSFORM_EPSILON)[:, np.newaxis]
+
+
+def _cardinality_linear_predict_transform(x, y):
```
malin1993ml, line 131: Eh sorry that I was not clear before. I didn't mean to remove the word "tuple" from all functions. I was only referring to this function, since this only transforms with the cardinality. The functions that transform with tuple num should still has "tuple_num" in their names... Or even better, we can name "tuple_num" to "num_rows" to be consistent with the csv header.

script/model/training_util/data_transforming_util.py
```diff
@@ -44,63 +45,98 @@ def _tuple_num_memory_cardinality_linear_predict_transform(x, y):
...
-    return y * (np.log2(tuple_num) + 1)[:, np.newaxis]
+    return y * (np.log2(tuple_num) + _TRANSFORM_EPSILON)[:, np.newaxis]
 
 
 # Transform the target in a log scale (logn) according to the tuple num
-_tuple_num_log_transformer = (_tuple_num_log_train_transform, _tuple_num_log_predict_transform)
+_num_log_transformer = (_num_log_train_transform, _num_log_predict_transform)
+
+
+def _num_cardinality_linear_log_train_transform(x, y):
```
malin1993ml, line 107: Also, move "log" right after "tuple_num"/"num_rows", which directly refers to how the transformation is done. I.e., this should be `num_rows_log_cardinality_linear_train_transform`.

malin1993ml (approved):
LGTM with only one minor thing.

util/execution/table_generator/table_generator.cpp
```diff
@@ -456,7 +457,7 @@ void TableGenerator::GenerateMiniRunnersData(const runner::MiniRunnersSettings &
             col_nums.emplace_back(dist.second);
```
malin1993ml, line 457: Somehow the comment is not added... I think there's this removing of the types with 0 column that is redundant right now with the new `GenerateTableName` refactor?

PR #1374, opened 2020-12-05
pg_attribute atttypmod and REAL/DECIMAL fixes

mbutrovich:
## Description

We've played fast and loose with how we handle REAL/DOUBLE/DECIMAL for too long. The front end shoved all REAL/DOUBLE/DECIMAL/NUMERIC types into our internal 8 byte floating point type, but called it `type::TypeId::Decimal` while the execution engine called the underlying type `Real`. This starts us back on the right track so that @rohanaggarwal7997 can get his fixed point DECIMAL stuff working end-to-end. We should make sure to use REAL or DOUBLE in SQL when we want floating point types now.

- Adds new `type::TypeId::REAL` that encompasses Postgres' REAL and DOUBLE types. This is an 8-byte double in our system. Execution engine wraps this as `Real`.
- `type::TypeId::DECIMAL` is now reserved for Postgres' NUMERIC and DECIMAL. This is a 16-byte double in our system. Execution engine wraps this as `DecimalVal`.
- Adds [pg_attribute's atttypmod](https://www.postgresql.org/docs/current/index.html) field to handle the max length for variable length types, and the precision for DECIMAL.
- Refactors `IndexSchema` to not use bit-twiddling to pack its fields.
- Renamed a few operating unit names from DECIMAL to REAL. @17zhangw @malin1993ml.
- Modified JUnit tests that were meant to exercise floating point types to define schemas that use DOUBLE instead of DECIMAL.

### Remaining Tasks
- [x] Doxygen
- [x] Probably fix end-to-end tests that might be misusing SQL DECIMAL/REAL types
- [x] Discuss the TODOs I added while writing this and potentially open some issues
- [x] Double check other documentation is still correct

## Further Work
- We should reconcile `type::TypeId` with `execution::sql::SqlTypeId`
- ~~The system **crashes** if a table is instantiated with an attribute of type DECIMAL or NUMERIC. Writing error-handling code when @rohanaggarwal7997 is working on supporting arbitrary precision types seemed like a misuse of time.~~ I'm gonna continue spoofing NUMERIC/DECIMAL as DOUBLE for now so we don't break oltpbench testing.


src/execution/sql/ddl_executors.cpp
```diff
@@ -79,7 +79,7 @@ bool DDLExecutors::CreateTableExecutor(const common::ManagedPointer<planner::Cre
     for (const auto &unique_col : unique_constraint.unique_cols_) {
       const auto &table_col = schema.GetColumn(unique_col);
       if (table_col.Type() == type::TypeId::VARCHAR || table_col.Type() == type::TypeId::VARBINARY) {
```
rohanaggarwal7997, line 81: Probably need decimal here too.

src/parser/expression/operator_expression.cpp
```diff
@@ -35,7 +35,8 @@ void OperatorExpression::DeriveReturnValueType() {
     return t1->GetReturnValueType() < t2->GetReturnValueType();
   });
   const auto &type = (*max_type_child)->GetReturnValueType();
-  NOISEPAGE_ASSERT(type <= type::TypeId::DECIMAL, "Invalid operand type in Operator Expression.");
+  NOISEPAGE_ASSERT(type <= type::TypeId::REAL, "Invalid operand type in Operator Expression.");
```
rohanaggarwal7997, line 38: I enountered this when i was making changes. should be <= Decimal too

src/include/type/type_util.h
```diff
@@ -36,12 +36,14 @@ class TypeUtil {
...
         return 4;
       case TypeId::BIGINT:
-      case TypeId::DECIMAL:
+      case TypeId::REAL:
       case TypeId::TIMESTAMP:
         return 8;
       case TypeId::VARCHAR:
       case TypeId::VARBINARY:
         return storage::VARLEN_COLUMN;
+      case TypeId::DECIMAL:
```
rohanaggarwal7997, line 45: Just doing this probably might be wrong. You need to change the constructor of the SQL table too. I have already done that in my PR, so that might work.

src/include/catalog/index_schema.h
```diff
@@ -190,35 +207,44 @@ class IndexSchema {
...
 
    private:
-    static constexpr uint32_t MASK_VARLEN = 0x00FFFF00;
-    static constexpr uint32_t MASK_NULLABLE = 0x00000080;
-    static constexpr uint32_t MASK_TYPE = 0x0000007F;
-    static constexpr uint32_t OFFSET_VARLEN = 8;
+    bool ShouldHaveTypeModifier() const {
+      return (attr_length_ == storage::VARLEN_COLUMN &&
+              (type_ == type::TypeId::VARCHAR || type_ == type::TypeId::VARBINARY)) ||
+             (attr_length_ == 16 && type_ == type::TypeId::DECIMAL);
```
rohanaggarwal7997, line 213: why attr_length_ used in this 
shouldnt it being decimal be enough

src/catalog/postgres/pg_type_impl.cpp
```diff
@@ -111,7 +111,8 @@ void PgTypeImpl::BootstrapTypes(const common::ManagedPointer<DatabaseCatalog> db
   insert_base_type(type::TypeId::SMALLINT, "smallint", sizeof(int16_t));
   insert_base_type(type::TypeId::INTEGER, "integer", sizeof(int32_t));
   insert_base_type(type::TypeId::BIGINT, "bigint", sizeof(int64_t));
-  insert_base_type(type::TypeId::DECIMAL, "decimal", sizeof(double));
+  insert_base_type(type::TypeId::REAL, "real", sizeof(double));
+  insert_base_type(type::TypeId::DECIMAL, "decimal", sizeof(int128_t));
```
rohanaggarwal7997, line 115: just to confirm is this the same as 

InsertType(txn, type::TypeId::FIXEDDECIMAL, "fixeddecimal", postgres::NAMESPACE_CATALOG_NAMESPACE_OID, sizeof(int128_t), true,
             postgres::Type::BASE);

rohanaggarwal7997 (changes requested):
Minor comments and one clarification requested.

src/execution/sql/ddl_executors.cpp
mbutrovich: I think we punt that type of change until the end-to-end DECIMAL support PR, especially if you've made those changes already. I wouldn't be able to exercise them yet.

src/parser/expression/operator_expression.cpp
mbutrovich: I think we punt that type of change until the end-to-end DECIMAL support PR, especially if you've made those changes already. I wouldn't be able to exercise them yet.

src/include/type/type_util.h
mbutrovich: I think we punt that type of change until the end-to-end DECIMAL support PR, especially if you've made those changes already. I wouldn't be able to exercise them yet.

src/include/catalog/index_schema.h
jkosh44, line 213: If you do keep `attr_length` here, you may want to change `16` to `type::TypeUtil::GetTypeSize(type::TypeId::DECIMAL)` just so it's a bit clearer and also so it's easier to change the underlying type if we ever need to.

src/catalog/postgres/pg_core_impl.cpp
```diff
@@ -82,8 +82,9 @@ void PgCoreImpl::BootstrapPRIsPgAttribute() {
   pg_attribute_all_cols_prm_ = columns_->ProjectionMapForOids(pg_attribute_all_oids);
 
   const std::vector<col_oid_t> get_columns_oids{PgAttribute::ATTNUM.oid_,     PgAttribute::ATTNAME.oid_,
-                                                PgAttribute::ATTTYPID.oid_,   PgAttribute::ATTLEN.oid_,
+                                                PgAttribute::ATTTYPID.oid_,   PgAttribute::ATTTYPMOD.oid_,
                                                 PgAttribute::ATTNOTNULL.oid_, PgAttribute::ADSRC.oid_};
+  // TODO(Matt): Maybe length should still be read back out from the table rather than relying ton Column's constructor?
```
jkosh44, line 87: `VAR_ARRAY` will throw a runtime exception if we try and get the size (since there's no case in `TypeUtil::GetTypeSize`). I don't think that's an issue because none of our columns in pg_attribute are an array type, but postgres does have a couple of columns in pg_attributes that have array types (https://www.postgresql.org/docs/9.3/catalog-pg-attribute.html: attacl, attoptions, attfdwoptions)

Probably doesn't matter, but I thought I'd point it out.

jkosh44 (approved):

src/catalog/postgres/pg_type_impl.cpp
mbutrovich: I believe so, yeah.

rohanaggarwal7997 (approved):
LGTM

PR #1377, opened 2020-12-05
Port intCast TPL builtin.

### LINKED ISSUE #1049: [TPL2] Should we use 32 bit or 64 bit integers by default?

lmwnshn:
# Add intCast TPL builtin.

## Description

Fix #1049 by porting the `intCast` builtin from [pmenon/tpl](https://github.com/pmenon/tpl).

This adds casting functions for integers of different types, and logic to select the casting functions accordingly by specifying the `Kind` of an integer literal.

~~**This removes support for implicit casts.**~~

~~Running check-tpl locally, I suspect that tpl got a lot slower, possibly due to those 200+ macro-generated cast opcodes. Or maybe I just need to reboot. We'll see how it does in CI.~~ I think I just needed to reboot. Both the main branch and this PR are about 14 seconds in release.

---

This should help unblock an issue with #1370 not being able to specify 128 bit integers.

---

Some thoughts on porting difficulty:

- Porting this one feature from the tpl repo, ~~I started at 5am and finished now 7am including running tests and tidy. Not bad.~~ I proceeded to go down rabbit holes related to implicit casting being removed, since TPL is mostly concerned with TPL but we're concerned with SQL. Still down a couple of rabbit holes.
- However, to do it, I was reminded of the main difficulty with the TPL2 port: not knowing what actually changed. Most people won't know what upstream added vs we added vs got fixed in upstream vs got fixed by us. The diffs are bigger than they should be due to formatting changes, different namespaces and macro names, renaming of stuff (`Expr` -> `Expression`), clang-tidy differences (we enforce the ptr-or-const-ref rule), and functions that we may or may not need (we added `CheckArgCountBetween`).
- If I tried to diff the branches to figure out what to change, I probably would still be porting. Instead, Prashanth pointed me to `sema_checking.cpp` and `IsRepresentable` and it was enough to work backwards from there to figure out what dependencies I needed.
- I think if we hope to incorporate upstream changes in the future, a guide list of "look at this to bring in that" would be very helpful. The alternative would probably be another multi-month project of staring at big diffs.

## Remaining tasks

See if it gets through CI. If yes, flip to ready-to-review.


lmwnshn:
I forgot that I had this sitting out here. While the core functionality works, this removes implicit casting that the current tests rely on and which the translators do not handle. Will poke @pmenon for how to proceed.

mbutrovich:
The coverage on the diff is pretty rough. How do we beef this up with tests? Will end-to-end exercise this stuff?

lmwnshn:
Let's punt this discussion to TPL3.

PR #1383, opened 2020-12-08
Concurrent B+ Tree

abhijithanilkumar:
# B+ Tree for NoisePage

## Description

This PR introduces a concurrent B+ Tree that can potentially replace the existing BwTree implementation. The implementation uses latch crabbing, with per-node shared latches for synchronization. For reads, shared latches are acquired throughout the path. For writes, exclusive latches are acquired, and all parent latches are released if the current node is safe (no overflow/underflow possible).

There are two types of nodes:
* InnerNode: Non-leaf nodes in the Tree. Stores a key and the pointer to its child node. The node itself acts as a vector of `<key, ptr>` pairs. This vector is manually managed by allocating and deallocating memory whenever necessary.
* LeafNode: Leaves of the Tree, storing actual values. Stores a key and a list of values for each key. The list of values is currently implemented as `std::unordered_set`~`std::list` (An implicit drawback of this is that the lookups are O(n), can be improved to O(logn) when there are multiple values in the same key. Can be potentially improved further by switching to an unordered_set, since ordering is not important for these values)~.

`ScanKey`, `ScanAscending`, `ScanDescending` and `ScanLimitDescending` are supported. The scans will abort and retry if the latch on a particular node cannot be acquired (in order to make sure that this acquisition pattern does not deadlock with concurrent scans and inserts/deletes).

The initial code for the B+ Tree has been ported over from 15721 Project-2 (Spring 2020), implemented by @rohanaggarwal7997 @gautam20197 and @preetansh .

## Remaining tasks

- [x] Use binary search to search among values in the LeafNode (potentially switch to unordered_set)

EDIT: Even though `std::unordered_set` improves performance of the system, it takes up more memory. Experiments suggest that `std::list` uses less memory while taking up slightly more time. Therefore, switched back to `std::list`.

## Performance

Will be updating this section with the actual time and memory footprint after running OLTPBench on the B+ tree.


abhijithanilkumar:
Review Request:
* `src/include/storage/index/bplustree.h` : 
  * Tree Structure (1-910) - @thepinetree @kjobanputra [Since both of reviewers will need background regarding tree structure]
  * Insert Functions + ScanAscending ScanDescending + ScanLimitDescending - @kjobanputra (1326-1537, 1540-2188)
  * Delete Functions + ScanAscending - @thepinetree (1200-1312, 2196-2498)
  * Helper functions - @thepinetree (2511-2552)
  * Structural Integrity Verification - @kjobanputra (921-1184) [This code is specific to tests, verifying structural integrity]
  
* `test/storage/bplustree_test.cpp` : 
  * 1-800 - @thepinetree 
  * 801-1468 - @kjobanputra 

* Remaining Files: @turingcompl33t [The remaining files are a direct replacement to the equivalent bwtree files, so should be easy to review] 
  * `bplustree_benchmark` and `bplustree_index_test` are basically the same as `bwtree_benchmark` and `bwtree_index_test`, with function name changes specific to B+ Tree.
  * `bplustree_index.*` is analogous to `bwtree_index.*`. The major change is that scan functions in B+ Tree has a retry in case it fails because a lock cannot be acquired.
  * The remaining changes are pretty minor.

turingcompl33t (commented):
None of the comments in this review should block merging this PR, except perhaps the question raised in the implementation of `BPlusTreeIndex::ScanKey()` regarding duplicate keys spread across multiple leaf nodes. Will explicitly approve (for my files at least) once that question is addressed.

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+  ~BPlusTreeBase() = default;
+};
+
+/**
+ * Implementation of a B+ Tree index using latch crabbing.
+ *
+ * Concurrency:
+ *  Each node (leaf and inner) contains one shared latch to gain access to the node.
+ *  Writers are prioritized in this lock. Additionally, to guarantee safety of the root pointer of the
+ *  tree there is an additional latch for the root (root has 2 latches now). Latches are acquired from
```
kjobanputra, line 111: Can you write something more specifically about why the root pointer needs two latches for safety? It's not clear to me which case only having one latch for the root is unsafe.

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+     *
+     * This function takes an element type and copy-construct it on the array
+     * which is invisible to the compiler. Therefore we must call placement
+     * operator new to do the job
+     */
+    void PushBack(const ElementType &element) {
+      // Placement new + copy constructor using end pointer
+      new (end_) ElementType{element};
+
+      // Move it pointing to the enxt available slot, if not reached the end
```
kjobanputra, line 490: nit: typo (next instead of exnt)

src/include/storage/index/bplustree.h
kjobanputra: It's in the case where the root is being replaced, right? If so, it'd be nice to make that clearer in the comment.

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+
+    bool finished_insertion = false;
+    // We maintain the element that we have to recursively insert up.
+    // This is the element that has to be inserted into the inner nodes.
+    auto node = reinterpret_cast<ElasticNode<KeyValuePair> *>(current_node);
+
+    auto location_greater_key_leaf = static_cast<LeafNode *>(node)->FindLocation(element.first, this);
+    if (location_greater_key_leaf != node->Begin()) {
+      if (KeyCmpEqual((location_greater_key_leaf - 1)->first, element.first)) {
+        // Key present in tree => insert into value list
```
kjobanputra, line 1646: Maybe when talking about the structure of the tree state that we allow duplicate keys

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+        index_pointer -= 1;
+        current_node = index_pointer->second;
+      } else {
+        current_node = node->GetLowKeyPair().second;
+      }
+    }
+
+    auto node = reinterpret_cast<ElasticNode<KeyValuePair> *>(current_node);
+    for (KeyValuePair *element_p = node->Begin(); element_p != node->End(); element_p++) {
+      if (element_p->first == key) {
```
mush-zhang, line 802: should we use `KeyCmpEqual` here?

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+      all_nodes.pop();
+      if (node->GetType() != NodeType::LeafType) {
+        auto current_node = reinterpret_cast<ElasticNode<KeyNodePointerPair> *>(node);
+        heap_size += sizeof(BaseNode) + current_node->GetSize() * sizeof(KeyNodePointerPair);
+      } else {
+        auto current_node = reinterpret_cast<ElasticNode<KeyValuePair> *>(node);
+        heap_size += sizeof(BaseNode) + current_node->GetSize() * sizeof(KeyValuePair);
+        for (KeyValuePair *element_p = current_node->Begin(); element_p != current_node->End(); element_p++) {
+          if (element_p->second != nullptr) {
+            // destroy the list of values in a key
```
mush-zhang, line 1174: Misleading documentation copied from FreeTree?

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+  /**
+   * This test checks if duplicate keys are correctly handled by the insert wrapper.
+   * For a B+ tree, with each key having multiple values, stored in a map keys_values
+   * check if the values_list in the tree have the same values as the map given
+   * in the function.
+   *
+   * NOTE: This function does not acquire any latches, used by tests
+   */
+  bool DuplicateKeyValuesCheck(std::unordered_map<KeyType, std::set<ValueType>> *keys_values) {
+    auto itr = (*keys_values).begin();
```
kjobanputra, line 1069: Style nit: Keep in loop guard

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+  /**
+   * The check gets a map of keys to vector of values that are whetted and
+   * removes any duplicates that might have been tried to be added by a
+   * transaction in insert unique. The check ensures that the values list
+   * in the B+ tree also have the same keys and the same order.
+   *
+   * NOTE: This function does not acquire any latches, used by tests
+   */
+  bool DuplicateKeyValueUniqueInsertCheck(std::unordered_map<KeyType, std::vector<ValueType>> *keys_values) {
+    auto itr = (*keys_values).begin();
```
kjobanputra, line 1108: Same small bit: keep in loop guard

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+
+  /**
+   * Returns current heap usage (tries to return the exact heap usage by traversing
+   * the tree).
+   *
+   * NOTE: This function acquires a coarse lock, on the root, and prevents access to
+   * any other thread. Not ideal behavior, but this function is not commonly used. To
+   * get an approximate heap usage, EstimateHeapUsage() should be used.
+   */
+  size_t GetHeapUsage() {
```
kjobanputra, line 1138: Is there a reason that a coarse lock is used?

mush-zhang (changes requested):
Generally LGTM.Very clear code and documentation. Thank you! I have left some minor comments and questions. You could probably add some testcases covering those edge cases pointed out by codecov. Another thing: Do we want the test functions in the same bplus tree header file? we could create a TestBplusTree class for those functions and put them into a separate file.

test/storage/bplustree_test.cpp
```diff
@@ -0,0 +1,1468 @@
...
+    BPlusTree<int, TupleSlot>::KeyElementPair p1;
+    p1.first = *(it);
+    it++;
+    EXPECT_EQ(bplustree->IsPresent(p1.first), true);
+  }
+
+  delete bplustree;
+}
+
+void StructuralIntegrityTestWithRandomInsertAndDelete2Reverse() {
```
kjobanputra, line 1218: How is this different than the test above?

test/storage/bplustree_test.cpp
```diff
@@ -0,0 +1,1468 @@
...
+  };
+
+  // Run the workload
+  for (uint32_t i = 0; i < num_threads_; i++) {
+    thread_pool_.SubmitTask([i, &workload] { workload(i); });
+  }
+  thread_pool_.WaitUntilAllFinished();
+
+  EXPECT_EQ(tree->GetSize(), key_num);
+
```
kjobanputra, line 1401: Also include the structural integrity check

test/storage/bplustree_test.cpp
```diff
@@ -0,0 +1,1468 @@
...
+  };
+
+  // Run the workload
+  for (uint32_t i = 0; i < num_threads_; i++) {
+    thread_pool_.SubmitTask([i, &workload] { workload(i); });
+  }
+  thread_pool_.WaitUntilAllFinished();
+
+  EXPECT_EQ(tree->GetSize(), deleted_keys);
+
```
kjobanputra, line 1456: Same comment as above: add structural integrity check

kjobanputra (changes requested):
My comments are mostly nitpicking: some commenting changes, a question about a design decision, and including calls to check the structural integrity of the tree in the later tests. Once those are addressed, I'll approve. Good job overall!

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+     * that the iterator begin and end can fetch the values easily.
+     */
+    const KeyNodePointerPair *low_key_p_;  /*!< Low Key */
+    const KeyNodePointerPair *high_key_p_; /*!< High Key */
+
+    /** The type of the node */
+    NodeType type_;
+
+    /** This is the height of the node */
+    int depth_;
```
tpan496, line 220: Would it be better to comment as "The depth of the node within the tree" instead of using height?

test/storage/bplustree_test.cpp
```diff
@@ -0,0 +1,1468 @@
...
+  void TearDown() override { thread_pool_.Shutdown(); }
+};
+
+void BasicNodeInitializationInsertReadAndFreeTest() {
+  auto bplustree = new BPlusTree<int, TupleSlot>;
+  BPlusTree<int, TupleSlot>::KeyNodePointerPair p1;
+  BPlusTree<int, TupleSlot>::KeyNodePointerPair p2;
+
+  // Get inner Node
+  auto node = BPlusTree<int, TupleSlot>::ElasticNode<BPlusTree<int, TupleSlot>::KeyNodePointerPair>::Get(
```
tpan496, line 31: Would it be better to create local variables for those magic numbers to improve readability?

test/storage/bplustree_test.cpp
```diff
@@ -0,0 +1,1468 @@
...
+void InsertElementInNodeTest() {
+  auto bplustree = new BPlusTree<int, TupleSlot>;
+  BPlusTree<int, TupleSlot>::KeyNodePointerPair p1;
+  BPlusTree<int, TupleSlot>::KeyNodePointerPair p2;
+
+  // Get inner Node
+  auto node = BPlusTree<int, TupleSlot>::ElasticNode<BPlusTree<int, TupleSlot>::KeyNodePointerPair>::Get(
+      10, BPlusTree<int, TupleSlot>::NodeType::LeafType, 0, 10, p1, p2);
+
+  for (unsigned i = 0; i < 10; i++) {
```
tpan496, line 77: Similar comment to the one above. I think it would be better to name "i" to something more meaningful.

test/storage/bplustree_test.cpp
```diff
@@ -0,0 +1,1468 @@
...
+  const uint32_t num_threads_ = 4;
+  common::WorkerPool thread_pool_{num_threads_, {}};
+
+ protected:
+  void SetUp() override { thread_pool_.Startup(); }
+
+  void TearDown() override { thread_pool_.Shutdown(); }
+};
+
+void BasicNodeInitializationInsertReadAndFreeTest() {
```
tpan496, line 25: Maybe some brief comments on what each test does?

test/storage/bplustree_test.cpp
```diff
@@ -0,0 +1,1468 @@
...
+  auto predicate = [](const TupleSlot slot) -> bool { return false; };
+
+  // Insert keys
+  auto bplustree = new BPlusTree<int, TupleSlot>;
+  bplustree->SetInnerNodeSizeUpperThreshold(5);
+  bplustree->SetLeafNodeSizeUpperThreshold(5);
+  std::set<int> keys;
+  for (unsigned i = 0; i < 100000; i++) {
+    BPlusTree<int, TupleSlot>::KeyElementPair p1;
+    int k = rand_r(&globalseed) % 500000;
```
tpan496, line 519: Similar style comments apply here

test/storage/bplustree_test.cpp
```diff
@@ -0,0 +1,1468 @@
...
+    p1.first = i;
+    bplustree->DeleteElement(p1);
+    EXPECT_EQ(bplustree->IsPresent(i), false);
+  }
+  EXPECT_EQ(bplustree->GetRoot() == nullptr, true);
+
+  delete bplustree;
+}
+
+void LargeKeySequentialInsertAndRetrievalTest() {
```
tpan496, line 677: Can we comment on why this is considered a "Larger" test? The numbers do seem large, but I think it would be more clear to clarify this.

test/storage/bplustree_test.cpp
```diff
@@ -0,0 +1,1468 @@
...
+    EXPECT_EQ(bplustree->IsPresent(*iter), true);
+    iter++;
+  }
+
+  // Free Everything
+
+  delete bplustree;
+}
+
+void StructuralIntegrityTestWithRandomInsertAndDelete2() {
```
tpan496, line 1177: Can we name it in a more meaningful way instead of just adding 2 at the end?

benchmark/storage/bplustree_benchmark.cpp
```diff
@@ -0,0 +1,288 @@
...
+};
+
+// NOLINTNEXTLINE
+BENCHMARK_DEFINE_F(BPlusTreeBenchmark, RandomInsert)(benchmark::State &state) {
+  common::WorkerPool thread_pool(BenchmarkConfig::num_threads, {});
+  thread_pool.Startup();
+
+  // NOLINTNEXTLINE
+  for (auto _ : state) {
+    auto *const tree = new storage::index::BPlusTree<int64_t, int64_t>;
```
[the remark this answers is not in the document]
abhijithanilkumar: Done

benchmark/storage/bplustree_benchmark.cpp
```diff
@@ -0,0 +1,288 @@
...
+      }
+    };
+
+    uint64_t elapsed_ms;
+    {
+      common::ScopedTimer<std::chrono::milliseconds> timer(&elapsed_ms);
+      MultiThreadTestUtil::RunThreadsUntilFinish(&thread_pool, BenchmarkConfig::num_threads, workload);
+    }
+    delete tree;
+    state.SetIterationTime(static_cast<double>(elapsed_ms) / 1000.0);
```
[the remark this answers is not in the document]
abhijithanilkumar: I believe all the microbenchmarks use this division as of now, we can probably make the change to ScopedTimer and change it everywhere, but I'm not sure if that is worth the effort.

src/include/storage/index/bplustree_index.h
```diff
@@ -0,0 +1,139 @@
...
+  void Delete(common::ManagedPointer<transaction::TransactionContext> txn, const ProjectedRow &tuple,
+              TupleSlot location) final;
+
+  /**
+   * Finds all the values associated with the given key in our index.
+   * @param txn txn context for the calling txn, used for visibility checks
+   * @param key the key to look for
+   * @param[out] value_list the values associated with the key
+   */
+  void ScanKey(const transaction::TransactionContext &txn, const ProjectedRow &key,
```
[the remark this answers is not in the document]
abhijithanilkumar: I feel you! This interface will be removed in the near future when we can stop materializing keys here. Instead use an index iterator in the execution engine.

src/include/storage/index/bplustree_index.h
```diff
@@ -0,0 +1,139 @@
...
+  /**
+   * @return type of the index. Note that this is the physical type, not extracted from the underlying schema or other
+   * catalog metadata. This is mostly used for debugging purposes.
+   */
+  IndexType Type() const final { return IndexType::BPLUSTREE; }
+
+  /**
+   * Invoke garbage collection on the index.
+   */
+  void PerformGarbageCollection() final;
```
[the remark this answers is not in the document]
abhijithanilkumar: This function is implemented in the generic index wrapper and a change here will have a bigger blast radius. This will need changes to the corresponding implementations in BwTree and Hash index. I'll keep this in mind and make the change in the next PR removing the BwTree.

src/include/storage/index/bplustree_index.h
```diff
@@ -0,0 +1,139 @@
...
+  size_t EstimateHeapUsage() const final;
+
+  /**
+   * Inserts a new key-value pair into the index, used for non-unique key indexes.
+   * @param txn txn context for the calling txn, used to register abort actions
+   * @param tuple key
+   * @param location value
+   * @return false if the value already exists, true otherwise
+   */
+  bool Insert(common::ManagedPointer<transaction::TransactionContext> txn, const ProjectedRow &tuple,
```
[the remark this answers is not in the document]
abhijithanilkumar: This change will require modification to all the index implementations. We will probably need a broader discussion on whether we should make this change.

src/include/storage/index/index_builder.h
```diff
@@ -33,6 +33,10 @@ class IndexBuilder {
 
   Index *BuildBwTreeGenericKey(IndexMetadata metadata) const;
 
+  Index *BuildBPlusTreeIntsKey(IndexMetadata metadata) const;
```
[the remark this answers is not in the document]
abhijithanilkumar: I've made the change for B+ Tree, not changing the other index implementations, we can probably do this in a later PR.

src/storage/index/bplustree_index.cpp
```diff
@@ -0,0 +1,217 @@
...
+
+#include "storage/index/compact_ints_key.h"
+#include "storage/index/generic_key.h"
+#include "transaction/deferred_action_manager.h"
+#include "transaction/transaction_context.h"
+
+namespace noisepage::storage::index {
+
+template <typename KeyType>
+BPlusTreeIndex<KeyType>::BPlusTreeIndex(IndexMetadata metadata)
```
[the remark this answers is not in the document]
abhijithanilkumar: Done

src/storage/index/bplustree_index.cpp
```diff
@@ -0,0 +1,217 @@
...
+    deferred_action_manager->RegisterDeferredAction([=]() {
+      const bool UNUSED_ATTRIBUTE result = bplustree_->DeleteElement(bplustree_->GetElement(index_key, location));
+
+      NOISEPAGE_ASSERT(result, "Deferred delete on the index failed.");
+    });
+  });
+}
+
+template <typename KeyType>
+void BPlusTreeIndex<KeyType>::ScanKey(const transaction::TransactionContext &txn, const ProjectedRow &key,
```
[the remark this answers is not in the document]
abhijithanilkumar: It's not possible to have multiple keys with the same value at the leaf level. Each key can hold multiple values though. So, when we scan for a key, it is essentially a point lookup and we are guaranteed that the key that pops out of the search is the only one present.

src/include/storage/index/bplustree.h
abhijithanilkumar: Done

src/include/storage/index/bplustree.h
abhijithanilkumar: Done

src/include/storage/index/bplustree.h
abhijithanilkumar: Done

src/include/storage/index/bplustree.h
abhijithanilkumar: Done

src/include/storage/index/bplustree.h
abhijithanilkumar: This function returns the **exact** heap usage. So technically, if we use fine-grained locking, the measurement will not be accurate since the tree can change dynamically. This function sort of captures the current snapshot. For estimates of heap usage, EstimateHeapUsage() function should be called.

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+ *                          Returns true for ValueTypes that are equal
+ */
+template <typename KeyType, typename ValueType, typename KeyComparator = std::less<KeyType>,
+          typename KeyEqualityChecker = std::equal_to<KeyType>,
+          typename ValueEqualityChecker = std::equal_to<ValueType>>
+class BPlusTree : public BPlusTreeBase {
+ public:
+  class BaseNode;
+
+  /** <KeyType, NodeID> pair */
```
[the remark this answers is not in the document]
abhijithanilkumar: Done

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+  using KeyNodePointerPair = std::pair<KeyType, BaseNode *>;
+  /** <KeyType, List of ValueType> pair */
+  using KeyValuePair = std::pair<KeyType, std::list<ValueType> *>;
+  /** <KeyType, ValueType> pair */
+  using KeyElementPair = std::pair<KeyType, ValueType>;
```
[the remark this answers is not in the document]
abhijithanilkumar: Done

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+      }
+    }
+
+   public:
+    /**
+     * InsertElementIfPossible - Returns true if inserted and false if node full
+     * Inserts at location provided.
+     */
+    bool InsertElementIfPossible(const ElementType &element, ElementType *location) {
+      if (GetSize() >= this->GetItemCount()) return false;
```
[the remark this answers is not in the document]
abhijithanilkumar: Done

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+      if (end_ - location > 0)
+        std::memmove(reinterpret_cast<void *>(location + 1), reinterpret_cast<void *>(location),
+                     (end_ - location) * sizeof(ElementType));
+      new (location) ElementType{element};
+      end_ = end_ + 1;
```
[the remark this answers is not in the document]
abhijithanilkumar: This is not a `PushBack` because this is to insert an element at a specific location while `PushBack` always inserts it at the end.

src/include/storage/index/bplustree.h
```diff
@@ -0,0 +1,2554 @@
...
+    }
+
+    /**
+     * Erase - remove the element at index 'i' from the list
+     * Returns False if empty
+     */
+    bool Erase(int i) {
+      if (this->GetSize() <= i) return false;
+
+      if (this->GetSize() == 1) {
```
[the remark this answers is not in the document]
abhijithanilkumar: This is a check to verify that the index is in bounds. There are some parts of delete code that relies on this check.

src/include/storage/index/bplustree.h
abhijithanilkumar: Done

src/include/storage/index/bplustree.h
abhijithanilkumar: Modified

test/storage/bplustree_test.cpp
abhijithanilkumar: This test deletes elements in the reverse order. Will include some meaningful documentation on what each test does while addressing @tpan496's comments below.

test/storage/bplustree_test.cpp
abhijithanilkumar: Done

test/storage/bplustree_test.cpp
abhijithanilkumar: Done

src/storage/index/bplustree_index.cpp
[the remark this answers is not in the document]
turingcompl33t: Great, thanks for the clarification!

turingcompl33t (approved):
All of the points in my initial review were addressed. Great work @abhijithanilkumar!

kjobanputra (approved):

mush-zhang (approved):
LGTM

src/include/storage/index/bplustree.h
abhijithanilkumar: Done

test/storage/bplustree_test.cpp
abhijithanilkumar: Done

test/storage/bplustree_test.cpp
abhijithanilkumar: Done

test/storage/bplustree_test.cpp
abhijithanilkumar: Done

test/storage/bplustree_test.cpp
abhijithanilkumar: Done

test/storage/bplustree_test.cpp
abhijithanilkumar: Done

test/storage/bplustree_test.cpp
abhijithanilkumar: Removed these 2 tests since they were redundant.

tpan496:
LGTM

lmwnshn (approved):
Rubber stamping, see `#pr-czar` in slack.

PR #1387, opened 2020-12-09
Implement a hacky version of SHOW for transaction_isolation.

lmwnshn:
# Implement a hacky version of SHOW for transaction_isolation.

## Description

This adds "support" for SHOW TRANSACTION ISOLATION LEVEL, which postgres translates as SHOW transaction_isolation.

This does **NOT** implement a general-purpose SHOW, which should talk to the settings manager.
Nor does it reconcile the fact that some SET and SHOW should be transactional (as found by @Poojita-Raj).

This is purely to unblock @AlanDZQ for his oltpbench stuff.
The rest of the work is left for #1188.

## Remaining tasks

If it gets through CI, then ready for review.

### DIFF
--- a/src/include/binder/sql_node_visitor.h
+++ b/src/include/binder/sql_node_visitor.h
@@ -21,6 +21,7 @@ class UpdateStatement;
 class CopyStatement;
 class AnalyzeStatement;
 class VariableSetStatement;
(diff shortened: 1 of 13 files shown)


apavlo (approved):
LGTM

PR #1390, opened 2020-12-10
Add ant version pin and openjdk11 to Mac.

lmwnshn:
# Heading

Add `ant` version pin and openjdk11 to Mac.

## Description

Latest version of `ant` has an `openjdk15` dependency.

This screws up compiling and running oltpbench.

If this works, I need to update the mac boxes `~/.profile`.


lmwnshn:
Obviated by @crd477 doing some jdk magic on the build machines.

PR #1392, opened 2020-12-14
Fix OpUnit that was renamed, causing modeling_test failures.

githubxxcc:
This PR fixes a bug in the `modeling_test` brought in by previous PR #1288:

The `OP_DECIMAL_COMPARE` was renamed in an earlier PR that wasn't incorporated into 1288 while merging.


test/self_driving/modeling_test.cpp
```diff
@@ -83,7 +83,7 @@ TEST_F(ModelServerTest, PipelineTest) {
   auto result = ms_manager->DoInference("OP_INTEGER_PLUS_OR_MINUS", save_path, features);
   ASSERT_TRUE(result.second);
   ASSERT_EQ(result.first.size(), features.size());
-  result = ms_manager->DoInference("OP_DECIMAL_COMPARE", save_path, features);
+  result = ms_manager->DoInference("OP_REAL_COMPARE", save_path, features);
```
malin1993ml, line 86: nit: in fact, can you actually use `OperatingUnitUtil::ExecutionOperatingUnitTypeToString` to convert `ExecutionOperatingUnitType::OP_REAL_COMPARE` to string? I think this would be a bit more stable against changes. The same applies to other valid operating unit types in this test.

githubxxcc: Oh yes, I did not know this function exist. This is the right way.

test/self_driving/modeling_test.cpp
```diff
@@ -38,14 +39,23 @@ class ModelServerTest : public TerrierTest {
 
     return db_main;
   }
+
+  /**
+   * Wrapper to hide long chain of function name resolution
+   * @param unit_type OpUnit
+   * @return  string representation of the opunit
+   */
+  static const std::string opunit_string(selfdriving::ExecutionOperatingUnitType unit_type) {
```
malin1993ml, line 48: nit: based on our naming convention, this should be `OpUnitToString`.

githubxxcc: Err right, what was I thinking...too much python/c

malin1993ml (approved):
LGTM!

Issue #1380, opened 2020-12-07
Fixed length VARCHAR isn't enforced


mbutrovich:
# Bug Report
## Summary
We don't check whether incoming VARCHAR values adhere to the length limit.

### Expected Behavior
Postgres:
```SQL
matt=# create table foo (name varchar(3));
CREATE TABLE
Time: 14.160 ms
matt=# insert into foo values ('andy');
ERROR:  value too long for type character varying(3)
Time: 1.514 ms
```

### Actual Behavior
NoisePage:
```SQL
noisepage=# create table foo (name varchar(3));
CREATE TABLE
Time: 126.967 ms
noisepage=# insert into foo values ('andy');
INSERT 0 1
Time: 64.881 ms
```
### Proposed Solution
I suspect this should be resolved in the binder. It has access to the query arguments (or prepared statement arguments) and access to the Catalog to check the schema. We should confirm what stage Postgres enforces this at if possible, and then return the same error code.

Issue #1158, opened 2020-09-06
INSERT INTO SELECT


tanujnay112:
`Insert into select` currently doesn't insert anything and does not generate code to do so for a few reasons.
1) InsertTranslator doesn't call prepare on its children if such children exist.

2) The InsertPlanNode needs to have DerivedValueExpressions in its values_ vector to signify tuples that are derived from its children for insertion. This should be done in `plan_generator.cpp`. Otherwise, there is nothing in the values_ vector during `INSERT INTO SELECT` and the insert translator generates nothing.

3) Also, in order for the children of the insert node to pass up tuples the InsertSelect node needs to request the required columns from the descendant scan node [here](https://github.com/cmu-db/terrier/blob/d60db2a543eed2c0463d2fae02eb4bc39628b4f8/src/optimizer/input_column_deriver.cpp#L233). Currently this just passes down required columns from above which is incorrect. You can't request columns from an Insert node. Also, at this point we must generate the required columns to request from the children. In the event that we have a `INSERT INTO yyy SELECT * FROM xxx;`, we can easily just query all the columns oids in xxx and request those columns. The tableoid of xxx is stored in the child of the concerned InsertSelect node.

4) The information stored in InsertSelect is incomplete. While the vanilla Insert physical operator node has [this](https://github.com/cmu-db/terrier/blob/e525b6ddde0598019d2a8d58e2d8907d0d3051b1/src/include/optimizer/physical_operators.h#L896) that allows us to know which columns we are inserting into, the InsertSelect operator has no equivalent even though this information is necessary.

PR #1394, opened 2020-12-16
Enforce fixed length VARCHAR insertions

eppingere:
## Description

Fix for #1380 

Add check to the binder that ensures that constant values inserted into fixed-length varchar columns, must fit within the columns. Check is in `bind_visitor_node.cpp` since is where we type check the inserted values and see if they can be cast to their column types. Implemented the error in the same form as [Postgres](https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/varchar.c#L634). 

## Future work
I'm not sure if this is the case, but since without this check it appears that such an insertion succeeds, do we have an issue if the values we insert are from a subquery or are computed and are larger than the column, we might not be checking for that either. This doesn't seem to work however since `INSERT INTO foo SELECT stuff FROM bar;` doesn't do anything (I think this is #1158?) I haven't been able to confirm this. Postgres is able to handle this when the values fit in the column and throws errors when the values from the select do not fit in the column.


src/binder/bind_node_visitor.cpp
```diff
@@ -440,6 +441,43 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::InsertStatement> node
           }
 
           ins_val->Accept(common::ManagedPointer(this).CastManagedPointerTo<SqlNodeVisitor>());
+
+          // if we have constant values of variable types being inserted into fixed size columns,
+          // we need to resolve that
+          if (ins_val->GetExpressionType() == parser::ExpressionType::VALUE_CONSTANT) {
+            auto const_val = ins_val.CastManagedPointerTo<parser::ConstantValueExpression>();
+            bool is_variable = const_val->GetReturnValueType() == type::TypeId::VARCHAR ||
+                               const_val->GetReturnValueType() == type::TypeId::VARBINARY;
```
mbutrovich, line 450: Does it make sense to apply this logic to a `VARBINARY`?

src/binder/bind_node_visitor.cpp
```diff
@@ -440,6 +441,43 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::InsertStatement> node
...
+            if (is_variable && !const_val->IsNull() && ins_col.TypeModifier() != -1 &&
+                const_val->GetStringVal().GetLength() > static_cast<size_t>(ins_col.TypeModifier())) {
+              // postgres truncates off trailing spaces from the end of the varchar only to the max size of the col
+              // if it is still too large it throws an error
+              // https://github.com/postgres/postgres/blob/master/src/backend/utils/adt/varchar.c
+              // issue #1380
+              size_t num_trailing_spaces = 0;
+              for (size_t idx = const_val->GetStringVal().GetLength() - 1;
+                   idx < const_val->GetStringVal().GetLength() && idx >= static_cast<size_t>(ins_col.TypeModifier());
```
mbutrovich, line 459: Why `idx < const_val->GetStringVal().GetLength()`? Can wraparound happen?

src/binder/bind_node_visitor.cpp
```diff
@@ -440,6 +441,43 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::InsertStatement> node
           }
 
           ins_val->Accept(common::ManagedPointer(this).CastManagedPointerTo<SqlNodeVisitor>());
+
+          // if we have constant values of variable types being inserted into fixed size columns,
+          // we need to resolve that
+          if (ins_val->GetExpressionType() == parser::ExpressionType::VALUE_CONSTANT) {
+            auto const_val = ins_val.CastManagedPointerTo<parser::ConstantValueExpression>();
+            bool is_variable = const_val->GetReturnValueType() == type::TypeId::VARCHAR ||
```
mbutrovich, line 449: Can we stash `ins_col.TypeModifier()` in a variable up here called `max_length`? It'll clean up the logic below a lot rather than the repeated called and make it more clear what that value is.

src/binder/bind_node_visitor.cpp
```diff
@@ -440,6 +441,43 @@ void BindNodeVisitor::Visit(common::ManagedPointer<parser::InsertStatement> node
...
+              size_t num_trailing_spaces = 0;
+              for (size_t idx = const_val->GetStringVal().GetLength() - 1;
+                   idx < const_val->GetStringVal().GetLength() && idx >= static_cast<size_t>(ins_col.TypeModifier());
+                   idx--) {
+                if (const_val->GetStringVal().GetContent()[idx] == ' ')
+                  num_trailing_spaces++;
+                else
+                  break;
+              }
+              size_t true_len = const_val->GetStringVal().GetLength() - num_trailing_spaces;
```
mbutrovich, line 466: `true_length` is fine. We try not to abbreviate variable names much.

mbutrovich (commented):
Minor changes so far. Thanks!

src/binder/bind_node_visitor.cpp
eppingere: yeah although you need to throw a different error in the `VARBINARY` case and not truncate

src/binder/bind_node_visitor.cpp
eppingere: Yeah if it wraps around then we have the whole thing as spaces? It can only happen if we have a size 0 column which idk if that can happen but will check.

src/binder/bind_node_visitor.cpp
eppingere: yep

src/binder/bind_node_visitor.cpp
eppingere: done

src/binder/bind_node_visitor.cpp
mbutrovich: Size 0 should be impossible. I think I tried in Postgres `VARCHAR(0)` and it yells at you.

eppingere: done

eppingere: We also do not handle that error that postgres throws, I added that to the pr

src/traffic_cop/traffic_cop.cpp
```diff
@@ -313,6 +313,9 @@ std::variant<std::unique_ptr<parser::ParseResult>, common::ErrorData> TrafficCop
                             common::ErrorCode::ERRCODE_SYNTAX_ERROR);
     error.AddField(common::ErrorField::POSITION, std::to_string(e.GetCursorPos()));
     result.emplace<common::ErrorData>(std::move(error));
+  } catch (const BinderException &e) {
```
mbutrovich, line 316: I don't love that it's a BinderException because we're at the parsing stage, but I see why you did it since you want a different error code and you don't have the cursor position. Is there any other approach you can think of?

eppingere: I can't really think of anything else. TBH i did this bug to learn about the parser and binder and things but I am totally not opposed to anything else. Do you think I should put a check that it is that exact binder exception?

mbutrovich:
In looking at this more, I think this should be handled in the binder's visit of a `ConstantValueExpression`. The problem with that is we don't have the schema anymore in that scope. We could stash metadata about the type modifier in the `BinderSherpa`, but that might be temporary fix until if/when we get a fat type support that @rohanaggarwal7997 was discussing for his decimal precision stuff. We should decide as a group what we think the best approach is.

apavlo:
Closing this because we need to push data validation into the EE.

mbutrovich:
Specifically: INSERT INTO SELECT could still violate this.

PR #1395, opened 2020-12-16
Add calls to lsof in Jenkins.

lmwnshn:
# Heading

Add calls to lsof in Jenkins.

## Description

Before and after everything that might spawn a NoisePage instance.

This is to catch any stray processes leftover from before our run, or lingering after our run.


lmwnshn:
Ok, the man page doesn't lie about `grep -q`, I simply can't read.
Though the behavior is unintuitive to me.

PR #1403, opened 2020-12-18
Workload forecasting script 

githubxxcc:
This script (and relevant changes) : 
- runs the OLTP bench to generate query trace
- perform simple LSTM training/testing on the query trace timeseries data.


script/model/forecast.py
```diff
@@ -0,0 +1,520 @@
+#!/usr/bin/env python3
```
malin1993ml, line 1: Where is this script executed? We need to add this test to the `End-to-End Self-Driving Test` stage. Probably as a separate sub test stage as the current modeling test.

script/model/forecast.py
```diff
@@ -0,0 +1,520 @@
...
+    plt.grid(True)
+    plt.autoscale(axis='x', tight=True)
+
+    for serie in series:
+        label, x, data = serie
+        plt.plot(x, data, label=label)
+    plt.legend()
+
+
+def gen_forecast_data(xml_config_file: str, rate_pattern: List[int]) -> None:
```
malin1993ml, line 100: This is just setting up the config file instead of actually generating the data. So I would rename this function.

script/model/forecast.py
```diff
@@ -0,0 +1,520 @@
...
+    def __init__(
+            self,
+            query_trace_file: str = DEFAULT_QUERY_TRACE_FILE,
+            test: bool = False,
+            eval_size: int = EVAL_DATA_SIZE,
+            seq_len: int = SEQ_LEN,
+            horizon_len: int = HORIZON_LEN) -> None:
+        """
+        :param query_trace_file: Where to process the query trace
+        :param test: If the Loader is for testing
```
malin1993ml, line 200: nit: `test` -> `test_mode`

script/model/forecast.py
```diff
@@ -0,0 +1,520 @@
+#!/usr/bin/env python3
+"""
+TODO(ricky):
+The modeling scripts should be under a project, with __init__
+What should be the best dependency between the modeling scripts and testing package?
```
malin1993ml, line 5: You're right that there should be a separation between the forecasting script and the testing script. Basically, the `DataLoader` and `LSTM` classes should be under `script/forecasting`. And everything else should be under `script/testing/forecasting`, where we invoke the test script for the Jenkins pipeline.

malin1993ml (changes requested):
First pass. Haven't gone to the details for `DataLoader` and `LSTM`

script/model/forecast.py
```diff
@@ -0,0 +1,520 @@
...
+        trace_start_ms = self.data[0][1]
+        trace_end_ms = self.data[-1][1]
+        trace_nsample = len(self.data)
+
+        print(
+            f"{trace_nsample} samples from {trace_start_ms} to {trace_end_ms}\n"
+            f"{(trace_end_ms - trace_start_ms) / 1000000}seconds")
+
+
+class LSTM(nn.Module):
```
malin1993ml, line 328: Create another `Model` wrapper class similar to `script/model/model.py` with `LSTM` only as one implementation option.  `LSTM` should just provide the simple sklearn style `fit(x, y)` and `predict(x)` interface. This way it's very easy to extend to other models.

script/model/forecast.py
```diff
@@ -0,0 +1,520 @@
...
+            data = np.array(
+                [[int(r['query_id']), int(r[' timestamp'])] for r in reader])
+
+            data = self._aggregate_data(data, INTERVAL_MS)
+            self.train_raw_data, self.test_raw_data = self._split_data(data)
+            LOG.info(
+                f"Loaded data successfully: num_test={len(self.test_raw_data)}, num_train={len(self.train_raw_data)}, "
+                f"seq_len={self.seq_len}, horizon={self.horizon_len}")
+
+    def _normalize_data(self, data: np.ndarray) -> np.ndarray:
```
malin1993ml, line 229: `_normalize_data` should belong to the `Model` class, as discussed below. Again, look at `script/model/model.py` for reference.

script/model/forecast.py
```diff
@@ -0,0 +1,520 @@
...
+        Forward propogation
+        :param input_seq:  1D FloatTensor
+        :return: A single value prediction
+        """
+        lstm_out, self.hidden_cell = self.lstm(
+            input_seq.view(len(input_seq), 1, -1), self.hidden_cell)
+        predictions = self.linear(lstm_out.view(len(input_seq), -1))
+        return predictions[-1]
+
+    def do_train(self, data_loader: DataLoader,
```
malin1993ml, line 360: Don't pass the `DataLoader` class here. Instead, generate the data outside and directly pass the data in here.

script/model/forecast.py
```diff
@@ -0,0 +1,520 @@
...
+                single_loss = loss_function(y_pred, labels)
+                single_loss.backward()
+                optimizer.step()
+
+            if i % 25 == 0:
+                LOG.info(f'epoch: {i+1:3} loss: {single_loss.item():10.8f}')
+
+        LOG.info(f'epoch: {epochs:3} loss: {single_loss.item():10.10f}')
+
+    def do_test(self, data_loader: DataLoader) -> List[float]:
```
malin1993ml, line 393: Similarly, generate the data outside.

script/model/forecast.py
```diff
@@ -0,0 +1,520 @@
...
+        norm_data = self.scaler.fit_transform(data.reshape(-1, 1))
+        return norm_data
+
+    def _aggregate_data(self, data: np.ndarray,
+                        interval_ms: int) -> Optional[np.ndarray]:
+        """
+        Aggregate multiple time-series data points into one bucket to generate a new time-series
+        data set.
+        It assumes the original time-series data has a Timestamp value, and the generated time-series
+        will be the number of data points fall into an interval for each interval in the original dataset
```
malin1993ml, line 239: Okay, so what we need is slightly different than this. Instead of aggregating all queries together right now, we should generate one time-series for each query (`query_id`). It's only when doing the training and inference we're going to further aggregate all the time-series together (assuming that all queries belong to the same cluster). But after the model inference, we need to split the single time-series prediction back to the predictions for each query, based on the ratio between the total occurrences of these queries in the training data. In other words, even though the model only predicts one time-series, we should generate the same number of time-series predictions as the number of queries at the end. We probably should have a `Forecaster` class that handles this logic. Is this explanation clear?

malin1993ml (changes requested):
The `DataLoader` and `LSTM` classes overall look good! There're a few re-structurings that we need to do.

script/model/forecast.py
githubxxcc: So something like this: 
A query trace file => multiple time-series of certain interval for each query id => aggregate multiple time-series into one time-series for each cluster and note down the ratio => for each cluster, a model will produce a prediction => convert the prediction into multiple time-series for different query id according to the ratio. 

And the ratio would just be the total number of events of a single query id over all events in the cluster? (Assuming queries in the same cluster would have the same pattern)

malin1993ml: Yes! Your summary is exactly right!

bialesdaniel (commented):
I am not very familiar with what the models are doing so take my comments with a grain of salt. Overall looks good. I mostly had some stylistic comments which you can feel free to ignore.

Jenkinsfile
```diff
@@ -418,45 +418,88 @@ pipeline {
...
-                docker {
-                    image 'noisepage:focal'
-                    args '--cap-add sys_ptrace -v /jenkins/ccache:/home/jenkins/.ccache'
-                }
-            }
-            steps {
-                sh 'echo $NODE_NAME'
+        stage('Self-Driving') {
+            parallel {
+                stage('Workload Forecasting'){
```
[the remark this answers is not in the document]
githubxxcc: It runs the OLTPbench to generate trace + train some models on the trace + does some dummy testing.

script/forecasting/forecaster.py
```diff
@@ -0,0 +1,292 @@
...
+        Split the raw data into a training set, and a testing(evaluation) set.
+        :param data: All the raw data
+        :return: traing, test raw data set
+        """
+        if self.test_mode:
+            self.test_set_size = len(data)
+        else:
+            self.test_set_size = self.eval_data_size
+
+        split_idx = len(data) - self.test_set_size
```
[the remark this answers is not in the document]
githubxxcc: good catch!

script/testing/self_driving/forecast.py
```diff
@@ -0,0 +1,135 @@
...
+    DEFAULT_OLTP_TEST_CASE,
+    DEFAULT_WORKLOAD_PATTERN)
+
+import psycopg2 as psql
+from pathlib import Path
+from xml.etree import ElementTree
+from typing import Dict, List, Optional, Tuple
+
+
+def config_forecast_data(xml_config_file: str, rate_pattern: List[int]) -> None:
```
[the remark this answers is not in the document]
githubxxcc: Oh cool, didn't know that. What's the function you are referring to?

script/testing/self_driving/forecast.py
```diff
@@ -0,0 +1,135 @@
...
+            work.append(elem)
+
+        works.append(work)
+
+    # Write back result
+    xml.write(xml_config_file)
+
+
+def gen_oltp_trace(
+        tpcc_weight: str, tpcc_rates: List[int], pattern_iter: int) -> bool:
```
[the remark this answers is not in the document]
githubxxcc: The testing code is well designed and written! Those DB and test cases management come in handy! Really great work.

script/testing/self_driving/forecast.py
```diff
@@ -0,0 +1,135 @@
...
+
+    if not Path(DEFAULT_QUERY_TRACE_FILE).exists():
+        LOG.error(
+            f"Missing {DEFAULT_QUERY_TRACE_FILE} at CWD after running OLTP TPCC")
+        return False
+
+    return True
+
+
+def turn_on_metrics_trace() -> None:
```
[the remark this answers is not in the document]
githubxxcc: Oh this is a good idea.

malin1993ml (commented):
I think the changes are good! Only one additional comment I have. I also re-requested @bialesdaniel's review since there're some changes in the testing code.

script/testing/self_driving/forecast.py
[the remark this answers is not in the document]
bialesdaniel: `config_xml_file()`. It looks like its a class function but it could probably take in parameters instead of using `self.abc`. https://github.com/cmu-db/noisepage/blob/master/script/testing/oltpbench/test_case_oltp.py#L173

script/testing/util/db_server.py
```diff
@@ -108,6 +109,46 @@ def print_db_logs(self):
...
+                        rows = cursor.fetchall()
+                        return rows
+
+                    return None
+        except Exception as e:
+            LOG.error(f"Executing SQL = {sql} failed: ")
+            # Re-raise this
+            raise e
+
+    def turn_on_metrics_trace(self):
```
bialesdaniel, line 142: You might want to name this `enable_metrics_trace` just to keep the terminology similar.

bialesdaniel (approved):
LGTM. Left two comments but I don't think they are a big deal.

malin1993ml (approved):
LGTM!

githubxxcc:
Changes to Jenkins also passed here: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/PR-33/8/pipeline

PR #1405, opened 2020-12-21
 Fix correlated subquery optimizer rules

### LINKED ISSUE #1404: Error in RewritePullFilterThroughAggregation rule for Correlated Subqueries
# Bug Report

## Summary
(NOTE: when referring to depth in this summary, the higher the depth is the more nested the expression is.) 

The logic for selecting the group by columns for the `RewritePullFilterThroughAggregation` rule is incorrect. Specifically the error is here: https://github.com/cmu-db/noisepage/blob/041fe8fb27a5d6db28fa037c20e7b9358e19a108/src/optimizer/rules/rewrite_rules.cpp#L504-L508
We select the left or right side of the predicate to become the group by column depending on whether or not the left side has a smaller depth than the predicate itself.

When we derive the depth of an expression we take the minimum depth of all it's children as seen here:
https://github.com/cmu-db/noisepage/blob/041fe8fb27a5d6db28fa037c20e7b9358e19a108/src/parser/expression/abstract_expression.cpp#L245-L253

The consequence is that an expression can never have a higher depth than it's children and the if statement `root_expr->GetChild(0)->GetDepth() < root_expr->GetDepth()` will always be false. So the left side of the predicate will always be chosen as the group by column. If the left side of the predicate can be found at same depth or higher as the aggregate then this is fine. If the left side of the predicate is at a lower depth, then it can't be accessed by the aggregate and will cause an error.

It's possible that one or some of the Nested Query issues opened by @gautam20197 are related to this. The way to test is see if the query has a correlated subquery where the predicate of the inner query has a column from the outer query on the left side. If flipping the predicate fixes the issue then it's probably related to this.

Consider the following two examples:
```
CREATE TABLE shipment (sno int, pno int, qty int, item_name VARCHAR(32));
CREATE TABLE supplier (sno int, quota int);
CREATE TABLE part (pno int, amount int, part_name VARCHAR(32));
SELECT item_name FROM shipment WHERE qty <= (SELECT SUM(amount) FROM part WHERE part.pno = shipment.pno) order by item_name;
```
The predicate in question is `WHERE part.pno = shipment.pno`. Here `part.pno` is the left side of the predicate so it will be selected as the group by column. Since `part` is at the same depth as the aggregate it won't cause an error and in fact if we run this we won't receive an error.

Now let's flip the predicate
```
CREATE TABLE shipment (sno int, pno int, qty int, item_name VARCHAR(32));
CREATE TABLE supplier (sno int, quota int);
CREATE TABLE part (pno int, amount int, part_name VARCHAR(32));
SELECT item_name FROM shipment WHERE qty <= (SELECT SUM(amount) FROM part WHERE shipment.pno = part.pno) order by item_name;
```
The predicate in question now is `WHERE shipment.pno = part.pno`. Here `shipment.pno` is the left side of the predicate so it will be selected as the group by column. Since `shipment` is a lower depth than the aggregate this will not work. In fact if we run this the DBMS will crash and output the following error
```
terminate called after throwing an instance of 'noisepage::ExecutionException'
  what():  Codegen: Cannot read column 1 from 'Aggregate' with output schema {"columns":[{"expr":{"alias":"","children":[],"depth":-1,"expression_name":"","expression_type":28,"has_subquery":false,"return_value_type":4,"tuple_idx":1,"value_idx":0},"name":"","type":4}]}
```

## Environment

**OS**: Ubuntu (LTS) 20.04

**Compiler**: GCC 7.0+

**CMake Profile**: `Debug`

**Jenkins/CI**: N/A

## Steps to Reproduce
Compile under any settings and run the queries below.

### Expected Behavior

```SQL
noisepage=# CREATE TABLE shipment (sno int, pno int, qty int, item_name VARCHAR(32));
CREATE TABLE
noisepage=# CREATE TABLE supplier (sno int, quota int);
CREATE TABLE
noisepage=# CREATE TABLE part (pno int, amount int, part_name VARCHAR(32));
CREATE TABLE
noisepage=# SELECT item_name FROM shipment WHERE qty <= (SELECT SUM(amount) FROM part WHERE shipment.pno = part.pno) order by item_name;
 item_name 
-----------
(0 rows)
```

### Actual Behavior
```
noisepage=# CREATE TABLE shipment (sno int, pno int, qty int, item_name VARCHAR(32));
CREATE TABLE
noisepage=# CREATE TABLE supplier (sno int, quota int);
CREATE TABLE
noisepage=# CREATE TABLE part (pno int, amount int, part_name VARCHAR(32));
CREATE TABLE
noisepage=# SELECT item_name FROM shipment WHERE qty <= (SELECT SUM(amount) FROM part WHERE shipment.pno = part.pno) order by item_name;
!> 
```

jkosh44:
## Description
Part of RewritePullFilterThroughAggregation and DependentSingleJoinToInnerJoin rules rule adds a group by column to the aggregation from one of the sides of the predicate. When selecting a column we need to make sure that the column is at the same depth as the aggregation or deeper. If the group by column is part of the outer query and the aggregation part of the inner query, then the aggregation has no way of accessing the column. A higher value for depth means deeper in the querer and a lower value for depth means more shallow in the query. The current code was always setting the column from the left side of the predicate as the group by column. This was sometimes causing the group by column to be more shallow than the aggregation itself which caused the query to error out.

Fixes #1404

### Remaining Tasks
N/A

## Performance
N/A

## Further Work
Check if this fixed any other nested query open issues

### DIFF
--- a/src/include/optimizer/rules/unnesting_rules.h
+++ b/src/include/optimizer/rules/unnesting_rules.h
@@ -1,6 +1,9 @@
 #pragma once
 
 #include <memory>
+#include <string>
+#include <tuple>
+#include <unordered_set>
 #include <vector>
 
 #include "optimizer/rule.h"
@@ -119,5 +122,21 @@ class DependentSingleJoinToInnerJoin : public Rule {
                  std::vector<std::unique_ptr<AbstractOptimizerNode>> *transformed,
                  OptimizationContext *context) const override;
 };
+
+/**
+ * Given predicates associated with some aggregate, this function will extract the predicates that are correlated with
+ * an outer query, as well as the columns from those predicates that aren't correlated.
+ * @param predicates vector of predicates associated with an aggregate
+ * @param child_group_aliases_set the table alias set of the predicate's child node
+ * @return
+ *      correlated_predicates - predicates which are correlated to an outer query
+ *      normal_predicates - predicates which are not correlated to an outer query
+ *      new_group_cols - columns from correlated predicates which are not correlated to an outer query. These will be
+ * used as group by columns in the nested aggregate
+ */
+std::tuple<std::vector<AnnotatedExpression>, std::vector<AnnotatedExpression>,
+           std::vector<common::ManagedPointer<parser::AbstractExpression>>>
+ExtractCorrelatedPredicatesWithAggregate(const std::vector<AnnotatedExpression> &predicates,
+                                         const std::unordered_set<std::string> &child_group_aliases_set);
 
 };  // namespace noisepage::optimizer
--- a/src/include/optimizer/rules/unnesting_rules.h
+++ b/src/include/optimizer/rules/unnesting_rules.h
@@ -128,15 +128,14 @@ class DependentSingleJoinToInnerJoin : public Rule {
  * an outer query, as well as the columns from those predicates that aren't correlated.
  * @param predicates vector of predicates associated with an aggregate
  * @param child_group_aliases_set the table alias set of the predicate's child node
- * @return
- *      correlated_predicates - predicates which are correlated to an outer query
- *      normal_predicates - predicates which are not correlated to an outer query
- *      new_group_cols - columns from correlated predicates which are not correlated to an outer query. These will be
- * used as group by columns in the nested aggregate
+ * @param[out] correlated_predicates predicates which are correlated to an outer query
+ * @param[out] normal_predicates predicates which are not correlated to an outer query
+ * @param[out] new_group_cols columns from correlated predicates which are not correlated to an outer query. These
+ * will be used as group by columns in the nested aggregate
  */
-std::tuple<std::vector<AnnotatedExpression>, std::vector<AnnotatedExpression>,
-           std::vector<common::ManagedPointer<parser::AbstractExpression>>>
-ExtractCorrelatedPredicatesWithAggregate(const std::vector<AnnotatedExpression> &predicates,
-                                         const std::unordered_set<std::string> &child_group_aliases_set);
+void ExtractCorrelatedPredicatesWithAggregate(
+    const std::vector<AnnotatedExpression> &predicates, const std::unordered_set<std::string> &child_group_aliases_set,
+    std::vector<AnnotatedExpression> *correlated_predicates, std::vector<AnnotatedExpression> *normal_predicates,
+    std::vector<common::ManagedPointer<parser::AbstractExpression>> *new_groupby_cols);
 
 };  // namespace noisepage::optimizer
--- a/src/optimizer/rules/rewrite_rules.cpp
+++ b/src/optimizer/rules/rewrite_rules.cpp
@@ -501,10 +501,11 @@ void RewritePullFilterThroughAggregation::Transform(common::ManagedPointer<Abstr
       // (outer_relation.a = (expr))
       correlated_predicates.emplace_back(predicate);
       auto root_expr = predicate.GetExpr();
-      if (root_expr->GetChild(0)->GetDepth() < root_expr->GetDepth()) {
-        new_groupby_cols.emplace_back(root_expr->GetChild(1).Get());
-      } else {
+      // See https://github.com/cmu-db/noisepage/issues/1404
+      if (root_expr->GetChild(0)->GetDepth() > root_expr->GetDepth()) {
         new_groupby_cols.emplace_back(root_expr->GetChild(0).Get());
+      } else {
+        new_groupby_cols.emplace_back(root_expr->GetChild(1).Get());
       }
     }
   }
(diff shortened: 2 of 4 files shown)


src/optimizer/rules/rewrite_rules.cpp
```diff
@@ -501,10 +501,10 @@ void RewritePullFilterThroughAggregation::Transform(common::ManagedPointer<Abstr
       // (outer_relation.a = (expr))
       correlated_predicates.emplace_back(predicate);
       auto root_expr = predicate.GetExpr();
-      if (root_expr->GetChild(0)->GetDepth() < root_expr->GetDepth()) {
-        new_groupby_cols.emplace_back(root_expr->GetChild(1).Get());
-      } else {
+      if (root_expr->GetChild(0)->GetDepth() > root_expr->GetDepth()) {
```
lmwnshn, line 504: 1. This is an awesome PR.
2. Could you please add a comment that links to your issue? Something like "<summary of 1404>, please see #1404 for more details."

jkosh44:
@lmwnshn I jumped the gun on this. sorry. While it removes the error and causes the query to execute successfully, the output is incorrect. Though it doesn't break any existing queries or tests, so it's a step in the right direction. There's probably a similar error in another rule causing an issue.

jkosh44:
@lmwnshn Excuse the force push, but this should be good now.

src/optimizer/rules/rewrite_rules.cpp
jkosh44: I borrowed the comment from the `DependentSingleJoinToInnerJoin`. I just need to double check that it's accurate.

lmwnshn (commented):
ignore this comment the github ui misled me

lmwnshn (approved):
LGTM. The copypasting is a bit unfortunate but I'm on the fence on whether it makes sense to refactor out that block of code to a helper function -- can you think of a suitable name for it?

jkosh44:
> LGTM. The copypasting is a bit unfortunate but I'm on the fence on whether it makes sense to refactor out that block of code to a helper function -- can you think of a suitable name for it?

Agreed on the copy pasting. I have no idea what these rules are actually doing so I didn't put in much effort into seeing how they could be refactored. I'll take a look today and see if anything makes sense.

lmwnshn:
Feel free to flip the label to ready-to-merge depending on that goes @jkosh44

jkosh44:
@lmwnshn @xinzhu-cai As far as I can tell the duplicate code in `DependentSingleJoinToInnerJoin` is actually all dead code. The separating of correlated and non correlated predicates happen in `RewritePullFilterThroughAggregation`, so there's nothing left for `DependentSingleJoinToInnerJoin` to do. I ran through all of our test nested queries as well as some of my own queries and I couldn't find any scenario where the code is actually run.

src/include/optimizer/rules/unnesting_rules.h
```diff
@@ -120,4 +123,20 @@ class DependentSingleJoinToInnerJoin : public Rule {
...
+ * an outer query, as well as the columns from those predicates that aren't correlated.
+ * @param predicates vector of predicates associated with an aggregate
+ * @param child_group_aliases_set the table alias set of the predicate's child node
+ * @return
+ *      correlated_predicates - predicates which are correlated to an outer query
+ *      normal_predicates - predicates which are not correlated to an outer query
+ *      new_group_cols - columns from correlated predicates which are not correlated to an outer query. These will be
+ * used as group by columns in the nested aggregate
+ */
+std::tuple<std::vector<AnnotatedExpression>, std::vector<AnnotatedExpression>,
```
jkosh44, line 137: Is it not good practice to use `tuple`s like this?

jkosh44: Would it be better to just make the input params to avoid moving the results?

src/optimizer/rules/rewrite_rules.cpp
```diff
@@ -1,5 +1,7 @@
 #include "optimizer/rules/rewrite_rules.h"
 
+#include <optimizer/rules/unnesting_rules.h>
```
lmwnshn, line 3: nit: `""` include

src/include/optimizer/rules/unnesting_rules.h
lmwnshn: I don't know if we have a codified good practice or not around `tuple`.
IIRC there's something about `tuple` potentially producing stupider code on gcc compared to clang.
I don't know enough C++ to have a good opinion here.

Perhaps input params are better in favor of readability though.

lmwnshn (approved):
Minor nit. I don't know enough C++ to have a good opinion on returning tuple vs having more input parameters + this is unlikely to be a performance bottleneck. Still happy to merge either way -- readability is much improved IMO.

src/optimizer/rules/rewrite_rules.cpp
jkosh44: fixed

src/include/optimizer/rules/unnesting_rules.h
jkosh44: I switched to the pointer approach

PR #1409, opened 2020-12-21
Better command line interaction when ModelServerManager enabled 

githubxxcc:
Pointed out by @rabbit721, this PR fixes some behaviour of ModelServer when invoked though the `noisepage` binary: 
1. INFO log when a ModelServer is connected
2. Graceful handling of KeyboardInterrupt in the ModelServer


src/self_driving/model_server/model_server_manager.cpp
```diff
@@ -69,6 +69,7 @@ ModelServerManager::ModelServerManager(const std::string &model_bin,
       case Callback::NOOP:
         break;
       case Callback::CONNECTED:
+        MODEL_SERVER_LOG_INFO("[PID={}] ModelServer connected", ::getpid());
```
lmwnshn, line 72: I kind of want to argue that this should be DEBUG or TRACE, but no strong enough opinion.

Just want to avoid having too many log messages when starting up the server, otherwise people will stop reading.

lmwnshn (approved):
LGTM with one very small protest

malin1993ml (approved):
LGTM. I don't have a strong opinion on whether the model server startup info should be INFO or DEBUG either. But I'm fine with INFO since it only gets printed out when the model server is enabled. And it's good to check that it connects in that scenario.

PR #1410, opened 2020-12-22
Add enum_defs and convert ExpressionType to be serialized to JSON as string.

lmwnshn:
# Heading

Add enum_defs and convert ExpressionType to be serialized to JSON as string.

## Description

Bring some execution-engine style macro magic to the frontend to automatically define enums with corresponding ToString and FromString macros.

## Questions

1. (Maintainability) Introduce macro magic? (see `enum_defs.h`) Essentially the same for what you need to navigate the execution engine.
2. (Performance, memory) Default to serializing human-readable representations?

## Motivation

- I'm trying to improve the overall debuggability of the system (assuming EXPLAIN merges).
- C++ enums serialize as a number. This is annoying for debugging purposes.
- Writing a ToString and FromString for every member of an enum also sucks (see diff for `parser::ExpressionType`).
  - The JSON library that we pull in can do this automatically for enums with code that looks like
   ```
   NLOHMANN_JSON_SERIALIZE_ENUM( TaskState, {
    {TS_INVALID, nullptr},
    {TS_STOPPED, "stopped"},
    {TS_RUNNING, "running"},
    {TS_COMPLETED, "completed"},
   })
   ```
   but in exchange you must include `json.hpp` instead of `json_fwd.hpp` which will blow up your compile times like crazy.
- Given that we don't expect serializing JSON to be on the path of important things, (or at least that was the plan some years back, we do deserialize stored expressions for some catalog stuff now), it seems reasonable to **default to serializing human-readable representations**, but maybe this is a bad idea.

Thoughts? Guinea pig is `parser::ExpressionType`.

(I also don't know why we need a `ExpressionTypeToShortString` but that's a separate yak.)

## Remaining tasks

This might blow up in self-driving mini-runners stuff or wherever else hardcoding of `ExpressionType` may have occurred. Running through CI to find out.


mbutrovich:
Can you add a short blurb to coding style in /docs mentioning how to handle serializable enums and then this is good to merge.

mbutrovich (approved):
LGTM!

PR #1415, opened 2020-12-24
add reporting to terrier build

bialesdaniel:
# CI Performance Reporting

## Description
This is just a small PR for some work that should have been part of #1368 but must have gotten lost on one of my merges. It just sends the performance data from the normal PR pipeline to our performance storage service. This is needed to enable the performance cop.

Here is proof of a passing build: http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/testing-team%2Fterrier/detail/ci-performance-reporting/3/pipeline/287

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -369,6 +369,11 @@ pipeline {
                 }
                 stage('Performance') {
                     agent { label 'benchmark' }
+                    environment {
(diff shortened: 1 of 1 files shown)


lmwnshn (approved):
LGTM.

Issue #1421, opened 2020-12-29
Optimizer framework improvements


thepinetree:
# Feature Request
## Summary
Optimizer framework rewrite tasks to support better operator choices and property pushdowns, in particular related to ORDER BY, LIMIT, DESCENDING scans.

## Solution
**Optimizer Related Changes**
- [ ] Implement LIMIT clause pushdown and functionality on Index Scans and Index Joins
- [ ] Implement LIMIT clause pushdown and functionality on sequential scans
- [ ] Convert LIMIT nodes into properties for the optimizer front end, as is done with (ORDER BY), which should also address issue #1414
- [ ] Support Descending index scans, which are never selected by the optimizer in `index_util.cpp`:https://github.com/cmu-db/noisepage/blob/520e0da7437d9e970e0fcb144a2f9939c35ba560/src/optimizer/index_util.cpp#L80

**Cost Model Related Changes**
- [ ] Add support for the cost model to choose the type of index scan based on the estimated cardinality of the scan itself (e.g. as observed by @tanujnay112, an Index Scan may match more predicate columns and create an open scan which runs more slowly compared to an Index Scan that matches fewer predicates but can perform an exact scan.)
- [ ] Add support for optional properties in the optimizer and make the (future) cost model aware of these for better operator choice (an example of optional properties for ORDER BY/sorting exists in https://github.com/cmu-db/noisepage/pull/1031)

**Index Related Changes**
- [ ] Support an index iterator that allows for predicate push down into the index, at which point, predicate filter nodes can be removed for index scans, and LIMITs can be used more robustly, as explained in comments in (https://github.com/cmu-db/noisepage/pull/1031)

PR #1422, opened 2020-12-29
Index Scan + Index Join Limit

thepinetree:
# Index Scan + Index Join Limit

## Description
Limit clauses are currently not propagated to the `IndexScanPlanNode` nor the `IndexJoinPlanNode` and as a result, the execution engine can't take advantage of pushing down the limit during operation. Instead, this is done in-post, with a `LimitPlanNode` doing so after an index scan is completed.

This PR adds functionality for the limit value to be pushed down to index scans, and is used in TPC-C. Limits values will be pushed down to their child `LogicalGet` via transformation rule and converted to values in the `PhysicalIndexScan` which are then set in the `IndexScanPlanNode`. The PR also moves the `OrderByOrderingType` from the optimizer to the catalog as a precursor to further changes to involve the sort direction of columns in creating/scanning an index.

## Further work
A description of further work is included in Issue #1421


docs/optimization.md
```diff
@@ -0,0 +1,153 @@
...
+
+- If the `SELECT` statement operates upon a table, the `TableRef` is accepted as well and added to the `output_expr_` as a `LogicalGet`. Otherwise, the `LogicalGet` is initialized to operate on an invalid table (used for queries like `SELECT 1 + 1`).
+- Predicates for the input expression, if they exist, are collected with an alias to each referenced column's associated table name in `std::vector<AnnotatedExpression> predicates_` and are stored in a `LogicalFilter` that is added as the parent of the existing `output_expr_`
+    - [ ]  Add information for subquery transformation above
+- [ ]  Add information about aggregations in `SELECT` here
+- [ ]  Add information about `SelectDistinct` in `SELECT` here
+- If a `LIMIT` exists in the select statement, the value of the limit and associated offset, along with any  `ORDER_BY` clauses, are stored in a `LogicalLimit` node that is added as the parent of the existing `output_expr_`
```
jkosh44, line 40: I think it would be helpful here to briefly explain why `ORDER_BY` information is stored within a `LIMIT` operator

docs/optimization.md
```diff
@@ -0,0 +1,153 @@
...
+- `OptimizerContext`: the structure which tracks the cost model, task pool, transaction context, statistics storage, catalog accessor, rule set, and tracked optimization contexts, and importantly the **memoized expressions for each group.**
+- `Memo`: a class to track groups and group expressions that allows for duplicate detection, rewrites, access, and erase procedures.
+- `GroupExpression`: a class to track a specific operator (logical or physical) in a group which tracks the transaction, group number, whether the stats have been derived, the lowest cost satisfying the required properties of an expression, and the details of the expression.
+- `Group`: a class to collect group expressions representing logically equivalent expressions including logical and physical transformations.
+- `LogicalExpression`: a group expression which represents a logical expression i.e. the logical operator that specifies the general behavior of a node
+- `PhysicalExpression`: a group expression which represents a logical expression i.e. the physical operator that specifies the implementation details of the expression
```
jkosh44, line 66: Did you mean to say "logical expression" here?

docs/optimization.md
```diff
@@ -0,0 +1,153 @@
...
+- `Rule`: definition of a transformation from one expression to another (may be logical transformation, physical implementation, or rewrite)
+- `Property`: interface to define a physical property, currently of which only sort properties exist
```
jkosh44, line 68: I think it would be helpful if we had a couple of sentences explaining what a property is and why we have them.

src/execution/compiler/operator/index_join_translator.cpp
```diff
@@ -202,6 +202,7 @@ void IndexJoinTranslator::FillKey(
     WorkContext *context, FunctionBuilder *builder, ast::Identifier pr,
     const std::unordered_map<catalog::indexkeycol_oid_t, planner::IndexExpression> &index_exprs) const {
   // For each key attribute,
+  uint32_t count = 0;
```
jkosh44, line 205: What is this used for?

src/optimizer/child_property_deriver.cpp
```diff
@@ -78,7 +81,7 @@ void ChildPropertyDeriver::Visit(UNUSED_ATTRIBUTE const HashGroupBy *op) {
 
 void ChildPropertyDeriver::Visit(UNUSED_ATTRIBUTE const SortGroupBy *op) {
```
jkosh44, line 82: I know you didn't really make changes to this, but should we remove `UNUSED_ATTRIBUTE` here because we are using `op`

src/optimizer/rules/implementation_rules.cpp
```diff
@@ -117,43 +113,85 @@ void LogicalGetToPhysicalIndexScan::Transform(common::ManagedPointer<AbstractOpt
...
+      // Check whether any index can fulfill in-order predicate evaluation
       if (IndexUtil::SatisfiesPredicateWithIndex(accessor, get->GetTableOid(), get->GetTableAlias(), index, preds,
                                                  allow_cves_, &scan_type, &bounds)) {
-        auto op = std::make_unique<OperatorNode>(IndexScan::Make(db_oid, get->GetTableOid(), index, std::move(preds),
-                                                                 is_update, scan_type, std::move(bounds))
-                                                     .RegisterWithTxnContext(context->GetOptimizerContext()->GetTxn()),
-                                                 std::vector<std::unique_ptr<AbstractOptimizerNode>>(),
-                                                 context->GetOptimizerContext()->GetTxn());
+        // Limit can only be pushed down in the following cases
+        // TODO(dpatra): Limit can be pushed down in the case of an exact scan as well, but the index framework
```
jkosh44, line 143: I don't really understand this comment. It seems like we are trying to push down limits for exact scans based on the predicate below ...`scan_type == planner::IndexScanType::Exact ||`...

src/optimizer/rules/implementation_rules.cpp
```diff
@@ -117,43 +113,85 @@ void LogicalGetToPhysicalIndexScan::Transform(common::ManagedPointer<AbstractOpt
...
+        // doesn't currently support it.
+        limit_exists = limit_exists && (scan_type == planner::IndexScanType::Exact ||
+                                        scan_type == planner::IndexScanType::AscendingOpenHighLimit ||
+                                        scan_type == planner::IndexScanType::AscendingClosedLimit ||
+                                        scan_type == planner::IndexScanType::DescendingLimit);
+        preds = get->GetPredicates();
```
jkosh44, line 149: why are we re-assigning `preds` here?

src/optimizer/rules/implementation_rules.cpp
```diff
@@ -487,11 +525,16 @@ void LogicalInnerJoinToPhysicalInnerIndexJoin::Transform(
...
   for (auto &pred : inner_join->GetJoinPredicates()) join_preds.push_back(pred);
 
+  // Get inner join limit information
+  auto limit_exists = inner_join->GetLimitExists();
+  auto limit = inner_join->GetLimit();
+
   std::vector<std::unique_ptr<AbstractOptimizerNode>> empty;
+  auto child_get = LogicalGet::Make(r_child->GetDatabaseOid(), r_child->GetTableOid(), join_preds,
+                                    r_child->GetTableAlias(), r_child->GetIsForUpdate());
+  if (limit_exists) child_get.GetContentsAs<LogicalGet>()->SetLimit(limit);
```
jkosh44, line 535: Just thinking out loud here, is it ever possible that we'll need more rows than the limit from the `child_get` to have enough rows in the join to satisfy the limit? I don't think so because we're also pushing the join predicates down and I haven't actually found any query that brakes because of this.

src/optimizer/rules/transformation_rules.cpp
```diff
@@ -158,4 +158,67 @@ void LogicalInnerJoinAssociativity::Transform(common::ManagedPointer<AbstractOpt
...
+void SetLimitInGet::Transform(common::ManagedPointer<AbstractOptimizerNode> input,
+                              std::vector<std::unique_ptr<AbstractOptimizerNode>> *transformed,
+                              UNUSED_ATTRIBUTE OptimizationContext *context) const {
+  auto get = input->GetChildren()[0]->Contents()->GetContentsAs<LogicalGet>();
+  size_t limit = input->Contents()->GetContentsAs<LogicalLimit>()->GetLimit();
+  get->SetLimit(limit);
+}
+
+///////////////////////////////////////////////////////////////////////////////
+/// SetLimitInGet
```
jkosh44, line 190: I think you probably meant to put SetLimitInLogicalInnerJoin here?

src/include/planner/plannodes/plan_node_defs.h
```diff
@@ -141,9 +141,12 @@ enum class SetOpType { INVALID = INVALID_TYPE_ID, INTERSECT = 1, INTERSECT_ALL =
 using IndexExpression = common::ManagedPointer<parser::AbstractExpression>;
 /** Type of index scan. */
 enum class IndexScanType : uint8_t {
+  Dummy,
```
jkosh44, line 144: If I'm understanding `Dummy` correctly then it may help to add a short comment saying something along the lines that `Dummy` is a placeholder value and shouldn't actually be used as a scan type.

src/include/optimizer/properties.h
```diff
@@ -19,10 +19,11 @@ class PropertySort : public Property {
    * Constructor for PropertySort
    * @param sort_columns vector of AbstractExpressions representing sort columns
    * @param sort_ascending Whether each sort_column is ascending or descending
+   * @param required Whether the sort property needs to be satisfied at the active level
    */
   PropertySort(std::vector<common::ManagedPointer<parser::AbstractExpression>> sort_columns,
-               std::vector<OrderByOrderingType> sort_ascending)
-      : sort_columns_(std::move(sort_columns)), sort_ascending_(std::move(sort_ascending)) {}
+               std::vector<catalog::OrderByOrderingType> sort_ascending, bool required = true)
+      : Property(required), sort_columns_(std::move(sort_columns)), sort_ascending_(std::move(sort_ascending)) {}
```
jkosh44, line 26: I think you may need to update `PropertySort::Copy` to copy `required_`.

src/optimizer/child_property_deriver.cpp
```diff
@@ -95,18 +96,15 @@ void ChildPropertyDeriver::Visit(UNUSED_ATTRIBUTE const Aggregate *op) {
...
     const std::vector<common::ManagedPointer<parser::AbstractExpression>> &exprs = op->GetSortExpressions();
     const std::vector<catalog::OrderByOrderingType> &sorts{op->GetSortAscending()};
-    provided_prop->AddProperty(new PropertySort(exprs, sorts));
-
-    child_input_properties[0]->AddProperty(new PropertySort(exprs, sorts, false));
+    child_input_properties[0]->AddProperty(new PropertySort(exprs, sorts));
   }
 
-  output_.emplace_back(provided_prop, std::move(child_input_properties));
+  output_.emplace_back(new PropertySet(), std::move(child_input_properties));
```
jkosh44, line 107: I think that if we push the sort down to the child, then the `Limit` should still have an output sort property.

src/optimizer/child_property_deriver.cpp
```diff
@@ -95,18 +96,15 @@ void ChildPropertyDeriver::Visit(UNUSED_ATTRIBUTE const Aggregate *op) {
 void ChildPropertyDeriver::Visit(const Limit *op) {
   // Limit fulfill the internal sort property
   std::vector<PropertySet *> child_input_properties{new PropertySet()};
-  auto provided_prop = new PropertySet();
 
   // Limit must satisfy output sort properties but child can attempt to satisfy sort property optionally
```
jkosh44, line 100: I think this comment no longer applies

src/optimizer/child_property_deriver.cpp
thepinetree: Hmm, how come? Wouldn't the Limit's sort already be fulfilled by the child since it gets pushed down?

src/optimizer/index_util.cpp
```diff
@@ -48,9 +48,15 @@ bool IndexUtil::SatisfiesSortWithIndex(
     if (tv_col_oid == mapped_cols[idx_ind]) {
       // Column is present in both sort and index so increment both
       sort_ind++, idx_ind++;
-    } else if (bounds != nullptr && bounds->find(lookup[mapped_cols[idx_ind]]) != bounds->end()) {
-      // If column is found in bounds but not sort, continue
-      idx_ind++;
+    } else if (bounds != nullptr) {
+      auto bound_column = bounds->find(lookup[mapped_cols[idx_ind]]);
+      if (bound_column != bounds->end() && bound_column->second[0] == bound_column->second[1]) {
```
jkosh44, line 53: From my understanding (let me know if this is incorrect) if the first N continuous columns of the index are not in the sort, then they don't have to be bound exactly. It's just the inner columns in the index but not in the sort that need to be bound exactly. I'm not sure how important it is to capture this scenario though.

thepinetree: I don't think that's always true:
c1 | c2 | c3
 1  |  3   | 4
 1  |  3   | 5
 2  |  3   | 4
 
 If you want the elements sorted by c2, c3 then since the first column isn't bound, the last tuple wouldn't be in the correct sort order (the inner loop runs faster than the outer loop is my way of thinking about it).

docs/optimization.md
thepinetree: Which line is this about?

src/optimizer/child_property_deriver.cpp
jkosh44: So maybe I'm have some misunderstanding with how this function behaves, but since the sort is fulfilled by the child and the Limit node doesn't break the sort, then it will also have a sorted output.  Otherwise `OptimizeExpressionCostWithEnforcedProperty` will see that the `Limit` node has a required sort property, but no output sort property and will enforce it with an OrderBy node.

jkosh44: The required properties from the `OptimizationContext` will only be enforced if they are not in the output properties derived from this function (Lines 340 of optimizer_task.cpp). So if a Limit has a required sort property, which it will if there's an order by, but it doesn't have an output sort property then it will generate an OrderBy node. Since the child will have already satisfied the sort property, this OrderBy node is unnecessary.

src/execution/compiler/operator/index_join_translator.cpp
thepinetree: Good catch, missed removing it

src/optimizer/child_property_deriver.cpp
thepinetree: Yeah, updated

src/optimizer/rules/implementation_rules.cpp
thepinetree: I'll move this comment to the index_scan_translator. The function used for exact scans doesn't take advantage of the limit right now.

src/optimizer/rules/implementation_rules.cpp
thepinetree: Yep, unnecessary + removed

src/optimizer/rules/transformation_rules.cpp
thepinetree: Yep, fixed

src/include/optimizer/properties.h
thepinetree: As noted, no longer needed with the removal of optional properties.

src/optimizer/child_property_deriver.cpp
thepinetree: Yep, removed

src/include/planner/plannodes/plan_node_defs.h
thepinetree: Updated

src/optimizer/child_property_deriver.cpp
thepinetree: Yeah, you're 100% right, my bad

thepinetree: Fixed

src/optimizer/rules/implementation_rules.cpp
thepinetree: I think it should be ok since this get is effectively a probe for after the join predicates, and the `LogicalGet` implementation rule should ignore the `Limit` in cases where all predicates aren't tracked by the index.

docs/optimization.md
jkosh44: "`PhysicalExpression`: a group expression which represents a logical expression..." I think you meant to put "physical expression" instead of "logical expression".

src/optimizer/index_util.cpp
jkosh44: Ah now I get it, thanks for the explanation

jkosh44:
Hey @thepinetree I'm just going to consolidate everything we discussed on slack here so that you have a single place to go to when you return to this PR, let me know if I'm missing anything. I'm not sure which of these belong in this PR, or should be done it later PRs, I guess that's up to you/Will/whoever makes those decisions.

1. We need someway of pushing down limits to `OrderBy` nodes that are generated through the `PropertyEnforcer`
2. We need to remove the sort information from `Limit` nodes, and remove the `OrderBy` generation for limit's in the `PlanGenerator`.
3. Since the `InputColumnDeriver` uses an `unordered_set` to store columns, the order of the columns returned is essentially "random". This can lead to unnecessary `ProjectionPlanNode`s being generated in the middle of the query plan to reorder columns.

PR #1425, opened 2021-01-01
Add a spin-until-initialized to messenger_test to avoid races.

### LINKED ISSUE #1319: jumbotest_messenger_messenger_test timesout on jenkins
# jumbotest_messenger_messenger_test timesout on jenkins

## Summary
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/master/791/tests
It seems that the test times out. Potentially due to deadlock or other hangup.

## Environment

**OS**: Ubuntu 20.04 (LTS)

**Compiler**: GCC 9.3 release

lmwnshn:
# Heading

Add a spin-until-initialized to messenger_test to avoid races.

## Description

Fix #1319 (I think).

Previously, it was possible in `BasicListenTest` for the replica to try to connect to the primary before the primary had opened an additional listening channel on port 8030.

This adds a dirty spin-on-shared-memory-until-things-are-ready hack. In general, I'm not too happy with how testing multiple processes in C++ works -- forking more than one child means that you lose the debugger autoattach (though you can still manually attach), coordinating between the processes is not so good either.

## Future work

I notice that I can't seem to set the SO_REUSEADDR property on the zeromq sockets.
The `DirtySleep` gets around that by waiting for what should be long enough for the test to work without address reuse.

### DIFF
--- a/test/messenger/messenger_test.cpp
+++ b/test/messenger/messenger_test.cpp
@@ -79,8 +79,8 @@ class MessengerTests : public TerrierTest {
     return db_main;
   }
 
-  /** A dirty hack that sleeps for a little while so that sockets can clean up. */
-  static void DirtySleep() { std::this_thread::sleep_for(std::chrono::seconds(5)); }
+  /** A dirty hack that sleeps for a little while so that sockets can wake up and clean up. */
+  static void DirtySleep() { std::this_thread::sleep_for(std::chrono::seconds(2)); }
 };
 
 // NOLINTNEXTLINE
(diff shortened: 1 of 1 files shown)


malin1993ml:
This looks reasonable but I'm wondering how about the `modeling_test`? Does that test time out for the same reason?

lmwnshn:
Don't think so. That seems to be a separate issue where the environment has leftover stuff from previous runs. I haven't looked into the kill_server etc scripts much, I've seen it hang there before too.

malin1993ml:
I see!

malin1993ml (approved):
LGTM.

jasonbak:
4d5f2680-4043-11eb-82b8-77d585223466

this is a straight banger. get this shit in prod right now. bless up🙏

PR #1426, opened 2021-01-04
Self-Driving Index/Knob Actions

malin1993ml:
This PR contains the code to generate candidate actions for self-driving planning. The main classes that the pilot will directly call are `ChangeKnobActionGenerator` and `IndexActionGenerator`. Actions should be applied with the SQL command by the `GetSQLCommand()` method in `AbstractAction`. You can reference the tests for their usage.


src/self_driving/pilot/action/change_knob_action.cpp
```diff
@@ -10,7 +10,9 @@ const std::string &ChangeKnobAction<T>::GetSQLCommand() {
   sql_command_ = "SET " + param_name_ + " ";
```
apavlo: Random thought: Should we have a strong typedef for SQL string instead of just using `std::string`?

src/include/self_driving/pilot/action/change_knob_action_generator.h
```diff
@@ -19,6 +19,7 @@ class AbstractAction;
  * Generate change knob candidate actions
  */
 class ChangeKnobActionGenerator {
```
apavlo, line 22: I feel like there should be an `AbstractGenerator` class. These can then go into a `generators` sub-directory.

apavlo (commented):
Minor comments.

src/self_driving/pilot/action/change_knob_action.cpp
malin1993ml: Sounds reasonable. I can add that.

src/include/self_driving/pilot/action/change_knob_action_generator.h
malin1993ml: Sure! I'll do the refactor.

src/self_driving/pilot/action/change_knob_action.cpp
malin1993ml: Eh I tried that but our strong typedef does not support string types: https://github.com/cmu-db/noisepage/blob/master/src/include/common/strong_typedef.h#L79

17zhangw (approved):
Generally LGTM! Just two small nits.

PR #1428, opened 2021-01-09
CPU Frequency Modeling Augmentation

17zhangw:
## Description
PR provides a framework for recording hardware context. Initially, we only record the `cpu frequency` (as recorded when the `execution::CpuInfo` is initialized as opposed to per-metric). However, this can be augmented in the future to capture other hardware context for modeling as needed.


malin1993ml (approved):
LGTM!

PR #1433, opened 2021-01-14
Monte Carlo Search Tree for action selection based on forecasted queries

rabbit721:
# Monte Carlo Search Tree

## Description

-  Created `MonteCarloSearchTree` that performs best action selection
-  Created `TreeNode` that records the cost and action applied so far
-  Added functions in PilotUtil to generate query plan from queries, compute the cost of query execution, and apply actions to relevant databases


malin1993ml (changes requested):
I think the overall structure is good! Thanks for the work! Please see my comments for the changes. You also need to merge PR #1445 to get some additional information in `AbstractAction`. And I added the "ready-for-ci" label for your PR so you also need to pass the CI checks.

src/self_driving/pilot/mcst/tree_node.cpp
```diff
@@ -0,0 +1,133 @@
...
+  for (auto &child : children_)
+    highest = std::max(child->cost_, highest);
+
+  // sample based on cost and num of visits of children
+  std::vector<common::ManagedPointer<TreeNode>> selected_children, out;
+  uint64_t best_value = 0;
+  for (auto &child : children_) {
+    auto child_obj = std::pow((highest + EPSILON) / (child->cost_ + EPSILON), 2) +
+                     std::sqrt(2 * std::log(number_of_visits_) / child->number_of_visits_ );
+    if (child_obj > best_value) {
```
malin1993ml, line 57: Oh forgot to mention this one, which is actually important:  we shouldn't directly pick the child with the highest objective. Instead, we should really *sample* from all the children using their objectives as the probability distribution. This way we don't need the tiebreaker either.

src/self_driving/pilot/mcst/tree_node.cpp
```diff
@@ -0,0 +1,133 @@
...
+  // compute max of children's cost
+  uint64_t highest = 0;
+  for (auto &child : children_)
+    highest = std::max(child->cost_, highest);
+
+  // sample based on cost and num of visits of children
+  std::vector<common::ManagedPointer<TreeNode>> selected_children, out;
+  uint64_t best_value = 0;
+  for (auto &child : children_) {
+    auto child_obj = std::pow((highest + EPSILON) / (child->cost_ + EPSILON), 2) +
```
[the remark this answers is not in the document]
rabbit721: This was adopted from the original one suggested here that uses the fraction of wins instead of the square of the inverse of normalized cost: 
https://en.wikipedia.org/wiki/Monte_Carlo_tree_search#Exploration_and_exploitation.

I was intending to use the square to favor smaller costs, but it's pretty flexible.

src/self_driving/pilot_util.cpp
```diff
@@ -65,7 +199,6 @@ const std::list<metrics::PipelineMetricRawData::PipelineData> &PilotUtil::Collec
           common::ManagedPointer(accessor), exec_settings, metrics_manager);
 
       exec_ctx->SetParams(common::ManagedPointer<const std::vector<parser::ConstantValueExpression>>(&params));
-
       execution::compiler::ExecutableQuery::query_identifier.store(qid);
       auto exec_query = execution::compiler::CompilationContext::Compile(*out_plan, exec_settings, accessor.get(),
```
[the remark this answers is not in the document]
rabbit721: Would it also work if we also reuse the `exec_ctx` and in the inner for loop only have only [`exec_ctx->SetParams`](https://github.com/cmu-db/noisepage/blob/5707742eb1264105c4d301c847248659517f46df/src/self_driving/pilot_util.cpp/#L201) and [`exec_query->Run`](https://github.com/cmu-db/noisepage/blob/5707742eb1264105c4d301c847248659517f46df/src/self_driving/pilot_util.cpp/#L205)?

I was using the assumption that we have pipeline data for the same query id nearby in pipeline metrics recorded and `pipeline_qids` recorded to adjust the qids [here](https://github.com/cmu-db/noisepage/blob/5707742eb1264105c4d301c847248659517f46df/src/self_driving/pilot_util.cpp/#L294-298). It was made quite implicit as a post-processing step since making a new copy of all pipeline metrics with the qids fixed directly here in [CollectPipelineFeatures](https://github.com/cmu-db/noisepage/blob/5707742eb1264105c4d301c847248659517f46df/src/self_driving/pilot_util.cpp/#L151) seem like too much effort.

src/self_driving/pilot/mcst/tree_node.cpp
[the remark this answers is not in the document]
malin1993ml: Got it! Could you document this as well?

src/self_driving/pilot_util.cpp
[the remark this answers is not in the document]
malin1993ml: Got it. But I still think that if you execute multiple sets of parameters for the same query template without reusing the `ExecutableQuery`, this won't really work because you only record one qid [here](https://github.com/cmu-db/noisepage/blob/5707742eb1264105c4d301c847248659517f46df/src/self_driving/pilot_util.cpp/#L176).

I think reusing `ExecutableQuery` should solve this problem. And you do need to create the `exec_ctx` for each query execution because there's specific information for single query execution in the `exec_ctx`, such as some memory allocators.

src/self_driving/pilot/mcst/tree_node.cpp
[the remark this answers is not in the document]
rabbit721: Sure!

src/self_driving/pilot/pilot.cpp
```diff
@@ -38,11 +40,27 @@ void Pilot::PerformPlanning() {
...
+                                          end_segment_index);
+  mcst.BestAction(simulation_number_, best_action_seq);
+
+  PilotUtil::ApplyAction(common::ManagedPointer(this), best_action_seq->begin()->first,
+                         best_action_seq->begin()->second);
+}
+
+void Pilot::ExecuteForecast(std::map<std::pair<execution::query_id_t, execution::pipeline_id_t>,
+                                     std::vector<std::vector<std::vector<double>>>> *pipeline_to_prediction,
+                            uint64_t start_segment_index, uint64_t end_segment_index) {
```
malin1993ml, line 63: Can we add some documentation to this function?

src/self_driving/pilot_util.cpp
```diff
@@ -12,19 +12,148 @@
...
+#include "self_driving/modeling/operating_unit.h"
 #include "self_driving/pilot/pilot.h"
+#include "settings/settings_manager.h"
 #include "traffic_cop/traffic_cop_util.h"
 #include "transaction/transaction_manager.h"
 
 namespace noisepage::selfdriving {
 
+void PilotUtil::ApplyAction(common::ManagedPointer<Pilot> pilot, const std::string &sql_query,
+                            catalog::db_oid_t db_oid) {
```
malin1993ml, line 33: Add some documentation to this function as well. In general, if the function is not super simple, we should always document the code logic along the way. So that other people (other than you and me) who happen to need to touch this function can know what's going on.

src/self_driving/pilot_util.cpp
```diff
@@ -115,13 +241,31 @@ void PilotUtil::InferenceWithFeatures(
     }
     inference_result.emplace(ou_map_it.first, res.first);
   }
+
+  // populate pipeline_to_prediction using pipeline_to_ou_position and inference_result
+  std::vector<std::vector<double>> ppl_res;
```
malin1993ml, line 247: nit: `ppl_res`->`pipeline_result`. Change other `ppl` in this function to `pipeline` as well. In general, avoid abbreviation if it's not super long.

src/self_driving/pilot_util.cpp
```diff
@@ -115,13 +241,31 @@ void PilotUtil::InferenceWithFeatures(
...
+  for (auto ppl_to_ou : pipeline_to_ou_position) {
+    for (auto ou_it : std::get<2>(ppl_to_ou)) {
+      ppl_res.push_back(inference_result.at(ou_it.first)[ou_it.second]);
```
malin1993ml, line 250: Document what's being pushed.

src/self_driving/pilot_util.cpp
```diff
@@ -115,13 +241,31 @@ void PilotUtil::InferenceWithFeatures(
...
+    }
+    auto ppl_key = std::make_pair(std::get<0>(ppl_to_ou), std::get<1>(ppl_to_ou));
+    if (pipeline_to_prediction->find(ppl_key) == pipeline_to_prediction->end()) {
+      pipeline_to_prediction->emplace(ppl_key, std::vector<std::vector<std::vector<double>>>());
```
malin1993ml, line 254: Document what's being pushed. Since `pipeline_to_prediction` is a bit complex it may be difficult to understand.

src/self_driving/pilot_util.cpp
```diff
@@ -115,13 +241,31 @@ void PilotUtil::InferenceWithFeatures(
...
+    pipeline_to_prediction->at(ppl_key).emplace_back(std::move(ppl_res));
+  }
 }
 
 void PilotUtil::GroupFeaturesByOU(
     std::list<std::tuple<execution::query_id_t, execution::pipeline_id_t,
                          std::vector<std::pair<ExecutionOperatingUnitType, uint64_t>>>> *pipeline_to_ou_position,
+    const std::vector<execution::query_id_t> &pipeline_qids,
     const std::list<metrics::PipelineMetricRawData::PipelineData> &pipeline_data,
     std::unordered_map<ExecutionOperatingUnitType, std::vector<std::vector<double>>> *ou_to_features) {
```
malin1993ml, line 265: Document this function as well.

src/self_driving/pilot_util.cpp
```diff
@@ -12,19 +12,149 @@
...
+        db_oid, common::ManagedPointer(txn), callback, out_plan->GetOutputSchema().Get(),
+        common::ManagedPointer(accessor), exec_settings, pilot->metrics_thread_->GetMetricsManager());
+
+    exec_query->Run(common::ManagedPointer(exec_ctx), execution::vm::ExecutionMode::Interpret);
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+    txn_manager->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+  }
+}
+
+std::unique_ptr<planner::AbstractPlanNode> PilotUtil::GetOutPlan(
```
malin1993ml, line 72: nit: rename to `GenerateQueryPlan`

src/self_driving/pilot_util.cpp
```diff
@@ -12,19 +12,149 @@
...
+      plan_vecs->emplace_back(std::move(out_plan));
+    }
+  }
+  txn_manager->Abort(txn);
+}
+
+uint64_t PilotUtil::ComputeCost(common::ManagedPointer<Pilot> pilot, common::ManagedPointer<WorkloadForecast> forecast,
+                                uint64_t start_segment_index, uint64_t end_segment_index) {
+  // Compute cost as average latency of queries weighted by their num of exec
+  // Keep track of the ou inference results (a vector of doubles) for each ous associated with the pipeline
```
malin1993ml, line 117: nit: First explain the map first: maps *** to ***. Then explain the meaning of the value (the vector). Otherwise, it may be a bit confusing.

src/self_driving/pilot_util.cpp
```diff
@@ -132,11 +277,21 @@ void PilotUtil::GroupFeaturesByOU(
...
+      ou_it.GetAllAttributes(&predictors);
       ou_to_features->at(ou_it.GetExecutionOperatingUnitType()).push_back(predictors);
     }
-    pipeline_to_ou_position->emplace_back(data_it.query_id_, data_it.pipeline_id_, std::move(ou_positions));
+
+    if (data_it.query_id_ != prev_qid) {
+      pipeline_idx += 1;
+      prev_qid = data_it.query_id_;
+    }
+    SELFDRIVING_LOG_INFO(fmt::format("Fixed qid: {}; ppl_id: {}", static_cast<uint>(pipeline_qids.at(pipeline_idx)),
```
malin1993ml, line 291: This should be `LOG_TRACE` after you finish debugging.

src/self_driving/pilot/pilot.cpp
```diff
@@ -38,11 +40,27 @@ void Pilot::PerformPlanning() {
...
 
-void Pilot::ExecuteForecast() {
+void Pilot::ActionSearch(std::vector<std::pair<const std::string, catalog::db_oid_t>> *best_action_seq) {
+  auto num_segs = forecast_->GetNumberOfSegments();
+  auto end_segment_index = std::min(action_planning_horizon_ - 1, num_segs - 1);
+  std::vector<std::unique_ptr<planner::AbstractPlanNode>> plans;
+  PilotUtil::GetQueryPlans(common::ManagedPointer(this), common::ManagedPointer(forecast_), end_segment_index, &plans);
+  auto mcst = pilot::MonteCarloTreeSearch(common::ManagedPointer(this), common::ManagedPointer(forecast_), plans,
+                                          end_segment_index);
+  mcst.BestAction(simulation_number_, best_action_seq);
```
malin1993ml, line 55: For now, print out information (e.g., which interval, the SQL string) about the selected actions with `LOG_INFO`.

src/self_driving/pilot_util.cpp
```diff
@@ -36,49 +166,45 @@ const std::list<metrics::PipelineMetricRawData::PipelineData> &PilotUtil::Collec
...
+        common::ManagedPointer(&(forecast->query_id_to_param_types_[qid])), common::ManagedPointer(stmt_list), db_oid,
+        pilot->stats_storage_, forecast->optimizer_timeout_);
 
+    execution::compiler::ExecutableQuery::query_identifier.store(qid);
+    auto exec_query = execution::compiler::CompilationContext::Compile(*out_plan, exec_settings, accessor.get(),
+                                                                       execution::compiler::CompilationMode::OneShot);
+    pipeline_qids->push_back(qid);
+    for (auto &params : forecast->query_id_to_params_[qid]) {
       auto exec_ctx = std::make_unique<execution::exec::ExecutionContext>(
           db_oid, common::ManagedPointer(txn), callback, out_plan->GetOutputSchema().Get(),
```
malin1993ml, line 196: Actually, we need to create a new transaction for each query and then abort the transaction right away as well. Forgot to mention that before. The `ExecutableQuery` can be reused across transactions. But the original transaction that creates the `ExecutableQuery` also needs to be aborted.

malin1993ml (changes requested):
The changes look good! I mostly just have some documentation requests. Also, the CI failure needs to be fixed.

src/include/self_driving/forecast/workload_forecast.h
```diff
@@ -42,7 +53,7 @@ class WorkloadForecast {
   uint64_t num_sample_{5};
 
   std::vector<WorkloadForecastSegment> forecast_segments_;
-  uint64_t num_forecast_segment_{0};
+  uint64_t num_forecast_segment_;
   uint64_t forecast_interval_;
   uint64_t optimizer_timeout_{10000000};
```
malin1993ml, line 58: Our convention is to use Getters and Setters for **all** member variable access unless it's a super simple data struct. This can give us control on whether the member is changeable, whether there should be any validation check before updating a member, etc. And for read-only access, we generally use a Getter with **const reference**. Could you refactor the access to all the members in `WorkloadForecast` with Getters?

src/include/self_driving/forecast/workload_forecast.h
```diff
@@ -25,8 +28,16 @@ class WorkloadForecast {
...
+  /**
+   * Get number of forecasted segments
+   * @return number of forecasted segments
+   */
+  uint64_t GetNumberOfSegments() { return num_forecast_segment_; }
+
  private:
   friend class PilotUtil;
+  friend class Pilot;
+  friend class pilot::MonteCarloTreeSearch;
```
malin1993ml, line 40: You can still keep the friend class if you want to restrict the Getters/Setters to a few specific classes, but don't use friend class as a way to access private members.

src/include/self_driving/forecast/workload_forecast_segment.h
```diff
@@ -21,6 +24,7 @@ class WorkloadForecastSegment {
 
  private:
   std::unordered_map<execution::query_id_t, uint64_t> id_to_num_exec_;
+  friend class PilotUtil;
```
malin1993ml, line 27: Same here, use a Getter that returns a const reference.

src/self_driving/pilot/pilot.cpp
```diff
@@ -38,11 +40,27 @@ void Pilot::PerformPlanning() {
...
 }
 
-void Pilot::ExecuteForecast() {
+void Pilot::ActionSearch(std::vector<std::pair<const std::string, catalog::db_oid_t>> *best_action_seq) {
+  auto num_segs = forecast_->GetNumberOfSegments();
+  auto end_segment_index = std::min(action_planning_horizon_ - 1, num_segs - 1);
+  std::vector<std::unique_ptr<planner::AbstractPlanNode>> plans;
+  PilotUtil::GetQueryPlans(common::ManagedPointer(this), common::ManagedPointer(forecast_), end_segment_index, &plans);
+  auto mcst = pilot::MonteCarloTreeSearch(common::ManagedPointer(this), common::ManagedPointer(forecast_), plans,
+                                          end_segment_index);
```
malin1993ml, line 54: I just realized another thing when I was testing the pilot against a workload with potential indexing actions. So when we get these plans with `GetQueryPlans`, we actually need to make sure that the transaction has not committed/aborted when we call `IndexActionGenerator().GenerateActions`. There're some transactional states that the action generator needs to use because of the current system implementation. So we need to push `GetQueryPlans` into `MonteCarloTreeSearch`, and pass a created `TransactionContext` into `GetQueryPlans`, and only abort that transaction after we have generated all the actions.

malin1993ml (changes requested):
Some additional changes.

src/self_driving/pilot_util.cpp
rabbit721: Sure, I put it in the header file.

src/self_driving/pilot_util.cpp
rabbit721: Sure, I have put it in the header file

src/self_driving/pilot_util.cpp
```diff
@@ -12,19 +12,149 @@
...
+
+void PilotUtil::GetQueryPlans(common::ManagedPointer<Pilot> pilot, common::ManagedPointer<WorkloadForecast> forecast,
+                              uint64_t end_segment_index,
+                              std::vector<std::unique_ptr<planner::AbstractPlanNode>> *plan_vecs) {
+  auto txn_manager = pilot->txn_manager_;
+  auto catalog = pilot->catalog_;
+  transaction::TransactionContext *txn = txn_manager->BeginTransaction();
+
+  for (auto idx = 0; idx <= end_segment_index; idx++) {
+    for (auto &it : forecast->forecast_segments_[idx].id_to_num_exec_) {
```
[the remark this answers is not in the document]
rabbit721: I see, but I think we would want to avoid getting query plans beyond the action planning horizon?

src/self_driving/pilot_util.cpp
malin1993ml: Eh, I was actually referring to the code logic in the cpp file. It's good to add a little bit of explanation along with the code so it's easier for others to understand if they ever need to refactor the implementation.

src/self_driving/pilot_util.cpp
malin1993ml: Same here. I was actually referring to the code logic of the function implementation in the cpp file.

src/self_driving/pilot_util.cpp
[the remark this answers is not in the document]
malin1993ml: Emm you're actually correct. Then maybe we should use a `set` to main the unique query ids up to the planning horizon.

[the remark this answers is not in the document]
rabbit721: Got it, that makes sense, thanks!

malin1993ml (approved):
LGTM! Thanks for the great work!

malin1993ml:
I don't think the performance decrease is real because this part of the code is not even invoked in any performance tests.

PR #1437, opened 2021-01-17
Skip recording the un-modeled LIMIT operating unit

malin1993ml:
We don't model the LIMIT operating unit since it does not have its own operator. But we still record this OpUnit in the `OperatingUnitRecorder`, which causes inference error unless we skip it manually. This PR skips recording the LIMIT OpUnit in `OperatingUnitRecorder`.


17zhangw (approved):
LGTM!

PR #1438, opened 2021-01-18
Add validation check for self-driving knob actions and add dynamic setting to wal_serialization_interval

malin1993ml:
Certain change knob actions may become invalid by exceeding the param min/max value, which can only be known at runtime. This PR adds the runtime check for that.

The PR also adds the ability to dynamically change the `wal_serialization_interval` knob (which has been modeled by the OU models), and removes immutable `wal_persist_threshold` from the `ChangeKnobValueConfig`.


rabbit721 (approved):
LGTM! Thanks for the great work!

PR #1442, opened 2021-01-19
Switch lightgbm to random forest in modeling_test

malin1993ml:
The self-driving `modeling_test` has been timing out frequently. @lmwnshn noticed that lightgbm has a known hanging issue when multithreading and using forking in Linux at the same time:
https://lightgbm.readthedocs.io/en/latest/FAQ.html#lightgbm-hangs-when-multithreading-openmp-and-using-forking-in-linux-at-the-same-time
which might be the reason for the time out.

Hence this PR switches lightgbm to random forest in the testing pipeline to see whether the time out is alleviated.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -135,7 +135,7 @@ pipeline {
                     agent {
                         docker {
                             image 'noisepage:focal'
-                            args '-v /jenkins/ccache:/home/jenkins/.ccache'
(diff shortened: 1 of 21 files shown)


malin1993ml:
@lmwnshn The PR passes the CI three times in a row! Never happened! Looks like the LightGBM hanging is indeed the issue.

PR #1446, opened 2021-01-25
Added wal_async_commit_enable setting, defaulted to false

mbutrovich:
### Summary
This PR implements something similar to Postgres' [asynchronous commit WAL setting](https://www.postgresql.org/docs/current/wal-async-commit.html), albeit not at a per-transaction granularity as Postgres allows.

### Background
Currently, if WAL is enabled then on transaction COMMIT the `TransactionManager` pushes a transaction's `CommitRecord` down to the `LogManager`, which in turn pushes it to the `DiskSerializerTask`. Once a transaction's records are serialized, it is removed from the running transactions table by the `DiskSerializerTask`, and then the callback is pushed to the `DiskLogConsumerTask`. Finally, once a transaction's records are persisted to disk (via `fdatasync`) its callback can be invoked and the client receives an acknowledgement of COMMIT. This coordination is expensive for short-lived queries, where the overhead of synchronizing data structures, threads scheduling, and the latency of the `fdatasync` system call can dominate transaction processing time. However, these steps are all necessary for full ACID guarantees (namely durability, though one could argue consistency because without synchronous commit, a client may see a result that is not reproducible after crash-and-recovery if its results were dependent on another COMMIT transaction that was not persisted to the WAL yet).

### Changes
There is a new command-line setting called `wal_async_commit_enable` that is defaulted to `false`, which maintains the current behavior of the system. If the setting is set to `true` (which only works if WAL is enabled) then on transaction COMMIT the `TransactionManager` still pushes a transaction's `CommitRecord` down to the `LogManager`, however the callback is swapped with the no-op `EmptyCallback` and the real callback is invoked immediately. This reduces the latency between when the system receives COMMIT and returns acknowledgement to the client.

### Caveats
Asynchronous commit in NoisePage comes with all of the caveats described in the Postgres documentation: if the system crashes after COMMIT but before the WAL worker syncs data to disk, there is the possibility of losing data.

### Performance

Some quick numbers from my NUC with an NVMe drive:

**oltpbench TPC-C, scale factor = 1, terminals = 1**
| WAL Configuration | Txns/sec |
|-|-|
| Disabled | 1655 |
| Enabled (wal_async_commit_enable=true) | 1623 |
| Enabled (wal_async_commit_enable=false) | 290 |

**oltpbench TPC-C, scale factor = 6, terminals = 6**
| WAL Configuration | Txns/sec |
|-|-|
| Disabled | 5755 |
| Enabled (wal_async_commit_enable=true) | 5312 |
| Enabled (wal_async_commit_enable=false) | 1048 |

**oltpbench YCSB-read only, scale factor = 1200, terminals = 1**
| WAL Configuration | Txns/sec |
|-|-|
| Disabled | 24328 |
| Enabled (wal_async_commit_enable=true) | 22999 |
| Enabled (wal_async_commit_enable=false) | 3934 |

**oltpbench YCSB-read only, scale factor = 1200, terminals = 6**
| WAL Configuration | Txns/sec |
|-|-|
| Disabled | 80595 |
| Enabled (wal_async_commit_enable=true) | 69375 |
| Enabled (wal_async_commit_enable=false) | 20972 |

### Discussion
- Is this useful for the full system or should I keep it for my experimental branches? My motivation for opening a PR is that Postgres exposes it as a setting, and SingleStore/MemSQL's default configuration is asynchronous durability.
- Should there be a configuration in CI for benchmarking this? My intuition is no, since WAL disabled and enabled both serve as the bounds for this configuration's performance.

### Remaining Tasks

- [ ] If we want to move forward then all of the tests and benchmarks will need a quick refactor to reflect the `TransactionManager`'s new signature.

### DIFF
--- a/src/include/main/db_main.h
+++ b/src/include/main/db_main.h
@@ -74,6 +74,9 @@ class DBMain {
     /**
      * @param buffer_segment_pool non-null required component
      * @param gc_enabled argument to the TransactionManager
+     * @param wal_async_commit_enable true if commit callbacks should be invoked by TransactionManager at commit time
+     * rather than waiting until durable on disk and being invoked by the WAL worker. Doesn't make sense to set to true
+     * if WAL is not enabled.
      * @param log_manager argument to the TransactionManager
      */
     TransactionLayer(const common::ManagedPointer<storage::RecordBufferSegmentPool> buffer_segment_pool,
--- a/src/include/main/db_main.h
+++ b/src/include/main/db_main.h
@@ -77,14 +77,17 @@ class DBMain {
      * @param log_manager argument to the TransactionManager
      */
     TransactionLayer(const common::ManagedPointer<storage::RecordBufferSegmentPool> buffer_segment_pool,
-                     const bool gc_enabled, const common::ManagedPointer<storage::LogManager> log_manager) {
+                     const bool gc_enabled, const bool wal_async_commit_enable,
+                     const common::ManagedPointer<storage::LogManager> log_manager) {
       NOISEPAGE_ASSERT(buffer_segment_pool != nullptr, "Need a buffer segment pool for Transaction layer.");
+      NOISEPAGE_ASSERT(!wal_async_commit_enable || (wal_async_commit_enable && log_manager != DISABLED),
+                       "Doesn't make sense to enable async commit without enabling logging.");
       timestamp_manager_ = std::make_unique<transaction::TimestampManager>();
       deferred_action_manager_ =
           std::make_unique<transaction::DeferredActionManager>(common::ManagedPointer(timestamp_manager_));
-      txn_manager_ = std::make_unique<transaction::TransactionManager>(common::ManagedPointer(timestamp_manager_),
-                                                                       common::ManagedPointer(deferred_action_manager_),
-                                                                       buffer_segment_pool, gc_enabled, log_manager);
+      txn_manager_ = std::make_unique<transaction::TransactionManager>(
+          common::ManagedPointer(timestamp_manager_), common::ManagedPointer(deferred_action_manager_),
+          buffer_segment_pool, gc_enabled, wal_async_commit_enable, log_manager);
     }
 
     /**
@@ -367,8 +370,9 @@ class DBMain {
         log_manager->Start();
       }
 
-      auto txn_layer = std::make_unique<TransactionLayer>(common::ManagedPointer(buffer_segment_pool), use_gc_,
-                                                          common::ManagedPointer(log_manager));
+      auto txn_layer =
+          std::make_unique<TransactionLayer>(common::ManagedPointer(buffer_segment_pool), use_gc_,
+                                             wal_async_commit_enable_, common::ManagedPointer(log_manager));
 
       auto storage_layer =
           std::make_unique<StorageLayer>(common::ManagedPointer(txn_layer), block_store_size_, block_store_reuse_,
@@ -784,6 +788,7 @@ class DBMain {
     int32_t wal_persist_interval_ = 100;
     uint64_t wal_persist_threshold_ = static_cast<uint64_t>(1 << 20);
     bool use_logging_ = false;
+    bool wal_async_commit_enable_ = false;
     bool use_gc_ = false;
     bool use_pilot_thread_ = false;
     bool pilot_planning_ = false;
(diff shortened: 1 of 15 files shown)


mbutrovich:
Ran on dev9 with the WAL on the NVMe drive:

**oltpbench TPC-C, scale factor = 20, terminals = 20**
| WAL Configuration | Txns/sec |
|-|-|
| Disabled | 20794 |
| Enabled (wal_async_commit_enable=true) | 20497 |
| Enabled (wal_async_commit_enable=false) | 16489 |

**oltpbench YCSB-read only, scale factor = 1200, terminals = 20**
| WAL Configuration | Txns/sec |
|-|-|
| Disabled | 277587 |
| Enabled (wal_async_commit_enable=true) | 263431 |
| Enabled (wal_async_commit_enable=false) | 74736 |

src/include/main/db_main.h
```diff
@@ -77,14 +77,17 @@ class DBMain {
      * @param log_manager argument to the TransactionManager
      */
     TransactionLayer(const common::ManagedPointer<storage::RecordBufferSegmentPool> buffer_segment_pool,
-                     const bool gc_enabled, const common::ManagedPointer<storage::LogManager> log_manager) {
+                     const bool gc_enabled, const bool wal_async_commit_enable,
```
thepinetree, line 80: What are your thoughts on stashing `wal_async_commit_enable` in the log_manager instead of in `TransactionLayer`?

src/transaction/transaction_manager.cpp
```diff
@@ -35,14 +35,23 @@ TransactionContext *TransactionManager::BeginTransaction() {
...
                                       timestamp_manager_.Get());
+  } else if (log_manager_ != DISABLED && wal_async_commit_enable_) {
+    // We still want to send this record to the LogManager, but we'll swap in the EmptyCallback to send with the
+    // LogRecord, and invoke the provided callback immediately. The WAL worker will still be responsible for removing it
+    // from the running transactions table once the record has been serialized. Otherwise the serializer might
+    // dereference bad memory if the GC prunes any varlens this transaction might be holding.
+    byte *const commit_record = txn->redo_buffer_.NewEntry(storage::CommitRecord::Size());
+    storage::CommitRecord::Initialize(commit_record, txn->StartTime(), commit_time, TransactionUtil::EmptyCallback,
+                                      nullptr, oldest_active_txn, txn->IsReadOnly(), txn, timestamp_manager_.Get());
+    commit_callback(commit_callback_arg);
```
thepinetree, line 54: This approach generally seems fine but I'm not sure that we account for all the same cases Postgres does:

> Certain utility commands, for instance DROP TABLE, are forced to commit synchronously regardless of the setting of synchronous_commit. This is to ensure consistency between the server's file system and the logical state of the database. The commands supporting two-phase commit, such as PREPARE TRANSACTION, are also always synchronous.

Do we need some sort of check on whether the query was a DDL command?

mbutrovich: We don't support per-transaction configuration of this setting anyway, so it's a bit moot. It's all or nothing for us.

src/include/main/db_main.h
mbutrovich: What do you mean by stashing it? The `TransactionManager` needs this dependency so it needs to be injected into its layer. How would `TransactionManager`'s constructor access the boolean value otherwise?

thepinetree: Nevermind, I thought the `LogManager` class had getters and setters since I saw `SetNumBuffers` but it doesn't do so for any other private variables.

mbutrovich:
Confirmed in meeting today that we do want to merge this. I will get the benchmarks and tests refactored, and then run it through CI, and then get it reviewed.

src/include/main/db_main.h
```diff
@@ -367,8 +373,9 @@ class DBMain {
         log_manager->Start();
       }
 
-      auto txn_layer = std::make_unique<TransactionLayer>(common::ManagedPointer(buffer_segment_pool), use_gc_,
-                                                          common::ManagedPointer(log_manager));
+      auto txn_layer =
+          std::make_unique<TransactionLayer>(common::ManagedPointer(buffer_segment_pool), use_gc_,
```
thepinetree, line 377: I'm completely unsure why this might not be covered by tests but every line in this function is... thoughts?

thepinetree (changes requested):
Minor codecov nit.

src/include/main/db_main.h
thepinetree: Also, codecov in general seems to be confused since it only says ~68% of your diff is hit

thepinetree (approved):
LGTM

mbutrovich:
Not sure where the performance-guard is failing. Numbers look good to me comparing to master.

PR #1450, opened 2021-01-26
Add ANALYZE query

### LINKED ISSUE #716: Add Support for Analyze Operators

jkosh44:
## Description

ANALYZE is a query taken from postgres, you can read more about it
here: https://www.postgresql.org/docs/current/sql-analyze.html.

The ANALYZE query will collect statistics about the contents of a
table and store that information in pg_statistic. You can read
about pg_statistic from postgres here:
https://www.postgresql.org/docs/current/catalog-pg-statistic.html
The optimizer can then take advantage of these statistics when
choosing an optimal query plan.

Note: while ANALYZE and pg_statistic in our system are motivated by
postgres, it does not conform exactly to postgres' specifications.

We currently support two different versions of ANALYZE.
  1. `ANALYZE <table-name>(<column-name-1>, <column-name-2>, ...)`
  2. `ANALYZE <table-name>`
For version 1. we will only analyze the provided columns from the
provided table. For version 2. we will analyze all columns from the
provided table. Postgres provides a third version, `ANALYZE`, that
analyzes all columns from all tables. We do not currently support
this.

Below I will breifly talk about various implementations details

Optimizer:
Within the optimizer we generate a sub SELECT query with aggregates
that will actually compute the relevant statistics over the provided
table. Many of the statistics can be computed using standard SQL
aggregates such as COUNT(*), which gives us the number of rows.
For more complex statistics, such as a histogram, we create our own
custon aggregates that create these data structures and returns a
serialized version of the data structure. We then store those
serialized versions in pg_statistic, so that the optimizer can
deserialize them.

Execution Engine:
The ANALYZE translator roughly performs the following steps
1. Execute the child aggregate query
2. For each column:
    a. Perform an index lookup in pg_statistic for the column
    b. Using a projected row update the row in pg_statistic and all
    indexes.
The implementation is heavily influenced by the index scan and update
implementations. Part 2. of ANALYZE is essentially a series of very
specific UPDATEs that use exact index scans. Note that this has the
implicit assumption that every existing column already has a row in
pg_statistic. We rely on the CREATE TABLE statements to add an empty
row for each column.

Some elements were taken from #900 which was a previous attempt at
implementing ANALYZE.

Closes #716

Co-authored-by: Rohan Aggarwal <rohanaggarwal7997@users.noreply.github.com>
Co-authored-by: Kevin Geng <gengkev@users.noreply.github.com>

## Remaining Tasks
Again, you should only create PR once you are reasonably confident you are near completion. However, if there are some tasks still remaining before the PR is ready to merge, please create a checklist to track active progress. An [example](https://github.com/cmu-db/noisepage/issues/1031) from a PR by @thepinetree follows:

:pushpin: TODOs:
- [x] Add tests

## Performance
I have not done any performance testing for this PR. One thing to mention is that in order to serialize and deserialize TopKElement objects we need to write it out to a file. 

## Further Work
- Add the functionality to `Analyze` that if you leave out a table name, then it will analyze all tables.
- Update the optimizer to use this information in cost modeling.


src/include/catalog/postgres/pg_statistic.h
```diff
@@ -30,13 +36,42 @@ class PgStatistic {
...
+    parser::ExpressionType aggregate_type_;
+    /** Whether the aggregate is distinct */
+    bool distinct_;
+    /** Oid of the column */
+    catalog::col_oid_t column_oid_;
+    /** Human readable name of the column */
+    std::string_view name_;
+  };
+  static constexpr uint8_t NUM_ANALYZE_AGGREGATES = 4;
+  static constexpr std::array<PgStatisticColInfo, NUM_ANALYZE_AGGREGATES> ANALYZE_AGGREGATES = {
```
jkosh44, line 66: I'm not sure if this is the best place for this. It's used by `AnalyzeTranslator`, and `InputColumnDeriver`. If you have a better suggestion please let me know.

src/include/optimizer/statistics/count_min_sketch.h
```diff
@@ -1,15 +1,23 @@
...
+#include <string>
+#include <utility>
 #include <vector>
 
+#include "common/error/exception.h"
+#include "common/json.h"
 #include "common/macros.h"
 #include "loggers/optimizer_logger.h"
 #include "madoka/madoka.h"
+#include "storage/write_ahead_log/log_io.h"
```
jkosh44, line 20: I think it might be better to move the Posix file I/O stuff out of the WAL directory and into it's own file I/O utility class. Thoughts?

apavlo: Yes, this sounds reasonable.

src/execution/compiler/operator/analyze_translator.cpp
```diff
@@ -0,0 +1,405 @@
...
+    auto *loop_init = codegen->MakeStmt(scan_call);
+    // @indexIteratorAdvance(&pg_statistic_index_iterator)
+    auto *advance_call =
+        codegen->CallBuiltin(ast::Builtin::IndexIteratorAdvance, {codegen->AddressOf(pg_statistic_index_iterator_)});
+
+    // for (@indexIteratorScanKey(&pg_statistic_index_iterator); @indexIteratorAdvance(&pg_statistic_index_iterator); )
+    Loop loop(function, loop_init, advance_call, nullptr);
+    {
+      // var pg_statistic_slot = @indexIteratorGetSlot(&pg_statistic_index_iterator)
+      auto pg_statistic_slot = codegen->MakeFreshIdentifier("pg_statistic_slot");
```
jkosh44, line 107: @pmenon I was experiencing a really weird issue with this. I first tried to use the same identifier for each loop iteration (I mean C++ loop iteration not TPL), but my query would keep failing. The specific failure is that on the second table delete the slot value would be identical to the slot value from the first table delete. Since it was the same slot the delete was failing. Specifically `!Visible` was returning true here: https://github.com/cmu-db/noisepage/blob/d60c55a979e200500ad11b50661ad338be2007d6/src/storage/data_table.cpp#L209
because `not_deleted` was returning false here:
https://github.com/cmu-db/noisepage/blob/d60c55a979e200500ad11b50661ad338be2007d6/src/storage/data_table.cpp#L297-L301

I was able to fix it by just declaring a fresh identifier each loop iteration. Looking at the generated TPL the only difference I was able to see was the name of the slot variable. I was never able to figure out why this happened because the slot variable is declared in the TPL loop so it should be a new scope every time, even if it's the same variable name. 

I don't think it's a big issue because it's working now, I just wanted to point it out in case I'm misunderstanding something fundamental here.

apavlo:
@arvindsaik @preetansh I realize that you guys are finished, but would you have time to look at this PR for Joe? Thanks.

src/include/self_driving/modeling/operating_unit_defs.h
```diff
@@ -108,6 +108,10 @@ enum class ExecutionOperatingUnitType : uint32_t {
    * translator (which requires an OpUnit type). We skip this OpUnit in the OperatingUnitRecorder.
    */
   LIMIT,
+  /**
+   * ANALYZE
+   */
+  ANALYZE,
```
jkosh44, line 114: I don't really understand how this is used

test/include/integration/end_to_end_test.h
```diff
@@ -0,0 +1,73 @@
...
+#include "binder/bind_node_visitor.h"
+#include "execution/compiler/compilation_context.h"
+#include "optimizer/cost_model/abstract_cost_model.h"
+#include "optimizer/cost_model/trivial_cost_model.h"
+#include "parser/postgresparser.h"
+#include "traffic_cop/traffic_cop_util.h"
+#include "execution/compiler/output_checker.h"
+#include "execution/sql_test.h"
+
+namespace noisepage::test {
```
jkosh44, line 18: Let me know if you have any thoughts on the placement of this file

test/integration/analyze_test.cpp
```diff
@@ -0,0 +1,492 @@
+
+#include "execution/compiler/output_checker.h"
+#include "integration/end_to_end_test.h"
+#include "optimizer/statistics/histogram.h"
+#include "optimizer/statistics/top_k_elements.h"
+#include "spdlog/fmt/fmt.h"
+#include "test_util/test_harness.h"
+
+namespace noisepage::test {
```
jkosh44, line 9: Let me know if you have any thoughts on the placement of this file

mbutrovich:
Review assignments: me (execution engine), Deepayan (optimizer)

src/include/execution/sql/runtime_types.h
```diff
@@ -415,6 +415,11 @@ class EXPORT Decimal {
   /** Underlying native data type. */
   using NativeType = T;
 
+  /**
+   * Empty constructor
+   */
+  Decimal() = default;
```
jkosh44, line 421: For the person reviewing this, the reason that I added this constructor is because `TopKElements` has a `min_key_` field that gets default constructed. https://github.com/cmu-db/noisepage/blob/1fdb71c57714260c1cc8105c072b365f46855209/src/include/optimizer/statistics/top_k_elements.h#L348-L352
Honestly it's a little suspect because it doesn't seem like there's any logic in that class to check if `min_key_` has been properly initialized or not, but it's an existing class with existing tests so I honestly didn't look too much into it.

jkosh44:
jumbotest_integration  is failing due to a memory leak that #1461 fixes

mbutrovich (commented):
Couple of comments, need to make another pass.

src/include/catalog/postgres/pg_statistic.h
```diff
@@ -30,13 +36,44 @@ class PgStatistic {
...
+   */
+  struct PgStatisticColInfo {
+    /** The type of aggregate to use */
+    parser::ExpressionType aggregate_type_;
+    /** Whether the aggregate is distinct */
+    bool distinct_;
+    /** Oid of the column */
+    catalog::col_oid_t column_oid_;
+    /** Human readable name of the column */
+    std::string_view name_;
```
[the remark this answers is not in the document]
jkosh44: Yes it's just for tpl variable names to make them slightly more readable in case anyone was reading/debuging the TPL. I think using the column OIDs is just as good though.

src/include/catalog/postgres/pg_statistic.h
jkosh44: Maybe if I give some background on to how this is used and why I put it here we can find a better place. The reason I created this object was to try and make it as easy as possible to add columns to `pg_statistic` and try and reduce to number of files that need to be changed. Right now the process looks something like this
1. Add column details to this class
2. Add column details to pg_statistic_impl
3. Add a new entry to this array

The reason I finally settled on this class is because this array is basically describing how to populate the pg_statistic table.

Maybe pg_statistic_impl.h is better since this is more of an implementation detail?

src/include/catalog/postgres/pg_statistic.h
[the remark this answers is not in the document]
jkosh44: Updated

src/optimizer/util.cpp
```diff
@@ -54,19 +55,42 @@ std::vector<parser::AbstractExpression *> OptimizerUtil::GenerateTableColumnValu
...
+  auto *col_expr = new parser::ColumnValueExpression(alias, column.Name());
+  col_expr->SetReturnValueType(column.Type());
+  col_expr->SetDatabaseOID(db_oid);
+  col_expr->SetTableOID(tbl_oid);
+  col_expr->SetColumnOID(col_oid);
+
+  col_expr->DeriveExpressionName();
+  col_expr->DeriveReturnValueType();
+  return col_expr;
+}
```
turingcompl33t, line 78: Newline.

src/optimizer/util.cpp
```diff
@@ -54,19 +55,42 @@ std::vector<parser::AbstractExpression *> OptimizerUtil::GenerateTableColumnValu
...
+parser::AbstractExpression *OptimizerUtil::GenerateAggregateExpr(const catalog::Schema::Column &column,
+                                                                 parser::ExpressionType aggregate_type, bool distinct,
+                                                                 const std::string &alias, catalog::db_oid_t db_oid,
+                                                                 catalog::table_oid_t tbl_oid) {
+  auto col_expr = std::unique_ptr<parser::AbstractExpression>(GenerateColumnValueExpr(column, alias, db_oid, tbl_oid));
+  std::vector<std::unique_ptr<parser::AbstractExpression>> agg_child;
+  agg_child.emplace_back(std::move(col_expr));
+  return new parser::AggregateExpression(aggregate_type, std::move(agg_child), distinct);
+}
```
turingcompl33t, line 87: Newline.

src/include/optimizer/statistics/top_k_elements.h
```diff
@@ -278,6 +241,82 @@ class TopKElements {
...
+    *(top_k_elements.sketch_) = CountMinSketch<KeyType>::FromJson(j.at("sketch"));
+    return top_k_elements;
+  }
+
+  /**
+   * Serialize TopKElements object into byte array
+   * @param[out] size length of byte array
+   * @return byte array representation of TopKElements
+   */
+  std::unique_ptr<byte[]> Serialize(size_t *size) const {
```
turingcompl33t, line 300: Love the use of the array overload for `std::unique_ptr` for dynamic arrays.

src/include/execution/sql/runtime_types.h
```diff
@@ -646,3 +651,115 @@ inline Timestamp Date::ConvertToTimestamp() const { return Timestamp(value_ * K_
 inline Date Timestamp::ConvertToDate() const { return Date(value_ / K_MICRO_SECONDS_PER_DAY); }
 
 }  // namespace noisepage::execution::sql
+
+namespace std {
```
turingcompl33t, line 655: Always a sweet day when you crack open `std`.

src/include/execution/sql/runtime_types.h
turingcompl33t: Thanks for calling this out. It definitely is concerning... The thing making it even a little more difficult to deal with is the default constructed `Decimal` is just going to have it's underlying value defaulted to `0`, so it is indistinguishable from an instance that was constructed with the non-default constructor and a `0` value.

I can't think of a good solution off the top of my head, but something we might be able to add soon once C++20 concepts are stabilized on both GCC and Clang is to do a little concepts-based polymorphism where we overload the constructor for `TopKElements` based on whether the `KeyType` is default constructible (already in `<type_traits>` I believe) and in the event that it is not require that an additional initializer be provided to the constructor.

src/include/execution/sql/aggregators.h
```diff
@@ -364,4 +367,195 @@ class AvgAggregate {
...
+  /**
+   * Return the TopK object.
+   * @return the TopK object.
+   */
+  const optimizer::TopKElements<CppType> &GetTopK() const { return top_k_; }
+
+ private:
+  // Histogram keeping track of the topK elements.
+  optimizer::TopKElements<CppType> top_k_;
+  bool null_{true};
```
turingcompl33t, line 449: Might just be me, and if this is the case definitely ignore this, but I'm not really a fan of the variable name `null_` just because of its similarity to type names (e.g. `NULL` and `nullptr`) - is there anything more descriptive that could be used here?

src/execution/compiler/operator/analyze_translator.cpp
```diff
@@ -0,0 +1,403 @@
...
+  }
+}
+
+void AnalyzeTranslator::PerformPipelineWork(WorkContext *context, FunctionBuilder *function) const {
+  const auto &plan = GetPlanAs<planner::AnalyzePlanNode>();
+  auto *codegen = GetCodeGen();
+
+  NOISEPAGE_ASSERT(plan.GetChild(0)->GetOutputSchema()->NumColumns() ==
+                       catalog::postgres::PgStatistic::NUM_ANALYZE_AGGREGATES * plan.GetColumnOids().size() + 1,
+                   fmt::format("We should 1 aggregate for the number of rows and then {} aggregates per column",
```
turingcompl33t, line 63: No worries, English is tougher when Java is your first language.

turingcompl33t (commented):
Comments in this review are largely superficial, and only a small number actually ask for changes. However I do think I did identify some bugs while testing; going to add those to the conversation on the PR separately.

turingcompl33t:
It looks like `ANALYZE` fails under compiled execution. Steps to reproduce:

```
# CREATE TABLE test (a INT, b INT);
# INSERT INTO test (a, b) VALUES (0, 1), (1, 2);
# ANALYZE test;
```

Output:

```
[2021-02-19 15:58:30.645] [execution_logger] [error] Resolved symbol has no address, you may need to EXPORT: _ZN6madoka6Sketch3addEPKvmm
noisepage: ../src/execution/vm/llvm_engine.cpp:83: virtual llvm::JITSymbol noisepage::execution::vm::LLVMEngine::TPLMemoryManager::findSymbol(const string&): Assertion `(symbol.getAddress().get() != 0) && ("Resolved symbol has no address!")' failed.
Aborted
```

src/execution/compiler/codegen.cpp
```diff
@@ -182,6 +188,8 @@ ast::Expr *CodeGen::Int32Type() const { return BuiltinType(ast::BuiltinType::Int
 
 ast::Expr *CodeGen::Int64Type() const { return BuiltinType(ast::BuiltinType::Int64); }
 
+ast::Expr *CodeGen::Uint32Type() const { return BuiltinType(ast::BuiltinType::Uint32); }
```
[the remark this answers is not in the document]
jkosh44: So I think I probably should have asked this while developing, instead of waiting for the review, but why isn't there already an unsigned integer type? I think I honestly might just be missing or misunderstanding something, I'm pretty sure TPL works with unsigned integers. The reason I added this was for column oids, which aren't unique to ANALYZE. Have we just casted to and from signed integers?

src/include/execution/sql/runtime_types.h
jkosh44: The other thing making me feel slightly better about this is that all the other runtime types have equally useless default constructors.

src/execution/compiler/codegen.cpp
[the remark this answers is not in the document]
jkosh44: Sorry, let me just answer my own question instead of being lazy. Here in the update translator we create an array of column oids (It's what I based the analyze translator code off of that uses TPL unsigned ints):
https://github.com/cmu-db/noisepage/blob/7c79856e4aa242c3747ddd642bec21fd9535d5f7/src/execution/compiler/operator/update_translator.cpp#L149-L160
Specifically this line `ast::Expr *rhs = GetCodeGen()->Const32(all_oids_[i].UnderlyingValue());` 
`all_oids_` is a vector of `col_oid_t`s which is an alias for `uint32_t`. Updates clearly work, but why are we using signed integers in TPL instead of unsigned? I don't really know.

src/execution/compiler/codegen.cpp
```diff
@@ -269,6 +277,40 @@ ast::Expr *CodeGen::AggregateType(parser::ExpressionType agg_type, sql::TypeId r
         return BuiltinType(ast::BuiltinType::IntegerSumAggregate);
       }
       return BuiltinType(ast::BuiltinType::RealSumAggregate);
+    case parser::ExpressionType::AGGREGATE_TOP_K:
+      if (child_type == sql::TypeId::Boolean) {
```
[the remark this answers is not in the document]
jkosh44: I'm sorry I don't really understand this comment (It's been a long day), can you explain what you mean?

src/execution/compiler/codegen.cpp
```diff
@@ -233,7 +241,7 @@ ast::Expr *CodeGen::TplType(sql::TypeId type) {
   }
 }
 
-ast::Expr *CodeGen::AggregateType(parser::ExpressionType agg_type, sql::TypeId ret_type) const {
+ast::Expr *CodeGen::AggregateType(parser::ExpressionType agg_type, sql::TypeId ret_type, sql::TypeId child_type) const {
```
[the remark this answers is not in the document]
jkosh44: So `agg_type` is of type `ExpressionType` which describes which subclass of `AbstractExpression` it is. I think in general the Expressions are meant to work on many different underlying children types. I think we would end up with way too many expression types if we made a separate one for each child type.

src/execution/sema/sema_builtin.cpp
```diff
@@ -601,15 +601,33 @@ void Sema::CheckBuiltinAggregatorCall(ast::CallExpr *call, ast::Builtin builtin)
...
-      // Argument must be a SQL aggregator
-      if (!IsPointerToAggregatorValue(args[0]->GetType())) {
-        GetErrorReporter()->Report(call->Position(), ErrorMessages::kNotASQLAggregate, args[0]->GetType());
-        return;
+
+      ast::BuiltinType::Kind type;
+      if (call->NumArgs() == 1) {
+        // Argument must be a SQL aggregator
+        if (!IsPointerToAggregatorValue(args[0]->GetType())) {
+          GetErrorReporter()->Report(call->Position(), ErrorMessages::kNotASQLAggregate, args[0]->GetType());
```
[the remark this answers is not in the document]
jkosh44: `ReportIncorrectCallArg()` will just call `GetErrorReporter()->Report(call->Position(), ErrorMessages::kIncorrectCallArgType, ...` I thought that `kNotASQLAggregate` was a slightly more specific error message than `kIncorrectCallArgType`, but I think they both work.

src/include/execution/sql/runtime_types.h
jkosh44: Is there a better way of doing this? I wasn't really sure.

src/execution/vm/bytecode_generator.cpp
```diff
@@ -899,6 +899,7 @@ void BytecodeGenerator::VisitBuiltinHashCall(ast::CallExpr *call) {
         break;
       case ast::BuiltinType::Timestamp:
         GetEmitter()->Emit(Bytecode::HashTimestamp, hash_val, input, hash_val.ValueOf());
+        break;
```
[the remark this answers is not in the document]
jkosh44: Yeah lol

src/include/common/posix_io_wrappers.h
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Modernized wrappers around Posix I/O sys calls to hide away the ugliness and use exceptions for error reporting.
+ */
+class PosixIoWrappers {
+ public:
+  /** This class cannot be instantiated. */
+  DISALLOW_INSTANTIATION(PosixIoWrappers);
+
+  // TODO(Tianyu): Use a better exception than runtime_error.
```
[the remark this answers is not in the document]
jkosh44: Yeah, this whole file is just a lift and shift from somewhere else anyway. I didn't actually touch the contents.

test/parser/expression_test.cpp
```diff
@@ -436,6 +436,14 @@ TEST(ExpressionTests, AggregateExpressionTest) {
   EXPECT_THROW(agg_expr_8->DeriveReturnValueType(), ParserException);
 #endif
 
+  // Testing DeriveReturnValueType functionality
+  auto agg_expr_9 = new AggregateExpression(ExpressionType::AGGREGATE_TOP_K, {}, false);
```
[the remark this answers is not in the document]
jkosh44: Classic Java Dev...I moved them to the stack for this test, but I don't think I'm going to switch the rest of them

test/include/integration/end_to_end_test.h
```diff
@@ -0,0 +1,76 @@
...
+    // Optimize
+    auto cost_model = std::make_unique<optimizer::TrivialCostModel>();
+    auto out_plan = trafficcop::TrafficCopUtil::Optimize(
+                        common::ManagedPointer(test_txn_), common::ManagedPointer(accessor),
+                        common::ManagedPointer(stmt_list), test_db_oid_, stats_storage_, std::move(cost_model), 1000000)
+                        ->TakePlanNodeOwnership();
+
+    // Execute
+    execution::compiler::test::OutputStore store{output_checker, out_plan->GetOutputSchema().Get()};
+    execution::exec::OutputPrinter printer(out_plan->GetOutputSchema().Get());
```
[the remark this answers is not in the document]
jkosh44: Removed

test/execution/compiler_test.cpp
```diff
@@ -3643,4 +3643,5 @@ TEST_F(CompilerTest, TPCHQ1Test) {
   executable->Run(common::ManagedPointer(exec_ctx), MODE); checker.CheckCorrectness();
 }
 */
+
```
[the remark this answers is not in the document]
jkosh44: Lol, fixed

src/optimizer/util.cpp
jkosh44: Fixed

src/optimizer/util.cpp
jkosh44: Fixed

src/include/execution/sql/aggregators.h
jkosh44: Switched to `is_null_`

src/include/common/action_context.h
```diff
@@ -16,8 +16,7 @@ namespace noisepage::common {
 STRONG_TYPEDEF_HEADER(action_id_t, uint64_t);
 
 /**
- * @enum common::ActionState
- * @brief The lifecycle states states of internal actions.
+ * The lifecycle states states of internal actions.
```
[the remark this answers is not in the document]
jkosh44: Fixed

src/execution/compiler/operator/analyze_translator.cpp
jkosh44: Fixed

jkosh44:
> It looks like `ANALYZE` fails under compiled execution. This is a `DEBUG` build, with AddressSanitizer enabled, although the latter is likely immaterial. Steps to reproduce:
> 
> Run the DBMS:
> 
> ```shell
> # in build/bin
> $ ./noisepage --compiled_query_execution=true
> ```
> 
> Issue queries:
> 
> ```
> # CREATE TABLE test (a INT, b INT);
> # INSERT INTO test (a, b) VALUES (0, 1), (1, 2);
> # ANALYZE test;
> ```
> 
> Output:
> 
> ```
> [2021-02-19 15:58:30.645] [execution_logger] [error] Resolved symbol has no address, you may need to EXPORT: _ZN6madoka6Sketch3addEPKvmm
> noisepage: ../src/execution/vm/llvm_engine.cpp:83: virtual llvm::JITSymbol noisepage::execution::vm::LLVMEngine::TPLMemoryManager::findSymbol(const string&): Assertion `(symbol.getAddress().get() != 0) && ("Resolved symbol has no address!")' failed.
> Aborted
> ```
> 
> The exact same error occurs in the event we run `ANALYZE test(a, b)` with the columns explicitly specified.

I'll look into this and see what's going on.

lmwnshn:
That usually happens when some symbols need to be exported (see https://github.com/cmu-db/noisepage/blob/master/src/execution/vm/llvm_engine.cpp#L76 )

src/execution/compiler/codegen.cpp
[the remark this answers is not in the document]
jkosh44: Copying what Prashanth said on Slack since it's relevant
> There was never a need for unsigned constant integers during SQL processing.
The update translator logic is wrong. But, it works because the type-checker figures out the constants are all positive anyways.
Feel free to add unsigned versions to the `CodeGen` class.
TPL supports both signed and unsigned integers.
Check out upstream [here](https://github.com/pmenon/tpl/tree/icg) to see if the problem you're having in master has been resolved already. No need to reinvent things.

So it seems like I should add an unsigned const function to codegen. As far as naming goes I'm really open to anything.

jkosh44:
> That usually happens when some symbols need to be exported (see https://github.com/cmu-db/noisepage/blob/master/src/execution/vm/llvm_engine.cpp#L76 )

@lmwnshn What exactly does exporting a symbol mean?

lmwnshn:
@jkosh44 Back when we supported Macs, we needed to [hide symbol visibility](https://github.com/cmu-db/noisepage/blob/master/CMakeLists.txt#L651) by default. For some reason not fully known to anyone afaik, this doesn't mesh well with the [LLVMEngine](https://github.com/cmu-db/noisepage/blob/master/src/execution/vm/llvm_engine.cpp#L76) and we need to add the [EXPORT](https://github.com/cmu-db/noisepage/blob/master/src/include/execution/util/execution_common.h#L7) annotation to various classes so that the engine can find them, for example [IndexIterator](https://github.com/cmu-db/noisepage/blob/ed86192c8ea5c7a075444888eefe833b8e398856/src/include/execution/sql/index_iterator.h#L19).

src/execution/compiler/codegen.cpp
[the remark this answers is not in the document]
turingcompl33t: It really isn't important haha, just mark this one as resolved, we both have better things to worry about.

src/execution/compiler/codegen.cpp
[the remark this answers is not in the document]
turingcompl33t: Is this something you want to bring in with ANALYZE? Or do you think we should:

- Add an issue for it
- Open a separate PR to address it

?

src/execution/compiler/codegen.cpp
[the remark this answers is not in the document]
turingcompl33t: Yep I failed to consider the distinction between the `ExpressionType` and the `child_type` which is a `sql::TypeId`. My mistake, this one is resolved.

src/execution/sema/sema_builtin.cpp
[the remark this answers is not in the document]
turingcompl33t: Dope, happy with that. Resolved.

src/execution/vm/bytecode_generator.cpp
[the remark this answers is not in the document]
turingcompl33t: Resolved.

src/include/common/posix_io_wrappers.h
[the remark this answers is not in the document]
turingcompl33t: Resolved.

src/include/execution/sql/runtime_types.h
turingcompl33t: Nope as far as I am aware this is relatively-common practice to overload `std::hash` for our own types. Resolved.

src/execution/compiler/codegen.cpp
```diff
@@ -85,6 +85,12 @@ ast::Expr *CodeGen::Const64(int64_t val) const {
   return expr;
 }
 
+ast::Expr *CodeGen::ConstU32(uint32_t val) const {
```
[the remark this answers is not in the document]
turingcompl33t: As alluded to above, this really isn't something to address in the ANALYZE PR. If others agree with me on the type naming consistency issue, I can open an issue for this to be refactored later. Otherwise, we can just leave it alone without any negative consequences.

src/execution/compiler/codegen.cpp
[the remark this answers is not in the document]
jkosh44: Let's open up a separate issue about it, we should probably take a look at the upstream and base any change off that.

src/include/optimizer/statistics/top_k_elements.h
```diff
@@ -52,31 +53,20 @@ class TopKElements {
...
     // Increment the count for this item in the sketch
-    sketch_->Increment(key, key_size, delta);
+    sketch_->Increment(key, delta);
 
     // If this key already exists in our top-k list, then
     // we need to update its entry
     auto entry = entries_.find(key);
     if (entry != entries_.end()) {
       entries_[key] += delta;
-      OPTIMIZER_LOG_TRACE("Increment Key[{0}] => {1} // [size={2}]", key, entries_[key], GetSize());
```
thepinetree, line 79: Is there a reason this log trace was removed? Similar with others.

src/include/optimizer/statistics/top_k_elements.h
```diff
@@ -278,6 +241,82 @@ class TopKElements {
...
   }
 
+  /**
+   * Merge Top K Elements with another Top K Elements
+   * @param top_k_elements Top K Elements to merge with this
+   */
+  void Merge(const TopKElements<KeyType> &top_k_elements) {
+    sketch_->Merge(*(top_k_elements.sketch_));
+
+    for (auto &[key, count] : top_k_elements.entries_) {
```
thepinetree, line 251: Hmm is it possible that the order in which the entries are added matters? Say the current min element will be incremented, but it is removed because a new element is added with a higher value first? Is this a concern even?

src/include/self_driving/modeling/operating_unit_defs.h
```diff
@@ -108,6 +108,10 @@ enum class ExecutionOperatingUnitType : uint32_t {
    * translator (which requires an OpUnit type). We skip this OpUnit in the OperatingUnitRecorder.
    */
   LIMIT,
+  /**
+   * ANALYZE
```
thepinetree, line 112: Can you make an issue for this an point it to Lin? I am not sure if they want to record metrics for Analyze queries, but if not, it would be nice to have a comment mentioning that Analyze OpUnits are being skipped in the recorder as above.

src/optimizer/child_property_deriver.cpp
```diff
@@ -257,4 +257,9 @@ void ChildPropertyDeriver::Visit(UNUSED_ATTRIBUTE const DropView *drop_view) {
   output_.emplace_back(new PropertySet(), std::vector<PropertySet *>{});
 }
 
+void ChildPropertyDeriver::Visit(UNUSED_ATTRIBUTE const Analyze *analyze) {
+  // Analyze does not provide any property
```
thepinetree, line 261: Can you make property->properties?

src/traffic_cop/traffic_cop.cpp
```diff
@@ -409,10 +410,11 @@ TrafficCopResult TrafficCop::RunExecutableQuery(const common::ManagedPointer<net
                    "Not in a valid txn. This should have been caught before calling this function.");
   const auto query_type = portal->GetStatement()->GetQueryType();
   const auto physical_plan = portal->OptimizeResult()->GetPlanNode();
-  NOISEPAGE_ASSERT(query_type == network::QueryType::QUERY_SELECT || query_type == network::QueryType::QUERY_INSERT ||
-                       query_type == network::QueryType::QUERY_CREATE_INDEX ||
-                       query_type == network::QueryType::QUERY_UPDATE || query_type == network::QueryType::QUERY_DELETE,
-                   "CodegenAndRunPhysicalPlan called with invalid QueryType.");
+  NOISEPAGE_ASSERT(
+      query_type == network::QueryType::QUERY_SELECT || query_type == network::QueryType::QUERY_INSERT ||
```
thepinetree, line 414: Can you make this one aligned with the other ones as well?

thepinetree (changes requested):
Things look good on the optimizer! A few minor changes requested on other parts. I haven't looked at the `test/` files on this pass, nor am I taking a look at anything in  `execution/` or `include/execution/`.

src/include/optimizer/statistics/count_min_sketch.h
```diff
@@ -40,25 +51,35 @@
...
+  CountMinSketch &operator=(CountMinSketch &&other) noexcept {
+    total_count_ = std::move(other.total_count_);
+    sketch_.copy(other.sketch_);
+    return *this;
+  }
 
   /**
    * Increase the count for a key by a given amount.
    * The key does not need to exist in the sketch first.
+   * This is a convenience method for those KeyTypes that have the
```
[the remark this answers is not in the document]
jkosh44: This comment was actually here before and the diff just makes it look like I wrote it. As far as I know `sizeof` should work for all types, so I think I'll just remove the comment.

src/include/optimizer/statistics/top_k_elements.h
jkosh44: They didn't compile with all the different key types, and I just thought it wasn't worth the effort to get it to work since they don't add a ton of value. 

I'd be happy to try and add them back though if you disagree.

src/include/optimizer/statistics/top_k_elements.h
jkosh44: I haven't traced through it to confirm this 100%, but I think the order could potentially lead to a different end state. However I think that's because it's a probabilistic data structure, and since it's a probabilistic data structure we can't expect it to be 100% correct. So I think that it's not a concern, what do you think?

src/include/self_driving/modeling/operating_unit_defs.h
jkosh44: https://github.com/cmu-db/noisepage/issues/1488

src/include/optimizer/statistics/top_k_elements.h
lmwnshn: I think tracing is something we generally moved away from after Peloton. Instead we now generally use debuggers, unless you have something multithreaded/over the network going on.

Otherwise if you enabled this, optimizer log trace is really not quite the granularity you want anyway -- if you're trying to trace the actual optimizer you're going to get a ton of spam.

src/include/optimizer/statistics/top_k_elements.h
lmwnshn: If I remember my suffering from taking Woodruff's class correctly, countminsketch is linear meaning (sketch each stream -> sum sketches) is the same as (concatenate streams -> sketch the concatenated stream).

Or in human, shouldn't be a concern.

src/include/optimizer/statistics/count_min_sketch.h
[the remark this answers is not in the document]
thepinetree: Wait, it looks like you only need to `sizeof` on a double, I think the question is more that can every datatype be converted into a double?

src/traffic_cop/traffic_cop.cpp
jkosh44: I keep trying to fix this, but clang format keeps putting it back into this format. When I change it manually, then it fails check-format.

thepinetree: Ah no worries then, just something that caught my eye.

src/include/optimizer/statistics/top_k_elements.h
thepinetree: Both sound reasonable to me!

src/optimizer/child_property_deriver.cpp
jkosh44: Fixed

src/include/optimizer/statistics/count_min_sketch.h
```diff
@@ -154,6 +214,185 @@
...
+
+  std::string CreateAndOpenTempFile(int *fd) const {
+    try {
+      auto temp_dir = std::filesystem::temp_directory_path();
+      // We use this CountMinSketch's address in memory in the file name to garauntee that the file name is unique in
+      // case multiple instances of CountMinSketch are being serialized at the same time
+      std::string file_name =
+          "noisepage_" + std::to_string(reinterpret_cast<const size_t>(reinterpret_cast<const void *>(this)));
+      temp_dir.append(file_name);
+      *fd = storage::PosixIoWrappers::Open(temp_dir.c_str(), O_CREAT | O_RDWR, S_IRWXU);
```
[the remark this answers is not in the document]
jkosh44: No, good catch, fixed

src/include/optimizer/statistics/count_min_sketch.h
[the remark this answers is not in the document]
jkosh44: Removed

src/include/self_driving/modeling/operating_unit_defs.h
malin1993ml: Please see my response in #1488. Basically adding a comment here would be fine.

jkosh44: Also added a comment

jkosh44:
@turingcompl33t and @thepinetree other than the symbol exporting issue, I believe I've addressed/responded to all of your comments, so please let me know if I missed anything.

jkosh44:
The linker issue is fixed thanks to @turingcompl33t

turingcompl33t (approved):
Updating my review to explicitly "approve".

test/include/integration/end_to_end_test.h
thepinetree: Personally, don't have a strong opinion, why not test_util though?

test/integration/analyze_test.cpp
thepinetree: Maybe a `sql` directory, again no strong thoughts though.

test/optimizer/histogram_test.cpp
```diff
@@ -147,4 +147,74 @@ TEST_F(HistogramTests, OutputTest) {
...
+  Histogram<int> h{100};
+  auto boundaries = h.Uniform();
+  EXPECT_EQ(boundaries.size(), 0);
+
+  EXPECT_EQ(h.EstimateItemCount(0), 0);
+  h.Increment(5);
+  EXPECT_EQ(h.EstimateItemCount(3), 0);
+  EXPECT_EQ(h.EstimateItemCount(4), 0);
+  EXPECT_EQ(h.EstimateItemCount(5), 1);
+  EXPECT_EQ(h.EstimateItemCount(6), 1);
```
thepinetree, line 180: Mind adding a short comment here why this is one?

thepinetree (commented):
Took a lighter pass on non-`execution` tests. Nothing particular stood out, only a few comments on some tests on location/documentation that is ok to merge without.

test/include/integration/end_to_end_test.h
jkosh44: I actually think test_util looks better, I probably just didn't see it

test/integration/analyze_test.cpp
jkosh44: @thepinetree what do you mean by a `sql` directory?

test/optimizer/histogram_test.cpp
jkosh44: Done

test/include/integration/end_to_end_test.h
jkosh44: Done

test/integration/analyze_test.cpp
jkosh44: Moved

lmwnshn (approved):
Rubber stamping.

PR #1451, opened 2021-01-26
Fix default values of mismatched types

mbutrovich:
### Overview
Postgres lets you do this:
```sql
matt=# create table foo (id int default '1');
CREATE TABLE
Time: 9.845 ms
matt=# insert into foo values (default);
INSERT 0 1
Time: 1.021 ms
matt=# select * from foo;
 id 
----
  1
(1 row)

Time: 0.248 ms
```
Basically you can have a default value that is not the same type as the schema column. It gets cast at insert time, which can be verified by setting a default that doesn't actually work. Postgres also lets you do this:
```sql
matt=# create table foo (id smallint default 3429089082);
CREATE TABLE
Time: 1.317 ms
matt=# insert into foo values (default);
ERROR:  22003: smallint out of range
LOCATION:  int82, int8.c:1313
Time: 1.315 ms
```
We now match this behavior. This was found while trying to run Wikipedia via oltpbench using the Postgres DDL.
### Changes
- Relax assertion in `Schema::Column` that default value's type matches column type, because Postgres doesn't require this.
- Set desired type in `BinderSherpa` for default values before visiting them in order to handle casting.
- Add tracefile test.

### DIFF
--- a/script/testing/junit/traces/insert.test
+++ b/script/testing/junit/traces/insert.test
@@ -267,3 +267,28 @@ SELECT c2 FROM insert13 WHERE c1 IS NOT NULL
 
 statement ok
 DROP TABLE insert13
+
+# Support default value expressions of compatible types
+statement ok
+CREATE TABLE insert14 (id integer default '1')
+
+statement ok
+INSERT INTO insert14 VALUES (DEFAULT)
+
+query I nosort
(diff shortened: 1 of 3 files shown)


lmwnshn (approved):
LGTM

PR #1453, opened 2021-01-29
Add predict_ou_only option to GlobalTrainer

malin1993ml:
Right now the GlobalTrainer script provides both the functionality of evaluating the OU models and training the interference model. This PR adds a `predict_ou_only` option to GlobalTrainer so that we can only do the OU model prediction without the interference model training when that's not necessary. This should be useful for @mbutrovich to accelerate the iterations on testing the online data collection for OU models.

I don't really think this needs to run through CI because this part of the code is not integrated into the tests yet anyway.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -135,7 +135,7 @@ pipeline {
                     agent {
                         docker {
                             image 'noisepage:focal'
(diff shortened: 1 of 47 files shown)


mbutrovich (approved):
LGTM! Thanks Lin!

PR #1455, opened 2021-01-29
Refactor LLVM Engine with Configurable Bytecode Handlers Path

turingcompl33t:
## Description

This PR makes a number of updates to support configuration of the bytecode handlers bitcode file within the LLVM engine:

- Adding the bytecode handlers file path to the settings manager
- Refactoring the LLVM engine itself to accept engine-wide settings as a parameter
- Updates to the build system to ensure that the bytecode handlers are generated for any of the targets that require access to them

As a short example, on `master` it is currently possible to execute queries in compiled mode by running the DBMS server in the following way:

```
# in build/bin/
./noisepage --compiled_query_execution=true
```

This works fine because the bytecode handlers file loaded by the LLVM engine is located in `build/bin/` and the path to the file is hardcoded to `./bytecode_handlers_ir.bc`. However, as soon as we try to run the DBMS like:

```
# in build/
./bin/noisepage --compiled_query_execution=true 
```
we receive an error because the LLVM engine cannot locate the bytecode handlers file in the current working directory for the DBMS server process (`build/`). With this PR, we remedy this situation by making the path configurable via the settings manager:

```
# in build
./bin/noisepage --compiled_query_execution=true --bytecode_handlers_path=./bin/bytecode_handlers_ir.bc
```

Now the LLVM engine is successfully able to locate the file at runtime. 

Essentially, this PR just makes this configuration parameter a first-class citizen of the system - like all of the other parameters configurable via the settings manager. This reduces the "hidden configuration" in the system and also makes it easier to run automated tests that depend on compilation.

## Remaining tasks

- Address `TODO`s in code; primary one being the way in which state (`LLVMEngine::Settings`) is managed in the context of the `LLVMEngine` class itself.
- Mark "ready for CI" and ensure that build system updates pass CI.

## Performance

Unit test runtime may be negatively affected by this PR because:

- The current setup searches for the bytecode handlers file in each of the test setups that require compilation (currently just a single test)
- Tests will now be run in both interpreted and compiled execution modes (currently just a single test)

## Further work

- Utilize the support this adds to run `compiler_test` in compiled execution mode


build-support/generate-bitcodes.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# generate-bitcodes.sh
+#
+# Generate bytecode handlers bitcode file, if necessary.
+# If the file is already present in the provided directory,
+# this script is a no-op.
```
lmwnshn, line 6: I think this changes the semantics of what currently happens.
Previously, as long as you rebuild `noisepage_static`, the bytecode handlers would get updated.
I think we still want that old behavior -- some of the oddest bugs are from having an outdated/incompatible bytecode handler.

I think keeping the old behavior would just be updating the comment + deleting the if test below.
But I may be missing some reasons for your change.

src/execution/vm/llvm_engine.cpp
```diff
@@ -21,6 +21,7 @@
 #include <llvm/Transforms/IPO/AlwaysInliner.h>
 #include <llvm/Transforms/IPO/PassManagerBuilder.h>
 
+#include <iostream>
```
lmwnshn, line 24: Temporary debug include?

src/include/execution/vm/llvm_engine.h
```diff
@@ -34,15 +35,17 @@ class LLVMEngine {
...
+  class Settings;
 
   // -------------------------------------------------------
   // Public API
   // -------------------------------------------------------
 
   /**
-   * Initialize the whole LLVM subsystem
+   * Initialize the whole LLVM subsytem and LLVM engine settings.
+   * @param settings
```
lmwnshn, line 46: Please add doc

src/include/execution/vm/llvm_engine.h
```diff
@@ -57,6 +60,11 @@ class LLVMEngine {
    */
   static std::unique_ptr<CompiledModule> Compile(const BytecodeModule &module, const CompilerOptions &options);
 
+  /**
+   * @return A reference to the LLVM engine settings.
```
lmwnshn, line 64: *constant pointer

src/include/execution/vm/llvm_engine.h
```diff
@@ -190,6 +193,59 @@ class LLVMEngine {
...
+  /**
+   * Engine-wide settings that apply for the entire process in which the LLVM Engine runs.
+   */
+  class Settings {
+   public:
+    /**
+     * Construct a settings instance from the relevant configuration parameters.
+     * @param bytecode_handlers_path The path to the bytecode handlers bitcode file.
+     */
+    explicit Settings(std::string_view bytecode_handlers_path) : bytecode_handlers_path_{bytecode_handlers_path} {}
```
lmwnshn, line 210: If we're going to construct a string anyway, would it make sense to have the caller construct the string and take ownership of it here?

src/include/execution/vm/llvm_engine.h
```diff
@@ -190,6 +193,59 @@ class LLVMEngine {
...
+  };
+
+  // -------------------------------------------------------
+  // Static Data Members
+  // -------------------------------------------------------
+
+  /**
+   * Process-wide LLVM engine settings.
+   *
+   * TODO(Kyle): I'm not particularly happy with this setup - an inline
```
lmwnshn, line 228: I'm also not a huge fan of this setup. No ideas currently come to mind though.

test/execution/atomics_test.cpp
```diff
@@ -11,13 +11,27 @@
...
 namespace noisepage::execution::test {
 
 class AtomicsTest : public TplTest {
  public:
-  AtomicsTest() : region_("atomics_test") { vm::LLVMEngine::Initialize(); }
+  static void SetUpTestSuite() {
+    // TODO(Kyle): We could just as easily locate the project root path here and then
+    // manually concatenate the path at which we expect the bytecode handlers file will
+    // be located, but this approach feels more robust, at the cost of some additional search.
+    bytecode_handlers_path =
```
lmwnshn, line 25: Seems like logic that you might want to extract into `TplTest` or even some new `TplBytecodeTest`.

lmwnshn: Setup + TearDown + getter in particular

test/include/test_util/fs_util.h
```diff
@@ -0,0 +1,26 @@
+#pragma once
+
+#include <string>
+#include <string_view>
+
+// TODO(Kyle): should this functionality be wrapped in another nested namespace?
+// Maybe its just me that feels putting fs stuff out into the top-level namespace
+// when std wraps it up into its own...
+namespace noisepage {
```
lmwnshn, line 9: `noisepage::common`?

test/include/test_util/fs_util.h
```diff
@@ -0,0 +1,26 @@
...
+ * @return The absolute path to the project root, as a string.
+ */
+std::string GetProjectRootPath();
+
+/**
+ * Determine the absolute path to the file with the given name by performing a
+ * recursive directory search from the specified root path.
+ * @param filename The name of the target file
+ * @param root_path The absolute path from which recursive search should begin
+ * @return The absolute path to a file with the specified name as a non-owning string reference.
```
lmwnshn, line 21: We return an actual string, I think?

test/test_util/fs_util.cpp
```diff
@@ -0,0 +1,42 @@
...
+// this check ensures that we fail at compile time in the event
+// that this definition is not present, rather than discovering
+// this at runtime and throwing an exception at test initialization.
+#ifndef NOISEPAGE_PROJECT_ROOT
+#error "NOISEPAGE_PROJECT_ROOT is not defined!"
+#endif
+
+// Necessary grossness for stringizing macro expansion.
+#define STR(x) #x
+#define XSTR(x) STR(x)
```
lmwnshn, line 16: Pick slightly longer names and undef at end of file?

Just in case it gets globbed into some other files with unity builds.

build-support/generate-bitcodes.sh
```diff
@@ -0,0 +1,31 @@
...
+# The path to clang
+CLANG_PATH=$3
+
+# Determine if the file is already present
+if [ -f "$BINARY_DIR/bin/bytecode_handlers_ir.bc" ]; then
+  exit 0
+fi
+
+# Run the compiler
+${PROJECT_ROOT}/build-support/tpl_bytecode_handlers_ir_compiler.py ${CLANG_PATH} ${BINARY_DIR} ${PROJECT_ROOT}/util/execution/bytecode_handlers_ir.cpp ${BINARY_DIR}/bin/bytecode_handlers_ir.bc || exit 1
```
lmwnshn, line 23: I'm generally a fan of printing out the bash commands that get executed by scripts. `set +x` or similar?

CMakeLists.txt
```diff
@@ -962,9 +975,16 @@ foreach (NOISEPAGE_TEST_CPP ${NOISEPAGE_SELF_DRIVING_E2E_TEST_SOURCES})
         set(EXCLUDE_OR_NOT "NO_EXCLUDE")
     endif ()
 
-    add_noisepage_test(${NOISEPAGE_TEST} ${NOISEPAGE_TEST_CPP} self_driving_e2e_test ${EXCLUDE_OR_NOT} NO_UNITY)
+    add_noisepage_test(${NOISEPAGE_TEST} ${NOISEPAGE_TEST_CPP} self_driving_e2e_test ${EXCLUDE_OR_NOT} NO_UNITY FALSE)
 endforeach ()
 
+# TODO(Kyle): this is not ideal because it requires manual synchronization with
```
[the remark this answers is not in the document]
turingcompl33t: Removed this `TODO`.

build-support/generate-bitcodes.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
+# generate-bitcodes.sh
+#
+# Generate bytecode handlers bitcode file, if necessary.
+# If the file is already present in the provided directory,
+# this script is a no-op.
+
+# TODO(Kyle): above we assume bash is an available shell, should we be doing this?
```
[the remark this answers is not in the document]
turingcompl33t: Removed the `TODO`.

src/execution/vm/llvm_engine.cpp
turingcompl33t: Good catch! Removed. This reminds me that the other day I was looking at how hard it would be to integrate [include-what-you-use](https://include-what-you-use.org/) into the project - I think it has native CMake support now so it might actually be pretty easy.

build-support/generate-bitcodes.sh
turingcompl33t: Agreed, the cost of re-compiling / re-optimizing is so low compared to a recompilation of the static lib anyway that any savings likely aren't even worth it. I've removed the test so now the script simply performs the three steps necessary to generate the optimized file. I think it still might be worth keeping it as a separate script though because:

- Slightly easier to grok when its in a bash script rather than the `CMakeLists.txt`, and more expressive than just having the free-floating `COMMAND` statements
- More extensible if something changes in the future

If you disagree with the above though I can just move the functionality back to the `CMakeLists.txt`.

src/include/execution/vm/llvm_engine.h
turingcompl33t: Added the documentation.

src/include/execution/vm/llvm_engine.h
turingcompl33t: Agreed, I overload the use of _reference_. Updated.

src/include/execution/vm/llvm_engine.h
turingcompl33t: I like generally like clean ownership transfers like that, but I don't think this is ideal with the current setup of `DBMain` - we have to declare the default path to the file as a member of `DBMain`, so we would have to move the string out of the `DBMain` instance, which works but seems... wrong. Like we shouldn't be moving members out of `DBMain`. There is the additional consideration that accepting the argument as a `std::string_view` enables us to pass string constants to it, which is nice for the API (I think). Interested to hear your thoughts on those two points.

src/include/execution/vm/llvm_engine.h
turingcompl33t: Yea... as the comment says I really tried a number of other approaches, but they all complicated the system so much more than should be necessary for such a small increase in functionality.

test/test_util/fs_util.cpp
turingcompl33t: Agreed. Updated.

test/include/test_util/fs_util.h
turingcompl33t: Indeed we do! Changed the API a couple times and didn't update comment to reflect. Thanks for the catch.

build-support/generate-bitcodes.sh
turingcompl33t: Good point. I think I am going to go with manually `echo`ing though because it just looks a little nicer.

test/include/test_util/fs_util.h
turingcompl33t: Good as any! Updated.

test/execution/atomics_test.cpp
turingcompl33t: Good idea. Pulled it out into a new `CompiledTplTest` because `TplTest` itself is relatively general purpose (used extensively, for many non-compiled tests).

turingcompl33t:
@lmwnshn made one other update in addition to your suggestions: whereas before, for unit test fixtures (e.g. `AtomicsTest`) we initialized the LLVM engine for every new test, I moved this functionality to the `CompiledTplTest` fixture and only invoke it once at the beginning of the test suite, rather than once for every test. I argue this makes sense because, while the LLVM API functions we use in initialization of the engine are safe to call multiple times from the same process, the API we call in `LLVMEngine::Shutdown()` can only be called once from the same process. After this, future calls to `LLVMEngine::Initialize()` don't actually have the desired effect, and the engine fails to completely re-bootstrap. This made me realize that we really should only ever be initializing the LLVM engine once for each process, and the entirety of the test suite runs in a single process, so performing this setup in the function that is invoked at suite-level initialization rather than test-level initialization is logical.

build-support/generate-bitcodes.sh
lmwnshn: Nah, sounds fine to me. Thanks!

src/include/execution/vm/llvm_engine.h
lmwnshn: Sounds good to me

lmwnshn (approved):
LGTM, thanks!

lmwnshn:
Not going to merge until standup where we can talk about performance guard.

PR #1456, opened 2021-01-31
Microbenchmark fix %change mistake

bialesdaniel:
# Microbenchmark %change fix + improved documentation

## Description
Matt found a mistake in the % change calculation on the dashboards so I audited all the percentage change calculations. I found I made the same mistake with the microbenchmark script. The impact is minimal as this is only really used when the results are printed out in the terminal. While I was fixing the mistake I figured I would improve the documentation for the files that I touched. Now they follow the numpydocs formatting. If anyone wants to double check that % change is calculated correctly here is a snippet of the output. 
Screen Shot 2021-01-31 at 12 26 11 PM

## Performance
No impact

## Further Work
None

### DIFF
--- a/script/testing/microbench/README.md
+++ b/script/testing/microbench/README.md
@@ -14,7 +14,7 @@ in TimescaleDB. The results will be visualized at [stats.noise.page](https://sta
 ## Requirements
 
 This script assumes that you have numactl package installed. If you are running the script locally you do not need to
- install `numctl` but you will need to comment out a line in `MicrobenchmarksRunner._build_benchmark_cmd()`. 
+ install `numctl` but you will need to specify `--local`. 
 
 ```
 sudo apt install numactl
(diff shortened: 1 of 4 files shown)


script/testing/microbench/README.md
```diff
@@ -14,7 +14,7 @@ in TimescaleDB. The results will be visualized at [stats.noise.page](https://sta
 ## Requirements
 
 This script assumes that you have numactl package installed. If you are running the script locally you do not need to
- install `numctl` but you will need to comment out a line in `MicrobenchmarksRunner._build_benchmark_cmd()`. 
+ install `numctl` but you will need to specify `--local`.
```
mbutrovich, line 17: Legacy typo but might as well fix it here: `numactl`.

mbutrovich (changes requested):
Minor typo.

PR #1458, opened 2021-02-04
Proper model caching in model server

malin1993ml:
We're not setting up the model map key correctly in the model server so we never cache the model maps. This PR fixes this problem.

This should accelerate the OU model inference through the model server significantly, which is especially useful for @rabbit721's work on the pilot.


malin1993ml:
I don't really think this needs review. It's pretty straight forward and only 4 lines of code.

PR #1459, opened 2021-02-04
Kill forked model server upon parent NoisePage process death.

malin1993ml:
Currently, if the main DBMS thread dies unexpectedly (e.g., crashes) when the model server is enabled, the Python model server process would become a zombie. And we have to kill it manually. This PR registers a parent death signal to the model server child process, so the model server is killed when the parent dies.

**Caveat**: This PR uses `prctl`, which only works on Linux. This won't work on macOS, but we don't support macOS anyway...


lmwnshn (approved):
LGTM.

Looks like the performance guard is quite flaky.

PR #1461, opened 2021-02-06
Fix memory leak associated with aggregators

jkosh44:
## Description
The codegen related to aggregators was never calling
AggregatorFree which frees the memory associated with the
aggregator datastructures. This commit adds the logic needed to
call AggregatorFree.

## Remaining tasks
None

## Performance
N/A

## Further work
None


jkosh44:
I'm slightly worried that the performance drop is real since aggregators now have to do some extra work to clean up when they're done.

jkosh44:
so it seems like the performance numbers are real, I've re-run the build multiple times and it keeps failing. What's weird to me is that the performance for the first commit was fine, but it wasn't until https://github.com/cmu-db/noisepage/pull/1461/commits/60a227ae302056c72c48b11098c28493fc59d2bc that the performance degraded so drastically. Which, if I'm understanding correctly, would suggest that the performance drop come from the call to `aggFree` and not he surrounding logic. When I was accidentally calling `aggReset` instead, the performance was fine. Not really sure what to do about this.

All the existing aggregators don't even have destructors which essentially makes this a no-op (further confusing me about the number). This may be why `aggFree` was never called in the first place. However for #1450 we add aggregators with destructors and may want to add other aggregators with destructors in the future. If `aggFree` really is that expensive we may need to add logic to only call it on aggregators where it's necessary.

pmenon:
Aggregators don't have (nor need) destructors because they don't allocate memory; this is by design. As you're seeing in the perf numbers, your approach has a non-negligible negative impact to performance. But those benchmarks don't stress your diff as much as an analytical benchmark would, e.g., TPC-H. The hit comes not from calling the destructor, but from the additional scan of the hash table.

One alternate solution is to only do the scan if an aggregate exists that requires destruction. But, this is side stepping the root issue. Can you think of other alternatives?

jkosh44:
> Aggregators don't have (nor need) destructors because they don't allocate memory; this is by design. As you're seeing in the perf numbers, your approach has a non-negligible negative impact to performance. But those benchmarks don't stress your diff as much as an analytical benchmark would, e.g., TPC-H. The hit comes not from calling the destructor, but from the additional scan of the hash table.
> 
> One alternate solution is to only do the scan if an aggregate exists that requires destruction. But, this is side stepping the root issue. Can you think of other alternatives?

@pmenon Thanks for the response and explanation. The performance hit coming from the hash table scan makes much more sense to me than from the call to the destructor. I guess the first commit's performance run was just an anomaly.

I'll have to think about this for a bit to see what the best solution is, we can also discuss at the meeting today. Only doing the scan if an aggregate exists that requires destruction might work, but then we get into a tricky situation when we have multiple aggregators and only some of them need destruction. I wonder if it would be possible to call the aggregators destructors from with the call to AggHashTableFree?

jkosh44:
I'm not sure if it's possible to rework the custom aggregators we want to work without allocating memory, though I'll also look and see if it's possible.

jkosh44:
Another crazy hacky idea that just might work, calling Serialize on the aggregator will free it's memory. Since once we Serialize it we no longer need any of it's contents. Though this seems a bit hacky.

jkosh44:
So I created the following branch which just logs to a file (the python script was eating my C++ std out). https://github.com/jkosh44/noisepage/commit/803610fb22c7bf31399e4946435b3799611c6f90
It just logs whether any of the code I wrote was actually called, as well as a single log when the DB starts up as a sanity check that logging is working. Here are the results from the following benchmarks

# TATP
python script: `python script/testing/oltpbench/run_oltpbench.py --config-file=script/testing/oltpbench/configs/end_to_end_performance/tatp_wal_disabled.json --build-type=release`
logs:
[tatp_agg.txt](https://github.com/cmu-db/noisepage/files/5954402/tatp_agg.txt)

# TPCC
python script: `python script/testing/oltpbench/run_oltpbench.py --config-file=script/testing/oltpbench/configs/end_to_end_performance/tpcc_wal_disabled.json --build-type=release`
logs:
[tpcc_agg.txt](https://github.com/cmu-db/noisepage/files/5954396/tpcc_agg.txt)


So TATP doesn't call any of this code. TPCC has a lot of aggregators, but they are all static aggregators not hash aggregators. Static aggregators don't have to iterate through any hash table, so they shouldn't have the performance penalty that we were talking about with regards to hash aggregators. So either the tatp numbers for this PR are wrong or something else that I'm not aware of is going on. Similarly the tpcc numbers are probably wrong or something else is going on.

I think next step is to get a breakdown/flame graph of the benchmarks for this branch and master and compare the two.

jkosh44:
I ran tatp and tpcc 5 times on this branch and master on a dev machine and compared the results. Below are the results

Master:
    TATP:
        Rate limited reqs/s: Results(nanoSeconds=60999484594, measuredRequests=30159) = 494.4140135073614 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60000699965, measuredRequests=30023) = 500.37749588776813 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60000414548, measuredRequests=31725) = 528.7463468209903 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60000458950, measuredRequests=30248) = 504.12947716294093 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60999752128, measuredRequests=31514) = 516.6250501128594 requests/sec
        AVG = 508.86
    
    TPCC:
        Rate limited reqs/s: Results(nanoSeconds=61000022184, measuredRequests=81374) = 1333.9995148615535 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60000093607, measuredRequests=80510) = 1341.831239920052 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60000393603, measuredRequests=84615) = 1410.2407487501762 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60998974411, measuredRequests=84962) = 1392.8430899106843 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60000050724, measuredRequests=79912) = 1331.8655407075385 requests/sec
        AVG = 1362.16

My Agg Free Branch:
    TATP:
        Rate limited reqs/s: Results(nanoSeconds=60999972805, measuredRequests=31461) = 515.7543282940812 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60999874379, measuredRequests=30517) = 500.27971878095985 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60000426596, measuredRequests=29869) = 497.81312724851944 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60000670421, measuredRequests=31238) = 520.6275160063349 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60999729831, measuredRequests=30012) = 492.0021790776511 requests/sec
        AVG = 505.29

    TPCC:
        Rate limited reqs/s: Results(nanoSeconds=61000115654, measuredRequests=86494) = 1417.9317378774228 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60999788673, measuredRequests=87580) = 1435.742678872018 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60999476855, measuredRequests=82590) = 1353.9460378704914 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60999589438, measuredRequests=79880) = 1309.5170104577517 requests/sec
        Rate limited reqs/s: Results(nanoSeconds=60000200581, measuredRequests=85258) = 1420.9619163672976 requests/sec
        AVG = 1387.62

TATP: (505.29 - 508.86)/508.86 = -0.70%
TPCC: (1387.62-1362.16)/1362.16 = +1.87%

This seems to indicate that TATP and TPCC performance isn't heavily affected. 

We still need to actually compare performance on a benchmark that does heavily use hash aggregators, basically anything with GROUP BYs.

jkosh44:
So I just decided to bite the bullet and implement the workaround that @pmenon mentioned, where we only clean up aggregators if we have any that allocate memory. The more I think about the better I feel about it as a solution. Cleaning up aggregator memory is going to be a rare occurrence, so we really shouldn't take the performance hit unless we need to. I also think that there isn't going to be a straight forward way for the ANALYZE related aggregators to not allocate memory.

src/planner/plannodes/aggregate_plan_node.cpp
```diff
@@ -136,6 +136,20 @@ std::vector<std::unique_ptr<parser::AbstractExpression>> AggregatePlanNode::From
   return exprs;
 }
 
+bool AggregatePlanNode::RequiresCleanup() const {
+  return std::any_of(aggregate_terms_.begin(), aggregate_terms_.end(),
+                     [](auto agg_term) { return agg_term->RequiresCleanup(); });
+}
```
turingcompl33t, line 142: Nit of all nits: add a blank line between these two functions.

src/planner/plannodes/aggregate_plan_node.cpp
```diff
@@ -136,6 +136,20 @@ std::vector<std::unique_ptr<parser::AbstractExpression>> AggregatePlanNode::From
...
+std::vector<size_t> AggregatePlanNode::GetMemoryAllocatingAggregatorIndexes() const {
```
turingcompl33t, line 143: Based on my understanding of this functionality, we are only ever going to call this member when tearing down the state for a particular pipeline, so it will only ever be called once by definition. If this is the case, it obviously makes sense to construct the vector when this is called, but would there ever be any situation where this might be called more than once for the same instance, making it worthwhile to build the vector once and cache it for later? Only alternative I can think of would be to make this return some lazy iterable, but that would almost certainly be overkill.

turingcompl33t (changes requested):
One change requested and one question, then we'll get this guy merged.

src/planner/plannodes/aggregate_plan_node.cpp
jkosh44: This method will should only ever get called once per instance. Though now that I'm looking at it it gets called multiple times, which I'm going to fix so it only gets called once. This whole method might be overkill tbh because we can just have the translators iterate through the expressions themselves and see which need cleanup. Let me know your thoughts if you think this method is necessary.

src/planner/plannodes/aggregate_plan_node.cpp
jkosh44: Done

src/planner/plannodes/aggregate_plan_node.cpp
turingcompl33t: I am partial to overkill so my vote is for keeping it! Anything that contributes to expressing what the code is actually doing is a net win in my book, and this member does that for me.

turingcompl33t (approved):
LGTM!

lmwnshn (approved):
Rubber stamping

PR #1466, opened 2021-02-11
Fix missing artifact stats import.

lmwnshn:
In a PR comment, changed inheritance from `object` to `ABC` but forgot to import it.


apavlo (approved):
LGTM

PR #1467, opened 2021-02-12
Proposal: Improve Commit Hash in Testing Reporting 

bialesdaniel:
# Change How Testing Scripts Report Commit Hash

### Problem
Currently, there is an issue with how the testing scripts record the git_commit_id (the commit hash). The commit hash comes from an environment variable set by Jenkins (GIT_COMMIT). When Jenkins checks out a pull requests code from the repository it first tries to merge it with master. If this is possible it creates a new merge commit hash. If it is already up to date with master it will not create a new commit hash. I'm not sure what happens if there are merge conflicts. I would guess that it does not create a merge commit.

The problem is that this merge commit becomes the GIT_COMMIT. Then the test results are stored in the database with that value. If you later search GitHub for that commit you won't find anything. 

This especially causes problems with the performance guard because the performance guard receives events saying Jenkins completed for a particular commit hash. This commit hash is not the merge commit hash. Therefore, the performance guard queries the database for `WHERE git_commit_id = 'hash'` and finds 0 results. Currently, the performance guard handles scenarios like this by creating a "Couldn't find results message". So, this doesn't really break the performance guard, but it makes it useless in certain cases.

### Solution
This is just my proposal for a solution. If someone has a better idea I'm all ears. 

I am proposing using the GitHub API to find the latest commit to the PR. This requires using a GitHub token to make the API call because unauthorized calls have a limit of 60 per hour. We could easily exceed that. The idea would be to pass this token into the scripts and then the scripts would lookup the commit hash when they go to report the test results. 

As I am writing this up I realize a problem with the solution. If a user pushes a commit while the build is running then the latest commit will be used as the git_commit_id instead of the commit that was used in Jenkins. One solution would be to run a script as the first step in the build to get the commit hash (this doesn't eliminate the problem but it makes it much less likely to occur). 
Another more complicated but more accurate solution would be (pseudo code below):

```python
def get_commit_hash():
    jenkins_commit_sha = os.environ['GIT_COMMIT']
    commits = get_all_commits_for_pr(pr_number)
    if any([(commit['sha'] == jenkins_commit_sha) for commit in commits]):
        # It is not a merge commit
         return os.environ['GIT_COMMIT']
    else:
        # it is a merge commit
        merge_commit = get_merge_commit_details_from_github(jenkins_commit_sha)
        latest_master_commits = get_last_10_commits_to_master()
    
        # The parent commits for a merge commit will be the commit to the PR and the master branch commit
        possible_commit_hashs = [parent['sha'] for parent in merge_commit['parents']]
        for hash in possible_commit_hashs:
            if hash not in [commit['sha] for commit in latest_master_commits]:
                # Find the hash that was not a commit to the master branch which means that it was the commit to the 
                # PR.
                return hash
```

## Remaining tasks

- [ ] Create a GitHub access token and store it as a credential in Jenkins. Andy/Chad would have to do this.

### DIFF
--- a/script/testing/artifact_stats/__main__.py
+++ b/script/testing/artifact_stats/__main__.py
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 import argparse
 import logging
+import os
 import sys
 
 from ..reporting.report_result import report_artifact_stats_result
@@ -75,10 +76,16 @@ if __name__ == "__main__":
                         type=str,
                         help="Performance Storage Service password")
 
+    parser.add_argument("--github-token", 
+                        help="GitHub token for calling REST API")
+
     args = parser.parse_args()
 
     if args.debug:
         LOG.setLevel(logging.DEBUG)
+
+    if args.github_token:
+        os.environ['GITHUB_TOKEN'] = args.github_token
 
     # Get the BaseBinaryMetricsCollector subclasses imported from binary_metrics.binary_metrics_collectors
     # Effectively this adds each binary metric collector class into an array to be instantiated later.
--- a/script/testing/reporting/parsers/parse_data.py
+++ b/script/testing/reporting/parsers/parse_data.py
@@ -2,6 +2,7 @@ import os
 import re
 from decimal import Decimal
 
+import requests
 import distro
 
 from ...util.constants import LOG
@@ -96,7 +97,7 @@ def _parse_jenkins_env_vars():
         },
         'github': {
             'git_branch': os.environ['GIT_BRANCH'],
-            'git_commit_id': os.environ['GIT_COMMIT'],
+            'git_commit_id': _get_git_commit_id(os.environ['GIT_BRANCH']),
         },
         'environment': {
             'os_version': ' '.join(distro.linux_distribution()),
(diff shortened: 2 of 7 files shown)


script/testing/artifact_stats/__main__.py
```diff
@@ -75,11 +76,17 @@ def collect_artifact_stats(collectors):
...
+    parser.add_argument("--github-token", 
+                        help="GitHub token for calling REST API")
+
     args = parser.parse_args()
 
     if args.debug:
         LOG.setLevel(logging.DEBUG)
 
+    if args.github_token:
+        os.environ['GITHUB_TOKEN'] = args.github_token
```
bialesdaniel, line 88: I'm not a big fan of using an environment variable but it seemed better than passing it down through a bunch of functions as a parameter.

script/testing/reporting/parsers/parse_data.py
```diff
@@ -106,6 +107,59 @@ def _parse_jenkins_env_vars():
...
+    str
+        The commit hash for the latest commit on the branch.
+    """
+    pr_number_match = re.search('PR-\d+', git_branch)
+    if pr_number_match:
+        # Get the hash from the last commit to the PR.
+        try:
+            pr_number = int(pr_number_match.group(1))
+            commits = _get_git_pr_commits(pr_number)
+            latest_commit = commits[-1]
```
bialesdaniel, line 131: Read the PR text for why this might not be the best solution and my proposed alternative.

lmwnshn:
Interesting. Re: GitHub API, I never noticed we had both [old](https://github.com/cmu-db/noisepage/tree/master/script/github) scripts and the [script I wrote](https://github.com/cmu-db/noisepage/blob/master/build-support/check_github_labels.py) for mucking around with GitHub. Maybe you could cannibalize some of those instead of using a token. Thinking about the other thing you mentioned right now.

bialesdaniel:
> script I wrote

It looks like the old scripts use a token and the new scripts do a combination of unauthenticated API calls and if it hits the rate limit it screen-scrapes. I think it would be hard to screen-scrape the data that I need but maybe it isn't.

jkosh44:
One thing to consider is that if your branch isn't updated with master then your performance numbers are somewhat irrelevant because 
1. Your code doesn't have the latest code, which may affect your branch's performance. If you see a performance difference it's unclear if the change is from your branch's new code, your branch's missing master code, or some combination.
2. If you ever want to merge your branch, then you'll have to update it with master which will re-kick off all the performance tests anyway
I can potentially see a scenario where someone is curious and wants to see performance metrics on a branch that isn't updated, but in that case they should probably just run the performance tests themselves instead of burdening the CI/CD pipeline with a commit that can't be merged.

It may be simpler to just fail or skip the performance guard and leave a message saying something like, "once the branch is updated with master performance guard will run" (if it's possible to check for something like this).

Just a suggestion though, I can definitely see reasons against this.

bialesdaniel:
It almost seems like maybe Jenkins shouldn't run if there are merge conflicts. I'm actually not sure what happens in Jenkins if that is the case.

crd477:
> 
> 
> It almost seems like maybe Jenkins shouldn't run if there are merge conflicts. I'm actually not sure what happens in Jenkins if that is the case.
> 

It doesn't build in that case.  See:
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-1422/26/pipeline
for an example.

bialesdaniel:
Oh awesome! So that means all the performance results are meaningful we just have to resolve the problem of the merge commit vs PR commit.

apavlo:
We want this but it's dead :-(

PR #1468, opened 2021-02-12
Fix nightly oltpbenchmark and microbenchmark builds.

lmwnshn:
# Heading

Fix nightly oltpbenchmark and microbenchmark builds.

## Description

Attempt to fix nightly build.

- Add wait to DBServer stop. I think it was racing between sending the terminate signal to the old process (and not waiting for that to die), and the next iteration of the OLTPBenchmark run, which would cause the new DB process to die since it couldn't grab the port.
- No longer log useless lines.
- Fix path issues for nightly microbenchmark.

### DIFF
--- a/script/testing/util/db_server.py
+++ b/script/testing/util/db_server.py
@@ -91,7 +91,7 @@ class NoisePageServer:
                 LOG.info(f'DB process is verified as running in {round(now - start_time, 2)} sec.')
                 self.db_process = db_process
                 return True
-            else:
+            elif log_line.strip() != '':
                 logs.append(log_line)
 
             if now - start_time >= 60:
@@ -123,6 +123,7 @@ class NoisePageServer:
         return_code = self.db_process.poll()
         if return_code is None:
             self.db_process.terminate()
+            self.db_process.wait()
             LOG.info("DBMS stopped successfully.")
             self.db_process = None
         else:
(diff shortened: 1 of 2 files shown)


script/testing/util/db_server.py
```diff
@@ -123,6 +123,7 @@ def stop_db(self, is_dry_run=False):
         return_code = self.db_process.poll()
         if return_code is None:
             self.db_process.terminate()
+            self.db_process.wait()
```
bialesdaniel, line 126: I'm wondering if its possible that wait could hang forever with this code. It looks like the default timeout on wait is `None` Should we do something like:
```
try:
    self.db_process.terminate()
    self.db_process.wait(60) #60 second timeout for terminating that throws an error if it times out
except:
    self.db_process.kill()
    self.db_process.wait(60) # If this one times out there are problems and we don't catch the error.
```

bialesdaniel (approved):
I added one comment that might be my misunderstanding of `wait()`. Otherwise LGTM!

script/testing/util/db_server.py
lmwnshn: Yeah, it would be possible for wait to hang. hmm. Sounds like a plan, incorporating.

lmwnshn: added, thanks!

Issue #1223, opened 2020-10-02
Run JUnit (SQL) Tests in Multiple Query Execution Modes


turingcompl33t:
## Summary

Currently, SQL tests with JUnit are only run with interpreted query execution (the default). Support should be added such that these tests are run in both interpreted and compiled query execution modes.

## Solution

We will need to add support in the unit testing framework for specifying system parameters each time the DBMS is restarted to run a new test. It looks like this should be simple enough by modifying `test_server.py`(in `script/testing/util/`) to pass commandline arguments when the DBMS subprocess is spawned. This functionality should be generally useful as we move forward and need to ensure correct operation of the system under all of our supported parameter combinations. 

As mentioned above, interpreted query execution is the default, and compiled query execution may be enabled by specifying the `-compiled_query_execution=true` flag when launching the DBMS.

PR #1388, opened 2020-12-09
Fix passing false arguments to testing scripts. Add dry-run functionality.

bialesdaniel:
# HotFix: Wal Disabled OLTPBench Tests

## Description
This fixes #1384. The server args string was not getting created correctly. This fixes the issue with boolean server args. Also, when the DB instance starts it prints the full start command along with the arguments. 

I also took a quick pass at adding `dry-run`, it ended up being much simpler than I thought. I implemented it so that the DB server starts and prints its starting command and then stops, without doing anything. Here is the output.
```
root@2adc8c73e54d:/repo/build# python3 ../script/testing/oltpbench/run_oltpbench.py --config-file=../script/testing/oltpbench/configs/nightly/nightly_wal_disabled.json --build-type=release --dry-run
12-09-2020 17:33:08,998 [common.py:36] INFO : ****** start killing processes on port 15721 ******
12-09-2020 17:33:08,998 [common.py:36] INFO : ****** finish killing processes on port 15721 ******
12-09-2020 17:33:09,009 [db_server.py:48] INFO : Server start: /repo/build/bin/noisepage -wal_file_path=wal.log -wal_enable=false -connection_thread_count=32 -record_buffer_segment_size=1000000 [PID=4703]
12-09-2020 17:33:09,091 [db_server.py:61] INFO : DB process is verified as running in 0.68 sec
12-09-2020 17:33:09,091 [db_server.py:86] INFO : Stopped DB successfully
12-09-2020 17:33:09,649 [common.py:36] INFO : ****** start killing processes on port 15721 ******
12-09-2020 17:33:09,649 [common.py:36] INFO : ****** finish killing processes on port 15721 ******
12-09-2020 17:33:09,662 [db_server.py:48] INFO : Server start: /repo/build/bin/noisepage -wal_file_path=wal.log -wal_enable=false -connection_thread_count=32 -record_buffer_segment_size=1000000 [PID=4743]
12-09-2020 17:33:09,751 [db_server.py:61] INFO : DB process is verified as running in 0.69 sec
12-09-2020 17:33:09,751 [db_server.py:86] INFO : Stopped DB successfully
```
If you prefer it to just print the command and not even start/stop that would only be a little additional work. Let me know.

## Further work
There is some further work to improve the documentation around the testing infrastructure. We are currently working on this.

PR #1470, opened 2021-02-13
Run JUnit CI Tests with Compiled Query Execution

turingcompl33t:
Closes [#1223](https://github.com/cmu-db/noisepage/issues/1223).

We have seen a number of bugs now that only surface under compiled or adaptive query execution modes, but our testing infrastructure fails to detect them. The reasons for this are manifold, but are primarily a result of the fact that:

- Apart from one exception, unit tests are not run in compiled or adaptive modes
- JUnit tests are not run run in compiled or adaptive modes

This PR takes a step towards increasing our coverage of compiled and adaptive query execution modes by adding support for running JUnit tests in CI with compiled query execution. This will hopefully decrease the probability of introducing new, uncaught bugs as we continue to modify and improve the execution engine and related aspects of the system.

I intended this PR to be solely an update to the CI infrastructure (i.e. the Jenkinsfile) but I also found it necessary to modify some of our testing infrastructure (Python implementation) to add the desired functionality.


turingcompl33t:
There may be some kinks in this PR that remain to be worked out, but I'm marking it `ready-for-ci` now so I can verify that the modifications in their current form work in the CI pipeline.

script/testing/util/db_server.py
```diff
@@ -227,21 +228,220 @@ def get_build_path(build_type):
...
+def handle_flags(value, meta):
+    """
+    Handle DBMS server arguments with no associated value.
+
+    Some arguments to the DBMS are flags that do not have an associated value;
+    in these cases, we do not want to format the complete argument as
+    `-attribute=value` and instead want to format it as `-attribute` alone.
+    This preprocessor encapsulates the logic for this transformation.
+
+    TODO(Kyle): Do we actually support any arguments like this?
```
lmwnshn, line 402: See #1388 and make sure it somewhat matches behavior.

script/testing/util/db_server.py
```diff
@@ -227,21 +228,220 @@ def get_build_path(build_type):
...
+# Utility
+
+def apply_all(functions, init_obj, meta):
+    """
+    Apply all of the functions in `functions` to object `init_obj` sequentially,
+    supplying metadata object `meta` to each function invocation.
+
+    TODO(Kyle): Initially I wanted to implement this with function composition
+    in terms of functools.reduce() which makes it really beautiful, but there
+    we run into issues with mutli-argument callbacks, and the real solution is
```
lmwnshn, line 428: nit: mutli

script/testing/util/db_server.py
```diff
@@ -227,21 +228,220 @@ def get_build_path(build_type):
...
+    to use partial application, but this seemed like overkill... maybe revisit.
```
lmwnshn, line 429: comment: simple is fine and arguably better, in case someone's not a functors fan

script/testing/util/db_server.py
```diff
@@ -227,21 +228,220 @@ def get_build_path(build_type):
...
+
+    Arguments
+    ---------
+    functions : List[function]
+        The collection of functions to invoke =
```
lmwnshn, line 434: nit: stray =

script/testing/util/db_server.py
lmwnshn: Incidentally, is this the expected behavior?
```
>>> construct_server_argument('wal_disabled', False, {})
'-wal_disabled=false'
>>> construct_server_argument('wal_disabled', True, {})
'-wal_disabled=true'
>>> construct_server_argument('wal_disabled', None, {})
'-wal_disabledNone'
```

script/testing/util/db_server.py
```diff
@@ -227,21 +228,220 @@ def get_build_path(build_type):
...
+    ---------
+    value : str 
+        The DBMS server argument value
+    meta : Dict
+        Dictionary of meta-information available to all preprocessors
+
+    Returns
+    -------
+    The preprocessed server argument value
+    """
```
lmwnshn, line 352: Want to add `assert value == True or value == False`?

lmwnshn (changes requested):
Generally LGTM, nice use of Python, question about the `None` case, minor typo.

script/testing/util/db_server.py
turingcompl33t: Fixed.

script/testing/util/db_server.py
turingcompl33t: Agreed, definitely not a design issue to stress over, but at the time it just seemed to tantalizingly close to being a practical use of some cool functional stuff.

script/testing/util/db_server.py
turingcompl33t: Fixed.

script/testing/util/db_server.py
turingcompl33t: Assertion added.

script/testing/util/db_server.py
turingcompl33t: No, that is a necessary case that I didn't consider. Have to do a little hacking around because we aren't allowed to manually specify `NoneType` (the type of the singleton `None`), but I updated to work around the issue such that we now get:

```
>>> construct_server_argument('wal_disabled', False, {})
'-wal_disabled=false'
>>> construct_server_argument('wal_disabled', True, {})
'-wal_disabled=true'
>>> construct_server_argument('wal_disabled', None, {})
'-wal_disabled'
```

script/testing/util/db_server.py
```diff
@@ -227,21 +228,220 @@ def get_build_path(build_type):
...
+    is_relative = str.startswith(value, "./") or str.startswith(value, "../")
+    # TODO(Kyle): This doesn't actually do any "resolving", it merely appends
+    # the absolute path to the binary directory to the relative path as it 
+    # presently exists; this works fine, but also means we might pass a needlessly
+    # long string to the DBMS server e.g. bin/././././whatever.txt is possible.
+    # It might be worth it to find some portable way to completely resolve these.
```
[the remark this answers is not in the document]
turingcompl33t: So I have an issue with this implementation that I have been mulling over, would appreciate an additional perspective. The functionality in this preprocessor is necessary, and I like having it there, but I don't like how 'loose' the constraint on it is - that is, we will attempt to apply this preprocessor to _any_ string that happens to start with `./` or `../` because we assume this means it is a relative path. This is _probably_ a safe assumption, but what I don't like is the fact that it applies to **all strings** and not just strings that are also paths.

I have been trying to come up with a way to express the fact that an argument is a path, rather than just a string. The best I have come up with so far (I think) is just to look at the suffix of the argument and declare that if it ends with `_path` it really is a string that corresponds to a path. We obey this naming scheme consistently for all of the arguments the DBMS currently supports, but of course this introduces a constraint that all future arguments must continue to observe this.

I even though about going so far as to parse the header file where the settings are defined (`settings_defs.h`) and derive type information from that file, but that doesn't really get us anywhere because its not like the string settings defined therein look any different (i.e. use a different macro, have a different type) when they are paths versus when they are not. Of course, this is a change I could make (refactor settings to add support for e.g. `SETTING_path`) but I fear I would be encroaching on over-engineering territory.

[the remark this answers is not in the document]
lmwnshn: If you can solve the problem in a single `_hidden_helper` function and <20 lines of code I'd say go for it. Otherwise, generally I do not think people muck around with infrastructure often, and those who do are generally prepared for a little bit of madness given the state of C++ infrastructure, so as long as the blowing up error is obvious then path issues aren't too bad to debug (xyz not found).

If you just want to apply it to paths, maybe you can enforce it at a type level?
https://docs.python.org/3/library/pathlib.html

malin1993ml:
The time to finish the junit tests under `compiled` mode seems really long. I don't believe it's an issue of this PR itself (nice work, @turingcompl33t). But it does seem alarming that it takes so long, which will significantly increase our Jenkins test time. For example, this build takes almost **4 hours** to finish:
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-1470/43/pipeline/142

turingcompl33t:
> The time to finish the junit tests under `compiled` mode seems really long. I don't believe it's an issue of this PR itself (nice work, @turingcompl33t). But it does seem alarming that it takes so long, which will significantly increase our Jenkins test time. For example, this build takes almost **4 hours** to finish:
> http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-1470/43/pipeline/142

Agreed @malin1993ml, was noticing the same thing, as you can see in the commits above where I had to bump the timeout for the JUnit tests twice in order to get them to run to completion before being terminated. I plan to look a bit closer at where exactly the huge jump in execution time is coming from, although its entirely possible that's just the current state of execution engine performance... do we have a robust comparison of the performance in interpreted vs compiled modes anywhere?

malin1993ml:
>  do we have a robust comparison of the performance in interpreted vs compiled modes anywhere?

I'm not sure. Maybe others have an idea of what should be the expected performance.

turingcompl33t:
Adding `ready-for-ci` tag because in theory all of the compiled JUnit tests should pass in compiled mode now, with compiled aggregations fixed AND the update where we fail the test in the event the DBMS exits with nonzero exit code.

script/testing/util/db_server.py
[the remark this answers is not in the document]
turingcompl33t: Ended up just implementing the `_path` check. My thinking is, at the worst, this will end up failing to resolve paths that some user may have thought would be resolved, which is better than the alternative. One has to explicitly opt-in to being resolved, rather than the other way around.

turingcompl33t:
Don't have a passing CI build to link to because the pipeline failed (in end-to-end performance, unrelated to this PR), but I think the modifications to the bytecode compilation settings may have done the trick. After making the change, compiled JUnit tests consistently finished in ~20 minutes (often far fewer). I am hoping this will be deemed 'good enough' for the time being to push the PR through.

turingcompl33t:
Per the discussion in standup, I am going to briefly investigate the breakdown of the runtime overhead for tests run in compiled execution mode. Related issue is [here](https://github.com/cmu-db/noisepage/issues/1173).

lmwnshn (approved):
Rubber stamp.

PR #1472, opened 2021-02-15
Synchronous Replication

### LINKED ISSUE #1319: jumbotest_messenger_messenger_test timesout on jenkins

lmwnshn:
# Synchronous Replication

This PR adds support for synchronous replication to `noisepage`. Replication depends on the messenger and recovery manager being enabled.

## Summary

- primary: `./noisepage  --messenger_enable=true --replication_enable=true --port=15721 --messenger_port=9022  --replication_port=15445 --network_identity=primary`
- replica1: `./noisepage  --messenger_enable=true --replication_enable=true --port=15722 --messenger_port=9023  --replication_port=15446 --network_identity=replica1`
- replica2: `./noisepage  --messenger_enable=true --replication_enable=true --port=15723 --messenger_port=9024  --replication_port=15447 --network_identity=replica2`

To add more primaries/replicas, see `build-support/data/replication.config`. Note that this is copied to your build/bin folder as a post-build of `noisepage`.

You should be able to run queries on the primary and run basic read-only things on the replicas.

Note that there is currently no "I'm missing logs xyz" mechanism, so replicas have to have seen all the primary's logs as they were generated and shipped out.

## Background

- Replication builds on previous work by @GustavoAngulo and @tpan496, specifically, we still use the old recovery manager logic essentially as-is. We keep the recovery task running in a loop. However, ITP internal terrier protocol is removed (and if you're not sure what ITP did, that's fine).
- Messages (which include the log buffers being replicated) are sent through the ZeroMQ-backed messenger, in what is essentially JSON/cbor format. This might prove to be a bottleneck someday, we can revisit the format if so. Since it is just JSON, buffer IDs, etc, are also tacked onto the JSON.

## Description

- Fix #1319. I believe this was caused by the Messenger's `ListenForConnection` being a non-blocking add to the list of routers. This changes it so that the `ListenForConnection` waits on a `cvar` which the messenger's main loop will signal whenever it grabs and adds new routers.

- Adds a new builtin `replication_get_last_record_id()`. On the primary, this returns the last record ID that was sent out, and on replicas, this returns the last record ID that was applied. This is used in testing to see if the replicas are up to date. This is implemented by pushing a pointer to the replication manager into the execution context.
- Most of the PR is in `replication/replication_manager` and `storage/replication_log_provider`.
- Adds a new test `script/testing/replication/tests_simple`, which will run under the simple unittests part of CI as well. No end to end oltpbench yet.
- Adds the idea of a transaction-wide `RetentionPolicy`, which is however not currently able to distinguish between "serialize locally, don't replicate" and "serialize locally, also replicate". We probably have to revisit the implementation. The idea is that at a per-transaction level, you describe whether (1) you serialize and replicate logs (2) you only serialize logs or (3) you don't serialize logs at all.

## Future work, future PR

- `RetentionPolicy` is basically just a flag right now, we need to figure out more details on doing it properly.
- Replication is now also on the critical path before a commit callback is invoked. This is probably inadvisable. Async replication has been buggy / racey, but that's the next step.
- OLTPBench e2e testing for replication


src/transaction/transaction_manager.cpp
```diff
@@ -35,30 +35,31 @@ TransactionContext *TransactionManager::BeginTransaction() {
...
-  } else if (log_manager_ != DISABLED && wal_async_commit_enable_) {
-    // We still want to send this record to the LogManager, but we'll swap in the EmptyCallback to send with the
-    // LogRecord, and invoke the provided callback immediately. The WAL worker will still be responsible for removing it
-    // from the running transactions table once the record has been serialized. Otherwise the serializer might
-    // dereference bad memory if the GC prunes any varlens this transaction might be holding.
-    byte *const commit_record = txn->redo_buffer_.NewEntry(storage::CommitRecord::Size());
-    storage::CommitRecord::Initialize(commit_record, txn->StartTime(), commit_time, TransactionUtil::EmptyCallback,
-                                      nullptr, oldest_active_txn, txn->IsReadOnly(), txn, timestamp_manager_.Get());
-    commit_callback(commit_callback_arg);
+  if (log_manager_ != DISABLED && txn->GetRetentionPolicy() == RetentionPolicy::RETENTION_ALL) {
```
lmwnshn, line 38: @mbutrovich You might want to take a quick look at this one.

script/testing/util/db_server.py
```diff
@@ -78,14 +78,19 @@ def run_db(self, is_dry_run=False):
...
-                                           stdout=subprocess.PIPE,
-                                           stderr=subprocess.PIPE)
+                                      stdout=subprocess.PIPE,
+                                      stderr=subprocess.PIPE)
         LOG.info(f'Ran: {db_run_command} [PID={db_process.pid}]')
 
         logs = []
+
+        check_line = f'[info] Listening on Unix domain socket with port {self.db_port} [PID={db_process.pid}]'
+        if self.server_args.get("messenger_enable", False):
```
lmwnshn, line 88: Our approach to detecting the server starting up is a little bit hacky...

src/include/storage/recovery/replication_log_provider.h
```diff
@@ -1,20 +1,143 @@
 #pragma once
```
lmwnshn, line 1: I would like to take a second pass on this file when I'm fresh. This is a "here be dragons" file to me that I haven't looked at much / there may be opportunities for refactoring here.

lmwnshn (commented):
quick self-review pass

src/replication/replication_manager.cpp
```diff
@@ -0,0 +1,273 @@
...
+    common::ManagedPointer<messenger::Messenger> messenger, const std::string &network_identity, uint16_t port,
+    const std::string &replication_hosts_path,
+    common::ManagedPointer<common::ConcurrentBlockingQueue<storage::BufferedLogWriter *>> empty_buffer_queue)
+    : messenger_(messenger),
+      identity_(network_identity),
+      port_(port),
+      provider_(std::make_unique<storage::ReplicationLogProvider>(std::chrono::seconds(1))),
+      received_buffer_queue_(CompareJsonBuffer),
+      empty_buffer_queue_(empty_buffer_queue) {
+  // TODO(WAN): provider_ arguments
```
lmwnshn, line 38: oh, and I need to fix this to not be a constant. But that's somewhat lumped together with the replication log provider.

src/replication/replication_manager.cpp
```diff
@@ -0,0 +1,273 @@
...
+  // TODO(WAN): Is it true that a replica will never want to send its buffers to the primary?
+  // TODO(WAN): Cache the identity check into a bool?
+  if (identity_ == "primary") {
+    common::json j;
+    // TODO(WAN): Add a size and checksum to message.
+    j["buf_id"] = next_buffer_sent_id_++;
+    j["content"] = nlohmann::json::to_cbor(std::string(buffer->buffer_, buffer->buffer_size_));
+
+    for (const auto &replica : replicas_) {
+      ReplicaSend(replica.first, MessageType::REPLICATE_BUFFER, j.dump(), true);
```
lmwnshn, line 153: the next goal is to figure out why everything catches on fire when `true` is flipped to `false`

Additionally, I don't think we want the `LogSerializerTask` thread running this code... ideally the buffer is handed off to the replication manager's queue and something executes this later. @jkosh44 thoughts on where this belongs?

src/include/storage/recovery/replication_log_provider.h
lmwnshn: I will probably take a pass on this tonight after submitting some other hw @jkosh44 maybe don't look at this until tomorrow Friday, thanks!

lmwnshn:
Actually, @jkosh44 give me tonight to figure out if I can make it proper true synchronous before you take a pass maybe

lmwnshn:
You can probably start taking a review pass now, but some minor cleanup / refactoring might happen. Would recommend using github's view file thingy.

script/testing/replication/utils_sql.py
```diff
@@ -0,0 +1,104 @@
...
+        False if the checks should be printed as they happen.
+    """
+    primary_rec_id = sql_get_last_record_id(primary)
+    for replica in replicas:
+        while True:
+            replica_rec_id = sql_get_last_record_id(replica)
+            if not quiet:
+                LOG.info(f"Syncing replica: [primary@{primary_rec_id}] [{replica.identity}@{replica_rec_id}]")
+            if replica_rec_id >= primary_rec_id:
+                return
```
jkosh44, line 102: Shouldn't this be a `break` or something similar? Otherwise it seems like we only wait for the first replica to sync up.

script/testing/replication/tests_simple.py
```diff
@@ -0,0 +1,67 @@
...
+
+if __name__ == "__main__":
+    aparser = argparse.ArgumentParser(description="Simple replication tests.")
+    aparser.add_argument("--build-type",
+                         default="debug",
+                         choices=["debug", "release", "relwithdebinfo"],
+                         help="Build type (default: %(default)s")
+    args = vars(aparser.parse_args())
+
+    servers = [NoisePageServer(build_type=args['build_type'], port=15721 + i, server_args={
```
jkosh44, line 48: Very pythonic, nice

src/include/main/db_main.h
```diff
@@ -393,6 +427,16 @@ class DBMain {
                                            common::ManagedPointer(log_manager), create_default_database_);
       }
 
+      std::unique_ptr<storage::RecoveryManager> recovery_manager = DISABLED;
```
jkosh44, line 430: Just for my own clarification, is replication the only current use for the `RecoveryManager`? Is there any other way to enable it?

src/include/replication/replication_manager.h
```diff
@@ -0,0 +1,204 @@
...
+  std::mutex blocking_send_mutex_;                             ///< Mutex used for blocking sends.
+  std::condition_variable blocking_send_cvar_;                 ///< Cvar used for blocking sends.
+
+  // TODO(WAN): I think it is better to use retention policies instead of enabling/disabling replication.
+  bool replication_enabled_ = false;  ///< True if replication is currently enabled and false otherwise.
+
+  uint64_t next_buffer_sent_id_ = 1;      ///< The ID of the next buffer to sent.
+  uint64_t last_record_received_id_ = 0;  ///< The ID of the last record to be received.
+
+  /** The received messages are queued up until the "next" received message is the right one. */
```
jkosh44, line 195: May want to include what "right one" means

script/testing/replication/utils_sql.py
lmwnshn: great catch, thanks!

script/testing/replication/tests_simple.py
lmwnshn: I actually don't know if it'll scale in the future, but it works for now...

src/include/main/db_main.h
lmwnshn: I believe `recovery_test` checks that it actually works, but otherwise, yes. `RecoveryManager` prior to this PR is dead code afaik.

src/include/replication/replication_manager.h
lmwnshn: Added (I hope).

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -93,6 +98,14 @@ class BufferedLogWriter {
   explicit BufferedLogWriter(const char *log_file_path)
       : out_(PosixIoWrappers::Open(log_file_path, O_WRONLY | O_APPEND | O_CREAT, S_IRUSR | S_IWUSR)) {}
 
+  /** Move constructor. */
+  BufferedLogWriter(BufferedLogWriter &&other) noexcept {
```
jkosh44, line 102: I can't remember exactly how these are used, but is it possible for someone to be writing to this as we are copying?

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -149,11 +162,44 @@ class BufferedLogWriter {
...
+   *
+   * Currently, BufferedLogWriter is used to serialize both over the disk and over the network.
+   * To avoid copying, we refcount the serialization usages of the BufferedLogWriter.
+   * This function initializes the refcount.
+   *
+   * TODO(WAN): Come up with a better way of "sharing" buffers?
+   */
+  void PrepareForSerialization() {
+    NOISEPAGE_ASSERT(serialize_refcount_.load() == 0, "This buffer is already being serialized.");
+    serialize_refcount_.store(NUM_SERIALIZERS);
```
jkosh44, line 176: I'm sort of thinking out loud here, but is it always true that a `BufferedLogWriter` will be consumed by all "Serializers"? 
The two specific situations I'm thinking of are the following
- If replication is disable then will this refcount every hit 0?
- If the buffer has read only logs then does the replication manager still consume it? Or does this distinction happen on a more granular level of individual logs?

jkosh44: Ok I think I see In `LogSerializerTask::HandFilledBufferToWriter`, if replication is disabled then you add an extra call to to `MarkSerialized` so the count will get down to 0. I think this is a bit not obvious. Is there a way to initialize `NUM_SERIALIZERS` to 1 if replication is disabled? It would prevent it from being a `constexpr` but I think it would be slightly cleaner or more obvious as to what's going on. That way you wouldn't need the extra call to `MarkSerialized` in `LogSerializerTask::HandFilledBufferToWriter`

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -149,11 +162,44 @@ class BufferedLogWriter {
...
+    auto count = serialize_refcount_.fetch_sub(1);
+    NOISEPAGE_ASSERT(serialize_refcount_.load() >= 0, "This buffer was serialized too many times?");
+    return count == 1;
+  }
+
  private:
+  friend class replication::ReplicationManager;
+
+  /** The number of callers of MarkSerialized(). See also PrepareForSerialization(). */
+  static constexpr int NUM_SERIALIZERS = 2;
```
jkosh44, line 196: Might be helpful to include in the comment who the callers are

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -136,6 +137,16 @@ BufferedLogWriter *LogSerializerTask::GetCurrentWriteBuffer() {
...
  */
 void LogSerializerTask::HandFilledBufferToWriter() {
+  // Mark the buffer as ready for serialization, if it exists. It may not exist for read-only transactions.
+  if (filled_buffer_ != nullptr) filled_buffer_->PrepareForSerialization();
+  // Replicate the buffer, if it exists.
+  if (filled_buffer_ != nullptr) {
+    if (replication_manager_ != DISABLED) {
+      replication_manager_->ReplicateBuffer(filled_buffer_);
+    } else {
+      filled_buffer_->MarkSerialized();
```
jkosh44, line 147: See my comment in `log_io.h`. If you end up not making that change, I think adding a comment here explaining why we call `MarkSerialized` would be helpful.

src/messenger/messenger.cpp
```diff
@@ -93,80 +93,31 @@
...
-    }
+ZmqMessage::ZmqMessage(std::string routing_id, std::string payload)
+    : routing_id_(std::move(routing_id)), payload_(std::move(payload)), message_(payload_) {
+  if (payload_.empty()) {
+    source_cb_id_ = 0;
+    dest_cb_id_ = 0;
+  } else {
+    // TODO(WAN): atoi, stoull, from_chars, etc? Error checking in general.
+    UNUSED_ATTRIBUTE int check =
+        std::sscanf(payload_.c_str(), "%" SCNu64 "-%" SCNu64 "-", &source_cb_id_, &dest_cb_id_);
```
jkosh44, line 113: This is my opinion, so feel free to ignore it if you disagree. I guess this was already here, but it makes me a little nervous seeing the expected format of the message hard coded into this constructor here and on lines 117, 118. It might be more obvious to make these constants and refer to those.

src/replication/replication_manager.cpp
jkosh44: Is this referring to the 1 second timeout for the provider? You could just add it as a setting or toss it in `include/common/constants.h

src/replication/replication_manager.cpp
```diff
@@ -0,0 +1,290 @@
...
+      empty_buffer_queue_(empty_buffer_queue) {
+  // TODO(WAN): provider_ arguments
+  auto listen_destination = messenger::ConnectionDestination::MakeTCP("", "127.0.0.1", port);
+  messenger_->ListenForConnection(listen_destination, network_identity,
+                                  [this](common::ManagedPointer<messenger::Messenger> messenger,
+                                         const messenger::ZmqMessage &msg) { EventLoop(messenger, msg); });
+  BuildReplicaList(replication_hosts_path);
+
+  // TODO(WAN): development purposes
+  replication_logger->set_level(spdlog::level::trace);
```
jkosh44, line 50: Just adding a comment here as a reminder, in case you forget to remove this.

src/replication/replication_manager.cpp
```diff
@@ -0,0 +1,290 @@
...
+        static_cast<uint64_t>(type));
+
+    if (block) {
+      std::unique_lock<std::mutex> lock(blocking_send_mutex_);
+      // If the caller requested to block until the operation was completed, the thread waits.
+      if (!blocking_send_cvar_.wait_for(lock, REPLICATION_MAX_BLOCKING_WAIT_TIME, [&completed] { return completed; })) {
+        // TODO(WAN): Additionally, this is hackily a messenger exception so that it gets handled by the catch.
+        throw MESSENGER_EXCEPTION("TODO(WAN): Handle a replica dying in synchronous replication.");
+      }
+      lock.unlock();
```
jkosh44, line 144: I thought that you didn't need to call this for `unique_lock`

src/replication/replication_manager.cpp
```diff
@@ -0,0 +1,290 @@
...
+    }
+  } catch (const MessengerException &e) {
+    REPLICATION_LOG_WARN(fmt::format("[FAILED] Send -> {} (type {} block {}): msg size {}", replica_name,
```
jkosh44, line 147: What are the ramifications of failing to send a log to a replica? If the replica must process the logs in order and one of the logs fails to send then won't this effectively hang the replica? Is this something we care about handling?

src/replication/replication_manager.cpp
```diff
@@ -0,0 +1,290 @@
...
+  }
+}
+
+void ReplicationManager::EventLoop(common::ManagedPointer<messenger::Messenger> messenger,
+                                   const messenger::ZmqMessage &msg) {
+  switch (static_cast<MessageType>(msg.GetDestinationCallbackId())) {
+    case MessageType::ACK: {
+      nlohmann::json json = nlohmann::json::parse(msg.GetMessage());
+      REPLICATION_LOG_TRACE(fmt::format("ACK: {}", json.dump()));
+      auto callback_id = json.at("callback_id").get<uint64_t>();
```
jkosh44, line 186: This applies to all the `json.at(...)` calls in this PR, but I think it would be better to extract the JSON keys out to some constants somewhere.

src/replication/replication_manager.cpp
```diff
@@ -0,0 +1,290 @@
...
+        // TODO(WAN): Sanity-check the received message.
+
+        uint64_t source_callback_id = msg.GetSourceCallbackId();
+        uint64_t msg_id = message.at("buf_id");
+        REPLICATION_LOG_TRACE(fmt::format("ReplicateBuffer from: {} {}", msg.GetRoutingId(), msg_id));
+
+        // Check if the message needs to be buffered.
+        if (msg_id > last_record_received_id_ + 1) {
+          // The message should be buffered if there are gaps in between the last seen buffer.
+          received_message_queue_.push(msg);
```
jkosh44, line 207: Why not push the parsed `message` to avoid parsing it again?

src/replication/replication_manager.cpp
```diff
@@ -0,0 +1,290 @@
...
+            // Otherwise, send the top buffer's contents along.
+            received_message_queue_.pop();
+            NOISEPAGE_ASSERT(msg_id == last_record_received_id_ + 1, "Duplicate buffer? Old buffer?");
+            content = nlohmann::json::from_cbor(message["content"].get<std::vector<uint8_t>>());
+            provider_->AddBufferFromMessage(source_callback_id, content);
+            last_record_received_id_ = msg_id;
+          }
+        }
+      }
+      // TODO(WAN): Is it actually necessary to buffer the message first? A crashed replica is indistinguishable from
```
jkosh44, line 233: Is this comment supposed to be here? It's a little unclear to me what it's referring to.

src/replication/replication_manager.cpp
```diff
@@ -0,0 +1,290 @@
...
+      // TODO(WAN): To support async, acknowledge receipt of the buffer to the sender here.
+      // TODO(WAN): Add a size and checksum to message.
+      break;
+    }
+    default:
+      break;
+  }
+}
+
+void ReplicationManager::ReplicaHeartbeat(const std::string &replica_name) {
```
jkosh44, line 245: It doesn't seem like we actually mark the replica dead as the Doxygen suggests. It seems like we just log and move on. Do we have a plan for dead replicas?

src/include/storage/recovery/replication_log_provider.h
```diff
@@ -1,20 +1,151 @@
...
+      arrived_buffer_queue_.emplace(primary_callback_id, std::move(buffer));
+      replication_cv_.notify_one();
+    }
+
+    // TODO(WAN): if you want sync/async I think this is where you do it
+  }
+
+  // TODO(gus): this method is not entirely correct, there could still be messages over network
+  /**
+   * TODO(WAN): Per Gus, tThis method is not entirely correct because there could still be messages pending over
```
jkosh44, line 57: Typo in "tThis"

src/include/storage/recovery/replication_log_provider.h
```diff
@@ -1,20 +1,151 @@
...
+   *  the network.
+   */
+  void WaitUntilSync() {
+    while ((curr_buffer_ != nullptr && curr_buffer_->HasMore()) || !arrived_buffer_queue_.empty())
```
jkosh44, line 61: I guess it doesn't really matter since this isn't correct but I think you could just replace this with
`while(NonBlockingHasMoreRecords()) {`

jkosh44 (changes requested):
Looks good overall, I had a couple of comments. Feel free to address the comments with a TODO or disregard them if you think they're irrelevant/wrong. I largely ignored the added TODOs because I figure that you know better than me which tasks are OK to leave pending for this to merge.

Also I wanted to add 1 more thing. This is 100% opinion so feel free to completely ignore this or address it at a later time. It felt a bit like the `ReplicationManager` was trying to do two things at once. That is the `ReplicationManager` is 1) responsible for sending logs to replicas from the primary and 2) responsible for receiving logs from the primary in the replica.
When reading through the code it was sometimes hard to reason about the context of the code, for example will this be running in the primary, replica, or both? 
Something to consider is to split  out the class into two classes  with a common parent. Something like `ReplicationManger`, `PrimaryReplicationManager`,`ReplicaReplicationManager`
This also somewhat applies to the `ReplicationManager` but not as much (where that is responsible for 1) recovering from crashes and 2) applying replicated logs).

src/include/storage/write_ahead_log/log_io.h
lmwnshn: This is a great point. Especially once Tianlei's RetentionPolicy stuff makes it in, the treatment is not quite uniform..

Currently thinking of using `PrepareForSerialization(RetentionPolicy ...)` so that it takes in an enum (disk, disk+network, neither) that sets the refcount accordingly.

src/messenger/messenger.cpp
lmwnshn: unfortunately it is also hardcoded into the model server message format, off in Python land, if I'm not mistaken. but yea, I can try to refactor -- I seem to recall facing some weird issue which ended with those SCNu64's there. you're right to point out hacks, please keep doing so!

src/replication/replication_manager.cpp
lmwnshn: The replica will hang yes. We currently do not do any kind of crash tolerance whatsoever / we assume things don't crash...

I definitely should add at least documentation about what happens, in the scenario you described. Ideally, I think we would add code here to check if the replica is dead, try to resend if it isn't. hmm. that might not be so bad to add.

src/replication/replication_manager.cpp
lmwnshn: yeah, you're right. The reason was that multiple fields were being read (msg_id, content). I could create a struct for all the fields that matter.

src/replication/replication_manager.cpp
lmwnshn: that was a stale comment from when i was in crazytown until andy pointed out that doing an ACK here wouldn't be synchronous replication (since the replica wasn't really seeing the changes yet), will remove. we would want to ack somewhere early on here for async though.

lmwnshn: but we're not worrying about async for now

src/replication/replication_manager.cpp
lmwnshn: ~~The current plan is to not have dead replicas~~

We're not particularly focused on fault tolerance.
The heartbeat stuff is stuff that hasn't been tested much. A lot of assumptions are being made about replicas always being available right now. I'll see how the code shakes out after the other refactors.

lmwnshn:
Thanks for the thorough review! Will get to addressing comments over the course of tonight. I agree re: splitting sending and applying logic. I'll see what can happen there

src/include/storage/write_ahead_log/log_io.h
lmwnshn: Changed, let me know what you think?

src/include/storage/write_ahead_log/log_io.h
lmwnshn: Removed.

src/storage/write_ahead_log/log_serializer_task.cpp
lmwnshn: Removed. It shouldn't be necessary if the Prepare call is setting the right count.

src/replication/replication_manager.cpp
lmwnshn: you are correct

src/include/storage/recovery/replication_log_provider.h
lmwnshn: It was actually never used, I just removed it.

src/include/storage/write_ahead_log/log_io.h
jkosh44: I like it, I think it's a lot cleaner/clearer

src/include/storage/write_ahead_log/log_io.h
lmwnshn: please see the new comment and see if it makes sense?

src/replication/replication_manager.cpp
lmwnshn: Refactored to do the struct, thoughts?

src/replication/replication_manager.cpp
lmwnshn: Refactored to do the struct which contains constants.

src/replication/replication_manager.cpp
lmwnshn: added a comment

src/replication/replication_manager.cpp
lmwnshn: I am inclined to punt on this issue and run with the assumption that replicas won't crash without a good reason (like segfaulting etc). thoughts? in particular, please stop me if you see something problematic with the design that we can't bolt on fault tolerant stuff later.

jkosh44: I think that's a fine approach, I don't think there's anything inherently wrong with the design.

src/include/main/db_main.h
```diff
@@ -395,18 +395,28 @@ class DBMain {
       if (use_replication_) {
         NOISEPAGE_ASSERT(use_messenger_, "Replication uses the messenger subsystem.");
         NOISEPAGE_ASSERT(use_logging_, "Replication uses logging.");
-        replication_manager = std::make_unique<replication::ReplicationManager>(
-            messenger_layer->GetMessenger(), network_identity_, replication_port_, replication_hosts_path_,
-            common::ManagedPointer(empty_buffer_queue));
+        if (network_identity_ == "primary") {
```
jkosh44, line 398: We should probably extract `"primary"` to a constant somewhere, but I don't think it's a huge deal (If you haven't been able to tell, I'm slightly obsessed with constants, so feel free to disregard this)

src/replication/replication_manager.cpp
jkosh44: I like it

jkosh44 (approved):
LGTM

src/include/main/db_main.h
lmwnshn: tada

lmwnshn:
@jkosh44 Sorry to keep bugging you, could you look at this commit?
https://github.com/cmu-db/noisepage/pull/1472/commits/649f05158f9b1869fd97eb01877d59b4a7bd113d

This came up before in the ModelServerManager.
The problem is that the Messenger is the owner of all the callbacks that get passed in with a `SendMessage(msg, ..., callback)` call ("send message callback").
This is distinct from the ServerLoops that are created during a `ListenForConnection` ("server loop callback").

I refactored it so that the Messenger exposed its callbacks, and the server loop callback became responsible for invoking the send message callback.
This is why my modeling_test kept failing CI, because the model server callback wasn't updated accordingly.

But upon further reflection, I _think_ it is preferable that the Messenger automagically handles it.
The main limitation is that, in the current implementation this means your send message callback will **always** be invoked before the model server callback.

src/messenger/messenger.cpp
```diff
@@ -418,25 +389,29 @@ void Messenger::RunTask() {
...
   is_messenger_running_ = false;
-  // Shut down the ZeroMQ context. This causes all existing sockets to abort with ETERM.
-  zmq_ctx_->shutdown();
-  MESSENGER_LOG_INFO(fmt::format("[PID={}] Messenger terminated.", ::getpid()));
 }
 
 void Messenger::ListenForConnection(const ConnectionDestination &target, const std::string &identity,
                                     CallbackFn callback) {
-  std::lock_guard lock(routers_add_mutex_);
+  std::unique_lock lock(routers_add_mutex_);
```
jkosh44, line 403: I'm a little confused by this part. If we're taking the time to get a latch, why not just have the latch be for `routers_` and insert it directly into that? Is there something I'm missing here?

src/messenger/messenger.cpp
```diff
@@ -418,25 +389,29 @@ void Messenger::RunTask() {
...
   // TODO(WAN): all this copying is stupid.
```
mbutrovich: These sorts of TODOs aren't terribly helpful 6 months in the future when someone else (or even you) is reading it. Can you give a bit more on what you'd do to fix it (if you know) and what the restriction on implementing it now is (it's okay if it's just time).

src/include/storage/write_ahead_log/log_manager.h
```diff
@@ -153,7 +166,7 @@ class LogManager : public common::DedicatedThreadOwner {
   std::vector<BufferedLogWriter> buffers_;
   // The queue containing empty buffers which the serializer thread will use. We use a blocking queue because the
   // serializer thread should block when requesting a new buffer until it receives an empty buffer
-  common::ConcurrentBlockingQueue<BufferedLogWriter *> empty_buffer_queue_;
```
lmwnshn, line 156: @mbutrovich This got pulled out to DBMain so that other components, e.g., replication manager, can return buffers to it.

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -94,11 +116,71 @@ class BufferedLogWriter {
    */
   bool IsBufferFull() { return buffer_size_ == common::Constants::LOG_BUFFER_SIZE; }
 
+  /**
```
lmwnshn, line 119: whoops, dead comment, I'll delete this.

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -94,11 +116,71 @@ class BufferedLogWriter {
...
+   * For example, the BufferedLogWriter may then be sent to any of the following destinations:
+   * - Serialized to disk.
+   * - Sent to replicas over the network.
+   *
+   * This function exists to avoid copying the BufferedLogWriter's buffers needlessly.
+   * Instead, a refcount is maintained depending on the retention policy.
+   *
+   * @param policy The retention policy that describes the destinations for this BufferedLogWriter.
+   */
+  void PrepareForSerialization(transaction::RetentionPolicy policy) {
```
lmwnshn, line 131: @mbutrovich The current retention policy thing.

src/include/storage/write_ahead_log/log_manager.h
mbutrovich: What do we believe the argument for this data structure is? Performance over `malloc`? Back-pressure on the serializer if the disk consumer falls behind?

lmwnshn: *was, I believe it was backpressure. This is more or less leaving things as they are.

src/messenger/messenger.cpp
lmwnshn: Oh, just saw this. I think this was an old comment that wanted to avoid the indirection of 

1. T1: invokes `ListenForConnection` trying to add a new connection endpoint, which enqueues the router to be added
2. dedicated messenger thread T2: grabs it off the queue, adds it

In addressing Joe's comment, I added documentation that notes why this is necessary (iirc zeromq requires that sockets are used on the thread where they are created)

lmwnshn:
All numbers on dev9.

## This PR

```
/bin/noisepage  -connection_thread_count=32 -record_buffer_segment_size=10000000

---

Benchmark:     TPCC {com.oltpbenchmark.benchmarks.tpcc.TPCCBenchmark}
Configuration: ./config/tpcc_config_postgres.xml
Type:          NOISEPAGE
Driver:        org.postgresql.Driver
URL:           jdbc:postgresql://localhost:15721/noisepage
Isolation:     TRANSACTION_READ_COMMITTED
Scale Factor:  32.0

16:39:42,651 (DBWorkload.java:271) INFO  - ======================================================================
16:39:42,663 (DBWorkload.java:533) INFO  - Creating new TPCC database...
16:39:42,733 (DBWorkload.java:535) INFO  - Finished!
16:39:42,734 (DBWorkload.java:536) INFO  - ======================================================================
16:39:42,734 (DBWorkload.java:559) INFO  - Loading data into TPCC database with 80 threads...
16:42:46,477 (DBWorkload.java:563) INFO  - Finished!
16:42:46,478 (DBWorkload.java:564) INFO  - ======================================================================
16:42:46,480 (DBWorkload.java:849) INFO  - Creating 32 virtual terminals...
16:42:46,528 (DBWorkload.java:854) INFO  - Launching the TPCC Benchmark with 1 Phase...
16:42:46,541 (ThreadBench.java:341) INFO  - PHASE START :: [Workload=TPCC] [Serial=false] [Time=60] [WarmupTime=0] [Rate=10000] [Arrival=REGULAR] [Ratios=[45.0, 43.0, 4.0, 4.0, 4.0]] [ActiveWorkers=32]
16:42:46,542 (ThreadBench.java:492) INFO  - MEASURE :: Warmup complete, starting measurements.
16:43:46,544 (ThreadBench.java:447) INFO  - TERMINATE :: Waiting for all terminals to finish ..
16:43:46,547 (ThreadBench.java:508) INFO  - Attempting to stop worker threads and collect measurements
16:43:46,548 (ThreadBench.java:247) INFO  - Starting WatchDogThread
16:43:46,804 (DBWorkload.java:860) INFO  - ======================================================================
16:43:46,808 (DBWorkload.java:861) INFO  - Rate limited reqs/s: Results(nanoSeconds=60000799209, measuredRequests=600017) = 10000.15012983358 requests/sec
```

```
/bin/noisepage  -connection_thread_count=32 -record_buffer_segment_size=10000000 -wal-enable=false

---

16:46:43,860 (DBWorkload.java:270) INFO  - ======================================================================

Benchmark:     TPCC {com.oltpbenchmark.benchmarks.tpcc.TPCCBenchmark}
Configuration: ./config/tpcc_config_postgres.xml
Type:          NOISEPAGE
Driver:        org.postgresql.Driver
URL:           jdbc:postgresql://localhost:15721/noisepage
Isolation:     TRANSACTION_READ_COMMITTED
Scale Factor:  32.0

16:46:43,861 (DBWorkload.java:271) INFO  - ======================================================================
16:46:43,873 (DBWorkload.java:533) INFO  - Creating new TPCC database...
16:46:43,936 (DBWorkload.java:535) INFO  - Finished!
16:46:43,936 (DBWorkload.java:536) INFO  - ======================================================================
16:46:43,936 (DBWorkload.java:559) INFO  - Loading data into TPCC database with 80 threads...
16:49:06,981 (DBWorkload.java:563) INFO  - Finished!
16:49:06,981 (DBWorkload.java:564) INFO  - ======================================================================
16:49:06,984 (DBWorkload.java:849) INFO  - Creating 32 virtual terminals...
16:49:07,035 (DBWorkload.java:854) INFO  - Launching the TPCC Benchmark with 1 Phase...
16:49:07,048 (ThreadBench.java:341) INFO  - PHASE START :: [Workload=TPCC] [Serial=false] [Time=60] [WarmupTime=0] [Rate=10000] [Arrival=REGULAR] [Ratios=[45.0, 43.0, 4.0, 4.0, 4.0]] [ActiveWorkers=32]
16:49:07,048 (ThreadBench.java:492) INFO  - MEASURE :: Warmup complete, starting measurements.
16:50:07,050 (ThreadBench.java:447) INFO  - TERMINATE :: Waiting for all terminals to finish ..
16:50:07,052 (ThreadBench.java:508) INFO  - Attempting to stop worker threads and collect measurements
16:50:07,053 (ThreadBench.java:247) INFO  - Starting WatchDogThread
16:50:07,267 (DBWorkload.java:860) INFO  - ======================================================================
16:50:07,271 (DBWorkload.java:861) INFO  - Rate limited reqs/s: Results(nanoSeconds=60000973926, measuredRequests=600021) = 10000.187675953626 requests/sec
```

lmwnshn:
## Main branch

```
./bin/noisepage  -connection_thread_count=32 -record_buffer_segment_size=10000000 -wal_enable=false

---

16:54:21,461 (DBWorkload.java:270) INFO  - ======================================================================

Benchmark:     TPCC {com.oltpbenchmark.benchmarks.tpcc.TPCCBenchmark}
Configuration: ./config/tpcc_config_postgres.xml
Type:          NOISEPAGE
Driver:        org.postgresql.Driver
URL:           jdbc:postgresql://localhost:15721/noisepage
Isolation:     TRANSACTION_READ_COMMITTED
Scale Factor:  32.0

16:54:21,462 (DBWorkload.java:271) INFO  - ======================================================================
16:54:21,474 (DBWorkload.java:533) INFO  - Creating new TPCC database...
16:54:21,539 (DBWorkload.java:535) INFO  - Finished!
16:54:21,539 (DBWorkload.java:536) INFO  - ======================================================================
16:54:21,539 (DBWorkload.java:559) INFO  - Loading data into TPCC database with 80 threads...
16:57:21,793 (DBWorkload.java:563) INFO  - Finished!
16:57:21,793 (DBWorkload.java:564) INFO  - ======================================================================
16:57:21,796 (DBWorkload.java:849) INFO  - Creating 32 virtual terminals...
16:57:21,848 (DBWorkload.java:854) INFO  - Launching the TPCC Benchmark with 1 Phase...
16:57:21,860 (ThreadBench.java:341) INFO  - PHASE START :: [Workload=TPCC] [Serial=false] [Time=60] [WarmupTime=0] [Rate=10000] [Arrival=REGULAR] [Ratios=[45.0, 43.0, 4.0, 4.0, 4.0]] [ActiveWorkers=32]
16:57:21,860 (ThreadBench.java:492) INFO  - MEASURE :: Warmup complete, starting measurements.
16:58:21,862 (ThreadBench.java:447) INFO  - TERMINATE :: Waiting for all terminals to finish ..
16:58:21,866 (ThreadBench.java:508) INFO  - Attempting to stop worker threads and collect measurements
16:58:21,867 (ThreadBench.java:247) INFO  - Starting WatchDogThread
16:58:22,092 (DBWorkload.java:860) INFO  - ======================================================================
16:58:22,096 (DBWorkload.java:861) INFO  - Rate limited reqs/s: Results(nanoSeconds=60000207603, measuredRequests=600015) = 10000.21539875471 requests/sec
```

```
./bin/noisepage  -connection_thread_count=32 -record_buffer_segment_size=10000000

---

Benchmark:     TPCC {com.oltpbenchmark.benchmarks.tpcc.TPCCBenchmark}
Configuration: ./config/tpcc_config_postgres.xml
Type:          NOISEPAGE
Driver:        org.postgresql.Driver
URL:           jdbc:postgresql://localhost:15721/noisepage
Isolation:     TRANSACTION_READ_COMMITTED
Scale Factor:  32.0

17:05:25,011 (DBWorkload.java:271) INFO  - ======================================================================
17:05:25,024 (DBWorkload.java:533) INFO  - Creating new TPCC database...
17:05:25,098 (DBWorkload.java:535) INFO  - Finished!
17:05:25,098 (DBWorkload.java:536) INFO  - ======================================================================
17:05:25,098 (DBWorkload.java:559) INFO  - Loading data into TPCC database with 80 threads...
17:07:49,983 (DBWorkload.java:563) INFO  - Finished!
17:07:49,984 (DBWorkload.java:564) INFO  - ======================================================================
17:07:49,987 (DBWorkload.java:849) INFO  - Creating 32 virtual terminals...
17:07:50,038 (DBWorkload.java:854) INFO  - Launching the TPCC Benchmark with 1 Phase...
17:07:50,050 (ThreadBench.java:341) INFO  - PHASE START :: [Workload=TPCC] [Serial=false] [Time=60] [WarmupTime=0] [Rate=10000] [Arrival=REGULAR] [Ratios=[45.0, 43.0, 4.0, 4.0, 4.0]] [ActiveWorkers=32]
17:07:50,050 (ThreadBench.java:492) INFO  - MEASURE :: Warmup complete, starting measurements.
17:08:50,051 (ThreadBench.java:447) INFO  - TERMINATE :: Waiting for all terminals to finish ..
17:08:50,055 (ThreadBench.java:508) INFO  - Attempting to stop worker threads and collect measurements
17:08:50,057 (ThreadBench.java:247) INFO  - Starting WatchDogThread
17:08:50,284 (DBWorkload.java:860) INFO  - ======================================================================
17:08:50,290 (DBWorkload.java:861) INFO  - Rate limited reqs/s: Results(nanoSeconds=60000236384, measuredRequests=600005) = 10000.043935826905 requests/sec
```

lmwnshn:
Corrected numbers since I forgot rate limits were a thing:

this branch no wal:
12:15:09,683 (DBWorkload.java:861) INFO  - Rate limited reqs/s: Results(nanoSeconds=60000934377, measuredRequests=1738649) = 28977.032075461677 requests/sec

main no wal:
12:09:11,836 (DBWorkload.java:861) INFO  - Rate limited reqs/s: Results(nanoSeconds=60000700282, measuredRequests=1693533) = 28225.220573101444 requests/sec

Within noise.

PR #1473, opened 2021-02-15
Favor std::mutex over SpinLatch for OutputWriter synchronization.

mbutrovich:
Found this will looking at instrumenting the output layer. We shouldn't be using `SpinLatch` for operations longer than ~20 instructions (per Intel). Since writing to the network layer for a large number of tuples is definitely going to involve sys calls and more instructions than that, we should favor a `std::mutex` for synchronization here.


lmwnshn (approved):
LGTM.

PR #1474, opened 2021-02-15
Hotfix for CI Builds.

turingcompl33t:
As the conversation below indicates, updates that I made in a previous PR lead to issues with parallel builds. Specifically, the changes I introduced made it so that multiple top-level targets depend on generation of the bytecode handlers file, `bytecode_handlers_ir.bc` (e.g. one of the execution jumbotests and the TPL runtime are not interdependent, but both depend on generating the bytecode handlers). This leads to concurrency issues in parallel builds wherein the script that is responsible for compiling the bytecode and subsequently optimizing it fails to locate a temporary file upon which it depends because it has been renamed / overwritten by a concurrently-running invocation of the custom command. Gross.

The (proposed) fix here is pretty simple, and follows in that great computer science tradition of solving problems by introducing an additional layer of indirection: construct a custom target (via `add_custom_target()`) for bytecode file generation that expresses dependency on the file itself, and have each of the top-level targets then depend on this custom target. This ensures that, once this target is "built", it is never built again because the dependency is already satisfied, thus avoiding the race entirely. You can confirm this is taking place by comparing the output from previous failing builds with the build progress output on this branch; you will observe that, whereas in prior builds the output from the `generate-bitcodes.sh` script appeared multiple times, it now always appears exactly once.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -751,11 +751,6 @@ add_custom_command(
 # (e.g. Make, Ninja) is now 'aware' of the target-level dependency.
 add_custom_target(bitcode_generator DEPENDS "${CMAKE_BINARY_DIR}/bin/bytecode_handlers_ir.bc")
 
-# A declarative convenience wrapper for expressing dependency on bitcode.
-function(target_requires_bitcode TARGET)
-    add_dependencies(${TARGET} bitcode_generator)
-endfunction()
-
 #######################################################################################################################
 # HEADER gen_opt_bc binary.
 # gen_opt_bc            :   Generates optimized bitcode from the LLVM IR file generated as a tpl post-build target.
@@ -767,7 +762,7 @@ target_include_directories(gen_opt_bc PRIVATE ${LLVM_INCLUDE_DIRS})
 target_link_libraries(gen_opt_bc PRIVATE ${LLVM_LIBRARIES})
 set_target_properties(gen_opt_bc PROPERTIES CXX_EXTENSIONS OFF ENABLE_EXPORTS ON)
 
-target_requires_bitcode(noisepage_static)
+add_dependencies(noisepage_static bitcode_generator)
 
 #######################################################################################################################
 # HEADER tpl binary.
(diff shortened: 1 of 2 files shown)


lmwnshn:
Do we understand why it broke though?

turingcompl33t:
I think it must be a concurrency issue in the build, so in retrospect now this won't actually address the root cause, which is that the stuff I added makes it so that multiple targets depend on this custom target that generates the bytecode handlers. So what we really need is some way to express: only run this external command once for any build invocation, regardless of how many targets may depend upon it.

CMakeLists.txt
```diff
@@ -727,17 +727,33 @@ add_library(hack_bytecode_handlers_ir EXCLUDE_FROM_ALL util/execution/bytecode_h
...
+
+# Here, we create a custom target that merely wraps the output file
+# from the custom command above; this ensures that the custom command
+# is actually only ever invoked once even when multiple indepdendent
+# targets that require bitcode are built because the build system 
+# (e.g. Make, Ninja) is now 'aware' of the target-level dependency.
+add_custom_target(bitcode_generator DEPENDS "${CMAKE_BINARY_DIR}/bin/bytecode_handlers_ir.bc")
+
+# A declarative convenience wrapper for expressing dependency on bitcode.
 function(target_requires_bitcode TARGET)
```
lmwnshn, line 755: Since this function doesn't do anything else, does it make sense to just directly write `add_dependencies(... bitcode_generator)` wherever necessary? Seems clearer.

lmwnshn (commented):
I am not really a fan of custom functions where cmake builtins will do, thoughts?

CMakeLists.txt
turingcompl33t: Only argument I have for maybe keeping it around was that it made this refactor much easier - all I had to do was hack around on stuff that would then populate the body of this function, leaving all of the "call sites" unmodified. This has the potential to make future updates less invasive.

But I see your point, because in it's current state it looks like this might be hiding some complexity, when in reality it actually does just the one thing. We might need a tiebreaker vote.

lmwnshn: I guess the question is, do you foresee future updates? If yes, happy to merge as is, otherwise I feel anticipating use cases has generally gone poorly / added indirection where it didn't help much.

lmwnshn: All code would be ctrl-f'able on `bitcode_generator`.

turingcompl33t: Pushed the refactor.

PR #1475, opened 2021-02-16
Create/Drop Index Action Fix and Costing Adjustment for Pilot

malin1993ml:
This PR has the following changes to the pilot:
1. Fix the application of create/drop index actions during the action search (we couldn't apply those actions before).
2. Adjust the cost type of MCTS tree node to `double` because the costs may not always be an integer.
3. Use the total cost (latency) for queries as the return of `PilotUtil::ComputeCost` instead of the average because the pilot may apply `PilotUtil::ComputeCost` with different numbers of segments (and different numbers of total queries). Just looking at the average cost is unfair to actions that benefit many segments (and queries).
4. Introduce a new option to calculate the cost of the internal tree nodes based on the minimum cost of all the leaf nodes (instead of the average). I found this useful since our costs are real values (instead of 0/1 in games like Go), averaging the leaf costs may penalize good actions that unluckily have very bad children (thus make the average cost super high).
5. (nit) Remove the `BUILD_ABS_PATH` environment variable from the pilot, so that we don't need an external script to set it.


rabbit721:
The changes look good! I agree that the min cost would be more helpful in some cases. Thanks a lot for the work!

rabbit721 (approved):
Approving the PR.

PR #1477, opened 2021-02-19
Delete domain socket on DBMS shutdown. Clarify junit tracefile reporting.

lmwnshn:
# Heading

Delete domain socket on DBMS shutdown. Clarify junit tracefile reporting.

## Description

Normally when you spin up a DBMS we get the network socket and then the domain socket.

```
[2021-02-19 11:50:54.637] [network_logger] [info] Listening on networked socket with port 15721 [PID=3372]
// my PR got stuck here
[2021-02-19 11:50:54.637] [network_logger] [info] Listening on Unix domain socket with port 15721 [PID=3372]
```

I noticed that one of my CI runs was failing because it couldn't get the domain socket, so this PR will try to delete the domain socket if it exists when the DBMS is stopped. In theory our DBMS startup code already does this but `¯\_(ツ)_/¯` maybe this helps.

Also, good news: we're _not_ running the tracefiles a million times, the scripts were just opening `junit_log.txt` in append "a+" mode and the entire file would get printed each time. Changed it to open the file in "w", seems to have the desired behavior locally when I introduce errors in two different trace files (only prints the relevant bit).

Other small qol fixes:

- stop Java tracefile stuff early if the DBMS crashes so that you don't see a huge list of "connection closed" errors
- add protection to some bare `except` handlers so that `KeyboardInterrupt` can go through
- sort tracefiles before running
- typo fix

### DIFF
--- a/script/testing/junit/__main__.py
+++ b/script/testing/junit/__main__.py
@@ -95,6 +95,8 @@ if __name__ == "__main__":
         set_env_vars()
         test_server = TestServer(args, quiet=True)
         errcodes = [run_tests_junit(test_server)] + run_tests_tracefiles(test_server)
+    except KeyboardInterrupt:
+        traceback.print_exc(file=sys.stderr)
     finally:
         unset_env_vars()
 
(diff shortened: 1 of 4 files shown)


mbutrovich (approved):
LGTM. Will merge once through CI.

Issue #1476, opened 2021-02-18
Possible race in atomics_test locating the bytecode.


lmwnshn:
# Bug Report

## Summary

`atomics_test` occasionally fails because 

```
[----------] 16 tests from AtomicsTest
unknown file: Failure
C++ exception with description "filesystem error: cannot increment recursive directory iterator: No such file or directory" thrown in SetUpTestSuite().
[ RUN      ] AtomicsTest.InterpretedAndOr1
...
```

## Environment

http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/PR-1472/24/tests

## Discussion

Per discussion on slack, this is probably because we are doing a root level walk of `noisepage`.
This would also have caused problems if you had multiple `bytecode_handers_ir.bc` files lying around.
We could completely construct the path to the `bytecode_handers_ir.bc` file instead.

PR #1478, opened 2021-02-19
Fix Directory Walk Race in Unit Tests

turingcompl33t:
Closes [#1476](https://github.com/cmu-db/noisepage/issues/1476).

The way we currently locate the bytecode handlers file in compiled tests (the `CompiledTplTest` fixture) performs a recursive directory walk that may sometimes fail when the directory that it enters is subsequently blown away by Ninja while running tests in parallel. This PR changes the way we locate the bytecode handlers file to always search for it in a known, stable location.


lmwnshn:
Do we still need noisepage project root?

turingcompl33t:
> Do we still need noisepage project root?

Indeed we do not! Removed the dead code.

test/include/test_util/fs_util.h
```diff
@@ -5,19 +5,20 @@
...
- * recursive directory search from the specified root path.
- * @param filename The name of the target file
- * @param root_path The absolute path from which recursive search should begin
- * @return The absolute path to a file with the specified name, as a string.
- * @throw std::runtime_error in the event the file cannot be found.
+ * Determine the absolute path to the build artifact identified by `name`.
+ *
+ * By using this function, you are expressing the assumption that the
+ * artifact identified by `name` is located in the binary directory for
+ * the current build; that is, `name` is located at `build/bin/<name>`.
```
lmwnshn, line 18: It doesn't have to be `build` right? Just whatever the build root is? `GetBuildRootPath()/bin/name` might be clearer, if so (because we do hardcode some assumptions about `build` elsewhere)

lmwnshn (approved):
LGTM, minor nit if you want to address, thanks!

flip to ready-to-merge when you've decided either way

test/include/test_util/fs_util.h
turingcompl33t: Thanks for the comment; you are right, the whole point of the function is to avoid backing-in logic that relies on a specific directory name / path. Updated it!

turingcompl33t:
Broke one of the microbenchmarks when I changed the API, fixing now...

PR #1494, opened 2021-02-23
Increase db_server start time from 60s to 600s.

malin1993ml:
We're experiencing lots of delays in the db_server start that make passing CI almost impossible (always time out when the server cannot start within 60s). I'm just trying to do a sanity check to see whether the server can start eventually and if so how long it would take (by increasing the timeout to 10 minutes).


lmwnshn (approved):
LGTM.

PR #1496, opened 2021-02-25
Fix bplustree.h redefined constant shared with bwtree.h.

lmwnshn:
# Heading

Fix bplustree.h redefined constant shared with bwtree.h.

## Description

Some PRs are failing CI due to `#define` definition leaking in some unity builds, and I think we'd like to avoid random define pollution in headers anyway.

### DIFF
--- a/src/include/storage/index/bplustree.h
+++ b/src/include/storage/index/bplustree.h
@@ -71,14 +71,14 @@ class BPlusTreeBase {
 
  protected:
   /** upper size threshold for inner node split */
-  int inner_node_size_upper_threshold_ = ((int)128);
+  int inner_node_size_upper_threshold_ = 128;
   /** lower size threshold for inner node removal */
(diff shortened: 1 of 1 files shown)


src/include/storage/index/bplustree.h
```diff
@@ -11,22 +11,12 @@
...
 #include "storage/index/index.h"
 #include "storage/index/index_defs.h"
 
-#include "loggers/index_logger.h"
-
 namespace noisepage::storage::index {
 
-// If node size goes above this then we split it
-// Thresholds for Inner Node
-#define INNER_NODE_SIZE_UPPER_THRESHOLD ((int)128)
```
mbutrovich, line 23: These should have been `constexpr` but I guess we missed that on review. We'll do a `#define` sweep of the codebase again at a later date.

PR #1499, opened 2021-02-25
Switch gcc/Debug/Coverage/unittest to only use h0.

lmwnshn:
Per @malin1993ml ,

> Another minor CI improvement I've been thinking is that maybe we can set ubuntu-20.04/gcc-9.3 (Debug/Coverage/unittest) to always run on h0 . It is building 3x more targets than other builds, and seems always to be the bottleneck in our Test stage.

This is done by introducing a new `h0-only` label to h0, and manually specifying that this specific job can only execute on nodes with the `h0-only` label.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -12,7 +12,6 @@ pipeline {
             agent {
                 docker {
                     image 'noisepage:focal'
-                    label 'h0-only'
                     args '-v /jenkins/ccache:/home/jenkins/.ccache'
                 }
             }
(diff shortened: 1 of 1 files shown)


malin1993ml (approved):
LGTM! I'm assuming you also need to change something on Jenkins to make it work?

lmwnshn:
Yes, changed already

malin1993ml:
It looks like this is working. Both the second and third builds for `gcc/Debug/Coverage/unittest` are scheduled at `h0`. I don't think we need later stages to finish.

PR #1479, opened 2021-02-20
ModelServer Abstraction and Integration

17zhangw:
## Description
PR adjusts the `ModelServerManager` and related scripts to support utilizing multiple models. This PR introduces the following high level changes:
- Restructured sub-part of script directory to accommodate different models
- Allow ModelServerManager to Train/Infer different models with different arguments
- QueryTrace metric records `db_oid` for both `query_text.csv` and `query_trace.csv` (for consistency)

## Further work
The following will be introduced in a separate PR:
- Performing workload forecast inference as part of pilot planning
- Storing inference results in internal tables via internal sql execution utility
- Collect query run stats (forecast segment granularity) and query parameter sampling (forecast interval granularity) directly from metrics thread to internal tables
- Storing inference results in internal tables
- Constructing `WorkloadForecast` from inference results + query information

Issue #1352, opened 2020-11-23
Caching of ExecutionSettings


17zhangw:
## Summary
As @mbutrovich mentions on #1351, PR #1208 creates a new `ExecutionSettings` by reading from the `SettingsManager` on every query execution. We should be caching the settings values in the correct class and relying on the settings callbacks to update any change to the `ExecutionSettings` parameters.

PR #1502, opened 2021-03-01
Forecasting Integration and Query Logging

17zhangw:
## Description

PR builds upon #1479 to add forecast integration and query trace logging. Major changes introduced in this PR are as follows:
- `QueryExecUtil` is an internal SQL execution utility for executing SQL internally. The execution utility currently wraps the optimization and execution logic to execute queries. If needed the `QueryExecUtil` implementation can be swapped out with a client (i.e., pqxx).
- `QueryInternalThread` is a background thread owned by `DbMain` that handles executing SQL queries asynchronously. Tasks are attached on a queue with lock-based synchronization. Currently used by the `Pilot` for recording forecast predictions and also by the `QueryTraceMetric` to record query segment information.
- Option for metrics to be logged to either CSVs or internal tables
- `QueryTraceMetric` logging of query segment frequency information and parameter sampling
- Ability to execute DDL on server startup
- `ReservoirSampling` implements https://en.wikipedia.org/wiki/Reservoir_sampling#With_random_sort
- `WorkloadForecast` can be constructed from logged query segments and forecast predictions

## Remaining tasks

- [x] Further testing


src/include/util/query_internal_thread.h
```diff
@@ -0,0 +1,88 @@
...
+class AbstractCostModel;
+}
+
+namespace noisepage::parser {
+class ConstantValueExpression;
+}
+
+namespace noisepage::util {
+
+class ExecuteRequest {
```
lmwnshn, line 26: Nit: documentation

src/include/util/query_internal_thread.h
```diff
@@ -0,0 +1,88 @@
...
+#include <chrono>  //NOLINT
+#include <condition_variable>
+#include <mutex>
+#include <queue>
+#include <thread>  //NOLINT
+
+#include "catalog/catalog_defs.h"
+#include "execution/compiler/executable_query.h"
+#include "optimizer/cost_model/abstract_cost_model.h"
+#include "planner/plannodes/output_schema.h"
```
lmwnshn, line 12: Is this used? In general, some includes can be maybe forward declared here

src/self_driving/planning/pilot_util.cpp
```diff
@@ -27,111 +28,76 @@
...
 
-  if (query_type == network::QueryType::QUERY_SET) {
-    // The set statements are executed differently (through settings_manager_)
-    const auto &set_stmt = statement->RootStatement().CastManagedPointerTo<parser::VariableSetStatement>();
-    pilot->settings_manager_->SetParameter(set_stmt->GetParameterName(), set_stmt->GetValues());
+    if (is_create_idx) {
+      bool success = true;
+      size_t idx = util.CompileQuery(sql_query, nullptr, nullptr, &success);
+      if (success) util.ExecuteQuery(idx, nullptr, nullptr, nullptr);
+      std::this_thread::sleep_for(std::chrono::seconds(1));
```
lmwnshn, line 63: A sleep?

src/self_driving/planning/pilot_util.cpp
```diff
@@ -27,111 +28,76 @@
...
-          db_oid, common::ManagedPointer(txn), callback, out_plan->GetOutputSchema().Get(),
-          common::ManagedPointer(accessor), exec_settings, pilot->metrics_thread_->GetMetricsManager());
-
-      exec_query->Run(common::ManagedPointer(exec_ctx), execution::vm::ExecutionMode::Interpret);
-    }
-    txn_manager->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+    bool success = true;
+    size_t idx = util.CompileQuery(sql_query, nullptr, nullptr, &success);
+    if (success) util.ExecuteQuery(idx, nullptr, nullptr, nullptr);
+    std::this_thread::sleep_for(std::chrono::seconds(1));
```
lmwnshn, line 70: Sleep?

src/self_driving/planning/pilot_util.cpp
```diff
@@ -166,7 +132,8 @@ double PilotUtil::ComputeCost(common::ManagedPointer<Pilot> pilot, common::Manag
       prev_qid = pipeline_to_pred.first.first;
     }
   }
-  double total_cost = 0, num_queries = 0;
+  double total_cost = 0;
+  double num_queries = 0;
```
lmwnshn, line 136: Can you have a fractional number of queries?

src/util/query_exec_util.cpp
```diff
@@ -0,0 +1,283 @@
...
+  auto accessor = catalog_->GetAccessor(txn, db_oid_, DISABLED);
+  auto exec_ctx = std::make_unique<execution::exec::ExecutionContext>(
+      db_oid_, txn, callback, schema, common::ManagedPointer(accessor), exec_settings_, metrics);
+
+  exec_ctx->SetParams(common::ManagedPointer<const std::vector<parser::ConstantValueExpression>>(params.Get()));
+
+  exec_queries_[idx]->Run(common::ManagedPointer(exec_ctx), execution::vm::ExecutionMode::Interpret);
+
+  if (require_commit) {
+    EndTransaction(true);
```
lmwnshn, line 264: Then actually it isn't really `EndTransaction` right, since it isn't called otherwise?

Maybe make `EndTransaction` private, expose a `CommitTransaction` and `AbortTransaction` instead?

src/util/query_exec_util.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+  return status;
+}
+
+size_t QueryExecUtil::CompileQuery(const std::string &statement,
+                                   common::ManagedPointer<std::vector<parser::ConstantValueExpression>> params,
+                                   common::ManagedPointer<std::vector<type::TypeId>> param_types, bool *success) {
+  auto txn = common::ManagedPointer<transaction::TransactionContext>(txn_);
+  bool require_commit = false;
+  if (txn == nullptr) {
```
lmwnshn, line 188: This code appears a couple of times, maybe it can be refactored into an internal helper function

src/util/query_exec_util.cpp
```diff
@@ -0,0 +1,283 @@
...
+
+void QueryExecUtil::BeginTransaction() {
+  NOISEPAGE_ASSERT(txn_ == NULL, "Nesting transactions not supported");
+  txn_ = txn_manager_->BeginTransaction();
+  own_txn_ = true;
+}
+
+void QueryExecUtil::EndTransaction(bool commit) {
+  NOISEPAGE_ASSERT(txn_ != NULL, "Transaction has not started");
+  if (own_txn_) {
```
lmwnshn, line 78: wait, so it does nothing if the txn is owned by someone else? maybe this should be checked before call, and assert here that you own the txn?

lmwnshn (changes requested):
I get that it isn't ready for review yet, just a quick first pass of comments. Will want to do another pass.

src/util/query_exec_util.cpp
```diff
@@ -0,0 +1,283 @@
...
+                             common::ManagedPointer<optimizer::StatsStorage> stats, uint64_t optimizer_timeout)
+    : db_oid_(db_oid),
+      txn_manager_(txn_manager),
+      catalog_(catalog),
+      settings_(settings),
+      stats_(stats),
+      optimizer_timeout_(optimizer_timeout),
+      cost_func_(nullptr) {
+  exec_settings_ = execution::exec::ExecutionSettings{};
+  exec_settings_.UpdateFromSettingsManager(settings_);
```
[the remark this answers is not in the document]
17zhangw: Correct. There is a function to change the settings on-demand if needed. Arguably, solving #1352 cleanly might make this much better. (i.e, settings manager callbacks update a "execution settings view"; view is copied when needed for compile/execution?)

src/include/self_driving/planning/pilot_thread.h
```diff
@@ -84,9 +84,9 @@ class PilotThread {
         if (forecaster_remain_period_ == std::chrono::microseconds::zero()) {
           pilot_->PerformForecasterTrain();
           forecaster_remain_period_ = forecaster_train_interval_;
-        } else {
-          pilot_->PerformPlanning();
         }
+
+        pilot_->PerformPlanning();
```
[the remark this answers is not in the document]
17zhangw: Yes.

src/include/metrics/query_trace_metric.h
```diff
@@ -10,7 +10,10 @@
 #include <vector>
 
 #include "catalog/catalog_defs.h"
+#include "common/container/chunked_array.h"
+#include "common/json.h"
 #include "common/managed_pointer.h"
+#include "common/reservoir_sampling.h"
```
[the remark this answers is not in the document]
17zhangw: I don't believe it's possible. The `std::unordered_map` needs the type definition.

src/metrics/metrics_manager.cpp
```diff
@@ -147,51 +162,72 @@ void MetricsManager::UnregisterThread() {
...
-        file.close();
+
+      if (output == MetricsOutput::DB || output == MetricsOutput::CSV_DB) {
+        ToDB(component);
       }
     }
   }
+
+  if (query_exec_util_ != nullptr) {
+    query_exec_util_->ClearPlans();
```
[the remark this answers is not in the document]
17zhangw: Yes. Generally, per "thing"/per "thread".

src/include/util/query_internal_thread.h
```diff
@@ -0,0 +1,88 @@
...
+      : query_exec_util_(std::move(query_exec_util)), query_thread_(std::thread([this] { QueryThreadLoop(); })) {}
+
+  ~QueryInternalThread() {
+    run_queries_ = false;
+    queue_cv_.notify_one();
+    query_thread_.join();
+  }
+
+  void AddRequest(ExecuteRequest &&request) {
+    if (!run_queries_) {
```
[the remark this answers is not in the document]
17zhangw: I think we should enforce it for now that `AddRequest` will succeed and add `Try` if needed.

src/include/self_driving/forecasting/workload_forecast_segment.h
```diff
@@ -15,19 +15,25 @@ namespace noisepage::selfdriving {
...
 class WorkloadForecastSegment {
  public:
+  explicit WorkloadForecastSegment() = default;
+
   /**
    * Constructor for WorkloadForecastSegment
    * @param id_to_num_exec Map from qids to number of execution of this query in this interval
    */
-  explicit WorkloadForecastSegment(std::unordered_map<execution::query_id_t, uint64_t> id_to_num_exec);
+  explicit WorkloadForecastSegment(std::unordered_map<execution::query_id_t, double> id_to_num_exec);
```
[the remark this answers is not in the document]
17zhangw: It's a `double` mostly since the forecast prediction can be non-integer.

src/self_driving/planning/pilot_util.cpp
17zhangw: Yes. It's possible from the forecasting.

src/self_driving/planning/pilot_util.cpp
17zhangw: Once upon a time, there was a merge conflict.

src/self_driving/planning/pilot_util.cpp
17zhangw: Then the merge conflict was resolved strangely.

src/util/query_exec_util.cpp
17zhangw: Restructured this slightly. This is meant to commit any "implicitly" started transactions.

17zhangw:
As discussed in a meeting on March 11th, 2021, the internal thread will be replaced with a preliminary task pool interface, from which jobs can be submitted to and executed.

src/self_driving/planning/pilot.cpp
```diff
@@ -32,27 +50,355 @@ Pilot::Pilot(std::string model_save_path, std::string forecast_model_save_path,
...
+  }
+
+  {
+    auto to_row_fn = [&metadata, cves_conv](const std::vector<execution::sql::Val *> &values) {
+      auto qid = execution::query_id_t(static_cast<execution::sql::Integer *>(values[1])->val_);
+      auto *param_val = static_cast<execution::sql::StringVal *>(values[2]);
+      {
+        // Read the parameters. In the worse case, we will have double the parameters, but that is
+        // okay since every parameter will be duplicated. This can happen since the parameters
+        // could already be visible by the time this select query runs.
```
malin1993ml, line 171: Why not do a duplication-checking here as well, just like above?

src/main/startup.sql
```diff
@@ -0,0 +1,14 @@
+-- File contains DDL statements that should be run on startup.
+-- Each DDL statement must be on its own line.
+-- A line that starts with '--' is considered to be a comment.
+
+CREATE TABLE noisepage_forecast_clusters(iteration INT, cluster_id INT, query_id INT, db_id INT);
+CREATE TABLE noisepage_forecast_forecasts(iteration INT, query_id INT, interval INT, rate REAL);
+CREATE TABLE noisepage_forecast_frequencies(iteration INT, query_id INT, interval INT, seen REAL);
```
malin1993ml, line 7: Seems that `noisepage_forecast_clusters` and `noisepage_forecast_forecasts` are predictions and `noisepage_forecast_frequencies` is history? Could you document this a bit?

test/self_driving/query_trace_logging.cpp
```diff
@@ -0,0 +1,234 @@
+#include "common/json.h"
```
malin1993ml, line 1: nit: rename the file to `query_trace_logging_test.cpp`

test/common/chunked_array_test.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+  size_t idx = 0;
+  auto it = array1.begin();
+  while (it != array1.end()) {
+    ASSERT_EQ(vec[idx], *it);
+    it++;
+    idx++;
+  }
+
+  ASSERT_EQ(idx, size1 + size2);
```
malin1993ml, line 43: We should use `EXPECT_EQ` etc. in tests instead of the asserts. Then it would integrate nicely with gtest. The same applies to the query trace logging test.

src/util/query_exec_util.cpp
```diff
@@ -0,0 +1,249 @@
...
+  try {
+    std::string query_tmp = query;
+    auto parse_tree = parser::PostgresParser::BuildParseTree(query_tmp);
+    statement = std::make_unique<network::Statement>(std::move(query_tmp), std::move(parse_tree));
+  } catch (std::exception &e) {
+    // Catched a parsing error
+    return {nullptr, nullptr};
+  }
+
+  // This is unfortunate, but we currently cannot optimize SET queries.
```
malin1993ml, line 100: I don't think `SET` queries need optimization anyway...

src/util/query_exec_util.cpp
```diff
@@ -0,0 +1,249 @@
...
+  if (statement->GetQueryType() == network::QueryType::QUERY_SET) {
+    return std::make_pair(std::move(statement), nullptr);
+  }
+
+  try {
+    // TODO(wz2): Specify params?
```
malin1993ml, line 106: Is this still a TODO?

src/util/query_exec_util.cpp
```diff
@@ -0,0 +1,249 @@
...
+}
+
+bool QueryExecUtil::ExecuteDML(const std::string &query,
+                               common::ManagedPointer<std::vector<parser::ConstantValueExpression>> params,
+                               common::ManagedPointer<std::vector<type::TypeId>> param_types, TupleFunction tuple_fn,
+                               common::ManagedPointer<metrics::MetricsManager> metrics,
+                               std::unique_ptr<optimizer::AbstractCostModel> cost,
+                               const execution::exec::ExecutionSettings &exec_settings) {
+  uint64_t idx = 0;
+  if (!CompileQuery(query, params, param_types, std::move(cost), exec_settings, &idx)) {
```
malin1993ml, line 242: We should actually cache the compiled internal query. This can be local to a `QueryExecUtil`. You just need to change the query vector to a map. I think the query strings should be sufficient to uniquely determine a query.

src/util/query_exec_util.cpp
```diff
@@ -0,0 +1,249 @@
...
+        // Drop index does not need execution of compiled query
+        status =
+            execution::sql::DDLExecutors::DropIndexExecutor(common::ManagedPointer<planner::AbstractPlanNode>(out_plan)
+                                                                .CastManagedPointerTo<planner::DropIndexPlanNode>(),
+                                                            common::ManagedPointer<catalog::CatalogAccessor>(accessor));
+        break;
+      case network::QueryType::QUERY_CREATE_INDEX:
+        // TODO(lin): We actually don't need to populate the index tuples after creating the index placeholder for the
+        //  "what-if" API. But since we need to execute the query to get the features, we need to compile and execute
+        //  the query for now.
```
malin1993ml, line 153: We actually do need to compile and execute `CREATE_INDEX` for now. See the implementation in the original `pilot_util.cpp`.

malin1993ml (changes requested):
Thanks for the great work, @17zhangw! I did the first pass on this, except the code related to the task manager, which I think you might still be working on? I think most of the infrastructure is already there. Just have some comments with tweaks and clarifications. But I'm actually not fully understanding the `iteration` related implementation and the handling of interval boundaries. Let's schedule a discussion for that.

src/include/task/task_manager.h
```diff
@@ -0,0 +1,55 @@
...
+class TaskRunner;
+
+class TaskManager : public common::DedicatedThreadOwner {
+ public:
+  TaskManager(common::ManagedPointer<common::DedicatedThreadRegistry> thread_registry,
+              std::unique_ptr<util::QueryExecUtil> util, uint32_t num_workers);
+
+  ~TaskManager();
+
+  void Flush();
```
malin1993ml, line 27: Seems that `Flush` only waits for things to finish, instead of trying to actively flush things? If so, it would be good to clarify.

src/include/task/task.h
```diff
@@ -0,0 +1,66 @@
...
+
+ private:
+  catalog::db_oid_t db_oid_;
+  std::string query_text_;
+};
+
+class TaskDML : public Task {
+ public:
+  TaskDML(catalog::db_oid_t db_oid, std::string query_text, std::unique_ptr<optimizer::AbstractCostModel> cost_model,
+          std::vector<std::vector<parser::ConstantValueExpression>> &&params, std::vector<type::TypeId> &&param_types)
```
malin1993ml, line 45: Is it by design that DML tasks that go through the `TaskManager` do not have any callback functions to process the tuples?

malin1993ml (changes requested):
I think the `TaskManager` also generally looks good. It seems that we execute asynchronous queries without retrieving the results through `TaskManager`, but we directly execute queries that read results with `QueryExecUtil`. I'm not sure whether we should unify them.

src/include/common/container/chunked_array.h
```diff
@@ -0,0 +1,157 @@
...
+#include <utility>
+#include <vector>
+
+namespace noisepage::common {
+
+/**
+ * Describes a chunked array. A chunked array batch allocates Size number of elements
+ * of type Type at a given time. The premise for this is to amortize comparably
+ * expensive memory allocation calls over a large number of inserts.
+ */
```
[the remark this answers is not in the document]
17zhangw: The issue I was trying to avoid with `std::vector` is that its iterator can get invalidated when adding new elements to the vector. `ChunkedArray` ensures that the iterator is still valid when new elements are added.

script/self_driving/forecasting/data_loader.py
```diff
@@ -31,8 +32,13 @@ def __init__(self,
...
 
-        data = self._load_data()
-        self._to_timeseries(data)
+        if query_trace_file is not None:
+            data = self._load_data()
+            self._to_timeseries(data)
+        else:
+            self._ts_data = {}
+            for item in query_sequence:
+                self._ts_data[item[0]] = item[1]
```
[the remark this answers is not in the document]
17zhangw: I am not too certain myself about this.

src/include/metrics/query_trace_metric.h
```diff
@@ -19,26 +23,162 @@
...
  * Raw data object for holding stats collected at logging level
  */
 class QueryTraceMetricRawData : public AbstractRawData {
  public:
-  void Aggregate(AbstractRawData *const other) override {
+  /** Parameter of how many query params to keep in sample */
+  static uint64_t query_param_sample;
+
+  /** Parameter controlling size of a query segment */
+  static uint64_t query_segment_interval;
```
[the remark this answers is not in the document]
17zhangw: `clang-tidy` didn't complain about this, but I think looking around in the system, these should be in caps.

src/include/metrics/query_trace_metric.h
```diff
@@ -82,13 +222,22 @@ class QueryTraceMetricRawData : public AbstractRawData {
   FRIEND_TEST(MetricsTests, QueryCSVTest);
 
   void RecordQueryText(catalog::db_oid_t db_oid, const execution::query_id_t query_id, const std::string &query_text,
-                       const std::string &type_string, const uint64_t timestamp) {
+                       const std::string &type_string, const std::string &type_json, const uint64_t timestamp) {
     query_text_.emplace_back(db_oid, query_id, query_text, type_string, timestamp);
+    metadata_.RecordQueryText(query_id, db_oid, query_text, type_json);
+
+    // Don't track range of time data for QueryText.
+    // QueryText is not linked with when the query is run.
```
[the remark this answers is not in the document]
17zhangw: Clarified with note that `RecordQueryText` not necessarily invoked when query (prepared) is run.

src/include/common/reservoir_sampling.h
```diff
@@ -0,0 +1,114 @@
...
+/**
+ * Implementation of reservoir sampling. This is done in a "streaming" style where
+ * we want to sample k elements from an unknown population, where elements are
+ * being inserted "streaming" style.
+ */
+template <class Value>
+class ReservoirSampling {
+ public:
+  /**
+   * Describes a key (comprising a priority and a data element)
```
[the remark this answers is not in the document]
17zhangw: Yes, the main advantage Is that it can be extended with sampling weights.

src/main/db_main.cpp
```diff
@@ -32,6 +73,12 @@ void DBMain::ForceShutdown() {
   if (recovery_manager_ != DISABLED && recovery_manager_->IsRecoveryTaskRunning()) {
     recovery_manager_->WaitForRecoveryToFinish();
   }
+
+  // Need to let internal thread flush through requests
+  (void)pilot_thread_.release();
+  (void)metrics_thread_.release();
+  (void)task_manager_.release();
```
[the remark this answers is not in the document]
17zhangw: The pilot thread and metrics thread have references to the task manager. This is more of just shutting down components in the correct order so the pilot/metrics thread don't try to submit tasks to a task manager being deleted.

src/util/query_exec_util.cpp
17zhangw: Old comment removed.

src/util/query_exec_util.cpp
17zhangw: Unified the strangely placed code inside of `pilot_util.cpp`.

src/task/task.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+void TaskDDL::Execute(common::ManagedPointer<util::QueryExecUtil> query_exec_util,
+                      common::ManagedPointer<task::TaskManager> task_manager) {
+  query_exec_util->BeginTransaction(db_oid_);
+  bool status = query_exec_util->ExecuteDDL(query_text_);
+  query_exec_util->EndTransaction(status);
+}
+
+void TaskDML::Execute(common::ManagedPointer<util::QueryExecUtil> query_exec_util,
+                      common::ManagedPointer<task::TaskManager> task_manager) {
```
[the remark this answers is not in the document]
17zhangw: `SET` queries are executable (they are currently lumped together in `ExecuteDDL`).

src/self_driving/planning/pilot.cpp
17zhangw: In the modified version, parameters are still logged whenever we are going to perform planning (i.e., parameters are logged at a batch timestamp). However, I've modified the range scan query to exclude the most recent batch of parameters written.

src/main/startup.sql
```diff
@@ -0,0 +1,14 @@
+-- File contains DDL statements that should be run on startup.
+-- Each DDL statement must be on its own line.
+-- A line that starts with '--' is considered to be a comment.
```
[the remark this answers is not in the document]
17zhangw: We can specify additional indexes (directly via CREATE INDEX).

src/include/common/container/chunked_array.h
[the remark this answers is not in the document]
malin1993ml: I see! Can you add that comment here since one may wonder?

src/include/settings/settings_defs.h
```diff
@@ -287,6 +310,14 @@ SETTING_bool(
     noisepage::settings::Callbacks::MetricsQueryTrace
 )
 
+SETTING_string(
+    query_trace_metrics_output,
+    "Output type for Query Traces Metrics (default: CSV)",
```
malin1993ml, line 315: Maybe document what are the valid options.

src/util/query_exec_util.cpp
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  const std::unique_ptr<planner::AbstractPlanNode> &out_plan = result.second;
+  NOISEPAGE_ASSERT(network::NetworkUtil::DMLQueryType(result.first->GetQueryType()), "ExecuteDML expects DML");
+  common::ManagedPointer<planner::OutputSchema> schema = out_plan->GetOutputSchema();
+
+  auto exec_query = execution::compiler::CompilationContext::Compile(*out_plan, exec_settings, accessor.get(),
+                                                                     execution::compiler::CompilationMode::OneShot);
+  schemas_[statement] = schema->Copy();
+  exec_queries_[statement] = std::move(exec_query);
```
malin1993ml, line 191: Do we actually use this query cache? It doesn't seem that we're checking whether this query is cached before recompiling it.

src/include/common/container/chunked_array.h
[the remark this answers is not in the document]
17zhangw: Added as a comment.

src/include/settings/settings_defs.h
17zhangw: Documented.

src/util/query_exec_util.cpp
17zhangw: Thanks for the catch!

src/metrics/query_trace_metric.cpp
```diff
@@ -0,0 +1,215 @@
...
+
+  // Since the high_timestamp_ is also controlled by the time the MetricsThread performs
+  // logging to disk or internal tables, it is possible for [freqs] to contain a full
+  // segment of data.
+  //
+  // This only happens in the following state:
+  // 1. The timeseries data has been fully read. If there is still remaining timeseries
+  //    data, then we should not be committing a segment here.
+  //
+  // 2. A segment interval has passed
```
malin1993ml, line 147: Thanks for the comments and explanation!

lmwnshn:
Per meeting, I'm going to take a review pass on the task manager

script/self_driving/model_server.py
```diff
@@ -420,36 +420,29 @@ class ForecastModel(AbstractModel):
...
+            input_sequence: Trace sequence or None
             save_path: PATH_TO_SAVE_MODEL_MAP
             interval_micro_sec: Interval duration for aggregation in microseconds
+            sequence_length: Number of intervals in a data sequence
+            horizon_length: Number of intervals ahead for the planning horizon
         }
         :return: if training succeeds, {True and empty string}, else {False, error message}
         """
-        input_path = data["input_path"]
+        input_path = data["input_path"] if "input_path" in data else None
```
lmwnshn, line 438: fyi Python has a `data.get("input_path", None)`, more generally `dict.get(key, default_val_if_not_exists)`

src/task/task.cpp
```diff
@@ -0,0 +1,52 @@
...
+  if (params_.empty()) {
+    result = query_exec_util->ExecuteDML(query_text_, nullptr, nullptr, tuple_fn_, nullptr,
+                                         std::make_unique<optimizer::TrivialCostModel>(), settings);
+  } else {
+    std::vector<parser::ConstantValueExpression> &params_0 = params_[0];
+    result = query_exec_util->CompileQuery(query_text_, common::ManagedPointer(&params_0),
+                                           common::ManagedPointer(&param_types_), std::move(cost_model_), settings);
+
+    // Execute with specified parameters only if compilation succeeded
+    if (!result) {
```
malin1993ml, line 33: This should be `result`, otherwise the query won't execute if compilation succeeds.

src/self_driving/planning/pilot.cpp
```diff
@@ -32,27 +50,310 @@ Pilot::Pilot(std::string model_save_path, std::string forecast_model_save_path,
...
+    }
+
+    // Record forecast into internal tables
+    RecordWorkloadForecastPrediction(timestamp, result.first, metadata_result.first);
+
+    // Construct workload forecast
+    forecast_ = std::make_unique<selfdriving::WorkloadForecast>(result.first, std::move(metadata_result.first));
+  } else {
+    auto sample = settings_manager_->GetInt(settings::Param::forecast_sample_limit);
+    forecast_ = std::make_unique<selfdriving::WorkloadForecast>(workload_forecast_interval_, sample);
```
malin1993ml, line 342: There should probably be three different options for `LoadWorkloadForecast`: (1) read from internal tables to do inference, (2) read from the disk to do inference, and (3) directly read the query trace from the disk without inference. All three scenarios could be useful for us. And right now if we're not in case (1), we'll do both (2) and (3) together and just use the results from (3). This could be confusing.

src/util/query_exec_util.cpp
```diff
@@ -0,0 +1,254 @@
...
+  auto txn = common::ManagedPointer<transaction::TransactionContext>(txn_);
+  auto accessor = catalog_->GetAccessor(txn, db_oid_, DISABLED);
+
+  std::unique_ptr<network::Statement> statement;
+  try {
+    std::string query_tmp = query;
+    auto parse_tree = parser::PostgresParser::BuildParseTree(query_tmp);
+    statement = std::make_unique<network::Statement>(std::move(query_tmp), std::move(parse_tree));
+  } catch (std::exception &e) {
+    // Catched a parsing error
```
malin1993ml, line 96: I don't know whether we want to crash the system here, but at least we should print out those exceptions with `LOG_ERROR`. Otherwise, it's pretty difficult to debug (e.g., when I'm trying to run the pilot with the new task manager today).

src/include/task/task.h
```diff
@@ -0,0 +1,160 @@
...
+   * @param cost_model Cost model to use for optimizing query
+   * @param params Relevant query parameters
+   * @param param_types Types of the query parameters if any
+   * @param tuple_fn Function for processing rows
+   * @param metrics_manager Metrics Manager to be used
+   * @param force_abort Whether to forcefully abort the transaction
+   * @param sync Future for the caller to block on
+   */
+  TaskDML(catalog::db_oid_t db_oid, std::string query_text, std::unique_ptr<optimizer::AbstractCostModel> cost_model,
+          std::vector<std::vector<parser::ConstantValueExpression>> &&params, std::vector<type::TypeId> &&param_types)
```
malin1993ml, line 85: Could you add a flag on whether to re-optimize the query? This could be useful during the pilot planning (as a "what-if" API).

src/task/task.cpp
17zhangw: Thanks for the catch!

src/self_driving/planning/pilot.cpp
17zhangw: Added 3 separate flags that essentially does:
- `INTERNAL_TABLES_WITH_INFERENCE`: pulls all data from internal tables and records inference results
- `DISK_WITH_INFERENCE`: inference from disk file and construct workload forecast from inference results (no recording to internal tables)
- `DISK_ONLY`: construct workload forecast without any inference

src/task/task.cpp
```diff
@@ -0,0 +1,65 @@
...
+    if (result) {
+      for (auto &param_vec : params_) {
+        if (!result) break;
+
+        result &= query_exec_util->ExecuteQuery(query_text_, tuple_fn_, common::ManagedPointer(&param_vec), nullptr,
+                                                settings);
+      }
+    }
+
+    // TODO(wz2): Require disciplined plan for clearing the plans
```
lmwnshn, line 50: So plans are just not cleared right now?

lmwnshn (commented):
Generally LGTM, just some questions, particularly:

1. our own custom future over the C++ future?
2. every DMLTask having its own cost model

And an unrelated comment related to reservoir sampling

src/execution/compiler/compilation_context.cpp
```diff
@@ -191,10 +191,14 @@ void CompilationContext::GeneratePlan(const planner::AbstractPlanNode &plan) {
...
-                                                             const CompilationMode mode,
+                                                             const CompilationMode mode, bool override_qid,
+                                                             execution::query_id_t override_qid_target,
                                                              common::ManagedPointer<const std::string> query_text) {
   // The query we're generating code for.
   auto query = std::make_unique<ExecutableQuery>(plan, exec_settings);
+  if (override_qid) {
+    query->SetQueryId(override_qid_target);
+  }
   // TODO(Lin): Hacking... remove this after getting the counters in
```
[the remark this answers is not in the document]
malin1993ml: Yes, this can be removed now. I'll just remove that in my own branch and submit that whenever I submit a PR.

src/task/task_runner.cpp
```diff
@@ -0,0 +1,29 @@
+#include "task/task_runner.h"
+#include "execution/compiler/executable_query.h"
+#include "planner/plannodes/output_schema.h"
```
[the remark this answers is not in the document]
17zhangw: This is actually needed to correctly size `QueryExecUtil`.

src/task/task.cpp
17zhangw: Currently plans aren't being cleared.
(this works because we don't execute a diverse set of queries).
(punting on this till there's an explicit need for plan invalidation).

src/include/task/task.h
```diff
@@ -0,0 +1,178 @@
...
+
+/**
+ * Abstract class for defining a task
+ */
+class Task {
+ public:
+  /**
+   * Executes the task
+   * @param query_exec_util Query execution utility
+   * @param task_manager TaskManager that task was submitted to
```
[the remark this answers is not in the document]
17zhangw: "The task was submitted to". Clarified the wording.

src/include/task/task_manager.h
```diff
@@ -0,0 +1,110 @@
...
+  void SetTaskPoolSize(int num_workers);
+
+  /**
+   * Submit a task to the TaskManager for execution
+   * @param task Task to execute
+   */
+  void AddTask(std::unique_ptr<Task> task);
+
+  /**
+   * TODO(wz2): These need to be revisited when the thread registry can scale
```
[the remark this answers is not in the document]
17zhangw: Clarified the wording. As far as I know right now, the thread registry doesn't actually alter threads?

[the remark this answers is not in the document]
17zhangw: Just a comment that if/when the thread registry can add/take away threads, to revisit this.

src/include/task/task.h
```diff
@@ -0,0 +1,178 @@
...
+  void Execute(common::ManagedPointer<util::QueryExecUtil> query_exec_util,
+               common::ManagedPointer<task::TaskManager> task_manager) override;
+  TaskType GetTaskType() override { return TaskType::DML_TASK; }
+
+  ~TaskDML() override = default;
+
+ private:
+  catalog::db_oid_t db_oid_;
+  std::string query_text_;
+  std::unique_ptr<optimizer::AbstractCostModel> cost_model_;
```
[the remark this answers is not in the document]
17zhangw: I think the issue is that the `Optimizer` as it stands expects the caller to specify the cost model that the optimizer should use. In a similar vein, the constructor of the `TaskDML` needs to specify the cost model that it should be using when optimizing a query.

src/include/common/future.h
```diff
@@ -0,0 +1,104 @@
...
+#include <atomic>
+#include <condition_variable>  // NOLINT
+#include <string>
+#include <thread>  // NOLINT
+#include <utility>
+
+namespace noisepage::common {
+
+/**
+ * Future is a wrapper over a condition_variable, and the condition.
```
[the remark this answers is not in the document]
17zhangw: Inserted a description. Principally it comes down to:
- No dependency on a std::promise<> and lifetime implications of that
- Wait() coordinates a result and success indicator, with access to an error message
- Don't depend on `std::exception_ptr` for the async task to indicate error

lmwnshn (approved):
LGTM if pass CI, with notes:

1. ReservoirSampling -- your call on whether that's ok, I don't have the context to know
2. For my own TaskManager purposes, need to make sure I always skip the plan caching, since there's no invalidation

malin1993ml (commented):
I think all my previous comments have been addressed as well! Besides the pilot issue I mentioned on Slack, the only thing I think we should add are some specific test cases on the `TaskManager` (include both DDL and DML). Maybe submitting tasks concurrently with multiple workers, and see whether things break. This way the tests for `TaskManager` are not entirely dependent on the query logging, in case we need to change that at some point.

src/self_driving/forecasting/workload_forecast.cpp
```diff
@@ -14,20 +14,50 @@
...
+  }
+  num_forecast_segment_ = forecast_segments_.size();
+}
+
+WorkloadForecast::WorkloadForecast(const WorkloadForecastPrediction &inference, WorkloadMetadata &&metadata) {
+  workload_metadata_ = std::move(metadata);
+  InitFromInference(inference);
+}
+
+WorkloadForecast::WorkloadForecast(const WorkloadForecastPrediction &inference) {
```
malin1993ml, line 47: I figured out why we're executing much more queries in this mode. Essentially we forgot to initialize `num_sample_` in this constructor, and we're loading all the parameters from `query_trace.csv`. We should hook this up with the `forecast_sample_limit` knob.

src/self_driving/planning/pilot.cpp
```diff
@@ -32,27 +50,338 @@ Pilot::Pilot(std::string model_save_path, std::string forecast_model_save_path,
...
   while (!model_server_manager_->ModelServerStarted()) {
   }
 }
 
+std::pair<uint64_t, uint64_t> Pilot::ComputeTimestampDataRange(uint64_t now) {
+  // Evaluation length is sequence length + 2 horizons
+  uint64_t eval_length = sequence_length_ + 2 * horizon_length_;
+
+  // Pull a range of the order of 5x (assuming classic 80% train/20% test split)
+  eval_length *= 5;
```
malin1993ml, line 68: This should only happen during training. During inference, we should just use `sequence_length_ + 2 * horizon_length_` amount of data.

src/self_driving/forecasting/workload_forecast.cpp
17zhangw: Ah. Thanks for the catch!

PR #1503, opened 2021-03-01
Interference Model Integration and Name Cleaning for Self-Driving Components

malin1993ml:
The major changes in this PR:
1. Add a new `InterferenceModel` class in `ModelServer`. Add a new `AbstractModel` class as the parent class for all the models that the `ModelServer` holds.
2. Extend the `ModelServerManager` to support the training and inference for the interference model.
3. Integrate the interference model training and inference into the `ModelServerTest`.
4. Making the names for self-driving components consistent across the codebase (especially across Python and C++ side) and consistent with the modeling paper. For example, "**mini**-trainer/model" is replaced with "**OU**-trainer/model". "**Global** model" is replaced with "**interference** model".
5. Also organize the module structure in `script/self_driving`. Add `__init__.py` files to all the packages. Under the new structure, we need to execute all the trainers with "python3 -m" under the `script/self_driving` directory (or above). For example, use `python3 -m modeling.ou_model_trainer` in `script/self_driving` to train the OU models. This is consistent with the `ModelServer` that loads all the trainers as modules. Also, remove the shebangs for all the trainers.


src/include/self_driving/model_server/model_server_manager.h
```diff
@@ -39,7 +39,7 @@ class ModelType {
   /**
    * Enum describing the specific type of model (i.e., forecast, minirunner, interference)
```
17zhangw, line 40: nit: should probably also change 'minirunner' -> 'operating unit'?

17zhangw (approved):
LGTM!

PR #1504, opened 2021-03-03
Add Statistics Cache

### LINKED ISSUE #657: Selectivity Calculations with regards to ParameterOffsets
As Lin mentioned in the comments on PR#600, the `StatsCalculator` uses the `ParameterOffset` value directly which makes the selectivity meaningless. 

Some high level discussion is needed to develop the necessary heuristics for calculating selectivities when dealing with `ParameterOffsets`. This issue is aimed at tracking/keeping a reminder of this issue within the Optimizer's Stats component.

### LINKED ISSUE #1338: Support for VARCHAR, DATE, and TIMESTAMP in selectivity calculation
# Feature Request
## Summary
We need to support `VARCHAR`, `DATE` and `TIMESTAMP` in selectivity calculation. Although the `NewColumnStats` class is templated, storing the underlying value of `DATE`, `TIMESTAMP` and `VARCHAR` in the histogram class causes issues. This is because the current histogram implementation requires that the underlying value for the above mentioned types support operators like '-' (subtract two dates), '*' (multiply a date by a number) and '/' (divide a date by a number). Any solution for a histogram class that supports all types can be easily plugged into the system as long as the APIs remain the same.

## Solution
To support all types:
- Search for an equi-depth histogram implementation that does not require such operators.

To support `DATE` and `TIMESTAMP`:
- The underlying values can be represented as `double` which is a supported type in the histogram implementation. This would require special treatment of `DATE` and `TIMESTAMP` in the templated `NewColumnStats` and `SelectivityUtil` classes.

### Alternatives
In the case that one histogram implementation cannot support all datatypes, have an abstract class for Histogram and have different implementations depending on the datatype of the column.

jkosh44:
## Description
This PR adds the ability for the optimizer to retrieve and cache column statistics. The first time the optimizer requests statistics for a column, statistics for the entire table are retrieved from pg_statistic and cached in StatsStorage. Subsequent requests for the same column, or any other column from that table, will be served from StatsStorage's cache. 

All ANALYZE queries now have a commit callback that will mark all analyzed columns as stale in StatsStorage. When StatsStorage is retrieving a column from it's cache it first checks if the column is stale. If it's stale then it will refetch the column from pg_statistic.

Closes #1338 
Closes #657 

## Remaining tasks

- [x] Add unit tests
    - ~~StatsStorage~~
    - ~~SelectivityUtil (Specifically Date, Timestamp, and StringVal columns. Also tables with 0 rows)~~
    - ~~StatsCalculator~~
- [x] Run performance tests
- [x] Compare TPC-H Join orderings of NoisePage and HyPer (https://hyper-db.de/interface.html#)
  - https://github.com/electrum/tpch-dbgen
- [x] Open issue about create column statistics for intermediate operators

## Performance
I ran TPCC 6 times, 3 with running ANALYZE and 3 without running ANALYZE. I used the configuration found below and got the results found below. The performance isn't that different which isn't surprising because I don't think there are many joins in TPCC.

Benchmark:     TPCC {com.oltpbenchmark.benchmarks.tpcc.TPCCBenchmark}
Configuration: config/noisepage_tpcc_config.xml
Type:          NOISEPAGE
Driver:        org.postgresql.Driver
URL:           jdbc:postgresql://localhost:15721/noisepage?preferQueryMode=extended
Isolation:     TRANSACTION_SERIALIZABLE
Scale Factor:  8.0

No ANALYZE
    1. Rate limited reqs/s: Results(nanoSeconds=60999992326, measuredRequests=50270) = 824.0984643300264 requests/sec
    2. Rate limited reqs/s: Results(nanoSeconds=60999876765, measuredRequests=50450) = 827.0508511739614 requests/sec
    3. Rate limited reqs/s: Results(nanoSeconds=61000020467, measuredRequests=51505) = 844.3439789969144 requests/sec

ANALYZE
    1. Rate limited reqs/s: Results(nanoSeconds=60000294750, measuredRequests=48560) = 809.3293575028646 requests/sec
    2. Rate limited reqs/s: Results(nanoSeconds=60999685330, measuredRequests=50443) = 826.9386920130855 requests/sec
    3. Rate limited reqs/s: Results(nanoSeconds=60999583704, measuredRequests=50890) = 834.26798856437 requests/sec

## TPC-H Join Order

Supported Queries (This list is just from personal testing. I'm not including any queries that would work with slight modifications)
- 3
- 5
- 10

#### 3
Screenshot from 2021-04-04 15-43-57

#### 5
Screenshot from 2021-04-04 15-56-06

#### 10
Screenshot from 2021-04-04 16-13-47

#### 3
Screenshot from 2021-04-04 15-52-52

#### 5
Bound Insert Into Select

#### 10
Bound Insert Into Select(1)

#### 3
Bound Insert Into Select

#### 5
Bound Insert Into Select(1)

#### 10
Bound Insert Into Select(2)

So it seems like `ANALYZE` doesn't have any impact on the join ordering right of those three queries right now.

## Further work
- Create eviction policy for StatsStorage


jkosh44:
I need to add some more unit tests, but it's passing all existing tests and the sql traces, so I think we can start reviewing.

Please let me know if I'm using `unique_ptr`s unnecessarily anywhere.

src/catalog/postgres/pg_statistic_impl.cpp
```diff
@@ -129,4 +130,137 @@ bool PgStatisticImpl::DeleteColumnStatistics(const common::ManagedPointer<transa
...
+std::unique_ptr<optimizer::ColumnStatsBase> PgStatisticImpl::CreateColumnStats(
+    common::ManagedPointer<storage::ProjectedRow> all_cols_pr, table_oid_t table_oid, col_oid_t col_oid,
+    type::TypeId type) {
+  auto num_rows = *PgStatistic::STA_NUMROWS.Get(all_cols_pr, pg_statistic_all_cols_prm_);
+  auto non_null_rows = *PgStatistic::STA_NONNULLROWS.Get(all_cols_pr, pg_statistic_all_cols_prm_);
+  auto distinct_values = *PgStatistic::STA_DISTINCTROWS.Get(all_cols_pr, pg_statistic_all_cols_prm_);
+  const auto *top_k_str = PgStatistic::STA_TOPK.Get(all_cols_pr, pg_statistic_all_cols_prm_);
+  const auto *histogram_str = PgStatistic::STA_HISTOGRAM.Get(all_cols_pr, pg_statistic_all_cols_prm_);
+
+  switch (type) {
```
jkosh44, line 214: Is there a better way to do this? I need to convert a `type::TypeId` to a `execution::sql::val`, but couldn't find a cleaner way other than hardcoding.

src/include/optimizer/statistics/column_stats.h
```diff
@@ -1,105 +1,261 @@
...
+#include "execution/sql/value.h"
+#include "optimizer/statistics/histogram.h"
+#include "optimizer/statistics/top_k_elements.h"
 
 namespace noisepage::optimizer {
+
+/**
+ * Base class for Column Statistics
+ */
+class ColumnStatsBase {
```
jkosh44, line 19: I'm still not great with templates, so if you have a better solution than using an abstract base class let me know.

src/optimizer/statistics/selectivity_util.cpp
```diff
@@ -71,6 +125,10 @@ double SelectivityUtil::Equal(common::ManagedPointer<NewColumnStats<T>> column_s
 
   size_t numrows = column_stats->GetNumRows();
 
+  if (numrows == 0) {
+    return 0;
```
jkosh44, line 129: I'm not actually sure if this should be 0 or 1. What's the selectivity of a column with no rows?

src/optimizer/statistics/stats_storage.cpp
```diff
@@ -6,39 +6,93 @@
...
 
-  if (table_it != table_stats_storage_.end()) {
-    return common::ManagedPointer<TableStats>(table_it->second);
+  if (table_it == table_stats_storage_.end()) {
+    InsertTableStats(database_id, table_id, accessor);
+    table_it = table_stats_storage_.find(stats_storage_key);
   }
-  return common::ManagedPointer<TableStats>(nullptr);
+
+  auto &stats_storage_value = table_it->second;
```
jkosh44, line 21: Does this make  copy?

jkosh44:
All the sql traces pass but OLTP fails, so there's probably some concurrency issue.

jkosh44:
So the issue is not actually a concurrency issue but is related to query params. Specifically, the error can be found here:
https://github.com/cmu-db/noisepage/blob/8896d8ec203ecf8019ecfa0dd5af449d64fdc144/src/optimizer/statistics/stats_calculator.cpp#L307-L309
There are two problems with this
1. `type::TypeId::PARAMETER_OFFSET` is not a valid type for a `ConstantValueExpression`.
2. We are trying to create a constant value expression from the parameter offset. This doesn't make much sense. Instead, we want the underlying value of the parameter.

As far as I'm aware we don't pass the parameter values into the optimizer, so this may take a bit of work. If anyone is reading this and has suggestions please let me know.

src/optimizer/statistics/selectivity_util.cpp
```diff
@@ -4,8 +4,62 @@
 #include "parser/expression_defs.h"
 
 namespace noisepage::optimizer {
+
+double SelectivityUtil::ComputeSelectivity(const TableStats &table_stats, const ValueCondition &condition) {
+  if (table_stats.GetNumRows() == 0) {
```
jkosh44, line 9: I'm not actually sure if this should be 0 or 1. What's the selectivity of a column with no rows?

test/optimizer/stats_calculator_test.cpp
```diff
@@ -0,0 +1,557 @@
...
+}
+
+// NOLINTNEXTLINE
+TEST_F(StatsCalculatorTests, TestNotPredicate) {
+  RunQuery("INSERT INTO " + table_name_1_ + " VALUES(1), (NULL), (3);");
+  RunQuery("ANALYZE " + table_name_1_ + ";");
+  txn_manager_->Commit(test_txn_, transaction::TransactionUtil::EmptyCallback, nullptr);
+  test_txn_ = txn_manager_->BeginTransaction();
+
+  parser::ColumnValueExpression col_a(table_name_1_, table_1_col_1_name_, test_db_oid_, table_oid_1_, table_1_col_oid_,
```
malin1993ml, line 105: Can you add a comment on what is the predicate that you're constructing? The same applies to all the other tests with predicates and joins. Otherwise, people really have to look through all the construction code to figure out what's tested.

test/optimizer/stats_calculator_test.cpp
```diff
@@ -0,0 +1,557 @@
...
+#include "parser/expression/column_value_expression.h"
+#include "parser/expression/comparison_expression.h"
+#include "parser/expression/conjunction_expression.h"
+#include "parser/expression/constant_value_expression.h"
+#include "parser/expression/operator_expression.h"
+#include "test_util/end_to_end_test.h"
+#include "test_util/test_harness.h"
+
+namespace noisepage::optimizer {
+class StatsCalculatorTests : public test::EndToEndTest {
```
malin1993ml, line 15: Can you add a test with multiple parameters as well?

malin1993ml (changes requested):
Generally looks pretty good! Tests look pretty comprehensive. Just have some small comments in addition to what we have discussed during the team meeting.

script/testing/junit/sql/analyze.sql
```diff
@@ -0,0 +1,47 @@
...
+
+SELECT a FROM foo WHERE a < 666;
+SELECT a FROM foo WHERE a <= 42;
+SELECT a FROM foo WHERE a > 999;
+SELECT a FROM foo WHERE a >= 24;
+SELECT a FROM foo WHERE a = 42;
+SELECT a FROM foo WHERE NOT a = 666;
+SELECT a FROM foo WHERE NOT a <> 666;
+SELECT a FROM foo WHERE b IS NULL;
+SELECT a FROM foo WHERE b IS NOT NULL;
```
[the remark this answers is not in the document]
jkosh44: My goal here was to try and target different stats calculator selectivity operations with empty statistics and with full statistics and make sure it didn't blow up. That's why I have a bunch of different predicates. I included it here because it's somewhat related to ANALYZE. Do you think there's a better place I can move that type of test?

src/catalog/postgres/pg_statistic_impl.cpp
```diff
@@ -129,4 +130,137 @@ bool PgStatisticImpl::DeleteColumnStatistics(const common::ManagedPointer<transa
...
+  std::vector<storage::TupleSlot> index_results;
+  {
+    auto *pr = oid_pri.InitializeRow(buffer);
+    auto *pr_hi = oid_pri.InitializeRow(key_buffer);
+
+    // Low key (class, min col_oid_t)
+    pr->Set<table_oid_t, false>(oid_prm.at(indexkeycol_oid_t(1)), table_oid, false);
+    pr->Set<col_oid_t, false>(oid_prm.at(indexkeycol_oid_t(2)), col_oid_t(std::numeric_limits<uint32_t>::min()), false);
+
+    // High key (class + 1, max col_oid_t)
```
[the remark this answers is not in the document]
jkosh44: Yes, good catch. Fixed.

src/catalog/postgres/pg_statistic_impl.cpp
```diff
@@ -129,4 +130,137 @@ bool PgStatisticImpl::DeleteColumnStatistics(const common::ManagedPointer<transa
...
+std::unique_ptr<optimizer::ColumnStatsBase> PgStatisticImpl::CreateColumnStats(
+    table_oid_t table_oid, col_oid_t col_oid, size_t num_rows, size_t non_null_rows, size_t distinct_values,
+    const storage::VarlenEntry *top_k_str, const storage::VarlenEntry *histogram_str, type::TypeId type) {
+  using CppType = decltype(T::val_);
+  auto top_k = top_k_str != nullptr
+                   ? optimizer::TopKElements<CppType>::Deserialize(top_k_str->Content(), top_k_str->Size())
+                   : optimizer::TopKElements<CppType>();
+  auto histogram = histogram_str != nullptr
+                       ? optimizer::Histogram<CppType>::Deserialize(histogram_str->Content(), histogram_str->Size())
+                       : optimizer::Histogram<CppType>();
```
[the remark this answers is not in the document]
jkosh44: When we first create a table all rows in pg_statistic for that table will have `NULL` in the histogram and topk columns. If you `ANALYZE` an empty table I'm actually not sure if the histogram and topk columns will be `NULL` or just a serialized empty data structure. Either way we get an empty top_k and histogram when the table is empty. The selectivity/cardinality calculations will handle this case because they just immediately return 0 on an empty table and never look at either data structure.

I'll add a comment in the code briefly explaining this.

src/include/catalog/catalog_accessor.h
```diff
@@ -355,6 +358,21 @@ class EXPORT CatalogAccessor {
...
+   * @return column statistics
+   */
+  std::unique_ptr<optimizer::ColumnStatsBase> GetColumnStatistics(table_oid_t table_oid, col_oid_t col_oid);
+
+  /**
+   * Gets the statistics of a table from pg_statistic
+   * @param table_oid table oid of table
+   * @return table statistics
+   */
+  optimizer::TableStats GetTableStatistics(table_oid_t table_oid);
```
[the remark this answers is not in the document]
jkosh44: This is due to my lack of experience with C++ templates and polymorphism. I didn't want to unnecessarily allocate extra heap memory so I made this function return a raw object. However, it's my understanding that since `ColumnStatsBase` is an abstract class, I need to return a pointer to it and I can't return the class directly. If you have any suggestions on how to improve this please let me know.

src/include/catalog/database_catalog.h
```diff
@@ -176,6 +179,14 @@ class DatabaseCatalog {
   common::ManagedPointer<execution::functions::FunctionContext> GetFunctionContext(
       common::ManagedPointer<transaction::TransactionContext> txn, proc_oid_t proc_oid);
 
+  /** @brief Get the statistics for the specified column. @see PgStatisticImpl::GetColumnStatistics */
+  std::unique_ptr<optimizer::ColumnStatsBase> GetColumnStatistics(
+      common::ManagedPointer<transaction::TransactionContext> txn, table_oid_t table_oid, col_oid_t col_oid);
+
+  /** @brief Get the statistics for the specified table. @see PgStatisticImpl::GetTableStatistics */
+  optimizer::TableStats GetTableStatistics(common::ManagedPointer<transaction::TransactionContext> txn,
+                                           table_oid_t table_oid);
```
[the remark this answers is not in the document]
jkosh44: See comment on catalog_accessor

src/include/execution/sql/aggregators.h
```diff
@@ -373,14 +373,11 @@ class TopKAggregate {
...
 
-  static constexpr size_t K = 16;
-  static constexpr uint64_t WIDTH = 64;
-
  public:
   /**
    * Constructor.
    */
-  TopKAggregate() : top_k_(K, WIDTH) {}
+  TopKAggregate() : top_k_() {}
```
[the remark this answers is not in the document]
jkosh44: We do need to know K and width, I just pushed the default values down into the `TopKElements` class directly instead of this class because I thought it was more relevant to the `TopKElements` class than the aggregator wrapping it.

src/include/optimizer/statistics/histogram.h
```diff
@@ -25,17 +26,56 @@ namespace noisepage::optimizer {
  *    A Streaming Parallel Decision Tree Algorithm
  *    http://www.jmlr.org/papers/volume11/ben-haim10a/ben-haim10a.pdf
  * Specifically Algorithm 1, 3, and 4.
+ *
+ * Histogram supports any key type that can be easily converted into a double. However in order to get meaningful
+ * results from the histogram the ordering of the keys should be preserved in this conversion. So for example if
+ * key1 > key2 then ConvertToPoint(key1) > ConvertToPoint(key2) where ConvertToPoint(k) converts k to a double.
+ *
+ * Histogram currently supports strings by hashing them, but since hashing doesn't preserve lexicographical ordering
+ * it's pretty much useless
```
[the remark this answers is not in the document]
jkosh44: As discussed, I made some adjustments to the psuedo code and implemented that. It actually works fairly well after some initial testing.

src/include/optimizer/statistics/selectivity_util.h
```diff
@@ -4,7 +4,8 @@
 #include <limits>
 
 #include "common/managed_pointer.h"
-#include "optimizer/statistics/new_column_stats.h"
+#include "optimizer/statistics/column_stats.h"
+#include "optimizer/statistics/table_stats.h"
```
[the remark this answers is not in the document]
jkosh44: Done

src/optimizer/statistics/stats_storage.cpp
```diff
@@ -6,39 +6,92 @@
...
-  table_stats_storage_.emplace(stats_storage_key, std::move(table_stats_ptr));
-  return true;
 }
 
-bool StatsStorage::DeleteTableStats(catalog::db_oid_t database_id, catalog::table_oid_t table_id) {
-  StatsStorageKey stats_storage_key = std::make_pair(database_id, table_id);
-  auto table_it = table_stats_storage_.find(stats_storage_key);
+void StatsStorage::InsertTableStats(catalog::db_oid_t database_id, catalog::table_oid_t table_id,
+                                    catalog::CatalogAccessor *accessor) {
+  std::unique_lock<std::mutex> latch(insert_latch_);
```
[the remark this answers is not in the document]
jkosh44: After doing a bit of research, you are right. I naively thought that since we're only inserting, it would be fine to have multiple concurrent readers and one writer, but this is not correct. This will have to be a `SharedLatch`.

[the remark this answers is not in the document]
jkosh44: @malin1993ml Can you take a look at the change I just did and see if it looks right? If so, just resolve this convo.

src/optimizer/statistics/table_stats.cpp
```diff
@@ -27,48 +18,44 @@ bool TableStats::AddColumnStats(std::unique_ptr<ColumnStats> col_stats) {
...
+std::vector<common::ManagedPointer<ColumnStatsBase>> TableStats::GetColumnStats() const {
+  std::vector<common::ManagedPointer<ColumnStatsBase>> column_stats;
+  for (const auto &[_, column_stat] : column_stats_) {
+    column_stats.emplace_back(common::ManagedPointer(column_stat));
   }
-  return false;
+  return column_stats;
+}
+
+std::unique_ptr<ColumnStatsBase> TableStats::RemoveColumnStats(catalog::col_oid_t column_id) {
```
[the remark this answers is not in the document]
jkosh44: Right now with all the copying, it is needed, however after the copying is removed we may no longer need to return the `unique_ptr`.

src/include/catalog/catalog_accessor.h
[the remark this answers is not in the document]
malin1993ml: I guess let's figure out whether we want to return a copy or a reference. If we need to return a copy, then I think this is fine.

script/testing/junit/sql/analyze.sql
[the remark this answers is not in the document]
malin1993ml: Ohoh I see. That makes sense! Maybe add a bit comment to explain this?

src/optimizer/statistics/stats_storage.cpp
```diff
@@ -6,39 +6,99 @@
...
 
 namespace noisepage::optimizer {
-common::ManagedPointer<TableStats> StatsStorage::GetTableStats(catalog::db_oid_t database_id,
-                                                               catalog::table_oid_t table_id) {
-  StatsStorageKey stats_storage_key = std::make_pair(database_id, table_id);
+
+StatsStorageValue &StatsStorage::GetStatsStorageValue(const catalog::db_oid_t database_id,
+                                                      const catalog::table_oid_t table_id,
+                                                      catalog::CatalogAccessor *accessor) {
+  stats_latch_.LockShared();
```
malin1993ml, line 13: nit: maybe call this `stats_storage_latch_` to distinguish it from the latch for the individual table.

src/optimizer/statistics/stats_storage.cpp
```diff
@@ -6,39 +6,99 @@
...
+  StatsStorageKey stats_storage_key{database_id, table_id};
   auto table_it = table_stats_storage_.find(stats_storage_key);
 
-  if (table_it != table_stats_storage_.end()) {
-    return common::ManagedPointer<TableStats>(table_it->second);
+  if (table_it == table_stats_storage_.end()) {
+    stats_latch_.UnlockShared();
+    InsertTableStats(database_id, table_id, accessor);
+    table_it = table_stats_storage_.find(stats_storage_key);
+    stats_latch_.LockShared();
```
malin1993ml, line 22: I think you need to move this one line up?

src/optimizer/statistics/stats_storage.cpp
```diff
@@ -6,39 +6,99 @@
...
+  common::SharedLatch::ScopedSharedLatch table_latch{&stats_storage_value.shared_latch_};
+  NOISEPAGE_ASSERT(stats_storage_value.table_stats_.HasColumnStats(column_oid), "Should have stats for all columns");
+  return stats_storage_value.table_stats_.GetColumnStats(column_oid)->Copy();
+}
 
-  if (table_it != table_stats_storage_.end()) {
-    OPTIMIZER_LOG_TRACE("There already exists a TableStats object with the given oids.");
-    return false;
+void StatsStorage::MarkStatsStale(catalog::db_oid_t database_id, catalog::table_oid_t table_id,
+                                  const std::vector<catalog::col_oid_t> &col_ids) {
```
malin1993ml, line 62: I think you need `stats_latch_` in this function as well, since it reads from `table_stats_storage_`.

test/optimizer/stats_storage_test.cpp
```diff
@@ -1,58 +1,116 @@
...
-
+#include "catalog/database_catalog.h"
 #include "gtest/gtest.h"
+#include "storage/sql_table.h"
+#include "test_util/end_to_end_test.h"
 #include "test_util/test_harness.h"
 
 namespace noisepage::optimizer {
-class StatsStorageTests : public TerrierTest {
+class StatsStorageTests : public test::EndToEndTest {
```
malin1993ml, line 10: Actually, is it possible to add a test that modifies the `StatsStorage` concurrently? Since there're two layers of latches and we've been back-and-forth on this a little bit, I feel like it would be good to have a test case for it.

jkosh44:
After some discussion, we have decided to remove the stored `ColumnStats` from the optimizer and open an issue to find a better way to create and store intermediate statistics.

src/optimizer/statistics/stats_storage.cpp
jkosh44: Done

src/optimizer/statistics/stats_storage.cpp
jkosh44: Done

src/optimizer/statistics/stats_storage.cpp
jkosh44: Done

script/testing/junit/sql/analyze.sql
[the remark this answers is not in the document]
jkosh44: Done

src/include/catalog/catalog_accessor.h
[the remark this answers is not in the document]
jkosh44: Even if we don't return a copy, I still don't know how to get around using a pointer for `ColumnStatsBase` since it's an abstract class.

test/optimizer/selectivity_util_test.cpp
```diff
@@ -9,111 +9,182 @@
 namespace noisepage::optimizer {
 class SelectivityUtilTests : public TerrierTest {
```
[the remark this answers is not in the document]
jkosh44: Done

test/optimizer/stats_calculator_test.cpp
jkosh44: Done

test/optimizer/stats_calculator_test.cpp
jkosh44: Done

src/traffic_cop/traffic_cop_util.cpp
```diff
@@ -55,6 +57,10 @@ std::unique_ptr<optimizer::OptimizeResult> TrafficCopUtil::Optimize(
     // immutability of all of these Optimizer inputs to reduce copies.
 
     CollectSelectProperties(sel_stmt, &property_set);
+  } else if (type == parser::StatementType::ANALYZE) {
```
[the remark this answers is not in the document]
jkosh44: Moved

test/optimizer/stats_storage_test.cpp
jkosh44: Added

src/catalog/postgres/pg_statistic_impl.cpp
```diff
@@ -129,4 +130,137 @@ bool PgStatisticImpl::DeleteColumnStatistics(const common::ManagedPointer<transa
...
+       */
+      return CreateColumnStats<execution::sql::StringVal>(table_oid, col_oid, num_rows, non_null_rows, distinct_values,
+                                                          top_k_str, nullptr, type);
+    default:
+      UNREACHABLE("Invalid column type");
+  }
+}
+
+template <typename T>
+std::unique_ptr<optimizer::ColumnStatsBase> PgStatisticImpl::CreateColumnStats(
```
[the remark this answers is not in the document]
jkosh44: @malin1993ml can you help me come up with a new name? I generally tend to overload methods when they perform roughly the same task but have different signatures, otherwise, I rename them. I'm having trouble coming up with good names for these methods that distinguish what they're doing.

src/include/optimizer/statistics/stats_storage.h
```diff
@@ -61,52 +84,92 @@ namespace noisepage::optimizer {
...
  public:
   /**
-   * Using given database and table ids,
-   * select a pointer to the TableStats objects in the table stats storage map.
+   * Returns a copy of the TableStats object for a specific table
+   *
+   * Currently all consumers of this method require their own copy of the statistics objects. Instead of them having to
+   * create their own copies, we just create a copy for them. This helps simplify thread safety and reduce contention.
+   * If a new consumer doesn't need a copy we may want to consider changing this to return a ManagedPointer to prevent
+   * unnecessary copying.
```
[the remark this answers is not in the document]
jkosh44: Removed

src/include/optimizer/statistics/table_stats.h
```diff
@@ -81,27 +71,47 @@ class TableStats {
...
+   */
+  bool HasStaleValues() const {
+    return std::any_of(column_stats_.begin(), column_stats_.end(), [](const auto &it) { return it.second->IsStale(); });
+  }
+
+  /**
+   * Make a deep copy of TableStats
+   * @return a copy of the underlying object
+   */
+  std::unique_ptr<TableStats> Copy() {
```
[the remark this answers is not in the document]
jkosh44: Removed

src/optimizer/statistics/stats_calculator.cpp
```diff
@@ -127,18 +110,18 @@ void StatsCalculator::Visit(const LogicalInnerJoin *op) {
...
     NOISEPAGE_ASSERT(col->GetExpressionType() == parser::ExpressionType::COLUMN_VALUE, "CVE expected");
     auto tv_expr = col.CastManagedPointerTo<parser::ColumnValueExpression>();
     auto col_name = tv_expr->GetFullName();
-    std::unique_ptr<ColumnStats> column_stats;
+    std::unique_ptr<ColumnStatsBase> column_stats;
 
     // Make a copy from the child stats
     if (left_child_group->HasColumnStats(col_name)) {
-      column_stats = std::make_unique<ColumnStats>(*left_child_group->GetStats(col_name));
+      column_stats = left_child_group->GetStats(col_name)->Copy();
```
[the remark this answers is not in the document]
jkosh44: Removed

src/optimizer/statistics/table_stats.cpp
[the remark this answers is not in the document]
jkosh44: Removed

src/include/optimizer/statistics/selectivity_util.h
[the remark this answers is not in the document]
jkosh44: Actually done

jkosh44:
@malin1993ml I've addressed everything we've discussed and all the comments except for the ones about the overloaded function (I'm still trying to think of good names). I think this PR is ready for another review. 

As discussed in the team meeting, removing Column Stats copying ended up being a pretty big change and probably requires a thorough review.

test/optimizer/stats_calculator_test.cpp
```diff
@@ -0,0 +1,639 @@
...
+}
+
+// NOLINTNEXTLINE
+TEST_F(StatsCalculatorTests, TestNotPredicate) {
+  RunQuery("INSERT INTO " + table_name_1_ + " VALUES(1), (NULL), (3);");
+  RunQuery("ANALYZE " + table_name_1_ + ";");
+  txn_manager_->Commit(test_txn_, transaction::TransactionUtil::EmptyCallback, nullptr);
+  test_txn_ = txn_manager_->BeginTransaction();
+
+  // Constructing Logical Get with NOT EQUALS predicate
```
malin1993ml, line 109: Can you make the comment to be specific? For example, in this case it can be:
`Constructing Logical Get with NOT EQUALS predicate "NOT colA=1" from "empty_nullable_table"`
The same applies to other tests in this file.

src/include/optimizer/statistics/stats_storage.h
```diff
@@ -61,52 +100,72 @@ namespace noisepage::optimizer {
...
-   * select a pointer to the TableStats objects in the table stats storage map.
+   * Returns a reference of a TableStats object for a specific table, and an acquired read lock on the entire Stats
+   * Storage
+   *
    * @param database_id - oid of database
    * @param table_id - oid of table
-   * @return pointer to a TableStats object
+   * @param accessor - catalog accessor
+   * @return reference to a TableStats object, latch for that TableStats object, and an acquired shared latch on
+   * StatsStorage
```
malin1993ml, line 117: Maybe document that the consumer of `empty_nullable_table` needs to take a shared latch when reading from `StatsStorageReference` and explain how to access that latch.

src/optimizer/statistics/stats_calculator.cpp
```diff
@@ -37,231 +35,112 @@ void StatsCalculator::Visit(const LogicalGet *op) {
...
-
-    // Use predicates to estimate cardinality. If we were unable to find any column stats from the catalog, default to 0
-    if (table_stats->GetColumnCount() == 0) {
-      root_group->SetNumRows(0);
-    } else {
-      auto est = EstimateCardinalityForFilter(table_stats->GetNumRows(), predicate_stats, op->GetPredicates());
-      root_group->SetNumRows(static_cast<int>(est));
-    }
-  }
+    const auto &stats_cache_reference = context_->GetStatsStorage()->GetTableStats(
```
malin1993ml, line 42: I don't think the `&` is safe here. The return value of `GetTableStats` is an `rvalue` (value on the stack) and just taking a `&` here would not store it anywhere. The value can be corrupted. I think we should use `const auto` here and in similar places (e.g., in `stats_storage_test`).

src/optimizer/statistics/stats_storage.cpp
```diff
@@ -6,39 +6,86 @@
...
+  UpdateStaleColumns(table_id, &stats_storage_value, accessor);
+
+  return StatsStorageReference(&stats_storage_value, std::move(shared_stats_storage_latch));
+}
+
+void StatsStorage::MarkStatsStale(catalog::db_oid_t database_id, catalog::table_oid_t table_id,
+                                  const std::vector<catalog::col_oid_t> &col_ids) {
+  /*
+   * It's ok to mark something stale while someone is reading it, the worst that happens is they end up using slightly
+   * stale statistics without realizing it.
```
malin1993ml, line 38: But what if someone is updating it through `UpdateStaleColumns`? Wouldn't it be possible that the value is corrupted? I think you still need to take the shared latch on the `stats_storage_value`.

malin1993ml (changes requested):
I think most of the comments are addressed. Thanks, @jkosh44!

I still have a few additional questions on the latches. Also, I need to test retrieving the cardinality information.

src/include/optimizer/statistics/stats_storage.h
jkosh44: @malin1993ml What do you think of this idea, I change `StatsStorageReference` to look something like this
```C++
struct StatsStorageReference {
 /** Table Statistics */
  const TableStats &table_stats_;
  /** Acquired Shared Latch on Table Stats */
  common::SharedLatchGuard table_stats_shared_latch_;
  /** Acquired Shared Latch on Stats Storage */
  common::SharedLatchGuard stats_storage_shared_latch_;
};
```
That when we return a table stats object it not only comes with an acquired latch on stats storage but it also comes with an acquired latch on the table stats itself.

I think the pro is that the consumer of this method doesn't need to do any additional locking which is probably less error-prone. The con is that if the consumer isn't using the table stats right away then they are unnecessarily holding a lock on it. Right now the consumer always consumes the table stats right away. Also it would probably be bad practice to get table stats from stats storage until right before you need it because you need a lock on stats storage anyway.

Also if you have a better name than `StatsStorageReference` please let me know.

malin1993ml: I actually like this idea a bit better than the original one. Please just document that in `GetTableStats` so that other people would notice the implication. Maybe call this `LockedTableStatsReference`.

test/optimizer/stats_calculator_test.cpp
jkosh44: Should be fixed now

src/include/optimizer/statistics/stats_storage.h
jkosh44: I'm going to push this change for now and in case there's an issue with it or you don't like it, then I'll revert it.

src/optimizer/statistics/stats_calculator.cpp
jkosh44: Fixed

src/optimizer/statistics/stats_storage.cpp
jkosh44: Thanks for catching this (and being patient with me not fixing this correctly multiple times). I meant to take a read lock on the table stats and this comment was explaining why I take a read lock instead of a write lock. It should be fixed now.

src/include/optimizer/statistics/stats_storage.h
malin1993ml: I think the changes look good! Thanks! I do think it would be a little bit better to rename `StatsStorageReference` to `LatchedTableStatsReference`, since that's what it really is. Also, it might be better to rename `StatsStorageKey` to `TableStatsKey` and `StatsStorageValue` to `TableStatsValue`. Again, those are what they really are.

malin1993ml:
I've looked at retrieving the cardinality information. I think the current way looks reasonable. I do need to retrieve the number of distinct value information as well, but I think that is out of this PR's scope. I think the PR should be good after renaming the last few things!

malin1993ml:
Oh lastly, I don't know whether you've run TPCH through OLTPBench before, but it would be nice to run TPCH and see whether we're picking the correct join order now (e.g., comparing the execution time with/without running `ANALYZE`).

jkosh44:
> Oh lastly, I don't know whether you've run TPCH through OLTPBench before, but it would be nice to run TPCH and see whether we're picking the correct join order now (e.g., comparing the execution time with/without running `ANALYZE`).

Not yet, but it's been on the TODO list for this PR, I just haven't had the bandwidth.

src/include/optimizer/statistics/stats_storage.h
jkosh44: Sorry, I did not see your last two messages until now. We must have been typing at the same time. I've updated everything with your suggestions.

malin1993ml (approved):
LGTM. Thanks for the great work, @jkosh44! Just curious whether this helps the TPCH join order...

PR #1505, opened 2021-03-05
Jenkins: Move microbenchmark building to pre-test and to any node. Add print_docker_info.sh.

lmwnshn:
# Heading

Jenkins: Move microbenchmark building to pre-test and on any node. Add print_docker_info.sh.

## Description

Since microbenchmarks are only built and are not run,
- We might as well build them earlier. This avoids people having their builds pass the test stage, only to fail some microbenchmark compilation error and have to go through CI again.
- We don't need a dedicated benchmark machine just to compile the code.

This PR also adds a script `build-support/print_docker_info.sh` to print image name, uptime, current time, to check if we're somehow executing on a different left-behind docker image. This hopes to provide data for weird `mkdir build` failures.

## Remaining tasks

Run through CI.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -45,6 +45,7 @@ pipeline {
                     }
                     steps {
                         sh 'echo $NODE_NAME'
+                        sh script: './build-support/print_docker_info.sh', label: 'Print image information.'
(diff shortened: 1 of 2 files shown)


apavlo (approved):
LGTM

PR #1507, opened 2021-03-08
Add guide to getting debug symbols on clang.

lmwnshn:
# Heading

Add guide to getting debug symbols on clang.

## Description

On GCC, you can view the values of a `std::string` in the debugger.
On clang, you can't. This is because clang assumes `libstdc++` will provide it.
You can invalidate that assumption by passing a compiler flag but that makes the binary bigger and linking slower.
So if you want to easily debug things like `std::string`s when compiling with clang, you should install debug symbol packages.


malin1993ml (approved):
LGTM!

PR #1511, opened 2021-03-09
Separate disk/replication queues for log serializer task

tpan496:
# Separate disk/replication queues for log serializer task

## Description
Adds two queues for log serializer task. Based on different retention policies, logs will be pushed into different queues and sent to separate destinations.

A separate queue is used for redirecting record buffers into separate locations based on retention policies. e.g. If the retention policy specifies only logging to disk, then the record buffer segment will only be added to the disk serialization queue and will not be added to the replication queue.

### Remaining Tasks

:pushpin: TODOs:


lmwnshn:
Appears to hang:
```
cmake-build-debug/bin [tianlei/replication-wan] » PYTHONPATH=../.. python3 -m script.testing.junit --build-type=debug --query-mode=extended
03-19-2021 12:22:46,714 [__main__.py:38] INFO : 
+++ ==================================================================================================== +++
+++ TEST JUNIT                                                                                           +++
03-19-2021 12:22:46,714 [db_server.py:78] INFO : Running: /home/kapi/CLionProjects/noisepage/cmake-build-debug/bin/noisepage -wal_file_path=wal.log
03-19-2021 12:22:46,715 [db_server.py:83] INFO : Ran: /home/kapi/CLionProjects/noisepage/cmake-build-debug/bin/noisepage -wal_file_path=wal.log [PID=10912]
03-19-2021 12:22:46,828 [db_server.py:91] INFO : DB process is verified as running in 0.11 sec.
03-19-2021 12:22:46,828 [common.py:48] INFO : Running: ant test-unit
03-19-2021 12:22:49,851 [test_server.py:178] INFO : Quiet and all tests passed (return code 0), skipping printing.
```

src/include/storage/write_ahead_log/log_serializer_task.h
```diff
@@ -115,22 +131,27 @@ class LogSerializerTask : public common::DedicatedThreadTask {
...
+  std::vector<std::pair<transaction::callback_fn, void *>> disk_commits_in_buffer_;
+  std::vector<std::pair<transaction::callback_fn, void *>> replication_commits_in_buffer_;
 
   // Used by the serializer thread to store buffers it has grabbed from the log manager
-  std::queue<RecordBufferSegment *> temp_flush_queue_;
+  std::queue<RecordBufferSegment *> temp_disk_flush_queue_;
+  std::queue<RecordBufferSegment *> temp_replication_flush_queue_;
 
   // We aggregate all transactions we serialize so we can bulk remove the from the timestamp manager
   // TODO(Gus): If we guarantee there is only one TSManager in the system, this can just be a vector. We could also pass
```
mbutrovich, line 145: Based on where @lmwnshn's design is going, can we make a decision about this? Would we ever want more than 1 TimestampManager in the system or was it a temporary decision for Gus' experiments?

src/include/storage/write_ahead_log/log_io.h
```diff
@@ -111,6 +111,11 @@ class BufferedLogWriter {
     return size;
   }
 
+  /**
+   * Empties the buffer.
+   */
+  void EmptyBuffer() { buffer_size_ = 0; }
```
mbutrovich, line 117: This should be called `ResetBuffer` to be more clear, and documentation should describe under what scenarios it should be used.

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -126,17 +153,26 @@ std::tuple<uint64_t, uint64_t, uint64_t> LogSerializerTask::Process() {
...
 }
 
 /**
  * Hand over the current buffer and commit callbacks for commit records in that buffer to the log consumer task
  */
-void LogSerializerTask::HandFilledBufferToWriter() {
+void LogSerializerTask::HandFilledBufferToWriter(SerializeDestination destination) {
   // Mark the buffer as ready for serialization, if it exists. It may not exist for read-only transactions.
 
   // TODO(WAN): Tianlei will be adding code that has different queues for different retention policies. When this
```
mbutrovich, line 178: Old TODO?

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -148,26 +184,44 @@ void LogSerializerTask::HandFilledBufferToWriter() {
   NOISEPAGE_ASSERT(primary_replication_manager_ != DISABLED ||
```
mbutrovich, line 184: Old assertion? See TODO

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -148,26 +184,44 @@ void LogSerializerTask::HandFilledBufferToWriter() {
...
+    // Mark that the task doesn't have a buffer in its possession to which it can write to
+    disk_commits_in_buffer_.clear();
+    disk_filled_buffer_ = nullptr;
+  } else if (destination == SerializeDestination::REPLICAS) {
+    if (replication_filled_buffer_ != nullptr) {
+      replication_filled_buffer_->PrepareForSerialization(retention_policy);
+
+      // Replicate the buffer if it exists.
+      // TODO(WAN): Note that this is effectively on the critical path to commit callbacks being invoked. Aka terrible.
+      if (primary_replication_manager_ != DISABLED) {
```
mbutrovich, line 204: These nested ifs are getting pretty spaghetti-code-like. It's hard to follow the logic. Can the replication manager even be disabled if you're trying to replicate? If not that should be an assertion. I feel like a lot of the permutations in these ifs aren't actually possible, but it's tough to follow the logic with no many nested ifs.

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -82,41 +84,66 @@ std::tuple<uint64_t, uint64_t, uint64_t> LogSerializerTask::Process() {
...
+
+        num_bytes += std::get<0>(num_bytes_records_and_txns);
+        num_records += std::get<1>(num_bytes_records_and_txns);
+        num_txns += std::get<2>(num_bytes_records_and_txns);
+      }
     }
 
     // Mark the last buffer that was written to as full
-    if (buffers_processed) HandFilledBufferToWriter();
+    if (disk_buffers_processed) HandFilledBufferToWriter(SerializeDestination::DISK);
```
jkosh44, line 134: I think there's a couple of logic errors with these boolean flags. Consider the following two examples

1. `disk_flush_queue` and `replication_flush_queue` both have log records.
2. The loop on line 99 completes, fills `disk_filled_buffer_` with logs, and marks `disk_buffers_processed` as true.
3. `temp_replication_flush_queue_` is not empty so the `if` on line 114 evaluates to false
4. The loop on line 118 completes, fills `replication_filled_buffer_` with logs, and DOES NOT mark `replication_buffers_processed` true.
5. The loop repeats `disk_flush_queue_.empty() && replication_flush_queue_.empty()` evaluates to true and we break out of the loop.
Now even though we have serialized log records for the replica `replication_buffers_processed` is false and we do not continue processing them.

Even if `disk_flush_queue_` is empty, we ALWAYS set `disk_buffers_processed` to true. I don't think this is correct. We end up trying to send logs to the disk log consumer even though we didn't serialize any.

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -148,26 +184,44 @@ void LogSerializerTask::HandFilledBufferToWriter() {
   NOISEPAGE_ASSERT(primary_replication_manager_ != DISABLED ||
                        retention_policy != transaction::RetentionPolicy::RETENTION_LOCAL_DISK_AND_NETWORK_REPLICAS,
                    "If replication is disabled, then you can't send buffers to replicas.");
+  if (destination == SerializeDestination::DISK) {
+    if (disk_filled_buffer_ != nullptr) {
```
jkosh44, line 188: This is probably echoing @mbutrovich comment's a bit, but it's unclear to me why we'd call `HandFilledBufferToWriter` with a destination of `DISK` if the disk filled buffer was null (i.e. not filled). If this is a legitimate scenario I think a comment would go a long way.

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -148,26 +184,44 @@ void LogSerializerTask::HandFilledBufferToWriter() {
...
+    }
 
-  if (filled_buffer_ != nullptr) {
-    filled_buffer_->PrepareForSerialization(retention_policy);
-    // Replicate the buffer if it exists.
-    // TODO(WAN): Note that this is effectively on the critical path to commit callbacks being invoked. Aka terrible.
-    if (primary_replication_manager_ != DISABLED) {
-      primary_replication_manager_->ReplicateBuffer(filled_buffer_);
+    // Hand over the filled buffer
+    if (replication_filled_buffer_ != nullptr) {
```
jkosh44, line 210: This is definitely echoing @mbutrovich comment, but you can merge this whole block with the previous one which has the same check in the if statement.

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -59,17 +59,19 @@ void LogSerializerTask::LogSerializerTaskLoop() {
   } while (run_task_);
   // To be extra sure we processed everything
   Process();
-  NOISEPAGE_ASSERT(flush_queue_.empty(), "Termination of LogSerializerTask should hand off all buffers to consumers");
+  NOISEPAGE_ASSERT(disk_flush_queue_.empty(),
```
jkosh44, line 62: I think we probably want to add an assert for the `replication_flush_queue_` as well.

jkosh44 (changes requested):
Overall looks I left a couple of comments. Also I think it would be helpful if you included in the PR description why we're making this change. That way when this is merged we'll have a log in the commit message as to why the change happened.

src/storage/write_ahead_log/log_serializer_task.cpp
tpan496: @lmwnshn

lmwnshn: From last last thursday, I think we said we're going to split up `RetentionPolicy` into something like the `TransactionPolicy` draft in #1525 (which does explicitly add an assertion of the above flavor). That would allow us to distinguish between our retention policy for the disk and our retention policy for the replica.

I'm not sure I get your question, sorry. Are you asking me if it is technically impossible?

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -82,41 +84,66 @@ std::tuple<uint64_t, uint64_t, uint64_t> LogSerializerTask::Process() {
...
-        RecordBufferSegment *buffer = temp_flush_queue_.front();
-        temp_flush_queue_.pop();
+      while (!temp_disk_flush_queue_.empty()) {
+        RecordBufferSegment *buffer = temp_disk_flush_queue_.front();
+        temp_disk_flush_queue_.pop();
 
         // Serialize the Redo buffer and release it to the buffer pool
         IterableBufferSegment<LogRecord> task_buffer(buffer);
-        const auto num_bytes_records_and_txns = SerializeBuffer(&task_buffer);
+        const auto num_bytes_records_and_txns = SerializeBuffer(&task_buffer, SerializeDestination::DISK);
```
lmwnshn, line 107: Comment A1/2: This (see A2/2) ends up serializing the same buffer twice right?

src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -82,41 +84,66 @@ std::tuple<uint64_t, uint64_t, uint64_t> LogSerializerTask::Process() {
...
+        replication_buffers_processed = true;
+      }
+
+      while (!temp_replication_flush_queue_.empty()) {
+        RecordBufferSegment *buffer = temp_replication_flush_queue_.front();
+        temp_replication_flush_queue_.pop();
+
+        // Serialize the Redo buffer and release it to the buffer pool
+        IterableBufferSegment<LogRecord> task_buffer(buffer);
+        const auto num_bytes_records_and_txns = SerializeBuffer(&task_buffer, SerializeDestination::REPLICAS);
```
lmwnshn, line 126: Comment A2/2: This (see A1/2) ends up serializing the same buffer twice right?

lmwnshn:
Per discussion yesterday, will cherry-pick changes out as necessary. Thanks!

PR #1514, opened 2021-03-10
fix Jenkinsfile-utils for self-driving e2e tests

mbutrovich:
jenkinsfile-utils wasn't updated when #1426 renamed the e2e test build option, so we've had:
```
CMake Warning:
  Manually-specified variables were not used by the project:

    NOISEPAGE_BUILD_SELF_DRIVING_TESTS
```
in every CI build since then.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -523,7 +523,6 @@ pipeline {
                         script{
                             utils = utils ?: load(utilsFileName)
                             utils.noisePageBuild(buildType:utils.RELEASE_BUILD, isBuildTests:false, isBuildSelfDrivingTests: true)
-                            utils.noisePageBuild(buildType:utils.RELEASE_BUILD, isBuildTests:false)
                         }
 
                         // This scripts runs TPCC benchmark with query trace enabled. It also uses SET command to turn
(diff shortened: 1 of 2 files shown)


malin1993ml:
Thanks! But I thought the two builds are for different purposes? One to build the NoisePage binary but the other to build the self-driving test. I could be wrong though...

mbutrovich:
Nah, the second one wasn't even doing anything. It literally couldn't: the utils file was never updated to set that variable anyway. If you look at this build:

http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/master/917/pipeline/464

The second build does nothing, and the test actually gets built much later at the invocation of `timeout 10m ninja self_driving_e2e_test`.

malin1993ml:
Oh I see! This makes sense!

malin1993ml (approved):
LGTM!

malin1993ml:
Well, it turns out that the second build is actually building the noisepage binary if you look closely. So I don't think just removing it would work. I actually had to fix it in #1509 along with renaming some build variables:
https://github.com/cmu-db/noisepage/pull/1509/files#diff-eda34502d530325f47d63da79567683c53d1a5f0181aab7af8bfc061f4e3992fR107-R110

PR #1520, opened 2021-03-17
Upgrade fast float to major version

jkosh44:
This upgrades fast float to the first major release version. I don't plan on monitoring the releases of fast float anymore, I just wanted to upgrade to a major version when one came out.


lmwnshn (approved):
LGTM if pass CI

PR #1521, opened 2021-03-18
Increase record_buffer_segment_size in TPCC performance test to 1e7; increase default record_buffer_segment_size and block_store_size in settings_defs.h to 1e6

malin1993ml:
We're failing the TPCC performance test on the new benchmark machines because we run out of record buffers with the machine's faster speed. This PR increases that setting in the test's JSON config file.

As discussed with @apavlo and @mbutrovich, this PR also increases the default `record_buffer_segment_size` and `block_store_size` since it's just an upper limit and we don't preallocate them anyway.


malin1993ml:
Look at that performance improvement:)

PR #1525, opened 2021-03-24
Add countdown latch to tcop commit callback.

lmwnshn:
# Heading

Add countdown latch to tcop commit callback.

## Description

Add commit callback counters to the traffic cop. The commit callback is invoked once per subsystem (disk, replication, etc) and therefore maintains a countdown latch so that the network thread is only woken up when all the subsystems have invoked the callback.

- Part of the replication fixes, figured breaking this out on its own would make it easier to review.
- **Replication tests are disabled** since more replication fixes are coming shortly.
- Slightly scope creeps into #1511 since I need to be able to have the replication policies on TempNamespace be used. Just a bandaid for now:

1. maintain `individual buffer -> TransactionPolicy` map in log serializer task, maintain current `filled_buffer_` (batch of buffers) policy
2. if you see an incompatible transaction policy as you're serializing new individual buffers, force the current batch of buffers to get flushed, start a new batch

## Remaining tasks

N/A

## Performance

Checking that wal_async_commit_enable still works, unscientifically on my machine,

```
NOOP terminals 6 rate 200k

./noisepage -connection_thread_count=6
30k req/s

./noisepage -connection_thread_count=6 -wal_async_commit_enable
173k req/s
```


src/traffic_cop/traffic_cop.cpp
```diff
@@ -39,9 +38,30 @@
...
+    }
+  }
+};
+
 static void CommitCallback(void *const callback_arg) {
-  auto *const promise = reinterpret_cast<std::promise<bool> *const>(callback_arg);
-  promise->set_value(true);
+  auto *const cb_arg = reinterpret_cast<CommitCallbackArg *const>(callback_arg);
+  --cb_arg->persist_countdown_;
+  if (cb_arg->persist_countdown_ == 0) {
```
mbutrovich, line 62: Is there a race here? I think the comparison should be against a value returned by a post-decrement operator. Otherwise, consider the scenario:

T1: Decrements `persist_countdown_` from 2 to 1
T2: Decrements `persist_countdown_` from 1 to 0. Reads `persist_countdown_` as 0, calls `set_value()`.
T1: Reads `persist_countdown_` as 0, calls `set_value()`.

It's possible that `std::promise::set_value` is idempotent, but I didn't spot that property in cppreference so the behavior of calling it multiple times is unclear to me. Maybe it generates an exception?

lmwnshn: yea, you're right! good catch

lmwnshn:
I thought something genuinely broke, but I can't seem to reproduce locally, so rerunning through CI.

mbutrovich:
This guy is tossing an exception in CI for tpcc_test:

https://github.com/lmwnshn/noisepage/blob/f4c2bc1dfb3eb30681aeecca39226e91bd0c996f/src/storage/write_ahead_log/log_serializer_task.cpp#L100

lmwnshn:
whoops, fixed that in my other PR, there's a race. thanks!

mbutrovich (approved):
Addressed my nits directly. LGTM!

PR #1526, opened 2021-03-25
Replication, except that this time it hopefully works.

### LINKED ISSUE #1515: script.testing.replication.tests_simple failing
Seen it on PRs and on master a few times since #1472 merged.

Example from most recent master build:

http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/master/921/pipeline

I know @lmwnshn is aware and watching it, but just want to track it here.

lmwnshn:
# Heading

Replication, except that this time it hopefully works.

## Description

Builds on #1525.

Fix #1515.

At a high level, there are _many_ issues with #1472. See `docs/design_replication.md`, `Replication (gone wrong)`.

This PR essentially rewrites the implementation of replication from the ground up. See `docs/design_messenger.md` and `docs/design_replication.md`. Some highlights:

- The `Messenger` is now responsible for performing guaranteed delivery. The `Messenger` achieves this by storing a list of pending messages and periodically retrying any messages that have not been acked by the recipient. The message format has changed from `senderCallback-receiverCallback-message` to `messageId-senderCallback-receiverCallback-message`.
- The `ReplicationManager` correctly handles edge cases for synchronous replication with `NotifyOATMsg`. Because buffers are shipped out to replicas as-is, and buffers are only ordered within-txn but not across-txn, the replica must not apply buffers until it is _safe_ to do so, but the safety trigger was the `OldestActiveTransaction` (OAT) time piggybacking off the `CommitRecord`. This would be fine back when our notion of _active_ was "transaction is active until LogCommit", but now the notion of active is "transaction is active until serialized".
- The buffering of received batches of log records was moved from the `ReplicationManager` (#1472 approach) to the `ReplicationLogProvider`.

## Performance

There should be no performance hit with replication disabled.

With replication enabled, currently there is a heavy performance hit. 

On dev10, default for everything, TPCC req/s, sync durability,
Baseline single-node: 1250
Async replication: 650 (2x slower)
Sync replication: 10 (100x slower)

Preliminary delete-profiling for async performance shows that performance is going to the following:

- `nlohmann::json`: Just the `ToJson().dump()` brings us from 1250 to 950.
- Messenger's pending message buffer: The copy into the pending messages buffer brings us from 950 to 650.

@jkosh44 will be speeding this up as part of his project.


src/catalog/postgres/pg_proc_impl.cpp
```diff
@@ -550,7 +550,7 @@ void PgProcImpl::BootstrapProcContexts(const common::ManagedPointer<transaction:
   create_fn("upper", VAR, {VAR}, execution::ast::Builtin::Upper, true);
 
   // Replication.
-  create_fn("replication_get_last_record_id", INT, {}, execution::ast::Builtin::ReplicationGetLastRecordId, true);
+  create_fn("replication_get_last_txn_id", INT, {}, execution::ast::Builtin::ReplicationGetLastTransactionId, true);
```
lmwnshn, line 553: Having to add two separate definitions is annoying, but I am overall very pleased with how adding something to pg_proc works now.

lmwnshn (commented):
Done with self-review.

lmwnshn:
I realize there are actually teardown dependency issues.
Since the log serializer sends things out to the replication manager, but the replication manager is torn down first.

src/include/replication/replica_replication_manager.h
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * The replication manager that should run on the replicas.
+ *
+ * TODO(WAN): If the primary fails, how does the replica switch its replication manager? I guess possible, but annoying.
+ */
+class ReplicaReplicationManager final : public ReplicationManager {
+ public:
+  /**
+   * On construction, the replication manager establishes a listen destination on the messenger and connects to all the
```
jkosh44, line 19: Does this mean that all the replicas connect to each other? If so, why is that?

src/include/storage/recovery/replication_log_provider.h
```diff
@@ -7,134 +7,195 @@
...
+    oats_.emplace(OATPair{oldest_active_txn, batch_id});
+    replication_cv_.notify_all();
+  }
 
-  // Current buffer to read logs from
-  std::unique_ptr<network::ReadBuffer> curr_buffer_ = nullptr;
+  /** @return The latest OAT to be applied, popped off the replication log. */
+  transaction::timestamp_t PopOAT() {
+    std::unique_lock lock(replication_latch_);
+    NOISEPAGE_ASSERT(!oats_.empty(), "Who's telling you to pop the OAT?");
```
jkosh44, line 95: Honestly, might be better to have a more descriptive message here

src/include/storage/recovery/replication_log_provider.h
```diff
@@ -7,134 +7,195 @@
...
+      return false;
+    }
+    NOISEPAGE_ASSERT(received_batch_queue_.top().GetBatchId() > last_batch_popped_, "Duplicate batch added?");
+    // The next batch is ready for application if the batch ID is consecutive.
+    bool top_batch_is_next =
+        received_batch_queue_.top().GetBatchId() == replication::RecordsBatchMsg::NextBatchId(last_batch_popped_);
+    return top_batch_is_next;
+  }
+
+  // bounded set of 100 elements
```
jkosh44, line 130: Seems like this comment is a little lost

src/include/storage/recovery/replication_log_provider.h
```diff
@@ -7,134 +7,195 @@
...
 
   /**
-   * @return true if log file contains more records, false otherwise. Can be blocking if waiting for message from
-   * master, or master has timed out.
+   * @return        If replication has ended, false.
+   *                If there are more log records that can be immediately applied, true.
+   *                If there are no such log records and replication has not yet ended, blocks.
    */
   bool HasMoreRecords() override {
```
jkosh44, line 137: I feel like there's probably a better name for this function. It doesn't really tell you if there or more records or not, but blocks the current until there are more records or replication has ended.

src/messenger/messenger.cpp
```diff
@@ -398,128 +404,225 @@ void Messenger::Terminate() {
...
+  return router_id;
 }
 
-ConnectionId Messenger::MakeConnection(const ConnectionDestination &target) {
-  return ConnectionId(common::ManagedPointer(this), target, identity_);
+connection_id_t Messenger::MakeConnection(const ConnectionDestination &target) {
+  // ZeroMQ is not thread-safe, and the actual creation of new sockets must be done from the same
+  // thread that is going to send and receive using those sockets. See ConnectionToBeAdded docstring.
+  std::unique_lock lock(connections_add_mutex_);
+  connection_id_t connection_id = next_connection_id_++;
```
jkosh44, line 422: I'm not sure if this matters, but since `next_connection_id_` is only accessed with a lock it might not need to be an atomic int.

src/messenger/messenger.cpp
```diff
@@ -398,128 +404,225 @@ void Messenger::Terminate() {
...
 
-void Messenger::ListenForConnection(const ConnectionDestination &target, const std::string &identity,
-                                    CallbackFn callback) {
+router_id_t Messenger::ListenForConnection(const ConnectionDestination &target, const std::string &identity,
+                                           CallbackFn callback) {
   // ZeroMQ is not thread-safe, and the actual binding of new connection endpoints must be done from the same
   // thread that is going to poll the endpoints. See RouterToBeAdded docstring.
   std::unique_lock lock(routers_add_mutex_);
-  routers_to_be_added_.emplace_back(RouterToBeAdded{target, identity, std::move(callback)});
+  router_id_t router_id = next_router_id_++;
```
jkosh44, line 412: I'm not sure if this matters, but since `next_router_id_` is only accessed with a lock it might not need to be an atomic int.

src/messenger/messenger.cpp
```diff
@@ -398,128 +404,225 @@ void Messenger::Terminate() {
...
+                               ZmqMessage::Build(msg_id, send_cb_id, remote_cb_id, router->identity_, message), true});
   }
 }
 
-common::ManagedPointer<ConnectionRouter> Messenger::GetConnectionRouter(const std::string &router_id) {
-  return common::ManagedPointer(routers_.at(router_id).get());
+callback_id_t Messenger::GetNextSendCallbackId() {
+  callback_id_t send_cb_id = next_callback_id_++;
+  // Check for wraparound.
+  if (0 == next_callback_id_.load().UnderlyingValue()) {
```
jkosh44, line 478: Does it matter that this `if` block is not necessarily atomic? When `next_callback_id_.store` is called the value isn't necessarily still 0. It would be nice if there was a test and set operation we could use, `compare-exchange` (https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange) seems like it might be the same thing?

src/messenger/messenger.cpp
```diff
@@ -398,128 +404,225 @@ void Messenger::Terminate() {
...
-    // Add any new routers that need to be added.
-    // TODO(WAN): I wonder how expensive this block is.
-    if (!routers_to_be_added_.empty()) {
-      std::lock_guard lock(routers_add_mutex_);
-      for (auto &item : routers_to_be_added_) {
-        // TODO(WAN): I think I am missing something very obvious on how to handle this. The types involved suck.
-        auto router = std::make_unique<ConnectionRouter>(common::ManagedPointer(this), item.target_, item.identity_,
-                                                         item.callback_);
-        routers_.try_emplace(item.identity_, std::move(router));
+void Messenger::ServerLoopSendMessages() {
```
jkosh44, line 510: Is it worth checking to see if `pending_messages_` is empty before acquiring the lock like the previous loop functions?

src/messenger/messenger.cpp
```diff
@@ -398,128 +404,225 @@ void Messenger::Terminate() {
...
+
+  bool first_time = false;
+  if (message_id > max_seen) {
+    for (message_id_t i = max_seen + 1; i < message_id; ++i) {
+      complement.emplace(i);
+    }
+    max_seen = message_id;
+    first_time = true;
+  } else {
+    if (complement.find(message_id) != complement.end()) {
```
jkosh44, line 615: Any particular reason that this isn't an `else if()`? Not really that important, but it could save you a couple curly braces.

src/replication/primary_replication_manager.cpp
```diff
@@ -0,0 +1,155 @@
...
+  const msg_id_t msg_id = msg.GetMessageId();
+  const std::string msg_string = msg.ToJson().dump();
+  for (const auto &replica : replicas_) {
+    Send(replica.first, msg_id, msg_string, messenger::CallbackFns::Noop, destination_cb);
+  }
+}
+
+record_batch_id_t PrimaryReplicationManager::GetNextBatchId() {
+  record_batch_id_t next_batch_id = next_batch_id_++;
+  if (next_batch_id_.load() == INVALID_RECORD_BATCH_ID) {
```
jkosh44, line 101: I don't think this `if` block is atomic, if multiple threads access it at the same time you might skip over a batch id. I think that shouldn't be a problem since I'm pretty sure only one thread is ever accessing this at a time. Though if that's the case it might not make sense to have `next_batch_id` be atomic in the first place. Also, I'm not sure if it matters if we skip over batch ids.

src/include/messenger/messenger.h
```diff
@@ -293,6 +322,8 @@ class Messenger : public common::DedicatedThreadTask {
    */
   void ProcessMessage(const ZmqMessage &msg);
 
+  bool UpdateMessagesSeen(const std::string &replica, message_id_t message_id);
```
jkosh44, line 325: I think this would benefit from a docstring, just something quick explaining that we keep track of the max message id seen and then every unseen message id below it.

src/replication/replication_manager.cpp
```diff
@@ -1,81 +1,63 @@
...
-const char *ReplicationManager::key_message_type = "message_type";
-
 ReplicationManager::~ReplicationManager() = default;
 
-void ReplicationManager::EnableReplication() {
-  REPLICATION_LOG_TRACE(fmt::format("[PID={}] Replication enabled.", ::getpid()));
-  replication_enabled_ = true;
+msg_id_t ReplicationManager::GetNextMessageId() {
+  msg_id_t next_msg_id = next_msg_id_++;
+  if (next_msg_id_.load() == INVALID_MSG_ID) {
```
jkosh44, line 34: I don't think this if statement is atomic (see my previous comments for more details).

src/storage/recovery/recovery_manager.cpp
```diff
@@ -48,10 +48,21 @@ void RecoveryManager::WaitForRecoveryToFinish() {
 void RecoveryManager::RecoverFromLogs(const common::ManagedPointer<AbstractLogProvider> log_provider) {
   // Replay logs until the log provider no longer gives us logs
   while (true) {
-    if (replication_manager_ != DISABLED &&
-        log_provider->GetType() == AbstractLogProvider::LogProviderType::REPLICATION) {
-      auto rep_log_provider = log_provider.CastManagedPointerTo<ReplicationLogProvider>();
-      if (!rep_log_provider->NonBlockingHasMoreRecords()) break;
+    if (replication_manager_ != DISABLED && replication_manager_->IsReplica()) {
```
jkosh44, line 51: I'm not sure if this matters, but the way this is set up it doesn't seem like it would be possible for a replica node to recover from a crash using the WAL.

src/storage/recovery/recovery_manager.cpp
```diff
@@ -104,24 +113,14 @@ void RecoveryManager::RecoverFromLogs(const common::ManagedPointer<AbstractLogPr
...
-      replication_manager_->ReplicaAck("primary", primary_cb_id, false);
-    }
-  }
 }
 
 void RecoveryManager::ProcessCommittedTransaction(noisepage::transaction::timestamp_t txn_id) {
   // Begin a txn to replay changes with.
   auto *txn = txn_manager_->BeginTransaction();
+  if (replication_manager_ != DISABLED && replication_manager_->IsReplica()) {
+    txn->SetReplicationPolicy(transaction::ReplicationPolicy::DISABLE);
```
jkosh44, line 122: Might be more efficient to just set the default replication policy to `DISABLE` in DB main for replica nodes.

src/storage/recovery/recovery_manager.cpp
```diff
@@ -145,6 +144,14 @@ void RecoveryManager::ProcessCommittedTransaction(noisepage::transaction::timest
 
   // Commit the txn
   txn_manager_->Commit(txn, transaction::TransactionUtil::EmptyCallback, nullptr);
+
+  last_applied_txn_id_ = std::max(last_applied_txn_id_, txn_id);
+  if (replication_manager_ != DISABLED) {
+    // Replicas have to send back their list of deferred transactions that were processed, periodically.
+    if (replication_manager_->IsReplica()) {
+      replication_manager_->GetAsReplica()->NotifyPrimaryTransactionApplied(txn_id);
```
jkosh44, line 152: Not for this PR, but in the future, it may be worth looking into sending back transaction ids to the primary in batches. Since we're usually applying multiple deferred transactions at once, it might be more efficient.

jkosh44 (changes requested):
Looks really good overall. I had a couple of small comments/questions, though I'm not sure if they're all valid.

src/include/replication/replica_replication_manager.h
lmwnshn: Yes. Currently, there is no reason to do so (replica1 never talks to replica2, for example). But you could imagine wanting to do fault tolerance, PAXOS, or something that requires talking to other replicas in the future.

lmwnshn: Marking resolved for now, feel free to reopen.

src/include/storage/recovery/replication_log_provider.h
lmwnshn: Resolved

src/include/storage/recovery/replication_log_provider.h
lmwnshn: good catch!

src/messenger/messenger.cpp
lmwnshn: good point, will add to comment

src/messenger/messenger.cpp
lmwnshn: algorithm used to have an extra if, but that's not a thing now. refactored per suggestion, thanks!

src/messenger/messenger.cpp
lmwnshn: same as above, changed

src/storage/recovery/recovery_manager.cpp
lmwnshn: Sounds good, added a todo. Might be particularly tricky to find a good batch size with sync replication / if we care about latency instead of throughput.

src/storage/recovery/recovery_manager.cpp
lmwnshn: The semantics of that are slightly different I think (disable replication for all txns vs disable replication for just txns that we are replaying via replication). We definitely want the latter, I'm not sure that we want the former further down the road (if we ever support multi-master, big doubts though).

lmwnshn: Well, by a similar token you can argue that those txns which should be replayed can request it explicitly. And you're right that it will save the if check. So I'm moving it anyway.

lmwnshn: done

src/messenger/messenger.cpp
lmwnshn: Added. Note that this check-without-lock is technically undefined behavior. I added a comment as well.

src/messenger/messenger.cpp
lmwnshn: yep you're right!

src/replication/primary_replication_manager.cpp
lmwnshn: 1. yes, only one thread is accessing it `LogSerializerTask` invoking `ReplicateBatchOfRecords`
2. you're right, the atomic is unnecessary, made it a regular non-atomic thing
3. just clarification, it is very important that batch IDs are not skipped. The replication log provider on the receiver will hold the batches until the "next" batch ID is seen, and if it is skipped, it won't make progress

src/include/messenger/messenger.h
lmwnshn: added!

src/replication/replication_manager.cpp
lmwnshn: I think you're correct. I'm actually not sure how you would fix it off the top of my head without making it more expensive (ie a latch); wraparound is expected to be very rare / never. Currently, a "gap" of missing message IDs at the front doesn't matter for any application, and I think message IDs are ultimately dumb identifiers / there is nothing relying on receiving consecutive message IDs. So I added a TODO but that's about it, if you're OK with that then mark as resolve otherwise hmm we should discuss it further.

lmwnshn: actually, I might just be a monkey.

src/include/messenger/messenger.h
```diff
@@ -356,10 +365,16 @@ class Messenger : public common::DedicatedThreadTask {
   std::atomic<message_id_t> next_message_id_{0};
   /** The source callback ID that gets automatically prefixed to messages. */
   std::atomic<callback_id_t> next_callback_id_{static_cast<uint8_t>(BuiltinCallback::NUM_BUILTIN_CALLBACKS) + 1};
-  /** The ID of the next listening router to be made from ListenForConnection(). */
+  /**
+   * The ID of the next listening router to be made from ListenForConnection().
+   * Note that this is always accessed while holding routers_add_mutex_.
+   */
   std::atomic<router_id_t> next_router_id_{0};
```
jkosh44, line 372: Is there a reason that this is atomic but `next_connection_id_` isn't? (Unless I'm misunderstanding I think you may have just forgotten to update this).

src/replication/replication_manager.cpp
```diff
@@ -31,10 +31,8 @@ ReplicationManager::~ReplicationManager() = default;
 
 msg_id_t ReplicationManager::GetNextMessageId() {
   msg_id_t next_msg_id = next_msg_id_++;
-  // TODO(WAN): This is not robust to multiple threads attempting to get a new message ID, and may result in message IDs
-  //            getting skipped over. For now, I think this is fine, but find a cheap way of doing this properly?
-  if (next_msg_id_.load() == INVALID_MSG_ID) {
-    next_msg_id_++;
+  if (next_msg_id == INVALID_MSG_ID) {
```
jkosh44, line 34: As far as I can tell, this will ensure that every message id is unique, no message ids are skipped, and no message id will be equal to `INVALID_MSG_ID`. It could end up assigning "out of order" message ids, in the sense that if thread1 requests a message id before thread2 then thread1 *might* get a higher message id than thread 2. But I don't think that's a problem here.

Also, If `++` is more efficient than `compare_exchange_strong`, it might be worth switching the implementation for `Messenger::GetNextSendCallbackId` to this approach. Though it's probably not too important.

src/replication/replication_manager.cpp
jkosh44: As far as I can tell, this will ensure that every message id is unique, no message ids are skipped, and no message id will be equal to `INVALID_MSG_ID`. It could end up assigning "out of order" message ids, in the sense that if thread1 requests a message id before thread2 then thread1 *might* get a higher message id than thread 2. But I don't think that's a problem here.

Also, If `++` is more efficient than `compare_exchange_strong`, it might be worth switching the implementation for `Messenger::GetNextSendCallbackId` to this approach. Though it's probably not too important.

src/include/messenger/messenger.h
lmwnshn: yep, you're absolutely right

src/replication/replication_manager.cpp
lmwnshn: 1. yea, that wouldn't be a problem.
2. I changed `GetNextSendCallbackId`, but that one is slightly more complicated (while instead of if) because it actually has a range of values reserved.

jkosh44 (approved):

Issue #1519, opened 2021-03-15
Prefix all temp/testing files with "noisepage-"


apavlo:
## Summary
Our testing scripts write a bunch of a temporary data files to `/tmp`. To make it easier to clean these things up, we should use the `noisepage-` prefix for these filenames. That way we know what files are created by our code.

We already do it for some files (e.g., [`compile_time.py`](https://github.com/cmu-db/noisepage/blob/master/script/testing/artifact_stats/collectors/compile_time.py#L16)) but it is not consistent in the repo.

## Solution
We will need to modify both the C++ and Python scripts that reference `/tmp`. We need to be careful to not break anything in Jenkins. You can use the following command to find the files:

```
$ grep -rn "/tmp" src test benchmark script util
```

* src/execution/util/file.cpp
* test/execution/util_file_test.cpp
* benchmark/benchmark_util/benchmark_config.cpp
* script/testing/util/constants.py
* script/testing/junit/constants.py
* script/testing/microbench/constants.py

PR #1529, opened 2021-03-31
Fix Failing Nightly Performance Benchmarks + Prefix for Temp Files

apavlo:
## Description
* Increase the `record_buffer_segment_size` parameter for all the OLTP-Bench configs so that we don't run out of memory and crash.
* Use the "noisepage-" prefix for all temp files that we generate during tests #1519 
* Fix a busy loop in the Python test code when checking for DBMS server output

https://youtu.be/CMr6_fsEekI


lmwnshn (approved):
LGTM

PR #1530, opened 2021-04-01
Pilot Preparation Fix: Update query trace loading in WorkloadForecast

rabbit721:
This PR is fixing the code for populating the forecasted queries in the pilot with recorded query traces.
Specifically, the indices for different columns are updated to align with the addition of db_oid as the first column.


malin1993ml (commented):
I'm just wondering why we change the `unordered_map` to `map`? Is that necessary? Other than that that looks good to me!

malin1993ml (approved):
LGTM.

PR #1532, opened 2021-04-02
WAL async commit no longer generates CommitRecord for read-only txns

mbutrovich:
This is an optimization that was missed when implementing #1446. Currently, with async commit enabled the callback that is sent with the `CommitRecord` is swapped out for an empty callback and the real callback is invoked immediately. The first step can be omitted entirely for read-only transactions, reducing the work performed by the logging components.


lmwnshn (approved):
LGTM! I don't think this breaks anything. Will merge once through CI.

mbutrovich:
Just as a point of reference for some point in the future. YCSB read-only, scale factor 1200, 20 terminals, dev9 went from about ~260,000 txns/sec to ~270,000 txns/sec.

PR #1534, opened 2021-04-06
Move metrics enabled check for logging

mbutrovich:
The `DiskLogConsumer` and `LogSerializer` tasks spin in infinite loops until completed. However, at each iteration they check if metrics should be recorded and thereby increment the sample count. Given enough iterations (the threads are periodically woken up even if no work is done) every event ends up getting recorded regardless of sampling rate. This has two adverse effects: generating more data than we want for a given sampling rate, and generating wrong data. We only want to enable metrics on the first iteration of an event's loop, never on subsequent idle loop iterations, which is what the current architecture does. Only at the start of a new event (the thread did work in the loop) do we test if we should enable metrics again.


src/storage/write_ahead_log/log_serializer_task.cpp
```diff
@@ -48,13 +49,20 @@ void LogSerializerTask::LogSerializerTaskLoop() {
     std::tie(num_bytes, num_records, num_txns) = Process();
```
jkosh44, line 49: I've noticed something thing about these metrics, I don't know if it's good or bad but I just wanted to point it out. `Process()` Has its own loop, which continuously loops as long as there are more log records to serialize. That means that `Process()` doesn't return until there's a break in log records. The two consequences of this are

1. If there's a steady stream of logs then you can end up with a very long interval that contains a lot of log records.
2. While there's a steady stream of logs then you won't get any metrics until there's a break in the logs.

So if you're running a workload with a lot of small transactions, you may end up with a single interval that contains the metrics of the entire workload.

If you wanted more granular metrics it might be a good idea to push all the metrics stuff into `Process()` itself.

malin1993ml (approved):
LGTM! Sorry that just noticed this.

PR #1540, opened 2021-04-11
Refactor Jenkinsfile scripts.

lmwnshn:
# Heading

Refactor Jenkinsfile scripts.

## Description

Some Jenkins changes in preparation for adding self-driving MCTS to CI. That will be done in a followup PR.

- Rename `Jenkinsfile-utils` -> `Jenkinsfile-utils.groovy`. Move _all_ test logic into `Jenkinsfile-utils.groovy`, so that the main `Jenkinsfile` only does configuration.
- Design goal: make cmake strings copy-pasteable from Jenkinsfile. Example

```
                stage('ubuntu-20.04/gcc-9.3 (Debug/ASAN/jumbotests)') {
                    agent       { docker { image 'noisepage:focal' ; args '--cap-add sys_ptrace -v /jenkins/ccache:/home/jenkins/.ccache' } }
                    steps       { script { utils = utils ?: load(utilsFileName) ; utils.stageTest(true, [cmake:
                        '-DCMAKE_BUILD_TYPE=Debug -DNOISEPAGE_BUILD_TESTS=ON -DNOISEPAGE_UNITY_BUILD=ON -DNOISEPAGE_USE_ASAN=ON -DNOISEPAGE_USE_JUMBOTESTS=ON'
                    ] ) } }
                    post        { always { script { utils = utils ?: load(utilsFileName) ; utils.stageArchive() } } ; cleanup { deleteDir() } }
                }
```

- Add a `Logging disabled (release build only)` stage corresponding to `Release` builds with `NOISEPAGE_USE_LOGGING=Off`.
- I personally favor a more condensed way of writing Jenkinsfile / I like the way all test stages fit on my screen. If people feel strongly otherwise, we can go back to the old very spaced out way.
- As a general note, this is not Jenkinsfile best practice which would involve a shared library instead of `script {}` escape hatches, but I think this approach (started by testing team) is actually the best for minimizing overall complexity.


Jenkinsfile-utils.groovy
```diff
@@ -0,0 +1,356 @@
...
+/** Build and run the default "ninja" target.  */
+void stageBuildDefault(Map args = [:]) {
+    stagePre()
+    installPackages()
+    buildNoisePage(args)
+    stagePost()
+}
+
+/** Run the C++ unit tests, TPL tests, optionally generate coverage. */
+void stageTest(Boolean runPipelineMetrics, Map args = [:]) {
```
lmwnshn, line 64: The boolean is a little ugly, but this is because we don't uniformly want to run pipeline metrics on all stages. I'm actually not sure if we have a principled way of picking when to run pipeline metrics.

lmwnshn: Added comment to main Jenkinsfile.

turingcompl33t (approved):
No questions here. I agree with the commentary in the PR conversation: this update makes the CI pipeline significantly more manageable and comprehensible (speaking from a small amount of experiencing in updating the CI pipeline). In other words, LGTM!

PR #1543, opened 2021-04-15
Nightly Microbenchmark JSON Fixes

apavlo:
This PR does three things:

* Strip debug statements from our system in the JSON output of the microbenchmarks.
* Fixes the environment variables that we weren't setting correctly for the # of CPU cores + log path.
* Adds a gitignore to hide the microbenchmark output files.

https://youtu.be/iin3ZVXot74

### DIFF
--- a/script/testing/microbench/micro_benchmarks_runner.py
+++ b/script/testing/microbench/micro_benchmarks_runner.py
@@ -2,6 +2,7 @@ import json
 import os
 import shutil
 import subprocess
+import re
 
 from ..util.constants import LOG
 from .constants import LOCAL_REPO_DIR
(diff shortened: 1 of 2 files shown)


script/testing/microbench/micro_benchmarks_runner.py
```diff
@@ -142,7 +143,11 @@ def _execute_benchmark(self, cmd):
         LOG.debug(f'Executing command [num_threads={self.config.num_threads}]: {cmd}')
         try:
             output = subprocess.check_output(cmd, stderr=subprocess.STDOUT, shell=True)
-            pretty_format_json = json.dumps(json.loads(output.decode('utf8').replace("'", '"')), indent=4)
+            
+            # Strip DBMS debug messages
+            output = re.sub(r"\[.*?\] \[.*?\] \[[\w]+\] .*\n", "", output.decode('utf8'))
```
lmwnshn, line 148: Technically, this might capture and eat something like `[[1] [2] [3] [4]]` which could appear in:
```
{
  'blah': ...,
  'raw_data': [[1] [2] [3] [4]],
  ...
}
```

But I ran a couple of quick tests against numpy and it doesn't seem to print that, so this is probably fine.

lmwnshn (approved):
LGTM!

script/testing/microbench/micro_benchmarks_runner.py
apavlo: I can make it specific to dates if you want

lmwnshn: I think it is probably fine, whoever adds more stuff will figure out that something is up when they realize their output is disappearing / it won't be flaky

PR #1545, opened 2021-04-15
Binder catches WHERE clause with incorrect return type

mbutrovich:
@abigalekim #1234 identified an issue with `WHERE` clauses that don't reference columns. If it's a literal or arithmetic expression things go kaboom, in a few different places. This fixes immediate errors in the `stats_calculator` for selectivity, adds a helper method in the Binder to verify a `WHERE`'s return type, and adds a couple of tests.

The simplest of literals are still broken though. I'll open an issue for that since it's a big rabbit hole.


jkosh44 (approved):
LGTM

PR #1548, opened 2021-04-16
Delete WAL in tests if it is specified as a relative path

jkosh44:
## Description
The WAL file location can be specified as either a relative path or an absolute path. In our test framework, we often start instances of the database with the WAL enabled, and after the test we attempt to delete the WAL. The test framework was only successfully deleting the WAL when it was specified as an absolute path, so this PR adds the functionality to delete the WAL when it's specified as a relative path.


apavlo (approved):
LGTM

PR #1549, opened 2021-04-16
Bplustree iterator

abhijithanilkumar:
# Iterators for BPlusTree.

## Description

This PR adds Iterator support to the BPlusTree in NoisePage. The PR adds two types of iterators: `BPlusTreeFwdIterator` and `BPlusTreeRevIterator` which are forward and reverse iterators respectively.

`Begin()`, `End()` functions can be used to perform forward iteration while `RBegin()` and `REnd()` can be used for reverse iteration. Since locks are being held during scan, whenever a lock cannot be acquired, a `Retry` iterator (`FwdRetry()` or `RevRetry()`) iterator is returned and the user must retry the operation.

## Further work

For the next PR, `IndexIterator` in the execution engine will be modified to directly operate on BPlusTree iterators, bypassing the index wrapper.


src/include/storage/index/bplustree.h
```diff
@@ -747,6 +749,501 @@ class BPlusTree : public BPlusTreeBase {
...
+     * Sets the state of the iterator to RETRY iterator
+     */
+    void SetRetryIterator() {
+      ResetIterator();
+      state_ = RETRY;
+    }
+
+   public:
+    /** Key at the current position of the iterator */
+    KeyType first_;
```
mbutrovich, line 920: Does this mean we do deep copies of current key and value into the iterator? Is there a way we can avoid that?

src/include/storage/index/bplustree.h
```diff
@@ -747,6 +749,501 @@ class BPlusTree : public BPlusTreeBase {
...
+        result = result && (curr_val_ == itr.curr_val_);
+      }
+      return !result;
+    }
+
+    /**
+     * Function to notify that iteration is complete. This is used to release any pending
+     * locks that is being held
+     */
+    void Done() {
```
mbutrovich, line 1075: I'm still not sold on it being in the destructor. We should chat in the standup.

src/include/storage/index/bplustree.h
abhijithanilkumar: Yeah, did this to make it work similar to a normal iterator. An alternative would be to not expose `first` and `second` as members, but have two public functions `first()` and `second()` or `key()` and `value()`. In this implementation, it will be dereference and copy on every call. 
I can switch to this implementation, technically it should decrease one copy. Currently for every key, we copy it to `first_` and for every value, we copy it to `second_` whenever the iterator moves.  Then there will be another copy to transfer this into `results` vector during a scan. If we switch to having a function instead, the first copy can be avoided.
Do you have any preference on function names? `first()`, `second()` vs `key()`, `value()`.

mbutrovich (approved):
This looks great! Awesome job and thanks for the quick changes based on feedback!

PR #1550, opened 2021-04-17
Add configurable timeout to run db script

jkosh44:
## Description
When running the database through our python scripts there is a hardcoded timeout of 600 seconds. After waiting for the DB to startup for 600 seconds the script times out and returns an error.

I think for some use cases 10 minutes is a bit excessive to wait and we may want to timeout much sooner. To solve this I added an optional timeout parameter that defaults to 600 seconds. All existing scripts will be unchanged, but those in the future who want to set a smaller timeout can do that.

Sorry for the formatting changes, I format all the files I'm working on out of habit. If it's an issue I can go through and revert all the formatting changes.

### DIFF
--- a/script/testing/util/db_server.py
+++ b/script/testing/util/db_server.py
@@ -2,19 +2,19 @@
 # Class definition for `NoisePageServer` used in JUnit tests.
 
 import os
-import time
-import shlex
 import pathlib
+import shlex
 import subprocess
+import time
+from typing import Dict, List
 
 import psycopg2 as psql
-
-from typing import Dict, List
 
 from .common import print_pipe
 from .constants import (DEFAULT_DB_BIN, DEFAULT_DB_HOST,
                         DEFAULT_DB_OUTPUT_FILE, DEFAULT_DB_PORT,
                         DEFAULT_DB_USER, DEFAULT_DB_WAL_FILE, DIR_REPO, LOG)
+
 
 # -----------------------------------------------------------------------------
 # NoisePageServer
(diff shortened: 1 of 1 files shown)


lmwnshn (commented):
I'm OK with merging it since it isn't changing anything in CI (see [0]), but it doesn't seem like you're adding a command line flag or something for it either

[0] https://github.com/cmu-db/noisepage/pull/1494

jkosh44:
> I'm OK with merging it since it isn't changing anything in CI (see [0]), but it doesn't seem like you're adding a command line flag or something for it either
> 
> [0] #1494

When you say I'm not adding a command line flag do you mean for a specific script or just all the CI scripts in general that use this file? I think this utility file is used by multiple different scripts each with their own command line arguments. For example the OLTP scripts have their own command line arguments and use this file to start the DB, and the replication scripts have their own command line arguments and use this file to start the DB. 

For those scripts I don't know if making a configurable timeout would be that useful especially since they're mainly run by Jenkins where they don't need to be babysat and we'd rather not have false negatives. However other scripts can have different timeouts especially if they aren't meant to be run by Jenkins and someone is running them manually. 

In general, I think the value of this PR is not to allow a single script to have a configurable timeout, but to allow multiple different scripts to have different timeouts from each other. Of course if someone thinks it's useful they can now also add a configurable timeout to a single script.

lmwnshn (approved):
LGTM

lmwnshn:
Sounds good to me

PR #1551, opened 2021-04-17
Rename port setting from Terrier to NoisePage

jkosh44:
Changes the description of the port setting from "Terrier port (default: 15721)" to "NoisePage port (default: 15721)". 

Note: The `model_save_path` setting still uses terrier for the default path. I don't know enough about this property to feel comfortable changing it though. If the reviewer thinks that it's fine to change that property, I can include it in this PR.


lmwnshn (approved):
LGTM

PR #1553, opened 2021-04-19
Port LLVMEngine::Optimize to skip module passes.

lmwnshn:
# Heading

Port LLVMEngine::Optimize to skip module passes.

## Description

Copies over Prashanth's commit here:
https://github.com/pmenon/tpl/commit/e2d3749032d81a9792d9b8a2e82569d063ebd1ad#diff-ed4b8c2805fa1b88ccfdd18b72ce7eebaf3485e63bf95990c7551716a9d79a5f

Unscientific timing on a separate branch showed that much
more time was spent running module passes than the singular
inlining function pass, so maybe this will help. In general, we
also would like to have function passes (as opposed to module)
wherever possible so that it is easier to time the passes.

And for a few hand-checked examples, function optimize time went up (as expected),
but overall optimize time went down.

Expect to see a speedup in the compiled tests on CI.

### DIFF
--- a/src/execution/vm/llvm_engine.cpp
+++ b/src/execution/vm/llvm_engine.cpp
@@ -953,8 +953,7 @@ void LLVMEngine::CompiledModuleBuilder::Simplify() {
 void LLVMEngine::CompiledModuleBuilder::Optimize() {
   llvm::legacy::FunctionPassManager function_passes(llvm_module_.get());
 
-  // Add the appropriate TargetLibraryInfo and TargetTransformInfo.
-  auto tli = std::make_unique<llvm::TargetLibraryInfoImpl>(target_machine_->getTargetTriple());
+  // Add the appropriate TargetTransformInfo.
   function_passes.add(llvm::createTargetTransformInfoWrapperPass(target_machine_->getTargetIRAnalysis()));
 
   // Build up optimization pipeline.
--- a/src/execution/vm/llvm_engine.cpp
+++ b/src/execution/vm/llvm_engine.cpp
@@ -20,6 +20,9 @@
 #include <llvm/Transforms/IPO.h>
 #include <llvm/Transforms/IPO/AlwaysInliner.h>
 #include <llvm/Transforms/IPO/PassManagerBuilder.h>
+#include <llvm/Transforms/InstCombine/InstCombine.h>
+#include <llvm/Transforms/Scalar.h>
+#include <llvm/Transforms/Scalar/GVN.h>
 
 #include <map>
 #include <memory>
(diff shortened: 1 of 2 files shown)


src/execution/vm/llvm_engine.cpp
```diff
@@ -948,29 +951,32 @@ void LLVMEngine::CompiledModuleBuilder::Simplify() {
 }
 
 void LLVMEngine::CompiledModuleBuilder::Optimize() {
-  llvm::legacy::PassManager module_passes;
   llvm::legacy::FunctionPassManager function_passes(llvm_module_.get());
 
-  // Add the appropriate TargetLibraryInfo and TargetTransformInfo
-  auto target_library_info_impl = std::make_unique<llvm::TargetLibraryInfoImpl>(target_machine_->getTargetTriple());
+  // Add the appropriate TargetLibraryInfo and TargetTransformInfo.
+  auto tli = std::make_unique<llvm::TargetLibraryInfoImpl>(target_machine_->getTargetTriple());
```
lmwnshn, line 957: @pmenon Is this unique pointer necessary? It doesn't seem to be used. I'm not sure whether it has side effects that we want.

pmenon: No, it isn't necessary. Good catch!

src/execution/vm/llvm_engine.cpp
```diff
@@ -948,29 +951,31 @@ void LLVMEngine::CompiledModuleBuilder::Simplify() {
...
-  auto target_library_info_impl = std::make_unique<llvm::TargetLibraryInfoImpl>(target_machine_->getTargetTriple());
+  // Add the appropriate TargetTransformInfo.
   function_passes.add(llvm::createTargetTransformInfoWrapperPass(target_machine_->getTargetIRAnalysis()));
-  module_passes.add(new llvm::TargetLibraryInfoWrapperPass(*target_library_info_impl));
-  module_passes.add(llvm::createTargetTransformInfoWrapperPass(target_machine_->getTargetIRAnalysis()));
 
-  // Build up optimization pipeline
+  // Build up optimization pipeline.
   llvm::PassManagerBuilder pm_builder;
   pm_builder.OptLevel = 3;
```
turingcompl33t, line 961: Perhaps in `CompiledModuleBuilder`:

`constexpr static const unsigned OPT_LEVEL = 3;`

src/execution/vm/llvm_engine.cpp
```diff
@@ -948,29 +951,31 @@ void LLVMEngine::CompiledModuleBuilder::Simplify() {
...
   pm_builder.Inliner = llvm::createFunctionInliningPass(3, 0, false);
```
turingcompl33t, line 962: As above, some descriptive constants here might be useful for the `OptLevel`, `SizeOptLevel`, and `DisableInlineHotCallSite` parameters - I personally hate seeing opaque boolean as formal parameters.

src/execution/vm/llvm_engine.cpp
```diff
@@ -948,29 +951,31 @@ void LLVMEngine::CompiledModuleBuilder::Simplify() {
...
+  function_passes.add(llvm::createReassociatePass());
+  function_passes.add(llvm::createGVNPass());
+  function_passes.add(llvm::createCFGSimplificationPass());
+  function_passes.add(llvm::createAggressiveDCEPass());
+  function_passes.add(llvm::createCFGSimplificationPass());
+
+  // Run optimization passes on all functions.
   function_passes.doInitialization();
-  for (auto &func : *llvm_module_) {
+  for (llvm::Function &func : *llvm_module_) {
```
turingcompl33t, line 975: Not a fan of [AAA](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)?

turingcompl33t (commented):
Looks good. One small nit. Excited to see if the changes to compilation time materialize!

src/execution/vm/llvm_engine.cpp
lmwnshn: While ported wholesale from Prashanth's commit, this reminds me of a conversation that I had with another masters student Y a long time ago (who advocated for less use of auto on readability grounds).

I _used_ to agree with that. I now disagree on readability; unless the RHS has an explicit type I will generally prefer not using auto.

I think there is a difference between writing code for experts and writing code for non-experts. Herb Sutter is definitely an expert. But for the average person picking up this project, they will probably be learning LLVM quirks and TPL quirks at the same time while reading this file, and the types are very helpful in learning what exactly is going on. Our interfaces are refactored rarely and having to type out the type doesn't make the interface any better or worse designed.

lmwnshn:
I kept the variables local to the Optimize() function because it isn't relevant to the rest of the CompiledModuleBuilder and I don't think we're exposing this as a tunable; if we do we will probably add it as a parameter to the Optimize() function anyway.

src/execution/vm/llvm_engine.cpp
turingcompl33t: This is actually _not_ one of the C++ style things I'm particularly attached to, and your point about making the LLVM types "more visible" to new people who come to this file convinces me. Also, there is the additional argument to be made that this is closer to "LLVM style" where it seems like they prefer to use explicit types, at least far more often than you might observe in other C++ codebases.

PR #1557, opened 2021-04-20
Add -async_replication_enable command line arg.

lmwnshn:
# Heading

Add -async_replication_enable command line arg.

## Description

Add -async_replication_enable command line arg.
This arg has no effect unless -replication_enable is also provided,
in which case it defaults to async replication instead of sync replication.

Minimum set of flags: `-messenger_enable -replication_enable -async_replication_enable`


jkosh44 (approved):
LGTM

Issue #1556, opened 2021-04-20
TaskManagerTests failure


malin1993ml:
The TaskManagerTests introduced with the TaskManager by #1502 failed on master (though it passed the tests for merging the PR):
http://jenkins.db.cs.cmu.edu:8080/blue/organizations/jenkins/terrier/detail/master/949/pipeline

lmwnshn:
Fixed by #1560 (hopefully)

PR #1560, opened 2021-04-22
Fix task_manager_test (Issue #1556)

17zhangw:
## Description

`util::QueryExecUtil` does not synchronize access to the callback row function. Patch introduces a `std::mutex` to guarantee that regardless of how the DML query is executed (i.e., parallel or not), the tuple output function can only be invoked by a single thread at a time.


src/util/query_exec_util.cpp
```diff
@@ -21,6 +22,7 @@
 #include "settings/settings_manager.h"
 #include "transaction/transaction_context.h"
 #include "transaction/transaction_manager.h"
+#include "util/query_exec_util.h"
```
malin1993ml, line 25: Our convention seems always put this at the top of the file?

malin1993ml (approved):
LGTM! Just a minor formatting comment.

src/util/query_exec_util.cpp
lmwnshn: Specifically if you use CLion's formatter (ctrl-alt-L) it does this, but I don't think the `clang-format` tool does this.

Issue #1555, opened 2021-04-20
Error catching for internal tasks (queries)


malin1993ml:
We added support for internal tasks in #1502, but we don't catch the errors when the internal tasks fail. Instead, we just print our error log messages from the command line. We should discuss how we really want to handle these errors.

malin1993ml:
Based on the discussion during today's team meeting, we're going to return the internal task execution status (e.g., success/failure) through a `Future` and let the caller decide how to handle that. So I think our infrastructure for `DML` tasks already supports that, but we should also support that for the `DDL` tasks, e.g., during the DBMS start time to create the internal tables.

PR #1561, opened 2021-04-22
Internal Query Execution Status (Issue #1555)

17zhangw:
## Description

PR resolves issue #1555. The PR augments `TaskDDL` with a future that can be used by the caller to identify whether or not the DDL query has succeeded or failed (along with obtaining an error message associated with the failure). PR also performs the following:
- `util::QueryExecUtil` now tracks a "most-recent" error that can be obtained.
- `util::QueryExecUtil::ExecuteQuery` returns false now if the transaction must abort due to execution


malin1993ml (approved):
Looks great. Thanks, @17zhangw!

src/include/util/query_exec_util.h
```diff
@@ -222,6 +228,11 @@ class QueryExecUtil {
    */
   std::unordered_map<std::string, std::unique_ptr<planner::OutputSchema>> schemas_;
   std::unordered_map<std::string, std::unique_ptr<execution::compiler::ExecutableQuery>> exec_queries_;
+
+  /**
+   * Stores the most recently encountered error.
+   */
+  std::string error_msg_;
```
lmwnshn, line 235: Should this be reset in between new queries? So you must use it or lose it before invoking new queries. Seems like a source of potential confusion otherwise.

17zhangw: Decided to reset the error when executing any operation that can throw (`PlanStatement`, `ExecuteDDL`, `CompileQuery`, and `ExecuteQuery`).

PR #1562, opened 2021-04-22
Allow sync durability with async replication

jkosh44:
## Description
This commit allows the database to run with synchronous durability but asynchronous replication. The benefit of this is that you can run with synchronous durability and replication turned on without taking some of the performance penalties associated with synchronous replication.

As the comment suggests there is a potential for data loss if the primary crashes and one of the replicas is promoted to primary (though this promotion functionality isn't implemented yet)

## Performance

All results performed under TPCC with a scale factor of 1

### Sync Durability and Sync Replication
13.198780362740948 requests/sec

### Sync Durability and Async Replication
96.29715975379193 requests/sec


lmwnshn (approved):
LGTM

PR #1565, opened 2021-04-24
Abstract replication message format and serialization benchmark

jkosh44:
## Description

This PR accomplishes two things

1. Abstracts away all usage of JSON in replication messages to a wrapper class called `MessageFacade`. This should make it much easier to switch between message formats as the underlying implementation is not exposed to any caller. Callers just need to call `Serialize()` and the messages will be serialized using whatever message format implements `MessageFacade`. It also makes it easier to benchmark because we can benchmark the `Serialize()` method for all implementations instead of having to switch between benchmarking third party library functions directly. 
I'm not sure if `MessageAdapter` or `MessageBridge` would have been more appropriate names, but `MessageFacade` definitely sounds the coolest out of all of them. Please let me know if I went overboard with the design patterns.
2.  Adds Microbenchmarks for replication message serialization and deserialization.

## Performance
Below is a sample from running the new benchmarks on dev10
```
+--------+------------+------------+----------------+-----------+--------+----------+----------------+-------------+------------------------------+--------------------------------+
| status | iterations | throughput | ref_throughput | tolerance | change | coef_var | reference_type | num_results | suite                        | test                           |
+--------+------------+------------+----------------+-----------+--------+----------+----------------+-------------+------------------------------+--------------------------------+
|  PASS  |  1639503   | 1.1684e+05 |   0.0000e+00   |     0     |   0    |    0     |      none      |      0      | ReplicationMessagesBenchmark | NotifyOATMsgDeserialization    |
|  PASS  |  2934420   | 2.0954e+05 |   0.0000e+00   |     0     |   0    |    0     |      none      |      0      | ReplicationMessagesBenchmark | NotifyOATMsgSerialization      |
|  PASS  |    6709    | 4.7924e+02 |   0.0000e+00   |     0     |   0    |    0     |      none      |      0      | ReplicationMessagesBenchmark | RecordsBatchMsgDeserialization |
|  PASS  |   15509    | 1.1117e+03 |   0.0000e+00   |     0     |   0    |    0     |      none      |      0      | ReplicationMessagesBenchmark | RecordsBatchMsgSerialization   |
|  PASS  |  1814614   | 1.2954e+05 |   0.0000e+00   |     0     |   0    |    0     |      none      |      0      | ReplicationMessagesBenchmark | TxnAppliedMsgDeserialization   |
|  PASS  |  3331124   | 2.3793e+05 |   0.0000e+00   |     0     |   0    |    0     |      none      |      0      | ReplicationMessagesBenchmark | TxnAppliedMsgSerialization     |
+--------+------------+------------+----------------+-----------+--------+----------+----------------+-------------+------------------------------+--------------------------------+
```

## Further work

Investigate other message formats and compare performance


src/include/replication/replication_messages.h
```diff
@@ -30,16 +33,105 @@ namespace noisepage::replication {
...
+class MessageFacade {
+ public:
+  /** The underlying format of messages used in replication */
+  using MessageFormat = common::json;
+
+  /** Default constructor */
+  MessageFacade() = default;
+
+  /** Constructor using underlying message format */
+  explicit MessageFacade(MessageFormat underlying_message) : underlying_message_(std::move(underlying_message)) {}
```
lmwnshn, line 49: thoughts on `MessageFormat &&underlying_message`?

src/replication/replication_messages.cpp
```diff
@@ -86,26 +95,27 @@ RecordsBatchMsg::RecordsBatchMsg(ReplicationMessageMetadata metadata, record_bat
...
+std::unique_ptr<BaseReplicationMessage> BaseReplicationMessage::ParseFromString(std::string_view str) {
+  MessageFacade message(str);
+  // BaseReplicationMessage switches on the message's key_message_type to figure out what type of message to create.
+  ReplicationMessageType msg_type = ReplicationMessageTypeFromString(message.Get<std::string>(key_message_type));
   switch (msg_type) {
-      // clang-format off
-    case ReplicationMessageType::NOTIFY_OAT:          { return std::make_unique<NotifyOATMsg>(json); }
-    case ReplicationMessageType::RECORDS_BATCH:       { return std::make_unique<RecordsBatchMsg>(json); }
-    case ReplicationMessageType::TXN_APPLIED:         { return std::make_unique<TxnAppliedMsg>(json); }
+    // clang-format off
```
lmwnshn, line 115: If I'm not mistaken, this line here still gets clang-format applied to it. That or CLion ctrl-alt-L format did it. but if CI is happy, I'm happy!

src/include/replication/replication_messages.h
```diff
@@ -3,8 +3,11 @@
 #include <limits>
 #include <memory>
 #include <string>
+#include <utility>
+#include <vector>
 
 #include "common/enum_defs.h"
+#include "common/json.h"
```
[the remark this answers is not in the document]
lmwnshn: ah... just saw the templated Put/Get. ok, this will look like ass. but I think it is worth the compilation times. can you move the template declarations to the .cpp file and explicitly instantiate them for the types that we use? sorry

benchmark/replication/replication_messages_benchmark.cpp
```diff
@@ -0,0 +1,151 @@
+#include "benchmark/benchmark.h"
+#include "benchmark_util/benchmark_config.h"
+#include "common/dedicated_thread_registry.h"
+#include "common/scoped_timer.h"
+#include "replication/replication_messages.h"
+#include "storage/write_ahead_log/log_io.h"
+
+namespace noisepage {
+
+class ReplicationMessagesBenchmark : public benchmark::Fixture {
```
lmwnshn, line 10: nice! This ended up looking much cleaner than I thought it would be

lmwnshn (changes requested):
Only request for change is on the inclusion of `json.h`. Because that file is so massive, please explicitly instantiate the templates for Put/Get and move those + the dump() to the cpp file. Trying to avoid including "json.h" in headers as far as is practical.

Otherwise, LGTM! The other comments are either "nice job" or shitposts.

src/include/replication/replication_messages.h
lmwnshn: essentially, makes the copy a bit more explicit for the caller. I don't think I have a strong opinion here since we're hopefully replacing the JSON.

src/replication/replication_messages.cpp
jkosh44: Yeah not really sure what happened here, I didn't explicitly move it. I think it was moved due to clang-format.

src/include/replication/replication_messages.h
jkosh44: Turns out this constructor wasn't used so I removed it.

src/include/replication/replication_messages.h
```diff
@@ -30,16 +33,105 @@ namespace noisepage::replication {
 ENUM_DEFINE(ReplicationMessageType, uint8_t, REPLICATION_MESSAGE_TYPE_ENUM);
 #undef REPLICATION_MESSAGE_TYPE_ENUM
 
+class MessageFacade;
+DEFINE_JSON_HEADER_DECLARATIONS(MessageFacade);
+
+/** Abstraction over the underlying format used to send replication messages over the network */
+class MessageFacade {
```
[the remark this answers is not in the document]
jkosh44: Updated to `MessageWrapper`

src/include/replication/replication_messages.h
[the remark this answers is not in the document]
jkosh44: Fixed

lmwnshn (approved):
LGTM! Thanks for suffering through some templates

PR #1566, opened 2021-04-26
Add pilot tests to CI.

lmwnshn:
# Heading

Add pilot tests to CI.

## Description

Blocked on / has already merged #1568.

- We hit an [import trap](https://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html) with how we were using Python by mixing absolute and relative imports, specifically, the singleton `data_info.instance` had multiple copies created and resolved differently in the Python import cache. To workaround this, we invoke the `ModelServer` with `python3 -m ...` and similarly when we run with coverage. Now, because we are running the python library as a module, the argument is always `python3 -m script.self_driving.model_server`, so the `model_server_path` setting has been removed. Instead, callers should take care that the PYTHONPATH can find that library.
- Speaking of coverage, do we want to run with coverage? Recompiling in DEBUG with COVERAGE is easy. But because we load with TPC-C, the load then takes a while.
- This refactors `forecast.py` to have a generic `run_oltpbench` function which takes in list of functions to be applied. Each function is a generic `func(TestOLTPBench, TestCaseOLTPBench) -> None`. I wonder if the None should be an `int` to return error codes, but this seems sufficient for now. These functions can be created by functors (`make_fn_blah`, "functions which create functions"). It probably makes more sense if you read the usage of the code directly, 

- In adding this, one limitation of the current testing framework is that we need a way to run multiple tests / processes in parallel. You essentially want the equivalent of bash background tasks, e.g., with multiprocessing in Python or something. This would allow you to have the NoisePage DBMS running in the background, something like

```
P1                                       P2
dbms_start()
                                           oltpbench_run()
dbms_stop()
```

Otherwise, your test logic will always be intricately tied to DBMS lifetime logic. There's probably 4-5 ways of separately invoking OLTPBench and 6-10 ways of invoking NoisePage right now with the test framework, each ever so slightly different.

- The correctness of the pilot test is based on the existence of indexes. The test is checking for success based on just timeouts -- it repeatedly queries `select relname from pg_class where relname like 'automated_index_%'` every 30s until a result pops up, and if no result pops up, Jenkins will bonk the test and declare it dead after 20 minutes.

## Remaining tasks

- Running through CI to see how fast it is with coverage.

---

also shoutout to @githubxxcc , I've poked around in model server stuff twice (once for replication, once for this, both in Python and C++ sides) and each time I could quickly find what I wanted to change and make my changes relatively easily


script/testing/self_driving/forecast.py
```diff
@@ -54,73 +52,200 @@ def config_forecast_data(xml_config_file: str, rate_pattern: List[int]) -> None:
...
+
+def fn_test_results_file(result_file_path: str) -> bool:
+    """Test if the specified results file exists (DEFAULT_PIPELINE_METRICS_FILE or DEFAULT_QUERY_TRACE_FILE."""
+    if not Path(result_file_path).exists():
+        LOG.error(f"Missing {result_file_path} at CWD.")
+        return False
+    return True
+
+
+def make_fn_test_case_run_pre_test(create: bool, load: bool) -> Callable[[TestOLTPBench, TestCaseOLTPBench], None]:
```
malin1993ml, line 116: Just curious why do you choose this functor design? What property do you want to achieve?

Jenkinsfile-utils.groovy
```diff
@@ -276,6 +276,66 @@ void stageModeling() {
...
+    PYTHONPATH=../.. python3 -m script.self_driving.modeling.ou_model_trainer --input_path=./ou_runner_input --model_results_path=./ou_runner_model_results --save_path=./ou_runner_trained_model
+    ''', label: 'Train OU models.'
+
+    buildNoisePage([buildCommand:'ninja noisepage', cmake:
+        '-DCMAKE_BUILD_TYPE=Debug -DNOISEPAGE_GENERATE_COVERAGE=ON'
+    ])
+
+    sh script :'''
+    cd build
+    PYTHONPATH=.. timeout 20m python3 -m script.testing.self_driving.jenkins
```
malin1993ml, line 326: I do think we want to enable the pilot planning c++ code coverage. So if we're going to load TPCC under DEBUG, how long it's going to take? Can we use multiple threads to load?

malin1993ml (commented):
Overall this makes sense. Besides a few minor questions, you can probably work with @17zhangw to:
1. Check that the expected index is created as an applied action instead of part of the search using the internal action table (#1575).
2. Besides loading the workload forecast from the disk, also test the scenario that the system trains forecasting models and predicts the future workload using query traces logged in internal tables. This also requires invoking the forecasting model training using `SET` (see below).
3. Invoke the OU/interference/forecasting model training using `SET` command through `psql`, instead of manually invoking them outside the DBMS. @17zhangw can probably help add that feature to the system. (Note that #1574 (which is probably going to be merged before this) is going to integrate the interference model into the pilot. So you also need to train that model to run pilot. Right now we have a [non-integrated test](https://github.com/cmu-db/noisepage/blob/master/test/self_driving_e2e/model_server_test.cpp#L126) for the interference model.)

script/testing/self_driving/forecast.py
lmwnshn: As mentioned in person the other day, but here for posterity:

It is unlikely that the script will always fit in all the functionality that we want. As we have changing requirements and different things that we want to add in the future, we have the following options:

- Modify the function being called into (then you must understand all the current callers and/or gate it behind additional flags)
- Duplicate the function being called into (then you have code duplication)

If it is a functor based design where the caller provides all the things that should happen, then it should be easier to extend functionality in the future (just add a new functor and have the caller invoke it, all existing call sites remain the same).

Jenkinsfile-utils.groovy
lmwnshn: We have since merged in OLTPBench to parallelize the loading of the item table. We'll see in CI if it is reasonable.

script/testing/self_driving/jenkins.py
```diff
@@ -0,0 +1,57 @@
+import sys
+
+from ..util.common import LOG
+from . import constants, forecast
+
+
+def _generate_data():
+    # 1. Create and load TPC-C via OLTPBench.
```
lmwnshn, line 8: TODO move to separate script

script/testing/self_driving/forecast.py
```diff
@@ -54,74 +53,240 @@ def config_forecast_data(xml_config_file: str, rate_pattern: List[int]) -> None:
...
+
+def fn_enable_pilot(oltpbench: TestOLTPBench, test_case: TestCaseOLTPBench) -> None:
+    """Enable pilot planning. Analyze must have been run!"""
+    db_server = oltpbench.db_instance
+    db_server.execute("SET pilot_planning=true", expect_result=False, quiet=False)
+
+
+def fn_enable_pilot_and_wait_for_index(oltpbench: TestOLTPBench, test_case: TestCaseOLTPBench) -> None:
+    """
+    TODO(WAN): This is a hack around the buggy interaction between the Pilot and the network layer requiring
```
lmwnshn, line 111: TODO WAN Update comment DDL lock

lmwnshn:
TODO 

> Besides loading the workload forecast from the disk, also test the scenario that the system trains forecasting models and predicts the future workload using query traces logged in internal tables. This also requires invoking the forecasting model training using SET (see below).

recall what this was

> Invoke the OU/interference/forecasting model training using SET command through psql, instead of manually invoking them outside the DBMS. @17zhangw can probably help add that feature to the system. (Note that #1574 (which is probably going to be merged before this) is going to integrate the interference model into the pilot. So you also need to train that model to run pilot. Right now we have a non-integrated test for the interference model.)

train forecast model over sql

lmwnshn:
Dead?

PR #1568, opened 2021-04-26
Refactor future.Wait() into future.WaitFor() and future.DangerousWait().

lmwnshn:
# Heading

Refactor future.Wait() into future.WaitFor() and future.WaitFootgun().

## Description

If a future is still waiting, the NoisePage DBMS does not shut down cleanly.
This might be OK in tests, so we have `WaitFootgun`.
The intent is that if you're typing `Footgun` explicitly, you know what you're signing up for and that this is a valid case where you won't end up waiting forever or you can reliably get your infinite wait resolved in some way.
Otherwise, per quick discussion with @malin1993ml, we will always wait with a caller-specified timeout in `WaitFor`.

Right now, the main callers are Pilot and ModelServerManager. The inference itself is usually milliseconds, but the timeout is set to 10 seconds to allow for possible Jenkins spikiness.

This does feel a little hacky..

---

Renamed WaitFootgun to DangerousWait

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1154,6 +1154,7 @@ file(GLOB_RECURSE NOISEPAGE_BENCHMARK_SOURCES
         "benchmark/integration/*.cpp"
         "benchmark/metrics/*.cpp"
         "benchmark/parser/*.cpp"
+        "benchmark/replication/*.cpp"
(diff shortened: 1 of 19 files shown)


lmwnshn:
This blocks #1566.

PR #1571, opened 2021-04-28
Minor Catalog refactors

mbutrovich:
The only real logic change is that we were dumping columns for a new table into `pg_statistic` before we validated that creating the new table would even succeed. We should try to be more defensive in the catalog and not assume that operations will succeed.

Otherwise I just did my usual style nit fixes, trying to to focus on `const &` where we're currently doing deep copies on STL data structures that aren't changed in the local scope. I wish clang-tidy would yell at us about that.

### DIFF
--- a/src/catalog/database_catalog.cpp
+++ b/src/catalog/database_catalog.cpp
@@ -334,14 +334,19 @@ void DatabaseCatalog::BootstrapIndex(const common::ManagedPointer<transaction::T
 bool DatabaseCatalog::CreateTableEntry(const common::ManagedPointer<transaction::TransactionContext> txn,
                                        const table_oid_t table_oid, const namespace_oid_t ns_oid,
                                        const std::string &name, const Schema &schema) {
-  // Create associated entries in pg_statistic.
-  {
-    col_oid_t col_oid(1);
-    for (auto &col : schema.GetColumns()) {
-      pg_stat_.CreateColumnStatistic(txn, table_oid, col_oid++, col);
-    }
+  if (pg_core_.CreateTableEntry(txn, table_oid, ns_oid, name, schema)) {
+    CreateTableStatisticEntry(txn, table_oid, schema);
+    return true;
   }
-  return pg_core_.CreateTableEntry(txn, table_oid, ns_oid, name, schema);
+  return false;
+}
+
+void DatabaseCatalog::CreateTableStatisticEntry(const common::ManagedPointer<transaction::TransactionContext> txn,
+                                                const table_oid_t table_oid, const Schema &schema) {
+  // Create associated entries in pg_statistic.
+  for (const auto &col : schema.GetColumns()) {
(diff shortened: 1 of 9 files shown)


lmwnshn (approved):
LGTM, thanks!

Technically, https://releases.llvm.org/8.0.0/tools/clang/tools/extra/docs/clang-tidy/checks/performance-unnecessary-copy-initialization.html is a thing. I'm not sure if copy elision is happening in the majority of old cases (C++17 https://en.cppreference.com/w/cpp/language/copy_elision with its mandatory copy elision in some cases), which might be why it isn't throwing any kind of error. Anyways, this is clearer.

lmwnshn:
It looks like something broke on CI though.

src/catalog/database_catalog.cpp
```diff
@@ -334,14 +334,19 @@ void DatabaseCatalog::BootstrapIndex(const common::ManagedPointer<transaction::T
...
+    return true;
+  }
+  return false;
+}
+
+void DatabaseCatalog::CreateTableStatisticEntry(const common::ManagedPointer<transaction::TransactionContext> txn,
+                                                const table_oid_t table_oid, const Schema &schema) {
   // Create associated entries in pg_statistic.
-  {
-    col_oid_t col_oid(1);
```
jkosh44, line 339: So I looked into it. The col oid for every column in `schema` is 0. That's because the table hasn't been created yet and col oids haven't been assigned yet. Even with your change of moving `pg_stat_.CreateColumnStatistic` to after the table creation, every col oid in `schema` is still 0. My guess is that the schema object doesn't actually get updated when you create a table and you'll have to pull a new one with the actual col oids.

You'll need to add something like `const auto &updated_schema = GetSchema(txn, table_oid)` and pull col oids from that.

jkosh44:
```suggestion
  const auto &updated_schema = GetSchema(txn, table_oid)
  for (const auto &col : updated_schema.GetColumns()) {
```

src/catalog/database_catalog.cpp
```diff
@@ -334,14 +334,19 @@ void DatabaseCatalog::BootstrapIndex(const common::ManagedPointer<transaction::T
...
-    for (auto &col : schema.GetColumns()) {
-      pg_stat_.CreateColumnStatistic(txn, table_oid, col_oid++, col);
-    }
+  for (const auto &col : schema.GetColumns()) {
```
jkosh44, line 347:
```suggestion
  const auto &updated_schema = GetSchema(txn, table_oid)
  for (const auto &col : updated_schema.GetColumns()) {
```

mbutrovich: Good catch, thanks!

mbutrovich:
For posterity, this looks to be about a 20% performance improvement in the relevant microbenchmarks:

Screen Shot 2021-05-04 at 1 33 52 PM

PR #1573, opened 2021-04-30
Add explicit `u` suffix to initialize TrafficCopResult for EXPLAIN.

lmwnshn:
# Heading

Add explicit cast to initialize TrafficCopResult for EXPLAIN.

## Description

Had to add this to make my compiler happy in Release mode.

`-- Compiler: /usr/bin/clang++-8 Clang 8.0.1`

Explicitly cast a 0 as unsigned to help some compilers along.

### DIFF
--- a/src/traffic_cop/traffic_cop.cpp
+++ b/src/traffic_cop/traffic_cop.cpp
@@ -364,7 +364,7 @@ TrafficCopResult TrafficCop::ExecuteExplainStatement(
   out->WriteDataRow(reinterpret_cast<const byte *const>(&plan_string_val), output_columns,
                     {network::FieldFormat::text});
 
-  return {ResultType::COMPLETE, static_cast<uint32_t>(0)};
+  return {ResultType::COMPLETE, 0u};
 }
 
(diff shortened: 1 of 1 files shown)


src/traffic_cop/traffic_cop.cpp
```diff
@@ -364,7 +364,7 @@ TrafficCopResult TrafficCop::ExecuteExplainStatement(
   out->WriteDataRow(reinterpret_cast<const byte *const>(&plan_string_val), output_columns,
                     {network::FieldFormat::text});
 
-  return {ResultType::COMPLETE, 0};
+  return {ResultType::COMPLETE, static_cast<uint32_t>(0)};
```
turingcompl33t, line 367: Hate to come back with this on a one-line diff, but perhaps just use an integer literal here to be a bit more concise?

```
return {ResultType::COMPLETE, 0U};
```

It looks like this is the approach we take elsewhere in the `TrafficCop`.

turingcompl33t (changes requested):

turingcompl33t (approved):
LGTM!

PR #1575, opened 2021-05-06
Internal Table Actions Recording

17zhangw:
## Description
Builds upon existing internal table and task manager infrastructure to support recording actions. This PR records the chosen action that is applied and also records a portion of the search tree. The recorded search tree recorded by this PR consists of the nodes on the "best action sequence path" and (k - 1) next best siblings of the chosen node at each depth of the path.

## Remaining tasks
- [x] Additional testing
- [x] Getting PR through CI

### DIFF
--- a/src/main/startup.sql
+++ b/src/main/startup.sql
@@ -15,3 +15,11 @@ CREATE TABLE noisepage_forecast_frequencies(ts BIGINT, query_id INT, seen REAL);
 -- entire noisepage_forecast_texts during forecasting.
 CREATE TABLE noisepage_forecast_texts(db_id INT, query_id INT, query_text VARCHAR, types VARCHAR);
 CREATE TABLE noisepage_forecast_parameters(ts BIGINT, query_id INT, parameters VARCHAR);
+
+-- There is a planning horizon where we plan 1 action per interval. noisepage_applied_actions
+-- records the action that is selected to be applied for the coming interval.
+CREATE TABLE noisepage_applied_actions(ts BIGINT, action_id INT, cost REAL, db_id INT, action_text VARCHAR);
+
+-- Following the "best" sequence of actions to apply in the horizon, noisepage_best_actions
+-- records that sequence of actions along with at each level, k-1 other high ranking actions.
+CREATE TABLE noisepage_best_actions(ts BIGINT, depth INT, node_id INT, parent_node_id INT, action_id INT, cost REAL, db_id INT, action_text VARCHAR);
--- a/src/main/startup.sql
+++ b/src/main/startup.sql
@@ -22,4 +22,4 @@ CREATE TABLE noisepage_applied_actions(ts BIGINT, action_id INT, cost REAL, db_i
 
 -- Following the "best" sequence of actions to apply in the horizon, noisepage_best_actions
 -- records that sequence of actions along with at each level, k-1 other high ranking actions.
-CREATE TABLE noisepage_best_actions(ts BIGINT, depth INT, node_id INT, parent_node_id INT, action_id INT, cost REAL, db_id INT, action_text VARCHAR);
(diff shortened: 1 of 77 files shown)


src/self_driving/planning/pilot.cpp
```diff
@@ -71,27 +72,40 @@ void Pilot::PerformPlanning() {
...
-                  settings_manager_->GetInt64(settings::Param::pilot_memory_constraint));
-  for (uint64_t i = 0; i < best_action_seq->size(); i++) {
+  mcst.RunSimulation(simulation_number_, settings_manager_->GetInt64(settings::Param::pilot_memory_constraint));
+
+  // Record the top 3 at each level along the "best action path".
+  // TODO(wz2): May want to improve this at a later time.
+  std::vector<std::vector<pilot::ActionTreeNode>> layered_action;
+  mcst.BestAction(&layered_action, 3);
+  for (size_t i = 0; i < layered_action.size(); i++) {
+    pilot::ActionTreeNode &atn = layered_action[i].front();
```
malin1993ml, line 94: nit: I prefer not to use abbreviations unless it's too long. Just use `action` here.

src/util/forecast_recording_util.cpp
```diff
@@ -170,4 +170,66 @@ void ForecastRecordingUtil::RecordForecastQueryFrequencies(uint64_t timestamp_to
   }
 }
 
+void ForecastRecordingUtil::RecordAppliedAction(uint64_t timestamp_to_record, selfdriving::pilot::action_id_t action_id,
```
malin1993ml, line 173: This is not just forecasting recording anymore. Maybe call this `SelfDrivingRecordingUtil`.

src/self_driving/planning/mcts/monte_carlo_tree_search.cpp
```diff
@@ -52,14 +50,25 @@ void MonteCarloTreeSearch::BestAction(uint64_t simulation_number,
...
+    for (size_t i = 0; i < order.size() && i < topk; i++) {
+      auto child = order[i];
+      auto action = child->GetCurrentAction();
+      auto &action_info = action_map_.at(action);
+      top.emplace_back(curr_node->GetTreeNodeId(), child->GetTreeNodeId(), action, child->GetCost(),
+                       action_info->GetDatabaseOid(), action_info->GetSQLCommand());
+    }
+
+    best_action_seq->emplace_back(std::move(top));
+    curr_node = order[0];
```
malin1993ml, line 71: We actually need to apply the top 1 actions (at least for `ChangeKnobAction`s) during the traversal and reverse them back. This is because the `ChangeKnobAction`s are generated based on deltas. If we don't modify the knob values, the actions will always generate commands relative to the knob values from the beginning.

malin1993ml (changes requested):
Generally looks good! A few minor comments.

src/main/startup.sql
```diff
@@ -15,3 +15,11 @@ CREATE TABLE noisepage_forecast_frequencies(ts BIGINT, query_id INT, seen REAL);
...
 CREATE TABLE noisepage_forecast_texts(db_id INT, query_id INT, query_text VARCHAR, types VARCHAR);
 CREATE TABLE noisepage_forecast_parameters(ts BIGINT, query_id INT, parameters VARCHAR);
+
+-- There is a planning horizon where we plan 1 action per interval. noisepage_applied_actions
+-- records the action that is selected to be applied for the coming interval.
+CREATE TABLE noisepage_applied_actions(ts BIGINT, action_id INT, cost REAL, db_id INT, action_text VARCHAR);
+
+-- Following the "best" sequence of actions to apply in the horizon, noisepage_best_actions
+-- records that sequence of actions along with at each level, k-1 other high ranking actions.
+CREATE TABLE noisepage_best_actions(ts BIGINT, depth INT, node_id INT, parent_node_id INT, action_id INT, cost REAL, db_id INT, action_text VARCHAR);
```
malin1993ml, line 25: After #1579, every level of the tree may represent multiple intervals. Could you add which interval the actions are planned to apply to this table as well?

17zhangw: Added a start and end columns to the table.

malin1993ml (approved):

PR #1577, opened 2021-05-09
Refactor workload forecasting logic in Pilot to a separate Forecaster class

malin1993ml:
We have quite some workload forecasting code that doesn't belong to the core `Pilot` logic, but is mixed in the `Pilot` class. This PR separates that code into its own `Forecaster` class so that the logic in `Pilot` is cleaner.


17zhangw (approved):
LGTM!

PR #1578, opened 2021-05-10
Fix waiting for commit callbacks to be invoked by non-existent replicas.

lmwnshn:
# Heading

Fix waiting for commit callbacks to be invoked by non-existent replicas.

## Description

If there are no replicas, but replication is synchronous, then replication will block indefinitely waiting for replicas to trigger the commit callback invocation. This adds a quick fix for that case.

This should pass CI without hitches.


lmwnshn:
Just going to go ahead and merge this one.

PR #1579, opened 2021-05-13
Monte Carlo Search Tree Fix: Varying action horizon for different levels

rabbit721:
- To account for the fact that actions don't need to be planned very frequently for segments forecasted to be far in the future, vary the granularity of action planning.
- As an initial attempt, the values for the number of next levels that will be influenced by only the current action being planned is hardcoded as a property of the search tree class


src/self_driving/planning/mcts/tree_node.cpp
```diff
@@ -112,13 +113,20 @@ common::ManagedPointer<TreeNode> TreeNode::Selection(
...
                                const std::unordered_set<action_id_t> &candidate_actions, uint64_t memory_constraint) {
-  auto start_segment_index = tree_start_segment_index + depth_;
+
+  auto action_plan_start_index = tree_start_segment_index + depth_;
+  auto action_plan_end_index = std::min(tree_start_segment_index + depth_ + action_horizon - 1, tree_end_segment_index);
   auto end_segment_index = tree_end_segment_index;
-  NOISEPAGE_ASSERT(start_segment_index <= end_segment_index,
-                   "start segment index should be no greater than the end segment index");
+
+  SELFDRIVING_LOG_INFO("action_plan_start_index: {} action_plan_end_index: {} end_segment_index: {}",
```
malin1993ml, line 124: This probably should be `LOG_DEBUG`.

malin1993ml (approved):
A minor `LOG` change. Otherwise, LGTM as long as CI passes!

src/self_driving/planning/mcts/tree_node.cpp
rabbit721: Sure, fixed!

malin1993ml:
Given the discussion with @rabbit721, will look at one issue in this PR again.

malin1993ml:
The changes make sense!

PR #1580, opened 2021-05-15
Small fix to compile on Mac (note that we do not support Mac builds).

abigalekim:
# Small Stats Optimizer Bug

## Description

There's a small bug in the stats optimizer that causes my build to not work locally--fixed by adding template to std::min. (putting in separate pull request for modularity's sake).

### Remaining Tasks

None.

## Performance

None.

## Further Work

None.


lmwnshn (changes requested):
LGTM. Had a chat out of band, this is actually because we dropped Mac support (the breaking config is brew LLVM8, Darwin 19.6.0 x86_64). But since it is a relatively trivial cast where `EstimateItemCount` has constrained type, sure. Note that in general we will not really accommodate Mac builds if they require involved changes, so it is recommended to work on a dev machine (e.g., CLion full remote mode) or dual boot instead.

lmwnshn (approved):
Hit the wrong button.

PR #1584, opened 2021-05-20
Add memory constraint support for Pilot

malin1993ml:
This PR adds to the ability to ensure memory constraint in `Pilot`'s planning. The pilot collects memory consumption information at the beginning of the planning and calculates the table growth rates based on the forecasted workload (stored in `MemoryInfo`). During the planning, the pilot maintains `ActionState` about what indexes have been hypothetically created/dropped. Then the `MonteCarloTreeSearch` estimates the memory consumption during the `TreeNode`'s `ChildrenRollout` step using the `MemoryInfo` and `ActionState`.

Following @jrolli's suggestion, this PR also refactors the pilot to create transaction contexts and catalog accessors for all the databases at the beginning of the planning, and put them in a `PlanningContext` to pass along. Those objects are destroyed when planning finishes. This way, the pilot doesn't violate the existing assumptions made by the catalog and transaction semantics.

Furthermore, the newly introduced `PlanningContext` contains all the information required during the planning. It replaces the original approach to pass the `Pilot` pointer as a dependency. `PilotUtil` and `MonteCarloTreeSearch` are no longer friend class of `Pilot` anymore.


src/optimizer/statistics/stats_calculator.cpp
```diff
@@ -162,6 +162,20 @@ void StatsCalculator::Visit(const LogicalDelete *op) {
   }
 }
 
+void StatsCalculator::Visit(const LogicalCreateIndex *op) {
+  auto *root_group = context_->GetMemo().GetGroupByID(gexpr_->GetGroupID());
+
+  // Compute selectivity at the first time
+  if (root_group->GetNumRows() == Group::UNINITIALIZED_NUM_ROWS) {
+    const auto latched_table_stats_reference = context_->GetStatsStorage()->GetTableStats(
+        op->GetDatabaseOid(), op->GetTableOid(), context_->GetCatalogAccessor());
```
jrolli, line 171: The `OptimizerContext` has pointers to both the `TransactionContext` and the `CatalogAcessor` which seem like more appropriate places from which fetch the `DatabaseOid` for storing stats.

src/self_driving/planning/pilot.cpp
```diff
@@ -71,6 +71,10 @@ void Pilot::PerformPlanning() {
     return;
   }
 
+  // Compute memory information used to satisfy the memory constraint
+  PilotUtil::ComputeTableSizeRatios(forecast_.get(), task_manager_, txn_manager_, catalog_, &memory_info_);
+  PilotUtil::ComputeTableIndexSizes(txn_manager_, catalog_, &memory_info_);
```
jrolli, line 76: Why do these take the `TransactionManager` rather than a single `TransactionContext`?

Note:  I think these should probably be merged into one function that does take the `TransactionManager`.  This would allow you to query the catalog (i.e., `pg_stats` and the indexes) inside of the same transaction that you resolved the OIDs (avoids TOCTOU concerns from concurrent DDL).

src/self_driving/planning/pilot_util.cpp
```diff
@@ -419,4 +425,233 @@ void PilotUtil::GroupFeaturesByOU(
...
+    memory_info->table_oids_.emplace_back(table_oid);
+  };
+
+  // Get <table_id, num_rows> pairs with num_rows > 0
+  auto query = fmt::format(
+      "select starelid, max(stanumrows) as num_rows from pg_statistic group by starelid "
+      "having max(stanumrows) > 0;");
+
+  common::Future<task::DummyResult> sync;
+  task_manager->AddTask(std::make_unique<task::TaskDML>(catalog::INVALID_DATABASE_OID, query,
```
jrolli, line 453: I think this is a bad query as `pg_statistic` lives inside of each `DatabaseCatalog` instance (or should based off the `pg_*` prefix).  If this does succeed, I'm curious as to which database/table is queried in the presence of multiple databases (and not just the default one created on first launch).

src/self_driving/planning/pilot_util.cpp
```diff
@@ -419,4 +425,233 @@ void PilotUtil::GroupFeaturesByOU(
...
+    // Parse the query
+    auto parse_result = parser::PostgresParser::BuildParseTree(query_text);
+    if (parse_result->NumStatements() == 0) continue;
+    auto statement = parse_result->GetStatement(0);
+
+    // Identify the table and the number of inserts/deletes
+    int64_t num_row_delta = 0;
+    catalog::table_oid_t table_oid;
+    auto accessor = catalog->GetAccessor(common::ManagedPointer(txn),
+                                         catalog::db_oid_t(forecast->GetDboidByQid(query_id)), DISABLED);
```
jrolli, line 479: I think this "works" for read-only transactions, but I don't know if there are consequences we need to be concerned about doing cross-database reads inside of a single transaction.  @lmwnshn or @mbutrovich?

Also, getting the accessor is probably expensive enough that doing this inside the loop for every query could be painful on performance.

src/self_driving/planning/pilot_util.cpp
```diff
@@ -419,4 +425,233 @@ void PilotUtil::GroupFeaturesByOU(
...
+      if (delete_statement->GetDeleteCondition() != nullptr) {
+        // TODO(lin): right now assuming only deleting one row when there's a condition. Need cardinality estimation
+        //  to get a more accurate estimation
+        num_row_delta = -1;
+      } else {
+        if (table_sizes.find(table_oid) != table_sizes.end())
+          // Delete all the tuples when without condition
+          num_row_delta = -table_sizes[table_oid];
+      }
+    }
```
jrolli, line 499: This seems like it is implementing a mini-binder.  Is there a reason to not callout to the real one?

src/self_driving/planning/pilot_util.cpp
```diff
@@ -419,4 +425,233 @@ void PilotUtil::GroupFeaturesByOU(
...
+    if (num_row_delta != 0 && table_sizes.find(table_oid) != table_sizes.end())
+      query_row_changes[query_id] = std::make_pair(table_oid, num_row_delta);
+  }
+  txn_manager->Abort(txn);
```
jrolli, line 503: Commit should be cheap for read-only transactions (no logs to flush).  What's the reason for aborting instead?

src/self_driving/planning/pilot_util.cpp
```diff
@@ -419,4 +425,233 @@ void PilotUtil::GroupFeaturesByOU(
...
+
+void PilotUtil::ComputeTableIndexSizes(common::ManagedPointer<transaction::TransactionManager> txn_manager,
+                                       common::ManagedPointer<catalog::Catalog> catalog,
+                                       pilot::MemoryInfo *memory_info) {
+  catalog::db_oid_t max_db_oid = catalog->GetNextOid();
+  auto txn = txn_manager->BeginTransaction();
+  // Traverse all databases to find the info (since we don't know which table belongs to which database)
+  for (auto db_oid = catalog::db_oid_t(1); db_oid < max_db_oid; db_oid++) {
+    auto accessor = catalog->GetAccessor(common::ManagedPointer(txn), db_oid, DISABLED);
+    if (accessor == nullptr) continue;
```
jrolli, line 535: This should probably be replaced with a method on the `Catalog` that takes a transaction and returns all current database OIDs (and names?).

src/self_driving/planning/pilot_util.cpp
```diff
@@ -419,4 +425,233 @@ void PilotUtil::GroupFeaturesByOU(
...
+        memory_info->segment_table_size_ratios_[idx][table_id] = new_table_size / table_sizes[table_id];
+    }
+  }
+}
+
+void PilotUtil::ComputeTableIndexSizes(common::ManagedPointer<transaction::TransactionManager> txn_manager,
+                                       common::ManagedPointer<catalog::Catalog> catalog,
+                                       pilot::MemoryInfo *memory_info) {
+  catalog::db_oid_t max_db_oid = catalog->GetNextOid();
+  auto txn = txn_manager->BeginTransaction();
```
jrolli, line 531: This transaction context is leaked (never closed).

jrolli (changes requested):
I have some significant concerns about how the catalog is being used here and think that we should hop on a call with @lmwnshn and/or @mbutrovich to reach an agreement and way forward.

src/include/optimizer/logical_operators.h
```diff
@@ -1272,15 +1272,17 @@ class LogicalCreateIndex : public OperatorNodeContents<LogicalCreateIndex> {
  public:
   /**
    * @param namespace_oid OID of the namespace
+   * @param database_oid OID of the database
```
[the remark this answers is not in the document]
jrolli: Based on the chat and giving the logical operators a second look, this is fine (appears this should have been passed the whole time).  However, please make the database OID the first parameter for consistency with the other operators.

src/self_driving/planning/pilot_util.cpp
malin1993ml: I figured out why this can work. Essentially this query [goes to the `DEFAULT_DATABASE`](https://github.com/cmu-db/noisepage/blob/master/src/util/query_exec_util.cpp#L59-L67), and right now we're inserting everything into the `DEFAULT_DATABASE` when we're executing TPC-C under OLTPBench. Nevertheless, it just works coincidently and I'll fix it to be the correct `db_oid`.

src/self_driving/planning/pilot_util.cpp
rabbit721: I second this.

src/self_driving/planning/pilot_util.cpp
malin1993ml: This is actually not needed anymore once we create transactions for every database at the beginning of the planning and pass them along.

jrolli (approved):
LGTM.  Thanks for the changes.

rabbit721 (approved):
The changes look good. Thanks for the great work!

PR #1585, opened 2021-05-20
Support training models from psql SET commands

17zhangw:
## Description
PR adds support to perform the training of the OU model, the interference model, and the forecast model over psql by exposing three dedicated `train_X_model` knobs that can be triggered with the `SET` command.

The following additional settings are introduced (although some could possibly be coalesced) that can also be altered over psql via the `SET` command to adjust the training of the models:
- `interference_model_input_path`: input path to directory for data to train interference model with
- `interference_model_train_methods`: comma-delimited methods to use for training interference model (i.e., "rf")
- `interference_model_train_timeout`: timeout for training the interference model
- `interference_model_pipeline_sample_rate`: sampling rate of pipeline metric OUs for interference model
- `ou_model_input_path`: input path for directory to data for training OU model
- `ou_model_train_methods`: comma-delimited methods to use to train OU model (i.e., "lr,rf")
- `ou_model_train_timeout`: timeout for training the OU model

### DIFF
--- a/src/include/self_driving/planning/pilot.h
+++ b/src/include/self_driving/planning/pilot.h
@@ -4,6 +4,7 @@
 #include <iostream>
 #include <map>
 #include <memory>
+#include <mutex>  // NOLINT
 #include <string>
 #include <unordered_map>
 #include <utility>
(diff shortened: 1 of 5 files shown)


malin1993ml (commented):
Looks good! I'm just wondering what would happen if we specify something like `--train_forecast_model` when we start the system? At least we should make sure that it doesn't crash. Then will the system train the model or just ignore it? I think either way it's probably fine. But we should document what would be the behavior.

17zhangw:
Documented the behavior. We just ignore it if it is specified during startup.

malin1993ml (approved):
LGTM!

PR #1586, opened 2021-05-20
Port TPL behavior for float literals. Change print behavior for REAL.

lmwnshn:
# Heading

Port TPL behavior for float literals. Change print behavior for REAL.

## Description

@tpan496 found a bug where values outside of float range were treated as 0 or inf.

We were casting doubles down into floats in the `bytecode_generator`.
Something we missed when porting TPL2 and making the "64-bit instead of 32-bit by default" change.

It's another one of those "fixed in TPL" things so I went and yoinked that code.

This also changes the default behavior of printing real values, e.g., `SELECT 0.3333;`.

- A. The old behavior of to_string is to always print to 6 decimal places.
  - Failure case: prints 0.333333333 as 0.333333.
- B. The fmt lib will print as many decimal places as are available, but will not be able to print trailing 0s.
  - Failure case: prints 0.333000 as 0.333.
- C. If we want to print trailing 0s, we can use stringstream and std::fixed.
  - Failure case: prints 0.333 as 0.333000.
    
In general, there is no good solution here. Given a REAL, we don't know if it was float or double initially. If precision matters, revive the decimal PR.
Going with B. for now.

I hope this does not blow up float SQL-based tests...

### DIFF
--- a/script/testing/junit/traces/functions.test
+++ b/script/testing/junit/traces/functions.test
@@ -1,6 +1,6 @@
 # Initialize the database and table for testing
 statement ok
-CREATE TABLE functions1 (int_val INT, neg_int_val INT,  mod_int_val INT, double_val DOUBLE, neg_double_val DOUBLE, small_double_a_val DOUBLE, small_double_b_val DOUBLE, str_i_val VARCHAR(32), str_a_val VARCHAR(32), str_b_val VARCHAR(32), is_null INT)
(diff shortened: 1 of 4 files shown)


tpan496 (approved):
LGTM. Should we be worried about the performance guard?

lmwnshn:
Nah, that thing is super noisy.

PR #1588, opened 2021-05-21
EXPLAIN (FORMAT TPL) and EXPLAIN (FORMAT TBC).

mbutrovich:
This PR adds support for `explain (format tpl) select * from foo;` and `explain (format tbc) select * from foo;`.

---

This branch has the front-end work (`PostgresParser`, `TrafficCop`, `ExplainStatement`, etc.) for `EXPLAIN (FORMAT TPL) ...` to get TPL of queries back via SQL. However, it still remains unclear how to accomplish the back-end work. `AstPrettyPrint` works on an AST, but that's ephemeral and discarded after compilation. Saving that with the `ExecutableQuery` every time seems wasteful (and maybe incorrect, not sure of the AST's memory life cycle). Instead we could build the string in the compilation phase, but then the question becomes how do we return it back? Is there an empty string in every `CompilationContext` that can optionally retain the TPL AST that's been pretty-printed to it? Maybe -- is there something more general purpose to store artifacts of compilation (@turingcompl33t might want this?). For now I'm opening this PR to document discussion before we figure out a path forward.

### DIFF
--- a/src/include/parser/explain_statement.h
+++ b/src/include/parser/explain_statement.h
@@ -7,6 +7,8 @@
 
 namespace noisepage::parser {
 
+enum class ExplainStatementFormat : uint8_t { JSON, TPL };
+
 /**
  * Represents the SQL "EXPLAIN ..."
  */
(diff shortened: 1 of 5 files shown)


lmwnshn:
image

PR #1590, opened 2021-05-24
Common Table Expressions

### LINKED ISSUE #1592: Integration Test Documentation Outdated
The documentation for integration tests / benchmarks (i.e. the things under `script/testing`) is outdated - the instructions that it provides for manually running local tests and benchmarks are out of date. The file `script/testing/README.md` should be updated to reflect changes to our integration testing setup.

I think I was the last one to touch this file, so this is at least partially on me.

turingcompl33t:
This PR continues the work from the original CTE PR [#1111](https://github.com/cmu-db/noisepage/pull/1111).

As the name suggests, this PR brings in support for Common Table Expressions in three variants:
- Simple (`WITH ... AS`)
- Recursive (`WITH RECURSIVE ... AS ...`)
- Iterative (`WITH ITERATIVE ... AS ...`)

In addition, we implement a few fixes / improvements to other parts of the system:
- Adds functionality for nested pipelines
- Adds support for duplicate aliases by making a new `AliasType` that is a class with essentially a string and a serial number
- Ensure we order output columns from `SELECT * ...` queries like Postgres does

The diff for this PR is large, but 44k of the lines are a result of editing the generated _pg_parse_ code to add support for the `ITERATIVE` keyword.

I updated some of the integration testing documentation while adding new tests, so this PR also resolves #1592.

Very little of the work on this PR is my own. The primary authors of the code on this PR are:
- @tanujnay112 
- @soumilm 

Additional contributions were made by:
- @gautam20197 
- @preetansh


turingcompl33t:
Currently failing in the SQL tracefile integration tests, in a couple of different cases:

From `cte.test`:

```
WITH RECURSIVE cte(x) AS (SELECT 1 UNION ALL SELECT tree.node FROM tree INNER JOIN cte ON tree.parent=cte.x) SELECT * FROM cte;
```

From `iterative_cte.test`:

```
WITH ITERATIVE cte(x) AS (SELECT 1 UNION ALL SELECT tree.node FROM tree INNER JOIN cte ON tree.parent=cte.x) SELECT * FROM cte;
```

In both cases, we fail during semantic analysis:

```
ERROR: Line: 110, Col: 0 => function 'indCteScanFree' expects argument of type '*IndCteScanIterator' in position '0', received type '*CteScanIterator'
```

so we have a bug somewhere in code generation.

turingcompl33t:
In `ind_cte_scan_leader_translator.cpp` we initialize the CTE type with:

```
auto cte_type = GetCodeGen()->BuiltinType(ast::BuiltinType::Kind::CteScanIterator);
```

I have updated this to:

```
auto cte_type = GetCodeGen()->BuiltinType(ast::BuiltinType::Kind::IndCteScanIterator);
```
Interestingly, making this change avoids the error during semantic analysis, but then we crash with a segmentation fault. So, either the implementation from before was intentional, OR we have further issues in the implementation of inductive CTEs.

turingcompl33t:
The following queries currently crash the DBMS with a segmentation fault.

From `cte.sql`:

```sql
CREATE TABLE tree (node INTEGER, parent INTEGER);
INSERT INTO tree VALUES (1,NULL), (10, 1), (11, 1), (100, 10), (101, 10), (110, 11), (111, 11);
WITH RECURSIVE cte(x) AS (SELECT 1 UNION ALL SELECT tree.node FROM tree INNER JOIN cte ON tree.parent=cte.x) SELECT * FROM cte;
```

From `iterative_cte.sql`:

```sql
CREATE TABLE tree (node INTEGER, parent INTEGER);
INSERT INTO tree VALUES (1,NULL), (10, 1), (11, 1), (100, 10), (101, 10), (110, 11), (111, 11);
WITH ITERATIVE cte(x) AS (SELECT 1 UNION ALL SELECT tree.node FROM tree INNER JOIN cte ON tree.parent=cte.x) SELECT * FROM cte;
```

turingcompl33t:
The bugfix in [my previous comment](https://github.com/cmu-db/noisepage/pull/1590#issuecomment-847358939) was on the right track, but ultimately incorrect. There, I was under the impression the state entry was declared as the wrong type. This solved the failure in semantic analysis, but led to a double free during teardown of the query state. It turns out that this state entry _was_ the correct type (I've added a comment within the translator to clarify the confusion that I ran into). Instead, the bug was actually in the teardown logic itself:

```c++
void IndCteScanLeaderTranslator::TearDownQueryState(FunctionBuilder *function) const {
  ast::Expr *cte_free_call =
      GetCodeGen()->CallBuiltin(ast::Builtin::IndCteScanFree, {cte_scan_ptr_entry_.Get(GetCodeGen())});
  function->Append(GetCodeGen()->MakeStmt(cte_free_call));
}
```

In this original code, we are calling the `IndCteScanFree` builtin on a pointer that is actually owned by the inductive CTE iterator itself, which led a situation in which we attempted to invoke the destructor for the `CteScanIterator` twice. The fix is just to ensure that we are invoking the `IndCteScanFree` operation on the outer inductive CTE entry: 

```c++
void IndCteScanLeaderTranslator::TearDownQueryState(FunctionBuilder *function) const {
  ast::Expr *cte_free_call =
      GetCodeGen()->CallBuiltin(ast::Builtin::IndCteScanFree, {cte_scan_val_entry_.GetPtr(GetCodeGen())});
  function->Append(GetCodeGen()->MakeStmt(cte_free_call));
}
```

src/include/catalog/catalog_accessor.h
```diff
@@ -397,6 +398,20 @@ class EXPORT CatalogAccessor {
    */
   common::ManagedPointer<transaction::TransactionContext> GetTxn() const { return txn_; }
 
+  /**
+   * Registers a temporary table in this accessor
+   * @warning For use in the execution engine only, such that cached modules function correctly
```
mbutrovich, line 403: Can we add a little more docs what registering means in this case? Why isn't there a corresponding unregister? Is there a bit more context for that warning?

turingcompl33t: I can add some documentation as well. I was actually avoiding it deliberately to minimize the amount of stuff I was touching on the PR, but honestly I love documenting things so this would not be a issue for me.

src/storage/sql_table.cpp
```diff
@@ -86,6 +66,59 @@ ProjectionMap SqlTable::ProjectionMapForOids(const std::vector<catalog::col_oid_
...
+        new_pr->SetNull(offset);
+        continue;
+      }
+      std::memcpy(new_pr_ptr, src_ptr, table_.layout_.AttrSize(cols.second.col_id_));
+
+      // copy over varlens contents
+      if (table_.layout_.IsVarlen(cols.second.col_id_)) {
+        auto varlen = reinterpret_cast<storage::VarlenEntry *>(src_ptr);
+        if (varlen->NeedReclaim()) {
+          byte *new_allocation = new byte[varlen->Size()];
```
mbutrovich, line 110: I don't think that's how we usually allocate the extra buffer for `VarlenEntry`. I think it's likely going through `AllocateAligned` but we should double check.

turingcompl33t: Agreed. It looks like `AllocateAligned` is the way we typically do this in those cases where the `VarlenEntry` is not backed by a string of some sort. I've updated the implementation.

turingcompl33t:
Further integration testing reveals that we fail on a large number of queries that Postgres handles correctly. Will re-evaluate at the standup meeting today how we want to proceed.

turingcompl33t:
Further exploration of the parser error reveals that all failures look something like:

```SQL
 WITH RECURSIVE t(n) AS (VALUES(1) UNION ALL SELECT n+1 FROM t WHERE n < 100) SELECT sum(n) FROM t;
```

The issue is in the use of `VALUES(1)` within the base case of the recursive CTE; it doesn't appear that we support this syntax while Postgres does. Switching the syntax slightly to

```
WITH RECURSIVE t(n) AS (SELECT 1 UNION ALL SELECT n+1 FROM t WHERE n < 100) SELECT sum(n) FROM t;
```

allows the query to run successfully. I imagine that adding support for this syntax is not a top priority. If this is the case, we can close the door on this particular source of errors and move onto the more pressing issues.

turingcompl33t:
**Binder Errors**

Reproduce this failure mode with the following query:

```SQL
WITH RECURSIVE t(n) AS (SELECT (SELECT 1) UNION ALL SELECT n+1 FROM t WHERE n < 5) SELECT * FROM t;
```

We get the following output from the DBMS:

```bash
noisepage: ../src/binder/binder_context.cpp:130: void noisepage::binder::BinderContext::AddCTETable(const string&, const std::vector<noisepage::common::ManagedPointer<noisepage::parser::AbstractExpression> >&, const std::vector<noisepage::parser::AliasType>&): Assertion `(select_list[i]->GetReturnValueType() != type::TypeId::INVALID) && ("CTE column type not resolved")' failed.
Aborted
```

turingcompl33t:
**Segmentation Faults**

Naturally the source of segmentation fault errors may vary across queries. The first query we can use to trigger such an error is:

```SQL
WITH RECURSIVE y (id) AS (SELECT 1), x (id) AS (SELECT * FROM y UNION ALL SELECT id+1 FROM x WHERE id < 5) SELECT * FROM x;
```

The relevant portion of the backtrace:

```bash
#0  0x0000555557e31faa in std::vector<noisepage::common::ManagedPointer<noisepage::parser::AbstractExpression>, std::allocator<noisepage::common::ManagedPointer<noisepage::parser::AbstractExpression> > >::size (this=0x10) at /usr/include/c++/9/bits/stl_vector.h:916
#1  0x00005555582b79ae in std::vector<noisepage::common::ManagedPointer<noisepage::parser::AbstractExpression>, std::allocator<noisepage::common::ManagedPointer<noisepage::parser::AbstractExpression> > >::operator= (this=0x7fffedff9f30, __x=<error reading variable: Cannot access memory at address 0x18>)
    at /usr/include/c++/9/bits/vector.tcc:223
#2  0x00005555582ae477 in noisepage::binder::BindNodeVisitor::Visit (this=0x7fffedffa300, node=...)
    at ../src/binder/bind_node_visitor.cpp:451
#3  0x0000555557e956ab in noisepage::parser::SelectStatement::Accept (this=0x7fffd4008d20, v=...)
    at ../src/include/parser/select_statement.h:351
#4  0x00005555582aa4a7 in noisepage::binder::BindNodeVisitor::BindNameToNode (this=0x7fffedffa300, 
    parse_result=..., parameters=..., desired_parameter_types=...)
    at ../src/binder/bind_node_visitor.cpp:61
#5  0x0000555558291099 in noisepage::trafficcop::TrafficCop::BindQuery (this=0x5555599bd7f0, 
    connection_ctx=..., statement=..., parameters=...) at ../src/traffic_cop/traffic_cop.cpp:420
```

turingcompl33t:
Per the discussion during standup yesterday (05/28) I am ignoring any errors resulting from the use of the `VALUES(x)` syntax in the base case of recursive CTEs, despite the fact that Postgres does support this syntax. This is not an important feature for research because it is so easily replicated with `SELECT x`. From what I can tell thus far, this is the source of all of the parser errors.

turingcompl33t:
I believe I fixed the issue with segmentation faults; the root cause was that in the binder we assumed that any CTE that was _declared_ as recursive (i.e. `WITH RECURSIVE ...`) must follow the recursive structure (i.e. has both a base case and an inductive case). However, this is not always the case. I've updated the logic in the binder to account for this, but this now exposes us to issues further down the stack:

This query used to crash the DBMS:

```SQL
WITH RECURSIVE y(id) AS (SELECT 1), x(id) AS (SELECT * FROM y UNION ALL SELECT id+1 FROM x WHERE id < 5) SELECT * FROM x;
```

but now triggers an assertion failure during binding which reports that "`x` has 0 columns available but 1 specified." This must be an issue in which the columns of the CTE `y` are not resolved as we are considering `x`?

This query used to crash the DBMS:

```SQL
WITH RECURSIVE cte(x) AS (SELECT 1) SELECT * FROM cte;
```

but now fails during code generation on an assertion failure that reports "index into children of plan node should be less than number of children" - the addition of the `RECURSIVE` specified makes it such that we get an index of out range. It must be the case that the `IndCteLeaderTranslator` assumes two children (base and inductive `SELECT`) where this clearly should have only one.

turingcompl33t:
The next issue I plan to address has to do with the "visibility" of one CTE within the definition of another CTE. Errors of this kind can be reproduced with relatively simple queries, which is why it seems important that we address these. Interestingly, we observe (at least) two distinct failure modes for queries of this kind:

The query

```SQL
WITH x(i) AS (SELECT 1), y(j) AS (SELECT * FROM x) SELECT * FROM y;
```

fails gracefully in the binder with the following error message:

```bash
noisepage=# WITH x(i) AS (SELECT 1), y(i) AS (SELECT * FROM x) SELECT * FROM y;
ERROR:  WITH query y has 0 columns available but 1 specified
```

While the query

```SQL
WITH x(i) AS (SELECT 1), y(j) AS (SELECT i FROM x) SELECT * FROM y;
```

dies on an assertion failure:

```bash
noisepage: ../src/include/catalog/schema.h:310: const noisepage::catalog::Schema::Column& noisepage::catalog::Schema::GetColumn(noisepage::catalog::col_oid_t) const: Assertion `(col_oid_to_offset_.count(col_oid) > 0) && ("col_oid does not exist in this Schema")' failed.
Aborted
```

turingcompl33t:
Working on the bug that causes the following query to fail:

```SQL
WITH x(i) AS (SELECT 1), y(j) AS (SELECT * FROM x) SELECT * FROM y;
```

As noted above, this query fails in the binder:

```SQL
noisepage=# WITH x(i) AS (SELECT 1), y(i) AS (SELECT * FROM x) SELECT * FROM y;
ERROR:  WITH query y has 0 columns available but 1 specified
```

The error is essentially reporting that CTE `y` declares an alias column `j` but that it does not actually "yield" any columns from the `SELECT` that defines the body of the CTE. This occurs because we fail to "find" any columns for CTE `x` while we are visiting the body of CTE `y`. The high-level flow that leads to this situation is something like:

- Visit the outermost `SELECT`
- Visit each of the table references (`WITH`) for the outermost `SELECT` (_Binder Context A_)
     - Visit the table reference for CTE `x` (_Binder Context A_)
          - Visit the `SELECT` that defines the body of `x` (_Binder Context B_)
     - Add the columns "yielded" by `x` to the alias map (_Binder Context A_)
     - Visit the table reference for CTE `y` (_Binder Context A_)
          - Visit the `SELECT` that defines the body of `y` (_Binder Context C_)
               - Attempt to locate the referenced table `x` in the current Binder Context (_Binder Context C_)
     - Add the columns "yielded" by `y` to the alias map (_Binder Context A_)

I'll be the first to admit the logic expressed above is a bit difficult to follow... The takeaway is that in the body of CTE `y` when we visit the `SELECT * FROM x`, we are in a `BinderContext` that is **different** from the one to which we previously added the information for CTE `x`, so the `SELECT` in CTE `y` **has no way** to "know about" the schema for the temporary table defined by CTE `x`.

turingcompl33t:
The following is an excerpt from a new document I added to the `docs/` directory ([discussion_cte_implementation.md](https://github.com/turingcompl33t/terrier/blob/feature-ctes/docs/discussion_cte_implementation.md)). I am including it here as well because it is relevant to the status of the PR and current attempts to merge it.

Our current implementation of CTEs suffers from some known limitations which limits the queries we are able to execute. This section provides a comprehensive overview of the queries on which the system currently fails, a best-estimate of the underlying reason for the failure, and what might be required to address it.

**Loops with `UNION`**

An example query that loops as a result of a `UNION` is:

```SQL
WITH RECURSIVE x(i) AS (SELECT 1 UNION SELECT (i+1)%10 FROM x) SELECT i FROM x LIMIT 20;
```

Here we can see that the modulus operation within the inductive case of the CTE will limit the number of unique values in the table `x` to the sequence `1..9,0` at which point the sequence repeats. Naturally, with `UNION`, the fact that only these 10 unique values will be present in the table limits the size of the temporary table to 10 rows, which are the final result of the query. In NoisePage, we currently sit in an infinite loop on this input query. It appears that Postgres is able to perform some higher-level reasoning about the possible values that might appear in the inductive case of the CTE? If this is actually how they handle these cases, this seems like it would constitute a more-involved fix.

**Loops with `UNION ALL`**

Infinite loops with recursive CTEs that involve `UNION ALL` have a different underlying issue than those that loop with `UNION` alone. An example query is:

```SQL
WITH RECURSIVE x(i) AS (SELECT 1 UNION ALL SELECT i+1 FROM x) SELECT * FROM x LIMIT 10;
```

This is another recursive CTE, but with the conspicuous absence of a predicate in the inductive case (the `SELECT` after the `UNION ALL`). Instead, the query relies on the outer `LIMIT` to limit the number of rows in the output set. This is a different underlying issue than the above because in this case the temporary table `x` does actually grow without bound, and it is _only_ the outer `LIMIT` that prevents this from being an inifinite loop.

We fail on this query as a result of the fact that we populate temporary tables for CTEs greedily, meaning that in this case we attempt to populate the temporary table `x` with an infinite number of rows and never break out of the resulting loop. It seems like there are two possible ways we might be able to address this issue:

- Push knowledge of the `LIMIT` _down_ into the iterator that is responsible for producing values from the temporary table. This way, the iterator will "know" that it should stop writing rows to the temporary table after 10 rows have been produced, elimintating the infinite loop.
- Pull tuples from the top of the query plan, rather than pushing them up from the bottom. This is a more fundamental redesign of the execution engine, and might be infeasible for that reason (i.e. it might be totally antagonistic to our implementation of pipelines). One might imagine, however, that in the "traditional" Volcano query processing style, this query would not pose an issue because the root `SELECT` operator with its `LIMIT` would only attempt to pull the next tuple from the temporary table `x` 10 times before completing the query. This would avoid the need to push knowledge of higher-level query constructs down into the CTE iterator.

**References to CTE Constructs in Outer Scope**

An example query that fails when we attempt to reference a CTE construct (e.g. a column of a temporary table)

```SQL
CREATE TABLE test (id INT PRIMARY KEY, name TEXT);
INSERT INTO test (id, name) VALUES (0, 'A')
INSERT INTO test (id, name) VALUES (1, 'B')

WITH x AS (SELECT * FROM test) SELECT * FROM x ORDER BY name;
```

This query fails with an error that reports the requested column does not exist. The column in question is the `name` column of the temporary table `x`. Interestingly enough, attempting to fix the issue with explicit column aliases for the temporary table `x` does not fix the issue:

```SQL
WITH x(id, name) AS (SELECT * FROM test) SELECT * FROM x ORDER BY name;
```

This leads me to believe that (as we will see again below) the issue lies in the order in which identifiers are visited during binding.

**Binder Temporary Table Resolution**

We fail in multiple places during binding as a result of our inability to resolve temporary tables. Example queries that highlight each of these distinct failure modes are:

```SQL
WITH RECURSIVE x(i) AS (SELECT * FROM y), y(j) AS (SELECT 1) SELECT * FROM x;
```

This query fails with an error that reports that the relation `y` does not exist.

```SQL
WITH RECURSIVE x(i) AS (SELECT 1), y(j) AS (SELECT * FROM x) SELECT * FROM y;
```

This query fails with an error that reports that the temporary table `y` has 0 columns available but 1 specified, implying that the columns of temporary table `x` have not been properly resolved.

```SQL
WITH RECURSIVE t(n) AS (SELECT (SELECT 1) UNION ALL SELECT n+1 FROM t WHERE n < 5) SELECT * FROM t;
```

This query fails with an error that reports that the CTE column type is not resolved.

In all of the above cases, the error appears because of the order in which binding of the temporary tables in the query is performed. Our current architecture for the binder performs a naive depth-first traversal of the parsed query statement during binding. However, once CTEs are introduced, this assumption is no longer valid because CTEs introduce more complex rules regarding the visibility of temporary tables across the statement.

We have an implementation in the binder now that allows us to compute a dependency graph among CTE table references within an input statement. Once this dependency graph is computed, we can establish a topoligical ordering of the temporary tables and resolve table references according to this ordering in order to avoid reference resolution failures (at least, that is the theory). Making use of this functionality will require more-invasive changes to the binder, which is why it has been deferred for the time being.

**`SELECT *` Column Ordering**

An example query that fails because of the order in which columns from `SELECT *` queries are produced is:

```SQL
WITH x(i) AS (SELECT 1), y(j) AS (SELECT 2) SELECT * FROM x INNER JOIN y ON x.i+1 = y.j;
```

We produce the correct result from this query, but fail during integration tests because we do not order the columns of the output in a way that is equivalent to the way that Postgres does it (Postgres is used as our baseline for correctness checks). This is not _really_ a correctness issue, but is still something that we might decide to take a look at in the future.

**Conclusion**

Despite these limitations, I believe we are in agreement that the time is long-passed that we should merge this PR. We have support for many of the common CTE use-cases, and these limitations can be addressed in future PRs in a more self-contained, systematic manner than would be the case if we continue to work on this one until ALL of the many bugs / edge cases are resolved.

script/testing/junit/__main__.py
```diff
@@ -2,14 +2,151 @@
```
lmwnshn, line 2: @turingcompl33t So I think I actually misread this, I think this file was an intended change right?
I like the way you nuked constants.py and in general it looks good to me

lmwnshn (approved):
I think I mistook intended changes for unintended changes when skimming the PR previously; I took a pass, there are some nits (e.g., I don't know if we use `using` that much in tests, some minor spelling typos, I feel some CTE code should have more asserts eg that it is a CTE plan node and whatnot) but nothing that can't be fixed at some point. So it could very well have merged on Friday, and so I'm just hitting the trigger now.

PR #1591, opened 2021-05-24
Add support for arbitrary SHOW. Add "set log_level_blah".

### LINKED ISSUE #1188: Implement support for SHOW.

lmwnshn:
# Heading

Add support for arbitrary SHOW. Add "set log_level_blah".

## Description

Fix #1188.

Added support for arbitrary SHOW.
This requires exposing GetParam from settings -- ok?

Added set log_level_(binder|catalog|etc) support.
e.g., `set log_level_binder = 'debug'`

Also fix set query_trace_metrics_output which would ASAN out if you actually tried to use it.

If anyone has objections to exposing `settings_manager_->GetParam(std::string)` please bring them up.

### DIFF
--- a/src/include/settings/settings_callbacks.h
+++ b/src/include/settings/settings_callbacks.h
@@ -28,204 +28,108 @@ class Callbacks {
   static void NoOp(void *old_value, void *new_value, DBMain *db_main,
                    common::ManagedPointer<common::ActionContext> action_context);
 
-  /**
-   * Changes the buffer segment pool size limit.
-   * @param old_value old settings value
-   * @param new_value new settings value
-   * @param db_main pointer to db_main
-   * @param action_context pointer to the action context for this settings change
-   */
+  /** Change the buffer segment pool size limit. */
   static void BufferSegmentPoolSizeLimit(void *old_value, void *new_value, DBMain *db_main,
                                          common::ManagedPointer<common::ActionContext> action_context);
 
-  /**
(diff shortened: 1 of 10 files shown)


src/include/settings/settings_callbacks.h
```diff
@@ -28,204 +28,108 @@ class Callbacks {
   static void NoOp(void *old_value, void *new_value, DBMain *db_main,
                    common::ManagedPointer<common::ActionContext> action_context);
 
-  /**
```
lmwnshn, line 31: I think there's limited value to copypaste comments. I left them for NoOp and removed the rest.

src/include/settings/settings_manager.h
```diff
@@ -152,6 +152,9 @@ class SettingsManager {
   /** @return The ParamInfo corresponding to the given parameter; throws exception if doesn't exist. */
   const ParamInfo &GetParamInfo(const settings::Param &param) const;
 
+  /** @return The Param corresponding to the given name; throws exception if doesn't exist. */
```
lmwnshn, line 155: We did not expose this before, and I don't remember why. If anyone has objections?

mbutrovich (approved):
LGTM, will merge once through CI.

PR #1594, opened 2021-05-27
Refactor Tasks to use builders. Do not replicate startup DDL.

### LINKED ISSUE #1576: Startup SQL interactions with replication.
Should startup SQL commands be replicated or not?

Currently, some SQL commands are just run on startup no matter what node you're on. Suppose the startup SQL included `create table foo (a int)`, for example. 

A. If the startup SQL is replicated, then this forces the following constraint: every replica must be brought up and ready before the primary is brought up.

B. If the startup SQL is not replicated, then weird things may happen if you refer to tables created by the startup process since the replica's RecoveryManager may be missing the oid mapping (haven't checked this). 

I am inclined to say that we should go with A. But it is something to watch out for.

lmwnshn:
# Heading

Refactor Tasks to use builders. Do not replicate startup DDL.

## Description

Fix #1576 per current understanding, we may decide in the future to add startup SQL that does get replicated, but right now everything is internal tables. Also create everything in the default `noisepage` database instead of in database 0 aka INVALID_DB_OID.

Mainly going from constructors like this

```
    task_manager->AddTask(std::make_unique<task::TaskDML>(
        catalog::INVALID_DATABASE_OID, "SELECT * FROM noisepage_forecast_frequencies",
        std::make_unique<optimizer::TrivialCostModel>(), std::move(params), std::move(param_types), freq_check, nullptr,
        settings, false, true, std::nullopt, common::ManagedPointer(&sync)));
```
to this
```
    task_manager->AddTask(task::TaskDML::Builder()
                              .SetDatabaseOid(task_manager->GetDatabaseOid())
                              .SetQueryText("SELECT * FROM noisepage_forecast_frequencies")
                              .SetFuture(common::ManagedPointer(&sync))
                              .SetParameters(std::move(params))
                              .SetParameterTypes(std::move(param_types))
                              .SetTupleFn(std::move(freq_check))
                              .SetShouldSkipQueryCache(true)
                              .SetExecutionSettings(settings)
                              .Build());
```

Fell down a rabbit hole of making it easier to specify additional properties for tasks in the future.

- Refactor TaskDDL and TaskDML to use builders.
- Track default database OID in catalog, expose via GetDefaultDatabaseOid.
- Refuse INVALID_DB_OID to force callers to do the right thing, i.e., explicitly specify what database they are interacting with.
- Change TryLoadStartupDDL to fail loudly instead of silently.
- Disable replication of TryLoadStartupDDL.
- Add TransactionPolicy arg to QueryExecUtil::BeginTransaction.
- Change a const unique_ptr<> & to a ManagedPointer.
- Allow SetReplicationPolicy(DISABLE) if DurabilityPolicy is also DISABLE.
- Nitpick at function docstrings in QueryExecUtil.
- Question: what should DurabilityPolicy be for internal table tasks?
- Found bug, see #1595.

## Remaining tasks

Seeing if it passes CI. I am surprised replication tests did not fail sooner.

### DIFF
--- a/src/include/task/task.h
+++ b/src/include/task/task.h
@@ -6,183 +6,200 @@
 #include <utility>
 #include <vector>
 
+#include "catalog/catalog_defs.h"
 #include "common/future.h"
+#include "execution/exec/execution_settings.h"
 #include "execution/exec_defs.h"
-#include "optimizer/cost_model/abstract_cost_model.h"
-#include "parser/expression/constant_value_expression.h"
-#include "util/query_exec_util.h"
+#include "transaction/transaction_defs.h"
+#include "type/type_id.h"
+#include "util/util_defs.h"
+
+namespace noisepage::metrics {
+class MetricsManager;
+}  // namespace noisepage::metrics
+
+namespace noisepage::optimizer {
+class AbstractCostModel;
+}  // namespace noisepage::optimizer
+
+namespace noisepage::parser {
+class ConstantValueExpression;
+}  // namespace noisepage::parser
+
+namespace noisepage::util {
+class QueryExecUtil;
+}  // namespace noisepage::util
 
 namespace noisepage::task {
 
 class TaskManager;
 
-/** Enum class used to describe the types of tasks available */
+/** The types of tasks available. */
 enum class TaskType : uint8_t { DDL_TASK, DML_TASK };
 
-/**
- * Type meant to represent a dummy result value.
- * A Future<> cannot be specified with (void), hence why
- * this dummy result type is used.
- */
-struct DummyResult {};
-
-/**
- * Abstract class for defining a task
- */
+/** An abstract representation of a Task to be executed. */
 class Task {
  public:
+  /** Base builder class for tasks. */
+  template <class ConcreteType>
+  class Builder {
+   public:
+    Builder() = default;
+    virtual ~Builder() = default;
+
+    /** Set the database to execute the task in. */
+    ConcreteType &SetDatabaseOid(const catalog::db_oid_t db_oid) {
+      db_oid_ = db_oid;
+      return *dynamic_cast<ConcreteType *>(this);
+    }
+    /** Set the query text to be executed. */
+    ConcreteType &SetQueryText(std::string &&query_text) {
+      query_text_ = std::move(query_text);
+      return *dynamic_cast<ConcreteType *>(this);
+    }
+    /** (Optional) Set the policy for the transaction that this Task will be executed in. */
+    ConcreteType &SetTransactionPolicy(const transaction::TransactionPolicy policy) {
+      policy_ = policy;
+      return *dynamic_cast<ConcreteType *>(this);
+    }
+    /** (Optional) Set the pointer to the Future that the caller is potentially blocking on. */
+    ConcreteType &SetFuture(const common::ManagedPointer<common::FutureDummy> sync) {
+      sync_ = sync;
+      return *dynamic_cast<ConcreteType *>(this);
+    }
+
+   protected:
+    std::optional<catalog::db_oid_t> db_oid_{std::nullopt};  ///< The database to execute the task in.
+    std::optional<std::string> query_text_{
+        std::nullopt};                       ///< The query text to be executed. Must be a DDL statement or SET command.
+    transaction::TransactionPolicy policy_;  ///< The policy for the transaction that this Task is executed in.
+    common::ManagedPointer<common::FutureDummy> sync_;  ///< Future* that the caller may block on.
+  };
+
+  /** Destructor. */
+  virtual ~Task() = default;
+
+  /** @return The type of this task. */
+  virtual TaskType GetTaskType() = 0;
+
   /**
-   * Executes the task
-   * @param query_exec_util Query execution utility
-   * @param task_manager TaskManager that the task was submitted to (i.e., TaskManager running the current task)
+   * Execute the specified task.
+   * @param query_exec_util Query execution utility.
+   * @param task_manager    TaskManager that the task was submitted to (i.e., TaskManager running the current task).
    */
   virtual void Execute(common::ManagedPointer<util::QueryExecUtil> query_exec_util,
                        common::ManagedPointer<task::TaskManager> task_manager) = 0;
 
-  /**
-   * @return task type
-   */
-  virtual TaskType GetTaskType() = 0;
+ protected:
+  const catalog::db_oid_t db_oid_;               ///< The database to execute the task in.
+  const std::string query_text_;                 ///< The query text to be executed.
+  const transaction::TransactionPolicy policy_;  ///< The policy for the transaction that this Task is executed in.
+  const common::ManagedPointer<common::FutureDummy> sync_;  ///< Future* that the caller may block on.
 
-  /**
-   * Default (virtual) destructor
-   */
-  virtual ~Task() = default;
+  /** Constructor. */
+  Task(catalog::db_oid_t db_oid, std::string &&query_text, transaction::TransactionPolicy policy,
+       common::ManagedPointer<common::FutureDummy> sync);
 };
 
-/**
- * Task executes a DDL or SET task.
- */
+/** Task representing a DDL command or a SET statement. */
 class TaskDDL : public Task {
  public:
-  /**
-   * TaskDDL constructor
-   * @param db_oid Database to execute task within
-   * @param query_text Query text of DDL/SET
-   * @param sync Future for the caller to block on
-   */
-  TaskDDL(catalog::db_oid_t db_oid, std::string query_text, common::ManagedPointer<common::Future<DummyResult>> sync)
-      : db_oid_(db_oid), query_text_(std::move(query_text)), sync_(sync) {}
-
-  void Execute(common::ManagedPointer<util::QueryExecUtil> query_exec_util,
-               common::ManagedPointer<task::TaskManager> task_manager) override;
-  TaskType GetTaskType() override { return TaskType::DDL_TASK; }
+  /** Builder for a TaskDDL. */
+  class Builder : public Task::Builder<Builder> {
+   public:
+    /** Constructor. */
+    Builder() = default;
+    /** No copying or moving allowed. */
+    DISALLOW_COPY_AND_MOVE(Builder);
+    /** @return The TaskDDL. */
+    std::unique_ptr<TaskDDL> Build();
+  };
 
   ~TaskDDL() override = default;
-
- private:
-  catalog::db_oid_t db_oid_;
-  std::string query_text_;
-  common::ManagedPointer<common::Future<DummyResult>> sync_;
-};
-
-/**
- * Task executes a DML
- */
-class TaskDML : public Task {
- public:
-  /**
-   * TaskDML constructor
-   * @param db_oid Database to execute task within
-   * @param query_text DML query to execute
-   * @param cost_model Cost model to use for optimizing query
-   * @param skip_query_cache Whether to skip retrieving pre-optimized and saving optimized plans
-   * @param params Relevant query parameters
-   * @param param_types Types of the query parameters if any
-   */
-  TaskDML(catalog::db_oid_t db_oid, std::string query_text, std::unique_ptr<optimizer::AbstractCostModel> cost_model,
-          bool skip_query_cache, std::vector<std::vector<parser::ConstantValueExpression>> &&params,
-          std::vector<type::TypeId> &&param_types)
-      : db_oid_(db_oid),
-        query_text_(std::move(query_text)),
-        cost_model_(std::move(cost_model)),
-        params_(params),
-        param_types_(param_types),
-        tuple_fn_(nullptr),
-        metrics_manager_(nullptr),
-        force_abort_(false),
-        skip_query_cache_(skip_query_cache),
-        override_qid_(std::nullopt),
-        sync_(nullptr) {}
-
-  /**
-   * TaskDML constructor
-   * @param db_oid Database to execute task within
-   * @param query_text DML query to execute
-   * @param cost_model Cost model to use for optimizing query
-   * @param params Relevant query parameters
-   * @param param_types Types of the query parameters if any
-   * @param tuple_fn Function for processing rows
-   * @param metrics_manager Metrics Manager to be used
-   * @param settings ExecutionSettings of this query (note: clang-tidy complains that ExecutionSettings is
-   * trivially-copyable so we can't use std::move())
-   * @param force_abort Whether to forcefully abort the transaction
-   * @param skip_query_cache Whether to skip retrieving pre-optimized and saving optimized plans
-   * @param override_qid Describes whether to override the qid with a value
-   * @param sync Future for the caller to block on
-   */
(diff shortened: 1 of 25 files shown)


src/task/task.cpp
```diff
@@ -1,51 +1,142 @@
...
+Task::Task(const catalog::db_oid_t db_oid, std::string &&query_text, const transaction::TransactionPolicy policy,
+           const common::ManagedPointer<common::FutureDummy> sync)
+    : db_oid_(db_oid), query_text_(std::move(query_text)), policy_(policy), sync_(sync) {}
+
+std::unique_ptr<TaskDDL> TaskDDL::Builder::Build() {
+  NOISEPAGE_ASSERT(db_oid_.has_value(), "Database OID must be specified.");
+  NOISEPAGE_ASSERT(*db_oid_ != catalog::INVALID_DATABASE_OID, "You're doing something hacky. Get a real database OID.");
+  NOISEPAGE_ASSERT(query_text_.has_value(), "Query text must be specified.");
+  NOISEPAGE_ASSERT(policy_.has_value(), "Transaction policy must be specified.");
+  return std::unique_ptr<TaskDDL>(new TaskDDL(*db_oid_, std::move(*query_text_), *policy_, sync_));
```
turingcompl33t, line 21: `std::make_unique()`?

src/task/task.cpp
```diff
@@ -1,51 +1,142 @@
...
+  NOISEPAGE_ASSERT(query_text_.has_value(), "Query text must be specified.");
+  NOISEPAGE_ASSERT(policy_.has_value(), "Transaction policy must be specified.");
+  NOISEPAGE_ASSERT(!override_qid_.has_value() || should_skip_query_cache_,
+                   "Cannot override query ID without skipping the query cache.");
+  NOISEPAGE_ASSERT(params_.empty() || (params_.at(0).size() == param_types_.size()),
+                   "Parameter types do not match parameters in length.");
+  // Set defaults.
+  cost_model_ = cost_model_ != nullptr ? std::move(cost_model_) : std::make_unique<optimizer::TrivialCostModel>();
+
+  return std::unique_ptr<TaskDML>(new TaskDML(*db_oid_, std::move(*query_text_), *policy_, sync_,
```
turingcompl33t, line 37: `std::make_unique()`?

src/task/task.cpp
```diff
@@ -1,51 +1,142 @@
...
+                                              std::move(cost_model_), std::move(params_), std::move(param_types_),
+                                              settings_, override_qid_, metrics_manager_, should_force_abort_,
+                                              should_skip_query_cache_, std::move(tuple_fn_)));
+}
+
+TaskDML::Builder &TaskDML::Builder::SetCostModel(std::unique_ptr<optimizer::AbstractCostModel> cost_model) {
+  cost_model_ = std::move(cost_model);
+  return *this;
+}
```
turingcompl33t, line 46: Mega-nit: empty lines between these function definitions.

src/task/task.cpp
```diff
@@ -1,51 +1,142 @@
...
     std::vector<parser::ConstantValueExpression> &params_0 = params_[0];
-    result = query_exec_util->CompileQuery(query_text_, common::ManagedPointer(&params_0),
+    dml_ok = query_exec_util->CompileQuery(query_text_, common::ManagedPointer(&params_0),
                                            common::ManagedPointer(&param_types_), std::move(cost_model_), override_qid_,
                                            settings_);
 
     // Execute with specified parameters only if compilation succeeded
-    if (result) {
+    if (dml_ok) {
       for (auto &param_vec : params_) {
```
turingcompl33t, line 136: We need some new construct that is a combination of both `for`-loop and `while`-loop...

turingcompl33t (approved):
LGTM. Some interesting C++ required to implement a builder when using inheritance-based runtime polymorphism!

src/main/db_main.cpp
```diff
@@ -4,53 +4,59 @@
...
+  for (auto &ddl : startup_ddls) {
+    COMMON_LOG_DEBUG("Executing startup DDL: {}", ddl);
+    common::FutureDummy sync;
+    task_manager_->AddTask(task::TaskDDL::Builder()
+                               .SetDatabaseOid(default_db_oid)
+                               .SetQueryText(std::move(ddl))
+                               .SetTransactionPolicy(policy)
+                               .SetFuture(common::ManagedPointer(&sync))
+                               .Build());
+    auto future_result = sync.DangerousWait();
```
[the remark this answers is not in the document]
malin1993ml: I think engineering-wise it's more straightforward to have all the internal tasks go through the same API, unless very beneficial to add a new API. Also this gives us control over how much resources that we want to allocate for internal tasks (e.g., we may be able to execute some internal queries in parallel and we can potentially control the resource consumption with a single `TaskManager`).

src/self_driving/planning/pilot_util.cpp
```diff
@@ -214,18 +221,28 @@ std::unique_ptr<metrics::PipelineMetricRawData> PilotUtil::CollectPipelineFeatur
...
-          db_oid, query_text, std::make_unique<optimizer::TrivialCostModel>(),
-          std::vector<std::vector<parser::ConstantValueExpression>>(*params), std::vector<type::TypeId>(*param_types),
-          nullptr, metrics_manager, settings, true, true, std::make_optional<execution::query_id_t>(qid),
-          common::ManagedPointer(&sync)));
+      planning_context.GetTaskManager()->AddTask(
+          task::TaskDML::Builder()
+              .SetDatabaseOid(db_oid)
+              .SetQueryText(std::move(query_text))
+              .SetFuture(common::ManagedPointer(&sync))
+              // TODO(WAN): Copying of params and param_types, is that intentional?
```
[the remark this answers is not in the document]
malin1993ml: Synchronous tasks don't need copying, but asynchronous tasks may go out of the scope of the original `params` and `param_types`. If this becomes too expensive, we may be able to avoid copying for synchronous tasks by adding a new API (see the comments at the top of this code block). Addressing the life cycle issue for asynchronous tasks may be challenging though.

src/self_driving/planning/planning_context.cpp
```diff
@@ -6,6 +6,28 @@
...
+    : ou_model_save_path_(std::move(ou_model_save_path)),
+      interference_model_save_path_(std::move(interference_model_save_path)),
+      catalog_(catalog),
+      metrics_thread_(metrics_thread),
+      model_server_manager_(model_server_manager),
+      settings_manager_(settings_manager),
+      stats_storage_(stats_storage),
+      txn_manager_(txn_manager),
+      query_exec_util_(std::move(query_exec_util)),
+      task_manager_(task_manager) {}
```
malin1993ml, line 27: Just curious, any particular reason to put this in the `.cpp` file? I don't have a strong opinion either way.

src/util/self_driving_recording_util.cpp
```diff
@@ -37,9 +38,15 @@ void SelfDrivingRecordingUtil::RecordQueryMetadata(
...
   if (!params_vec.empty()) {
     std::string query = SelfDrivingRecordingUtil::QUERY_TEXT_INSERT_STMT;
-    task_manager->AddTask(std::make_unique<task::TaskDML>(catalog::INVALID_DATABASE_OID, query,
-                                                          std::make_unique<optimizer::TrivialCostModel>(), false,
-                                                          std::move(params_vec), std::move(param_types)));
+    task_manager->AddTask(
+        task::TaskDML::Builder()
+            .SetDatabaseOid(task_manager->GetDatabaseOid())
+            .SetQueryText(std::move(query))
+            // TODO(WAN): #1595
```
malin1993ml, line 45: Is #1595 the correct issue number? It's already closed...

malin1993ml (commented):
The refactoring looks great! Appreciate it, @lmwnshn! Just some small clarifications and comments. The main thing that I'm wondering though, is that have you tried running the internal query trace recording and the pilot with your refactor? Since we don't have tests for them in CI yet, the only way to make sure that they're not broken is to run them locally.

src/include/task/task.h
```diff
@@ -6,183 +6,199 @@
...
-  /**
-   * TaskDDL constructor
-   * @param db_oid Database to execute task within
-   * @param query_text Query text of DDL/SET
-   * @param sync Future for the caller to block on
-   */
-  TaskDDL(catalog::db_oid_t db_oid, std::string query_text, common::ManagedPointer<common::Future<DummyResult>> sync)
-      : db_oid_(db_oid), query_text_(std::move(query_text)), sync_(sync) {}
+  /** Builder for a TaskDDL. */
+  class Builder : public Task::Builder<Builder> {
```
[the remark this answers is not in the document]
lmwnshn: so it did

src/include/task/task.h
```diff
@@ -6,183 +6,199 @@
...
+namespace noisepage::util {
+class QueryExecUtil;
+}  // namespace noisepage::util
 
 namespace noisepage::task {
 
 class TaskManager;
 
-/** Enum class used to describe the types of tasks available */
+/** The types of tasks available. */
```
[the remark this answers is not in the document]
lmwnshn: If you find a way to get format to reformat
```
/**
 * oneliner
 */
```

into
```
/** oneliner */
```

I would be very happy

src/main/db_main.cpp
```diff
@@ -4,53 +4,59 @@
...
+    auto startup_ddl_location = settings_manager_->GetString(settings::Param::startup_ddl_path);
+    std::ifstream ddl_file(startup_ddl_location);
+    if (!(ddl_file.is_open() && ddl_file.good())) {
+      throw EXECUTION_EXCEPTION(fmt::format("Error locating startup DDL at \"{}\".", startup_ddl_location),
+                                common::ErrorCode::ERRCODE_INTERNAL_ERROR);
+    }
+    std::string input_line;
+    while (std::getline(ddl_file, input_line)) {
+      bool is_empty = input_line.empty();                                                       // Empty lines.
+      bool is_comment = input_line.size() > 2 && input_line[0] == '-' && input_line[1] == '-';  // -- comments.
```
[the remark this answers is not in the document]
lmwnshn: Very far punt, not at all.. I can add a comment here if you'd like, I asked the original PR introducing this to add a comment in the sql file https://github.com/cmu-db/noisepage/blob/master/src/main/startup.sql

src/metrics/query_trace_metric.cpp
```diff
@@ -29,27 +30,29 @@ void QueryTraceMetricRawData::SubmitFrequencyRecordJob(uint64_t timestamp,
...
-    param_vec[0] = parser::ConstantValueExpression(type::TypeId::BIGINT, execution::sql::Integer(timestamp));
-
-    // Record the query identifier.
-    param_vec[1] =
-        parser::ConstantValueExpression(type::TypeId::INTEGER, execution::sql::Integer(info.first.UnderlyingValue()));
-
-    // Record how many queries seen
-    param_vec[2] =
-        parser::ConstantValueExpression(type::TypeId::REAL, execution::sql::Real(static_cast<double>(info.second)));
+    std::vector<parser::ConstantValueExpression> param_vec = {
```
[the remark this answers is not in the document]
lmwnshn: Can I? I thought that would annoy the `move` below

src/self_driving/planning/pilot_util.cpp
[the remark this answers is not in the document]
lmwnshn: I replaced the TODO with your comment, thanks!

src/task/task.cpp
lmwnshn: This would require making the ctor public, are you in favor of that approach?
```
/usr/include/c++/9/bits/unique_ptr.h:857:30: error: ‘noisepage::task::TaskDDL::TaskDDL(noisepage::catalog::db_oid_t, std::string&&, noisepage::transaction::TransactionPolicy, noisepage::common::ManagedPointer<noisepage::common::Future<noisepage::common::FutureDummyResult> >)’ is private within this context
```

src/task/task.cpp
lmwnshn: Same as above, requires public ctor

src/task/task.cpp
lmwnshn: I'm surprised format didn't do it, actually.. done!

src/task/task.cpp
lmwnshn: specifically, range-based for and while loop, you could have stuffed dml_ok into the cond

src/self_driving/planning/planning_context.cpp
lmwnshn: I think this let me do forward declarations

src/util/self_driving_recording_util.cpp
lmwnshn: Yes, at the time I was not sure if it should be disabled durability (disabled durability didn't work). Good catch, refactoring to push it down

src/include/task/task.h
```diff
@@ -6,183 +6,199 @@
...
+      return *dynamic_cast<ConcreteType *>(this);
+    }
+    /** (Optional) Set the pointer to the Future that the caller is potentially blocking on. */
+    ConcreteType &SetFuture(const common::ManagedPointer<common::FutureDummy> sync) {
+      sync_ = sync;
+      return *dynamic_cast<ConcreteType *>(this);
+    }
+
+   protected:
+    std::optional<catalog::db_oid_t> db_oid_{std::nullopt};  ///< The database to execute the task in.
```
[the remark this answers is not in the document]
turingcompl33t: Nope, I usually just thing using `std::nullopt` directly looks a bit strange, but that's a personal taste thing. This is resolved.

src/main/db_main.cpp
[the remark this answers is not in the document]
turingcompl33t: I don't even think this is worth a comment at this point. As you say that is something that would be wayyy down the road, if ever. This is resolved.

src/metrics/query_trace_metric.cpp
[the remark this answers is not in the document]
turingcompl33t: This is more a curiosity than anything else but it does appear you can move from a `const std::vector`  in the call to `emplace_back()`: https://godbolt.org/z/sxjvW117q

src/task/task.cpp
turingcompl33t: Whoops, I need to pay closer attention to the context, that is the whole point of the builder... my bad! No I like the current approach. This is resolved.

src/task/task.cpp
turingcompl33t: Resolved.

src/util/self_driving_recording_util.cpp
lmwnshn: Refactored

src/metrics/query_trace_metric.cpp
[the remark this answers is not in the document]
lmwnshn: You can, but clang-tidy yells at me to take the move away. 
```
Clang-Tidy: Std::move of the const variable 'param_vec' has no effect; remove std::move() or make the variable non-const
```
I guess I'll mark this as resolved barring objections

lmwnshn:
Thanks for reviews! Going to call this blocked on #1566 as I hope to get that in.. for real.. in the near future..

lmwnshn:
While I would like this to get in, I guess this will not get in until we get the pilot in CI.

Issue #1598, opened 2021-05-27
CHbenchmark Q2 Performance


lmwnshn:
With 0 tuples loaded, only schemas created, CHBenchmark Q2 takes a long time to execute (debug mode):

```
noisepage=# \pset pager off
Pager usage is off.
noisepage=# \timing on
Timing is on.
noisepage=# SELECT su_suppkey,
       su_name,
       n_name,
       i_id,
       i_name,
       su_address,
       su_phone,
       su_comment
FROM item, supplier, stock, nation, region,
  (SELECT s_i_id AS m_i_id, MIN (s_quantity) AS m_s_quantity
   FROM stock,
        supplier,
        nation,
        region
   WHERE MOD ((s_w_id*s_i_id), 10000)=su_suppkey
     AND su_nationkey=n_nationkey
     AND n_regionkey=r_regionkey
     AND r_name LIKE 'Europ%'
   GROUP BY s_i_id) m
WHERE i_id = s_i_id
  AND MOD ((s_w_id * s_i_id), 10000) = su_suppkey
  AND su_nationkey = n_nationkey
  AND n_regionkey = r_regionkey
  AND i_data LIKE '%b'
  AND r_name LIKE 'Europ%'
  AND i_id=m_i_id
  AND s_quantity = m_s_quantity
ORDER BY n_name,
         su_name,
         i_id;
 su_suppkey | su_name | n_name | i_id | i_name | su_address | su_phone | su_comment 
------------+---------+--------+------+--------+------------+----------+------------
(0 rows)

Time: 4692.921 ms (00:04.693)
noisepage=# explain (format tpl) SELECT su_suppkey,
       su_name,
       n_name,
       i_id,
       i_name,
       su_address,
       su_phone,
       su_comment
FROM item, supplier, stock, nation, region,
  (SELECT s_i_id AS m_i_id, MIN (s_quantity) AS m_s_quantity
   FROM stock,
        supplier,
        nation,
        region
   WHERE MOD ((s_w_id*s_i_id), 10000)=su_suppkey
     AND su_nationkey=n_nationkey
     AND n_regionkey=r_regionkey
     AND r_name LIKE 'Europ%'
   GROUP BY s_i_id) m
WHERE i_id = s_i_id
  AND MOD ((s_w_id * s_i_id), 10000) = su_suppkey
  AND su_nationkey = n_nationkey
  AND n_regionkey = r_regionkey
  AND i_data LIKE '%b'
[...]

PR #1599, opened 2021-05-27
Small fixes to explain json and tpl for chbenchmark q2.

lmwnshn:
# Heading

Small fixes to explain json and tpl for chbenchmark q2.

## Description

Will allow #1598 to be reproduced with #1588.

### DIFF
--- a/src/self_driving/modeling/operating_unit_recorder.cpp
+++ b/src/self_driving/modeling/operating_unit_recorder.cpp
@@ -353,7 +353,7 @@ void OperatingUnitRecorder::AggregateFeatures(selfdriving::ExecutionOperatingUni
   std::vector<execution::translator_id_t> translator_ids;
   common::ManagedPointer<execution::compiler::OperatorTranslator> translator = current_translator_;
   translator_ids.emplace_back(translator->GetTranslatorId());
-  while (translator->IsCountersPassThrough()) {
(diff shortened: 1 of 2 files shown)


src/self_driving/modeling/operating_unit_recorder.cpp
```diff
@@ -353,7 +353,7 @@ void OperatingUnitRecorder::AggregateFeatures(selfdriving::ExecutionOperatingUni
   std::vector<execution::translator_id_t> translator_ids;
   common::ManagedPointer<execution::compiler::OperatorTranslator> translator = current_translator_;
   translator_ids.emplace_back(translator->GetTranslatorId());
-  while (translator->IsCountersPassThrough()) {
+  while (translator->GetParentTranslator() != nullptr && translator->IsCountersPassThrough()) {
```
lmwnshn, line 356: @17zhangw I have no recollection of this code, do you know if the first condition would be masking a more serious issue?

lmwnshn: It _seems_ fine to me...

lmwnshn: > It seems fine? I’m a little skeptical of the plans being generated though, cause this seems to imply a series of top-level projections

-- @17zhangw

PR #1607, opened 2021-06-02
Add support for aliases in GROUP BY. Add Postgres-style date_part.

lmwnshn:
# Heading

Add support for aliases in GROUP BY. Add Postgres-style date_part.

## Description

Add support for aliases in GROUP BY. Add Postgres-style date_part.

Note that the Postgres parser rewrites EXTRACT(YEAR FROM some_timestamp)
as date_part("year", some_timestamp).

Note that the optimizer gives us a bad OutputSchema when the GROUP BY expression
is on an aggregated column of the table, so this does not fix chbenchmark. It is currently
unclear if this is primarily a problem with the binder, optimizer, or translator.

Reasons:

- Translator: This is where the output schema kaboom happens.
- Optimizer: This produced the output schema. Maybe it never accounted for `select abs(a) as x ... group by x`.
- Binder: But maybe the optimizer would be fine if the binder wasn't swapping out expressions (quite literally, the group by `x` child becomes the expression `abs(a)` in the above example. This seems to work for ORDER BY fwiw.)

## Further work

This is enough to merge on its own, seeing if it gets through CI.

### DIFF
--- a/script/testing/junit/traces/groupby.test
+++ b/script/testing/junit/traces/groupby.test
@@ -1,0 +1,30 @@
+statement ok
+CREATE TABLE foo (a INT, b INT);
+
+statement ok
+INSERT INTO foo VALUES (1, 2);
+
+statement ok
+INSERT INTO foo VALUES (1, 3);
+
+statement ok
+INSERT INTO foo VALUES (2, 2);
+
+query I rowsort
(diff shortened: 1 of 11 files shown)


apavlo:
@lmwnshn This is a hot PR!

lmwnshn:
@lmwnshn fix up poke matt

PR #1609, opened 2021-06-06
Enable changing the compiled_query_execution setting at runtime

malin1993ml:
This PR adds a `CompiledQueryExecution` `SettingsCallback` to change the `execution_mode_` setting in `TrafficCop`. So that at run time, we can use `SET` command to change the `compiled_query_execution` setting.

Note that with the existing system architecture, the system only compiles the query when it **executes** the query as `Compiled` mode for the first time, instead of when constructing the `ExecutableQuery`. Once the query has been executed as `Compiled` once, it can switch between the `Interpert` and `Compiled` modes without more compilation.

### DIFF
--- a/src/settings/settings_callbacks.cpp
+++ b/src/settings/settings_callbacks.cpp
@@ -2,6 +2,7 @@
 
 #include <memory>
 
+#include "loggers/loggers_util.h"
 #include "main/db_main.h"
 
 namespace noisepage::settings {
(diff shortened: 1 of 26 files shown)


src/settings/settings_callbacks.cpp
```diff
@@ -186,6 +186,14 @@ void Callbacks::MetricsQueryTraceOutput(void *const old_value, void *const new_v
   action_context->SetState(common::ActionState::SUCCESS);
 }
 
+void Callbacks::CompiledQueryExecution(void *const old_value, void *const new_value, DBMain *const db_main,
+                                       common::ManagedPointer<common::ActionContext> action_context) {
+  action_context->SetState(common::ActionState::IN_PROGRESS);
+  bool new_status = *static_cast<bool *>(new_value);
```
lmwnshn, line 192: Nit: maybe rename new_status to is_compiled?

lmwnshn (approved):
LGTM

PR #1613, opened 2021-06-13
Multi part log record batch message

jkosh44:
### Summary of changes
Currently when we send a batch of log records form the primary to a replica the message looks something like this
```
{
    type: RECORDS_BATCH,
    batch_id: <batch-id>,
    contents: <serialized-batch-of-log-records>,
    metadata: {
        msg_id: <msg-id>
    }
}
```
We use MessagePack which is a binary form of JSON so in reality it's less human readable.

This change would split the message up into two messages. The first one looking something like this
```

{
    type: RECORDS_BATCH,
    batch_id: <batch-id>,
    metadata: {
        msg_id: <msg-id>
    }
}
```
and the second one would be the raw batch of log records.

The idea behind this is that we don't have to copy the batch of log records into the MessagePack before sending, which would save us time. We would increase the total number of messages sent, but the actual number of bits sent over the wire would be roughly the same. 

### Discussion
See the performance section below before reading this. The TPCC numbers are roughly the same before and after this change. The only difference is that log throughput on the primary node has increased by slightly more than 2x. 

After some thought and analysis I think this actually isn't that surprising. 
1. The TPCC numbers for disabled replication and async replication are roughly the same. This means that with MessagePack, async replication is no longer slowing down the DB performance.
2. Synchronous replication with and without the multi message change has roughly the same performance, even though the primary is generating log records much faster.
Both of these seem to suggest that the replica has become the bottleneck. Even though we are generating log records at a much faster pace the replica is too slow to take advantage of this.

One thing to consider is how I'm measuring log throughput. A log record is counted once we are done serializing it and we copy it into the messenger's queue. This doesn't take into consideration the time spent in the messenger itself. Since we are increasing the total number of messages sent by the messenger, we might be over estimating log throughput.

I'm a little hesitant to merge this change since it adds some complexity but doesn't actually speed up any queries. I think these results suggest that in order to speed up replication we're going to have focus on speeding up the replica. On the other hand it might be worth it to merge this because we may start to see advantages from it once the replica becomes faster and can handle more log records. I thought this PR would be a good place to discuss it.

### TPCC 

All TPCC metrics use scale factor 1, 8 threads, 60 seconds and dev machine 10

| Replication 	| Durability 	| Requests per Second |
| -------------------|-------------------|------------------------------|
| DISABLED   	| Sync        	| 216.7929849           	| 

#### Current
| Replication 	| Durability 	| Requests per Second |
| -------------------|-------------------|------------------------------|
| Sync   	| Sync        	| 16.09950183           	| 
| Async   	| Sync        	| 230.3881886           	| 
| Async   	| Async        	| 223.0805671           	| 

#### Multi Message
| Replication 	| Durability 	| Requests per Second |
| -------------------|-------------------|------------------------------|
| Sync   	| Sync        	| 16.71637636           	| 
| Async   	| Sync        	| 231.4920633           	| 
| Async   	| Async        	| 211.164187           	| 

### Log Throughput

All metrics for the primary node use the load stage of YCSB, scale factor 3000, 140 threads, and dev machine 10

All metrics for the replica node use the load stage of YCSB, scale factor 1000, a single thread, and dev machine 10

#### Current
| Primary or Replica | Replication 	| Durability 	| Log Throughput |
| -------------------------|-------------------|-------------------|-----------------------|
| Primary | Sync   	| Sync        	| 3.860618863          	| 
| Primary | Async   	| Sync        	| 29.38279985           	| 
| Primary | Async   	| Async        	| 28.47136216           	| 
| Replica | N/A   	| Sync        	| 3.887471136           	| 
| Replica | N/A   	| Async        	| 3.91150189           	| 

#### Multi Message
| Primary or Replica | Replication 	| Durability 	| Log Throughput |
| -------------------------|-------------------|-------------------|-----------------------|
| Primary | Sync   	| Sync        	| 3.984384733          	| 
| Primary | Async   	| Sync        	| 79.40114202           	| 
| Primary | Async   	| Async        	| 70.03673426           	| 
| Replica | N/A   	| Sync        	| 3.986637738           	| 
| Replica | N/A   	| Async        	| 3.994576457           	|


jkosh44:
Going to hold off on this until replica is sped up

Issue #1229, opened 2020-10-09
TPC-H : Generate incorrect operator tree for nested query 


xinzhu-cai:
# Bug Report

## Summary
TPCH Query 2 does not stop executing. The transformer.convertToOpExpression() generates an incorrect op_tree that contains the InnerJoin operator repeatedly (InnerJoin -> InnerJoin -> InnerJoin -> ...). Thus, OptimizeGroup tasks never stop.

## Environment
I don't think it's environment-specific. I run on macOS 10.14+ with Clang 8.0+

## Steps to Reproduce
- Compile with the following args: -DCMAKE_BUILD_TYPE=Release ..
- Run terrier executable
- Connect to the terrier terminal by executing : psql -h localhost -p 15721 -d terrier
- Create the tables from "https://github.com/oltpbenchmark/oltpbench/blob/master/src/com/oltpbenchmark/benchmarks/tpch/ddls/tpch-postgres-ddl.sql" .
- Run the query
```
-- TPC-H Query 2

select
        s_acctbal,
        s_name,
        n_name,
        p_partkey,
        p_mfgr,
        s_address,
        s_phone,
        s_comment
from
        part,
        supplier,
        partsupp,
        nation,
        region
where
        p_partkey = ps_partkey
        and s_suppkey = ps_suppkey
        and p_size = 15
        and p_type like '%BRASS'
        and s_nationkey = n_nationkey
        and n_regionkey = r_regionkey
        and r_name = 'EUROPE'
        and ps_supplycost = (
                select
                        min(ps_supplycost)
                from
                        partsupp,
                        supplier,
                        nation,
                        region
                where
                        p_partkey = ps_partkey
                        and s_suppkey = ps_suppkey
                        and s_nationkey = n_nationkey
                        and n_regionkey = r_regionkey
                        and r_name = 'EUROPE'
        )
order by
        s_acctbal desc,
        n_name,
        s_name,
        p_partkey
limit 100;
```
This query never stops running. We expect it to return an empty result.

lmwnshn:
2020-11-13: this query seems to run (though yes, I think it hits an optimizer timeout).

Patch IndexJoinTranslator to only initialize the table_pm_ if table columns are actually used:
```
diff --git a/src/execution/compiler/operator/index_join_translator.cpp b/src/execution/compiler/operator/index_join_translator.cpp
index 72277602b..ab8914745 100644
--- a/src/execution/compiler/operator/index_join_translator.cpp
[...]

PR #1619, opened 2021-06-18
Unique table alias

jkosh44:
## Description
SQL allows queries to have duplicate table aliases as long as they are at different query depths. For example the following is valid SQL
    SELECT * FROM foo AS f WHERE f.a IN (SELECT a FROM foo AS f)
Additionally if a table is not given an alias then we assign it an alias equal to it's name. Meaning that the following query will result in "foo" being a duplicate table alias
    SELECT * FROM foo WHERE a IN (SELECT a FROM foo)
The optimizer assumes that all table aliases will be unique, so these types of queries fail.

The solution is to change the type of the table alias in TableRef from string to AliasType, and assign a unique serial number to all TableRefs in a query.

The table name attribute of ColumnValueExpression actually corresponds to the table alias. So one consequence of this change was that we need to also update ColumnValueExpression to have an AliasType attribute for it's table alias. We also need to make sure that the serial number in the ColumnValueExpression's AliasType matches the serial number in the TableRef's AliasType so they could be matched up during optimization and query execution. This is handled in the binder.

This PR does not add supports for duplicate table aliases in CTEs.

Works towards fixing #1229 

## Remaining tasks
N/A

## Performance
N/A

## Further work
Address duplicate table aliases in CTEs


src/include/parser/expression/abstract_expression.h
```diff
@@ -6,6 +6,8 @@
 #include <vector>
 
 #include "common/hash_defs.h"
+#include "common/hash_util.h"
```
mbutrovich, line 9: We want to avoid including headers in other headers as much as possible to keep build times down. Can you move any implementation that needed these to the corresponding .cpp?

src/include/parser/expression/abstract_expression.h
```diff
@@ -76,6 +84,15 @@ class AliasType {
    */
   size_t GetSerialNo() const { return serial_no_; }
 
+  /**
+   * Set the serial number of this alias
+   * @param serial_no serial number to set alias to
+   */
+  void SetSerialNo(size_t serial_no) {
```
mbutrovich, line 91: We should STRONG_TYPEDEF something for this. We don't like using plain number types for unique identifiers since it makes it easy to mix them up. Maybe an alias_oid_t?

src/include/parser/expression/abstract_expression.h
jkosh44: Done

src/include/parser/expression/abstract_expression.h
jkosh44: Done

src/include/optimizer/abstract_optimizer_node_contents.h
```diff
@@ -66,7 +66,13 @@ class AbstractOptimizerNodeContents {
    * equal -- simply checks whether OpType and ExpressionType match
    */
   virtual bool operator==(const AbstractOptimizerNodeContents &r) {
-    return GetOpType() == r.GetOpType() && GetExpType() == r.GetExpType();
+    bool equal = GetOpType() == r.GetOpType() && GetExpType() == r.GetExpType();
```
jkosh44, line 69: It's been almost a year since I've worked on this, but this method confused me a little bit. Specifically why was it only checking the types and not the actual contents? I forget why but my changes didn't work without this change.

src/include/parser/expression/abstract_expression.h
```diff
@@ -66,6 +66,12 @@ class AliasType {
    */
   const std::string &GetName() const { return name_; }
 
+  /**
+   * Set the name of this alias
+   * @param name name to set alias to
+   */
+  void SetName(const std::string &name) { name_ = std::string(name); }
```
turingcompl33t, line 73: `name_` is a `std::string` so there is no need to manually call the `std::string` constructor here.

src/include/parser/expression_defs.h
```diff
@@ -3,9 +3,13 @@
 #include <string>
 
 #include "common/enum_defs.h"
+#include "common/strong_typedef.h"
 
 namespace noisepage::parser {
 
+// It's important that alias_oid_t is the same size as pointers because we often convert a pointer to an alias_oid_t
+STRONG_TYPEDEF_HEADER(alias_oid_t, size_t);
```
turingcompl33t, line 11: namespace `std` re-exports the C library types like `size_t`, and real C++ devs always use `std` where possible.

src/parser/postgresparser.cpp
```diff
@@ -581,10 +582,21 @@ std::unique_ptr<AbstractExpression> PostgresParser::ColumnRefTransform(ParseResu
       if (all_columns)
         result = std::make_unique<TableStarExpression>(table_name);
       else if (alias != nullptr)
+        /*
```
turingcompl33t, line 585: Thanks for this comment here, I was pretty confused above where we were using `reinterpret_cast` to go from an expression pointer to an `alias_oid_t`.

turingcompl33t (approved):
None of the comments are in desperate need of attention, you can take them or leave them prior to merging. Otherwise looks good to me, thanks for this update! I will take care of any fallout as it pertains to CTEs, but I actually have internal logic that addresses most issues I can think of with duplicate references.

src/include/binder/binder_sherpa.h
```diff
@@ -29,7 +31,10 @@ class BinderSherpa {
   explicit BinderSherpa(const common::ManagedPointer<parser::ParseResult> parse_result,
                         const common::ManagedPointer<std::vector<parser::ConstantValueExpression>> parameters,
                         const common::ManagedPointer<std::vector<execution::sql::SqlTypeId>> desired_parameter_types)
-      : parse_result_(parse_result), parameters_(parameters), desired_parameter_types_(desired_parameter_types) {
+      : parse_result_(parse_result),
+        parameters_(parameters),
+        desired_parameter_types_(desired_parameter_types),
+        unique_table_alias_serial_num_(1) {
```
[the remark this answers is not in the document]
jkosh44: I'm actually a professional MATLAB programmer which is why I prefer the convention.

Actually though it's so we know that when it goes to 0 we've wrapped around and have overflow.

src/include/binder/binder_sherpa.h
```diff
@@ -113,11 +118,25 @@ class BinderSherpa {
    */
   bool HasCTETableName(const std::string &cte_table_name) const { return cte_table_names_.count(cte_table_name) > 0; }
 
+  /**
+   * Get a unique table serial number for this query. This will help ensure that all table aliases are unique.
+   * @return unique serial number
+   */
+  parser::alias_oid_t GetUniqueTableAliasSerialNumber() {
+    auto unique_serial_number = unique_table_alias_serial_num_++;
+    if (unique_serial_number.UnderlyingValue() == 0) {
```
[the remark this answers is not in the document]
jkosh44: Do you think I should remove this then? I was actually a little conflicted here because I'll be very slightly slowing down every single query to check for an error condition that will likely never happen. I was just hoping branch prediction would make this a no-op.

src/include/parser/expression_defs.h
jkosh44: Can you dumb this down for me, you're saying I should use `std::size_t` instead of `size_t`?

src/parser/postgresparser.cpp
jkosh44: Thanks, but I don't think this comment actually addresses that. This comment is explaining that for SELECT queries at some point we'll have to add a serial number to the `ColumnValueExpression`'s table `AliasType`, so that it matches the serial number of the corresponding `AliasType` that belongs to the `TableRef`.

Later in the binder `ColumnValueExpression`s are matched to `TableRef`s using the `AliasType` so they need to be equivalent.

On a separate note, we are using `reinterpret_cast` to go from an expression ponter to an `alias_oid_t` to create the serial number for the `ColumnValueExpression`'s column `AliasType`. The assumption here is that the pointer will be unique. That was already there though so I didn't really add a comment for it. Though I can now.

src/include/parser/expression_defs.h
jkosh44: I think that's what you're saying, I made the change.

src/include/parser/expression/abstract_expression.h
jkosh44: Changed

src/include/binder/binder_sherpa.h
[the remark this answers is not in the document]
turingcompl33t: Dang I didn't actually connect the dots on that... nice!

PR #1620, opened 2021-06-19
Fix trace files that end in comments

### LINKED ISSUE #1604: TracefileTest Breaks if Trace Ends with Comment
If the last line in the input SQL file is a comment (i.e. begins with `#`) no `statement ok` message is produced in the resulting tracefile. As a consequence, the `TracefileTest.java` program throws an `IndexOutOfBounds` exception when it attempts to run the associated test.

jkosh44:
closes #1604


jkosh44:
I was originally planning on cleaning up the Java logic to handle this scenario, but I think this hack is more appropriate. I'm probably going to open a separate PR to clean up the Java since there were a couple of issues I noticed, such as unclosed resources.

lmwnshn (approved):
LGTM

turingcompl33t (approved):
@lmwnshn already reviewed / approved but since my review was requested: LGTM. Would also be very interested to see the Java refactor mentioned in the conversation above, as performed by a real, professional Java programmer.

PR #1621, opened 2021-06-20
Add ClearQueryCache setting

malin1993ml:
This PR introduces a `ClearQueryCache` setting that forces the `TrafficCop` to clear the query plans and executable queries cached in the portals and re-generate them. Similar to the settings that trigger model training, the value of this setting is irrelevant and does not have any impact during the database startup.

Since the DBMS may still be executing the query when `ClearQueryCache` is triggered, the settings callback cannot directly remove the `ExecutableQuery` from the cache. Rather, it advances the minimum timestamp (stored in `TrafficCop`) allowed for when an `ExecutableQuery` is generated. When the `TrafficCop` executes queries, it checks whether the `ExecutableQuery`'s generation timestamp is outed. If so, it re-optimizes the query and re-generates the `ExecutableQuery`. The `ExecutableQuery` class stores this generation timestamp as a member variable.


src/include/settings/settings_defs.h
```diff
@@ -413,6 +413,15 @@ SETTING_bool(
     noisepage::settings::Callbacks::CompiledQueryExecution
 )
 
+SETTING_bool(
+    clear_query_cache,
```
lmwnshn, line 417: I feel we are also generally trending towards adding a lot of self-driving "settings" that should really be functions

src/include/traffic_cop/traffic_cop.h
```diff
@@ -282,6 +289,7 @@ class TrafficCop {
   common::ManagedPointer<optimizer::StatsStorage> stats_storage_;
   uint64_t optimizer_timeout_;
   const bool use_query_cache_;
+  uint64_t query_cache_timestamp_;
```
lmwnshn, line 292: Should this be a strong typedef timestamp_t?

src/include/network/postgres/portal.h
```diff
@@ -59,12 +59,20 @@ class Portal {
    * @return params for this query
    */
   common::ManagedPointer<const std::vector<parser::ConstantValueExpression>> Parameters() {
+    return common::ManagedPointer(reinterpret_cast<const std::vector<parser::ConstantValueExpression> *>(&params_));
+  }
+
+  /**
+   * @return modifiable params for this query (Some existing API requires this version. Use with caution.)
+   * TODO(lin): If we have time, we probably should refactor other APIs to always use the const version
```
lmwnshn, line 67: I don't know if this will realistically ever happen if we don't do it now -- do we have an idea of how bad the modification would be?

src/include/settings/settings_defs.h
lmwnshn: An option is to invest the work now in implementing just one bytecode, `self_driving_action(action_enum)`
To implement all of the self driving functions based on the `action_enum` value passed in?

src/include/traffic_cop/traffic_cop.h
lmwnshn: I realize this is actually metrics timestamps and not txn::timestamps. They have no relationship I guess?

src/include/network/postgres/statement.h
```diff
@@ -148,6 +156,7 @@ class Statement {
   // same query text. The exception to this that DDL changes can break these cached objects.
   std::unique_ptr<optimizer::OptimizeResult> optimize_result_ = nullptr;              // generated in the Bind phase
   std::unique_ptr<execution::compiler::ExecutableQuery> executable_query_ = nullptr;  // generated in the Execute phase
+  uint64_t executable_query_timestamp_;                                               // generated with executable_query
```
lmwnshn, line 159: Actually, if the executable query timestamp is a property of the executable query, should it simply be stored in the `ExecutableQuery` class? Then you also don't need a default timestamp, if you have a valid `ExecutableQuery` then you have a valid timestamp.

lmwnshn (commented):
Some quick high-level questions:

- Is this on the critical path of something? I wonder if it is worth the time to just write one builtin function now, `select self_driving_action('enable_pilot')`, `select self_driving_action('reset_cache')`, instead of continuing to implement custom "settings" that are not settings.
- I think there's potential for mixing up metrics timestamps and txn timestamps, but I realize MetricsUtil is also just returning `uint64_t`.
- Is it possible to push the executable query timestamp into the executable query? Since it seems a valid timestamp doesn't exist without the executable query.

src/include/network/postgres/portal.h
malin1993ml: As discussed over Slack, this would require casting a non-const managed pointer to a const one, which can be either hacky or verbose. We might eventually need the `ManagedPointer` class to support this behavior.

src/include/settings/settings_defs.h
malin1993ml: We'll punt on this and I've created an issue to track this (#1622).

lmwnshn (approved):
LGTM

PR #1623, opened 2021-06-23
Index Knobs Support

17zhangw:
## Description
Supports specifying index knobs (inner node split/merge threshold) through SQL statements. The following are the big changes in this PR that enable such support:
- Pipeline Metrics now have 2 additional fields (split/merge for B+trees on OUs that touch indexes)
- pg_class now has `reloptions` which contains a string of the `key=value` options specified
- `CREATE INDEX ... WITH (key=value)` parsing support to pass down the options

## Remaining Work

- [x] Reconstruct on recovery
- [x] Documentation
- [x] PR through CI


src/include/catalog/index_schema.h
```diff
@@ -280,6 +503,30 @@ class IndexSchema {
...
+   * Instantiates a new catalog description of an index
+   * @param columns describing the individual parts of the key
+   * @param type backing data structure of the index
+   * @param is_unique indicating whether the same key can be (logically) visible repeats are allowed in the index
+   * @param is_primary indicating whether this will be the index for a primary key
+   * @param is_exclusion indicating whether this index is for exclusion constraints
+   * @param is_immediate indicating that the uniqueness check fails at insertion time
+   * @param index_options that are options for building the index
+   */
+  IndexSchema(std::vector<Column> columns, const storage::index::IndexType type, const bool is_unique,
```
malin1993ml, line 516: I'm a little bit hesitant about adding a new `IndexSchema` constructor. Can we remove the old constructor and then passing in an empty `index_options` when not applicable?

malin1993ml (commented):
Looks good! Thanks a lot, William! I only have a minor comment. There are just two additional things:
(1) Could you add a test case for this? I think this could be a SQL test, similar to the `analyze_test`. We could test that the thresholds for the indexes are set correctly. Ideally, we also want to test that the index is built with the correct number of threads, but it's fine if that's too difficult.
(2) Could you add the OU recording for the parallel create index? I left you some detailed comments on Slack.
Thanks!

malin1993ml (approved):
LGTM! Thanks a lot, William!

Issue #1359, opened 2020-11-30
pg_proc PROALLARGTYPES does not follow Postgres semantics.


lmwnshn:
In #1344, @jrolli points out [here](https://github.com/cmu-db/noisepage/pull/1344/files/7de2d3cece60a1e4d10ff288c323bb50a5f1606f#diff-fd3f09b3d960acd569446954b3a5cf1ecfc05d54a3da11fdb19cb537bd8924f4R717) that pg_proc's PROALLARGTYPES is not following what Postgres does:

> Does not follow semantics of Postgres' column. From the catalog documentation
> [...] however, if all the arguments are IN arguments, this field will be null.[...]
> Specifically, this column should be NULL for the vast majority of standard functions in the table which ties back to a logic bug in the implementation.

lmwnshn:
The same is true for proargmodes.

Issue #1183, opened 2020-09-15
Fix made up types


jkosh44:
Custom types `PARAMETER_OFFSET` and `VARIADIC` are not real types and should be refactored.

https://github.com/cmu-db/noisepage/blob/master/src/include/type/type_id.h

Brought up in #1139

PR #911, opened 2020-05-21
Remove type::TransientValue and related classes

mbutrovich:
While reviewing object life cycles in the front-end, I decided to finally remove this stopgap solution. `type::TransientValue` and its related factory and peeker were temporary ports from Peloton until we got the execution engine in and could use its `Value` type. We built more and more on top of `type::TransientValue`, but it's time to go.

`parser::ConstantValueExpressions` now own a `execution::sql::Val` pointer (and possible extra allocated buffer if it's a non-inlineable variable length type) instead of a `type::TransientValue`. Components (`SettingsManager`, `TrafficCop` parameters) that used `type::TransientValue` now use `parser::ConstantValueExpression` instead.

**Major Changes:**
- Remove `type::TransientValue`
- Remove `type::TransientValueFactory`
- Remove `type::TransientValuePeeker`
- Add `execution::sql::Val` member to `parser::ConstantValueExpression`
- New `execution::sql::ValueUtil` class to simplify construction/allocation of `StringVal`

**Future Work:**
- Does `ConstantTranslator` need to exist in code generation anymore?

**Misc:**
- Fix #407, apparently

PR #1628, opened 2021-07-15
Type cleanup

mbutrovich:
### Background:
`type::TypeId` has been another bee in my bonnet like `TransientValue` was, which was also a stopgap as other parts of the system were built without the execution engine or optimizer yet. As of right now we have about 5 different type definitions in the system: `PostgresValueType` in the network layer, Peloton-like `type::TypeId` in `/include/type`, other Postgres types in the parser, and `SqlTypeId` and `TypeId` in `/include/execution/sql`. 

### Overview:
Remove `type::TypeId` and replace with `execution::sql::SqlTypeId`.

### Changes:
- Finishes what #911 started by removing the rest of the `/type` folder.
- Remove `TypeId::VAR_ARRAY`. Doesn't seem like it was used and not a real type.
- Remove `ColumnMapInfo` from `ColumnMap` in the `SqlTable`. This went in with TPL2 and stashed type info at the storage layer. That's a huge footman for how we plan to do schema change in the future, and the catalog is the sole authority of transactional schema information.
- CTEs stash their temp tables' schemas in the `CatalogAccessor` just like their temp tables. `TableVectorIterator` was relying on the `ColumnMapInfo` and now requests schema for all tables from `CatalogAccessor`, so this is transparent to the execution engine.
- Remove inserting of INVALID type into pg_type since that's not a real type.
- pg_type_impl uses the type sizes defined in sql.h/.cpp rather than hardcoded.

### Future work:
- This partially addresses #1183, but we still have `SqlTypeId::Variadic` which is used in pg_proc incorrectly. This will take a larger rewrite of pg_proc behavior to actually match Postgres (see #1359). I tried to do it in this PR, but the blast radius gets bigger than I'd like for an already-large PR.
- Discuss REAL type. TPL supports it, but the front-end doesn't and neither does the storage layer. Current behavior just does a magic trick and uses DOUBLE, but maybe we should explicitly tell the user they're not getting REALs?
- Discuss renaming TPL's Real to Double since that's what it actually is, and it's confusing in the current context.
- Discuss merging `PostgresValueType` with `SqlTypeId` to reduce type translation. This would bump the latter to being an `int32_t` and has a large blast radius.
- Discuss fixing the whole "NULLs come in as INVALID type" until fixed much later by the binder and execution engine. Large blast radius on this one and not in the scope of this PR.


mbutrovich:
Ran TATP and YCSB read only with 20 terminals on dev9 and didn't see a performance change.

PR #1629, opened 2021-07-23
End-to-end (JUnit and oltpbench) only for coverage

mbutrovich:
This is a test PR just to see what coverage looks like if we only run JUnit and oltpbench.


mbutrovich:
60% is way better than I expected. We should discuss putting together a real PR for this next week.

Issue #1361, opened 2020-12-01
Build a unique index on pg_proc's argtypes.


lmwnshn:
# Build a unique index on pg_proc's argtypes.

## Summary

In #1344, @jrolli [suggests](https://github.com/cmu-db/noisepage/pull/1344/files#diff-fd3f09b3d960acd569446954b3a5cf1ecfc05d54a3da11fdb19cb537bd8924f4R767) that we might be able to remove [this](https://github.com/cmu-db/noisepage/pull/1344/files#diff-cbdefe66528d610c036d9c7be5c1daac083c4d4bafc31198d30fee46f16432adR378) `BINDER_EXCEPTION` in the catalog by adding the argtypes to the index.

Issue #1354, opened 2020-11-26
Refactor pg_proc CreateProcedure signature by packing things into structs.


lmwnshn:
# Refactor pg_proc CreateProcedure signature by packing things into structs.

## Summary

The `CreateProcedure` signature is pretty messy:

```
  bool CreateProcedure(common::ManagedPointer<transaction::TransactionContext> txn, proc_oid_t oid,
                       const std::string &procname, language_oid_t language_oid, namespace_oid_t procns,
                       const std::vector<std::string> &args, const std::vector<type_oid_t> &arg_types,
                       const std::vector<type_oid_t> &all_arg_types,
                       const std::vector<postgres::PgProc::ProArgModes> &arg_modes, type_oid_t rettype,
                       const std::string &src, bool is_aggregate);
```

In #1344, @jrolli suggested encapsulating this information for both creators and consumers in a struct.

> The description of a procedure probably deserves its own class/struct which should be used to pass this information instead (and probably be used on the consumer side of this information as well). Alternatively, how much of this information can we obtain from the FunctionContext when it is passed in?

lmwnshn:
Additionally, see this review comment chain: https://github.com/cmu-db/noisepage/pull/1344/files/7de2d3cece60a1e4d10ff288c323bb50a5f1606f#diff-55d79f738f783bcf4e7188b5280fbf0f1e05efcde4a4acfe689f0343b6f4ee34R1757-R1759

PR #1630, opened 2021-07-24
pg_proc cleanup

### LINKED ISSUE #1183: Fix made up types

### LINKED ISSUE #1359: pg_proc PROALLARGTYPES does not follow Postgres semantics.

mbutrovich:
### Background:
Fix #1183. Fix #1359. pg_proc doesn't really respect Postgres' semantics. We are denoting variadic args using a dedicated type. We populate fields like `proallargtypes` and `proargmodes` when they should be NULL-able and in most cases be NULL since almost all of our procedures right now have `IN` args only.

### Overview:
Clean up insertion and lookup in pg_proc to adhere more closely to Postgres' behavior. We still have some work to do there. 

### Changes:
- Remove the `VARIADIC` type which isn't real and isn't how Postgres denotes a variadic function signature.
- `proargtypes`, `proallargtypes`, `proargmodes`, `proargnames` are now NULL-able like Postgres since these are not required fields.
- The aforementioned fields now mirror Postgres behavior as much as possible, though I don't think we really support `OUT` or `INOUT` function arguments right now.

### Comments:
In general, we brought a lot of columns into pg_proc that are just dummy fields for us right now. I don't like that approach because it leaves us with more code to maintain that isn't getting us anything. I'm a proponent of bring in the minimum number of Postgres tables and columns of those tables as necessary to accomplish what we want, and only add more later. Copying the entire Postgres schema and dumping in placeholder data isn't useful unless we have a use case (specific test/benchmark).

### Future work:
- #1354 wants to see the function signature for creating a procedure simplified to take a struct instead of the (now worse) long function signature.
- #1361 proposes an index to fix some duplicate entries.


src/common/strong_typedef.cpp
```diff
@@ -64,7 +64,5 @@ template class StrongTypeAlias<noisepage::selfdriving::pilot::tags::tree_node_id
 template class StrongTypeAlias<noisepage::storage::tags::col_id_t_typedef_tag, uint16_t>;
 template class StrongTypeAlias<noisepage::storage::tags::layout_version_t_typedef_tag, uint16_t>;
 template class StrongTypeAlias<noisepage::transaction::tags::timestamp_t_typedef_tag, uint64_t>;
-// template class StrongTypeAlias<noisepage::type::tags::date_t_typedef_tag, uint32_t>;
```
mbutrovich, line 67: These were commented out in #1628 but I forgot to go back and actually delete the lines.

PR #1631, opened 2021-07-26
Refactor support for SHOW commands in DESCRIBE.

lmwnshn:
# Heading

Refactor support for SHOW commands in DESCRIBE.

## Description

Previously, DESCRIBE was writing NO DATA for SHOW.
This fix makes the HikariCP connection pool in the new version of OLTPBench happy.

## Remaining tasks

- Get through CI.


mbutrovich (approved):
That code to generate an OutputSchema for SHOW living in postgres_network_commands feels funky to me, but I don't have a clear argument for where that bespoke code should live otherwise. LGTM.

PR #1640, opened 2021-11-05
remove junit from microbenchmark stage

crd477:
This stage does not produce any junit results anyway.  The junit
Jenkins plugin prior to version 1.53 would silently ignore missing &
empty test result files.  As of version 1.53, it produces an error that
fails the stage.

Signed-off-by: Chad Dougherty <crd@acm.org>


crd477:
For a while, I was able to manually hold back the update to the junit plugin in Jenkins.  As time went on, other plugins started specifying the newer version as a dependency so it's just easier to do the proper fix in the pipeline here.

